Issue #3, opened 2021-03-21
Welcome screen 


bumi:
After the user installs the extension we need some kind of onboarding wizzard that gives the user some introduction and helps the user getting started connecting an account. 

### Figma screens:
https://www.figma.com/file/xEPJFGFdc3HD0JS1ax2IpH/Lightning-Browser-Extension

### Current page: 
https://github.com/bumi/lightning-browser-extension/blob/master/source/Welcome/Welcome.js

### Goals

- Guide the user through the initial required account configuration
- Validate and test the account 
- Explain the user the concept of lightning payments in the browser

### Steps

- Intro
- Choose an account type (LND, Lnbits, etc.) 
- Configure account (e.g. enter LND macaroon and URL) (later we can also add some onramp tools here)
- Validate and review the connection details (can we connect to the node, get node details)
- Confirm and save the account
- Success: explain what can be done now. give an example of a payment to test. - maybe list pages that support webln

### ToDos:

- [ ] automatically open the page after the user has installed the extension 

### Inspiration: 
https://metamask.zendesk.com/hc/en-us/articles/360015489531-Getting-Started-With-MetaMask-Part-1-
<img width="269" alt="image" src="https://user-images.githubusercontent.com/318/111898379-57236c00-8a2e-11eb-8396-5a9a142e2264.png">
<img width="360" alt="image" src="https://user-images.githubusercontent.com/318/111898384-5c80b680-8a2e-11eb-81d7-81c65e070f53.png">
<img width="323" alt="image" src="https://user-images.githubusercontent.com/318/111898386-5ee31080-8a2e-11eb-8e8c-225911cecce1.png">
<img width="318" alt="image" src="https://user-images.githubusercontent.com/318/111898387-60143d80-8a2e-11eb-9636-5a7e020a5eb9.png">
<img width="392" alt="image" src="https://user-images.githubusercontent.com/318/111898389-64d8f180-8a2e-11eb-8578-f77bd0aa6712.png">

### Later

johnsBeharry:
what are going to be the supported account types for the hacksprint?

bumi:
I think we should focus on LND first. (currently I only added LndHub and lnbits as a proof of concept)

bumi:
we got some amazing onboarding screens now! thanks @johnsBeharry @ShawkiS @dylancom

PR #21, opened 2021-03-28
[WIP] Welcome wizard

bumi:
The welcome wizard helps the user to configure the account and get started. 

See miro and figma boards for flow and design details

[ref #3]

### DIFF
--- a/source/Welcome/Welcome.js
+++ b/source/Welcome/Welcome.js
@@ -1,45 +1,42 @@
 import React from "react";
 import browser from "webextension-polyfill";
 
+import { createHashHistory } from "history";
+import { HashRouter, Switch, Route } from "react-router-dom";
+
 import utils from "../lib/utils";
+
+import Home from "./pages/Home";
+import LndSetup from "./pages/LndSetup";
+import LndConfirm from "./pages/LndConfirm";
+import Success from "./pages/Success";
+
 import Accounts from "../lib/accounts";
-import Allowances from "../lib/allowances";
 import Settings from "../lib/settings";
 
 class Welcome extends React.Component {
   constructor(props) {
     super(props);
     this.state = {};
-    this.accounts = new Accounts();
-    this.settings = new Settings();
-    this.allowances = new Allowances();
+    this.history = createHashHistory();
+    this.accountsStore = new Accounts();
   }
 
-  componentDidMount() {}
-
-  reset() {
-    //browser.storage.sync.set({ accounts: {} });
-    //browser.storage.sync.set({ settings: {} });
-    //browser.storage.sync.set({ allowances: {} });
-    this.accounts.reset();
-    this.allowances.reset();
-    this.settings.reset();
-    alert("Done, you can start over");
-    utils.openPage("Options.html");
+  componentDidMount() {
+    this.accountsStore.reset();
   }
 
   render() {
     return (
-      <div>
-        <p>Hallo</p>
-        <span
-          onClick={() => {
-            this.reset();
-          }}
-        >
-          Reset Everything
-        </span>
-      </div>
+      <HashRouter>
+        <Switch>
+          <Route exact path="/" render={(props) => <Home />} />
+          <Route exact path="/password" render={(props) => <Home />} />
+          <Route exact path="/lnd/setup" render={(props) => <LndSetup />} />
+          <Route exact path="/lnd/confirm" render={(props) => <LndConfirm />} />
+          <Route exact path="/success" render={(props) => <Success />} />
+        </Switch>
+      </HashRouter>
     );
   }
 }
--- a/source/Welcome/Welcome.js
+++ b/source/Welcome/Welcome.js
@@ -17,9 +17,37 @@ import Settings from "../lib/settings";
 class Welcome extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {};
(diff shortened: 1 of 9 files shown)


bumi:
@reznord do you have a suggestion on what's an easy way to keep state between two components? e.g. in this case between two router steps (LndSetup and LndConfirm for example?)

chris-aeviator:
@bumi https://redux.js.org/

bumi:
@chris-aeviator thanks, do you recommend we should introduce redux?

chris-aeviator:
It is possible to go completely without a store when refactoring so that the extension's main entrypoint (e.g. a not yet existing `app.js`) holds all state and passes it via props (https://reactjs.org/docs/components-and-props.html). You can also pass setter functions like `const myFancyProp, setMyFancyProp = useState()`, defined by state hooks (https://reactjs.org/docs/hooks-state.html)...I believe this works the same with class based state hooks as used in this extension.

I think it boils down to the question if we will further down the road have more state to manage. Nuxt/Vue uses a system like redux natively and it works pretty good. Redux is the go to solution for many in react.

bumi:
hard for me to decide because I lack the react experience. 

the use case that I have here is the following:
1. screen: allow the user to set a password  
2. screen: ask for node config
3. then encrypt the node config with the password given in 1. screen. and store the encrypted config

chris-aeviator:
I'll have some free time in the early morning tomorrow and will take a look, it sounds to me that with some refactoring we can go without.

source/Welcome/Welcome.js
```diff
@@ -1,45 +1,42 @@
...
+    this.accountsStore.reset();
   }
 
   render() {
     return (
-      <div>
-        <p>Hallo</p>
-        <span
-          onClick={() => {
-            this.reset();
```
chris-aeviator, line 37: reset functionality has gone nirvana - any replacement?

chris-aeviator:
@bumi no refactoring was necessary. I showcased how to pass data (that's hold by a central state in https://github.com/bumi/lightning-browser-extension/blob/0135057d9e5dc24346803a6bc105f7c0702afb09/source/Welcome/Welcome.js ) from the _Setup_ step to the _Confirm_ step of the welcome wizard. If you agree we can discuss what data really should be passed between which screens and implement the single screens.

reznord:
> @reznord do you have a suggestion on what's an easy way to keep state between two components? e.g. in this case between two router steps (LndSetup and LndConfirm for example?)

Sorry for the late reply, but since the content passing is between two different routes of the same page, it is better to go with redux as @chris-aeviator suggested earlier.

bumi:
@reznord why do you suggest redux instead of @chris-aeviator's simpler solution to pass state to the central Welcome.js?

reznord:
The reason why I suggested to use redux is to avoid prop-drilling through the components. I would totally agree with just passing it as a prop to the component provided we write the entire view logic for the page in one single component. But that makes readability a lot harder.

So, to avoid issues like this, there are two options:
- Pass the prop to the page as a `prop` and then on the page level add it on the context provider where you can share the required prop across all components by avoiding prop drilling. The problem with this approach is that you will have a lot of manual boilerplate to handle for each and every page. 
- Do the store logic manipulations on the redux store and then you can access the data from the store directly in any component. This advantages of this particular approach would be less boilerplate code (since you just need to wrap the redux  provider only on the top level where you render the app) and there won't be any extra logic on each screen to handle the prop drilling etc.

This is the reason why I suggested to go with redux.

chris-aeviator:
I agree with @reznord 's view in partial. Since we're just passing a minimal amount of information here it will not grow anymore unreadable as of now though.

Bumi defined 

> 
1.  screen: allow the user to set a password
2.    screen: ask for node config
3.   then encrypt the node config with the password given in 1. screen. and store the encrypted config

and that's it. I'd say if it comes to a point where we need to work with more state logic (and I don't see it at the moment) we could re-discuss redux but for now my view is to save our resources and concentrate on coding the screens?!

We can further reduce the props passed from e.g.

```
      <LndConfirm
                lndName={this.state.lndName}
                lndMacaroon={this.state.lndMacaroon}
                lndUrl={this.state.lndUrl}
              />
```

to 

```
      <LndConfirm
                lndConfig={this.state.lndConfig}
              />
```

chris-aeviator:
btw. for better explanation to @bumi - phraseology like "prop drilling" boils down to paradigms around communities.

The react community (not react itself) discourages the concept and calls it prop drilling, emberJS (and I believe also vue) for example encourages it and calls it "data down – actions up".

 I think it more boils down to a "believe". Following that we'd also need to discuss to rewrite everything to functional components instead of class based components, since this is where the community seems to be heading.

bumi:
thanks for that explanation! puh... :)  
Generally I just think it is important to follow best practices and standards as much as possible. 

but that means we go for now with passing in the props and if it becomes too much and complicated we move to redux? - and that should be rather simple to do then?

chris-aeviator:
the only thing I find in the react docs about it is

"A component may choose to pass its state down as props to its child components:" (https://reactjs.org/docs/state-and-lifecycle.html#the-data-flows-down)

chris-aeviator:
will continue with the welcome wizard, open to any refactoring shortly thereafter.

bumi:
closing in favor of #44

Issue #1, opened 2021-03-20
Initial popup frontend


bumi:
Initial frontend for the extension popup including: 

* unlock option if the account is locked
* balance in satoshis (and fiat - which later should be configurable)
* receive / send (links to create an invoice or pay an invoice from within the popup)
* switch accounts (link to list and select an account - if multiple ones arr configured)

## Current status:
https://github.com/bumi/lightning-browser-extension/blob/master/source/Popup/Popup.js

## Inspiration
<img width="300" src="https://user-images.githubusercontent.com/318/111884098-eba6b380-89c7-11eb-92c3-4191e9191cd9.png">
<img width="300" src="https://user-images.githubusercontent.com/318/111884020-7f2bb480-89c7-11eb-8b54-4435811b6bad.png">

chris-aeviator:
- [ ]  unlock option if the account is locked

- [ ] balance in satoshis (and fiat - which later should be configurable)

> receive / send (links to create an invoice or pay an invoice from within the popup)
@bumi is this relevant or better the job of a wallet extension like joule
- [ ] switch accounts (link to list and select an account - if multiple ones arr configured)

bumi:
what you you mean by "better the job of a wallet extension like joule"?

chris-aeviator:
I understood that you wanted this extension to not represent a full wallet , that's why I'm saying wallet extension like joule

bumi:
sorry, I still don't understand how that is related to `receive / send (links to create an invoice or pay an invoice from within the popup)`

with "send / receive" options I understand that the user can manually initiate a payment (by copy and paste an invoice, or an lnurl text or something like that) and also can create an invoice to receive a payment (that then needs to be provided to a website or uploaded somewhere)


For now I think this extension should not try to run a full lightning node in the browser just because I think that adds a lot of complexity and probably requires a lot of work. I'd rather for now focus on the UX of using lightning in the web. 
Later I think it could be cool to have a non-custodial lighting wallet in the browser (just like one on the mobile phone)

chris-aeviator:
> sorry, I still don't understand how that is related to receive / send (links to create an invoice or pay an invoice from within the popup)

I think it's overkill for this extension, otherwise it would just be copying what the others are doing. I believe if we focus purely on letting users pay for stuff online with our extension, and do not offer manual payment/sending we might reach a different set of users, that find 128 character hex invoices way too confusing. *opensTheDiscussion* 🎬

bumi:
[...]

PR #22, opened 2021-03-31
#1 fix 'browser is not defined' error on chrome

motorina0:


reznord (approved):

bumi:
nice! that's great. thanks for the fix and improvement!

PR #41, opened 2021-04-09
Basic Wallet Info with Transactions

chris-aeviator:
- Implements a basic transaction history
- optimized on zero transactions also
- shows account name (with edit button -> `options.html`)
- shows USD amount calculated from current BTC rate 

## Transacting a lot
image
**cursor is hovering the time entry, dunno why not visible**
## No transactions yet

image


chris-aeviator:
TODO:

- [ ]  remove wallet reference from connect src

bumi:
:shipit: merging this, and we do further updates in new PRs.

src/extension/background-script/connectors/lnd.js
```diff
@@ -61,6 +61,12 @@ class Lnd extends Base {
     });
   };
 
+  getTransactions = () => {
+    return this.request("GET", "/v1/transactions", undefined, {
```
bumi, line 65: oh and this should be "list payments"
https://api.lightning.community/#v1-payments
https://api.lightning.community/#listpayments

The payments are the lightning payments. Transactions are the on-chain bitcoin transactions.

chris-aeviator: thanks for making that clear - was confused already that they seemed not up to date

chris-aeviator: https://github.com/bumi/lightning-browser-extension/issues/45

PR #54, opened 2021-04-20
Issue 39 & restructure app

motorina0:
- [x] abstract the messaging logic in one simple service. All components should only use this service for messaging.
  - a `msg` util was already present, but the logic is too heavy (will probably be replaced by `messaging.svc`)
- [x] abstract the browser storage in `data-store`. Other components/services should only use persistence via this module.
  - [x] implement key value store (easier for other services to persist) 
  - [x] data encrypt/decrypt encapsulated in `data-store`
  - [x] password is required for CRUD operations
  - [x] password is never persisted to browser storage. Use JS closure to cache password in memory
- [x] decide  between storage.sync or storage.local
   - [x] sync: available across all instances of that browser that the user is logged into
   - local: to the machine the extension was installed on
- [x] when the user provides the password, then both the **Popup** and **Options** views are unlocked (not required to enter the password twice) 
- [x] it should not be the responsibility of the `connectors` to do `lock`, `unlock`, `isLocked`, but instead the `password.svc` service should take care of this
- [x] add services layer (`account.svc`, `password.svc`, `messaging.svc`)
  -  here it is where the bulk on the logic should reside and this part should have serious unit tests
- [x]  spit the `background-script` in smaller components. One big script file is hard  to understand & maintain.
   - each component should have one and only one responsibility (handling a particular event, connecting to a service, etc) 
   - [x] extract `password-handler`
   - [x] move previous logic to `old` folder and later extract components one by one
- [x] make it easy to add new connectors (minimal changes to existing components)
  - [x] UI: create Connector React component & add connector in the Account view
  - BE: add new connector
- [x] account management (add/remove/update/set current)
  - [x] add UUID to accounts (do not use name as key, avoid name colision)
- [x] account details view
- [x] Loading view: add `antd` Spinner 
- [x] decouple Options view from Configurations from Accounts
- [x] do not force `SetPassword` and `Unlock` views to be always in a Modal (can be used as standalone)
- [x] do not use `salt` for encryption (at least for now)
- [x] create architecture diagram
  - extension architecture diagram(1)
- some screenshots can be found here: https://github.com/bumi/lightning-browser-extension/issues/39


chris-aeviator:
I'll need some days to get on track with this PR

src/common/services/password.svc.js
```diff
@@ -0,0 +1,75 @@
...
+    return isPasswordSet === true;
+  } catch (err) {
+    console.log(err);
+    return false;
+  }
+}
+
+async function _checkCachedPassword() {
+  return new Promise(async (resolve) => {
+    await messagingSvc.onMessage("cached-password", async (msg) => {
```
bumi, line 60: I think this listener would never get removed and it will also receive all future messages? is that correct?

bumi: and you introduce this pattern to receive a reply:

1. add a listener 
2. send a message 
3. handler of the sent message, in return sends a message
4. listener receives the message of the handler

Do I understand this correct?

bumi (commented):
Did you look into redux? 
e.g. https://github.com/tshaddix/webext-redux

I feel like some problems that this PR addresses can be solved with the redux patterns? 
Would this help us and avoid using our own messaging pattern, that is introduced here?

bumi:
Here are a few links that talk about the redux pattern: 
https://hacks.mozilla.org/2017/06/introducing-filterbubbler-a-webextension-built-using-reactredux/ (old)
https://levelup.gitconnected.com/using-redux-in-event-driven-chrome-extensions-problem-solution-30eed1207a42

https://github.com/tshaddix/webext-redux/network/dependents
extension using webext-redux:
https://github.com/Pocket/extension-save-to-pocket
https://github.com/blockstack/stacks-wallet-web
https://github.com/tshaddix/webext-redux-examples

src/app/components/Account/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  const handleConnectorTypeChange = (type) => {
+    setAccountType(type);
+  };
+
+  const handleTestAccount = async () => {
+    try {
+      const values = connectorForm && (await connectorForm.validateFields());
+      const lndConnector = new connectors.lnd({
```
bumi, line 69: so far I did not expect the connectors to be called from the UI, but only from the background script. Not sure if that always works? 

I would see testing the connection as part of some validation. (then the user also does not need to extra "klick" test.
In Johns [UI suggestion](https://www.figma.com/file/xEPJFGFdc3HD0JS1ax2IpH/Lightning-Browser-Extension?node-id=1%3A22152) we have two steps: first the form, then the confirmation. 
what do you think about that?

src/app/components/Account/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+    fetchAccount();
+  }, [accountId]);
+
+  let connectorForm = null;
+  const submitHook = (innerForm) => {
+    connectorForm = innerForm;
+  };
+
+  const handleSubmit = async () => {
```
bumi, line 43: this method is kinda long and does a lot, maybe we should split it it? 

validate(), then save(), and save does either an update or a create?

src/app/components/Configurations/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+    await accountSvc.setCurrentAccount(accountId);
+    await load();
+  };
+
+  const removeAccount = async (accountId) => {
+    await accountSvc.remove(accountId);
+    await load();
+  };
+
+  const goToAddAccount = () => {
```
bumi, line 58: could this be a [`<Link>`](https://reactrouter.com/web/api/Link) from react router? then we would not need to have a manual function and manually do a history.push? (this is also for several cases)

src/app/components/Options/Options.jsx
```diff
@@ -1,232 +1,110 @@
...
+    await this.checkDataStoreState();
+  }
+
+  async handleUnlock() {
+    await this.checkDataStoreState();
+  }
+
+  async updateRoute() {
+    if (this.state.isInitialized === false) {
+      return this.history.replace("/init");
```
bumi, line 39: If nothing has been configured we might want to send the user to welcome.html which includes the onboarding flow.

src/app/components/Options/Options.jsx
```diff
@@ -1,232 +1,110 @@
...
+                <Route
+                  exact
+                  path="/unlock"
+                  render={() =>
+                    this.state.isUnlocked === false ? (
+                      <Modal closeable={false} visible={true} footer={null}>
+                        <Unlock onUnlock={this.handleUnlock.bind(this)} />
+                      </Modal>
+                    ) : (
+                      <Loading />
```
bumi, line 68: why does a route (/unlock) render different components?

src/app/components/SetPassword/index.jsx
```diff
@@ -1,68 +1,66 @@
...
+import messagingSvc from "../../../common/services/messaging.svc";
+
+const SetPassword = ({ onOk }) => {
   const [form] = Form.useForm();
 
-  const onDone = (values) => {
+  const onDone = async (values) => {
     if (values.password === values.confirm) {
-      onOk(values.password);
+      await passwordSvc.init(values.password, values.confirm);
```
bumi, line 12: I still struggle with the password management. 
why is there a password Service and then the messaging service still has to send some cache password message?

src/common/data-store/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Encripted storage module. A passord is required to initialize this module.
+ * The storage is under the form of key-value pairs.
+ */
+let cachedStorage = null;
+
+function init(password) {
+  if (!password) {
+    throw new Error("A passord is required to initialize the storage module!");
```
bumi, line 13: I see quite some exceptions of this in my console. Should this really be an error?

src/common/data-store/index.js
```diff
@@ -0,0 +1,95 @@
...
+      }
+    });
+  }
+
+  function set(key, data) {
+    if (!key || typeof key !== "string") {
+      throw new Error("Invalid key: " + key);
+    }
+
+    const encryptedData = encryptData({ data }, pwd);
```
bumi, line 39: as discussed in the chat. 
I think encrypting every single storage data will prevent some features and will make data management harder. 
The only sensitive data we have are the ln account credentials, but allowances or settings would not need to be encrypted I think?

src/common/data-store/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+  return {
+    set,
+    get,
+    isInitialized: async function () {
+      // TODO: recheck
+      const isPasswordSet = await browser.storage.sync.get("isPasswordSet");
+      return (
+        isPasswordSet !== undefined &&
+        JSON.stringify(isPasswordSet) !== JSON.stringify({})
```
bumi, line 54: what is this doing?

src/common/data-store/index.js
```diff
@@ -0,0 +1,95 @@
...
+      );
+    },
+    isUnlocked: async function () {
+      return true;
+    },
+  };
+}
+
+function initNullDataStore() {
```
bumi, line 63: why do we need the null data store that throws errors? 
can't we prevent accessing some kind of storage and not have a store at all if it is not decrypted? (besides that I think we need to discuss what to encrypt as mentioned above)

src/common/services/password.svc.js
```diff
@@ -0,0 +1,75 @@
+import dataStore from "../data-store";
+import messagingSvc from "./messaging.svc";
+
```
bumi, line 3: I don't really understand this service and what it is supposed to do.

src/extension/background-script/index.js
```diff
@@ -1,106 +1,5 @@
...
-  browser.runtime.onMessage.addListener(handleConnectorCalls);
-
-  browser.runtime.onInstalled.addListener(handleInstalled);
-
-  if (settings.enableLsats) {
-    await browser.storage.sync.set({ lsats: {} });
-    initLsatInterceptor(connector);
-  }
-}
+import init from "./old/init";
```
bumi, line 1: I don't think we should have something `old` in the master. either we refactor it, or it is not old.

bumi (commented):
I tried to add some more thoughts to talk about. 

is it correct that this PR mainly solves the issue to cache the password across different pages (popup, options, etc.)?

I think the main topics are: 

#### 1. Messaging
I think there is a lot of complexity here that is not easy to understand (for me). And I actually think there are some problems in there, e.g. you add a message listener every time and never remove them. Those message listeners stick around and will receive ALL messages.
I've never seen the pattern to add a handler, send a message to receive some data. Messages can be answered (e.g. with returning a promise or the sendResponse function).

I somehow think we should consider there redux to handle actions and global state. 
There was a discussion in some other PR/issue about that and we tried to avoid it. But this basically solves what this PR tries to solve with a manual custom messaging solution. 

 #### 2. Storage
Should really everything be encrypted? Or do we only encrypt sensitive account data?



--- need to come back to it later.

bumi:
that said! lot's of great stuff and good to push these discussion forward. thanks a lot for it!!!

src/common/services/password.svc.js
chris-aeviator: another thought: I'd personally appreciate removing the `svc` duplication in the filename, since we have `services` already in the path.

src/app/components/Account/index.jsx
motorina0: agree

src/app/components/Configurations/index.jsx
motorina0: agree

src/app/components/Options/Options.jsx
motorina0: agree

src/common/data-store/index.js
motorina0: agree, log&return

src/extension/background-script/index.js
motorina0: main-handler?

PR #56, opened 2021-04-27
WIP: Minimal Redux Setup

chris-aeviator:
Minimal redux example that introduces a redux store and development tooling.

After starting the development server as usual, the redux devtools are available via `http://localhost:8000` or via the redux-devtools browser extension remote button.

An inspector shows all changes to the store in realtime. e.g. the state of `options/isInitialized` changes from false to true here. This is dispatched (as an example) inside https://github.com/bumi/lightning-browser-extension/commit/28db4ee95a682dc6b136555cd0a2d64ddc4defe8#diff-575a080cf195a8bc235775b8d938e801e0fcb7df31f822c36e67222ba0de00b5L47

image

The inspectors "Tree" View features a top-down view on all store-slices (accounts, options,...). Slice is a naming convention picked by redux and represents a "slice of the store(information)"
image


src/app/store/features/accounts/accountsSlice.js
```diff
@@ -0,0 +1,29 @@
...
+function nextAccountId(accounts) {
+  const maxId = accounts.reduce((maxId, account) => Math.max(account.id, maxId), -1)
+  return maxId + 1
+}
+
+
+export default function accountsReducer(state = initialState, action) {
+  switch (action.type) {
+    case 'accounts/addAccount': {
+      // Can return just the new accounts array - no extra object around it
```
bumi, line 16: would I implement certain business logic here, too? or is this only for data storage?

bumi:
do you have an example on how an action(?) would look like that does not change the state? 

e.g. testing a connection?

src/app/store/features/accounts/accountsSlice.js
chris-aeviator: https://medium.com/@jeffbski/where-do-i-put-my-business-logic-in-a-react-redux-application-9253ef91ce1

chris-aeviator:
I believe testing the connection would save the connections last known state (e.g. functional)

bumi:
actually any call to to a connector would be something like that. 
e.g. getting the balance from the node which is displayed in the popup.

src/app/store/features/accounts/accountsSlice.js
bumi: which option is your favorit? joule used sagas.

chris-aeviator:
> ctually any call to to a connector would be something like that.
e.g. getting the balance from the node which is displayed in the popup.

ok I will check for this

src/app/store/features/accounts/accountsSlice.js
chris-aeviator: without much refactoring we'd be at "Dispatching from action creators", meaning we'd call the store to inform it about new data after we've invoked a connector call for example. I'd very much vote for this approach as much as possible in  a single PR - we'd be able to refactor introduce new stuff with PR's further down the road.

PR #60, opened 2021-05-26
Remotes/origin/add tailwind and tailwind UI

ShawkiS:


package.json
```diff
@@ -15,18 +15,21 @@
...
   "scripts": {
     "dev:chrome": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=chrome webpack --watch",
+    "build:style": "postcss src/styles/tailwind.css -o src/styles/index.css",
     "dev:firefox": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=firefox webpack --watch",
     "dev:opera": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=opera webpack --watch",
     "build:chrome": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=chrome webpack",
     "build:firefox": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=firefox webpack",
     "build:opera": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=opera webpack",
-    "build": "yarn run build:chrome && yarn run build:firefox && yarn run build:opera",
+    "build": "craco build && npm run build:style && yarn run build:chrome && yarn run build:firefox && yarn run build:opera",
```
bumi, line 24: why do we need that?
can this be part of webpack? 
and if we need it we probably also need it in the other `build` tasks?

src/styles/tailwind.css
```diff
@@ -0,0 +1,5 @@
+@tailwind base;
```
bumi, line 1: I think `styles` should go under `app/styles` as the `app` holds the user facing application

package.json
```diff
@@ -84,6 +99,7 @@
     "copy-webpack-plugin": "^8.1.1",
     "cross-env": "^7.0.3",
     "css-loader": "^5.2.0",
+    "dotenv-webpack": "^7.0.2",
```
bumi, line 102: this is new, isn't it? where do we use env/.env configurations?

bumi (commented):
nice! super cool. 
I really like storybook I have to say :)

package.json
ShawkiS: I made a change for it. It generates the tailwind css file, we don't need it in every build, just once or when adding new things

src/styles/tailwind.css
ShawkiS: right. just moved it.

package.json
ShawkiS: hmmmmm yeah it seems came by a mistake

package.json
bumi: ah ok, thanks.

and the `craco build` ? is this also only needed for the styles? (if so, can we add it to the `build:style` command?)

package.json
```diff
@@ -15,18 +15,21 @@
...
   "scripts": {
     "dev:chrome": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=chrome webpack --watch",
+    "build:style": "tailwind build src/app/styles/tailwind.css -o src/app/styles/index.css",
     "dev:firefox": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=firefox webpack --watch",
     "dev:opera": "cross-env NODE_ENV=development cross-env TARGET_BROWSER=opera webpack --watch",
     "build:chrome": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=chrome webpack",
     "build:firefox": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=firefox webpack",
     "build:opera": "cross-env NODE_ENV=production cross-env TARGET_BROWSER=opera webpack",
-    "build": "yarn run build:chrome && yarn run build:firefox && yarn run build:opera",
+    "build": "craco build && npm run build:style && yarn run build:chrome && yarn run build:firefox && yarn run build:opera",
```
bumi, line 24:
```suggestion
    "build": "yarn run build:style && yarn run build:chrome && yarn run build:firefox && yarn run build:opera",
```

craco.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
bumi, line 1: this file is not needed, correct?

ShawkiS: yes

package.json
```diff
@@ -41,23 +44,30 @@
   "dependencies": {
     "@ant-design/icons": "^4.6.2",
     "@babel/runtime": "^7.13.10",
+    "@craco/craco": "^6.1.2",
```
bumi, line 47: I missed this before. Are we using craco? 

can you review that only required dependencies are added?

bumi:
thanks for the PR! 

btw. we use yarn. so we should not commit the package-lock.json file afaik .

PR #61, opened 2021-06-01
Design onboard pages

ShawkiS:


ShawkiS:
Storybook (5)
Storybook (4)
Storybook (3)
Storybook (2)
Storybook (1)

bumi (commented):
we should not commit the package-lock.json. best use yarn instead of npm. 
and maybe we should add the package-lock to gitignore then. afik we should either commit the yarn.lock or package-lock

src/app/components/Onboard/Shared/template.jsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import Steps  from "../Shared/steps";
+
+const steps = [
```
namxam, line 4:
```suggestion
const steps = [
```

src/app/utils/index.js
```diff
@@ -0,0 +1,3 @@
+export function classNames(...classes) {
```
namxam, line 1: There is a pretty awesome `classnames` package which offers a few more use cases. Perhaps it would make sense to use it instead https://www.npmjs.com/package/classnames

src/app/components/Shared/button.jsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+export default function Button({ label }) {
```
namxam, line 3: Why nest it in a `Shared` folder. From my understanding `src/app/components` is for shared components. So no need to nest it even deeper.

src/app/components/Shared/input.jsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+export default function Input({name, id, placeholder}) {
```
namxam, line 3: Same as above… Why nest it in a Shared folder. From my understanding src/app/components is for shared components. So no need to nest it even deeper.

src/app/components/Shared/input.jsx
```diff
@@ -0,0 +1,13 @@
...
+  return (
+    <input
+      type="text"
+      name={name}
+      id={id}
+      className="shadow-sm focus:ring-orange-bitcoin h-14 focus:border-orange-bitcoin block w-full sm:text-sm border-gray-300 rounded-md"
+      placeholder={placeholder}
+    />
+  );
+}
```
namxam, line 13: I would only override the `className` property. Or even better extend it…

```suggestion
export default function Input({className, ...props}) {
  return (
    <input
      {...props} 
      className={`shadow-sm focus:ring-orange-bitcoin h-14 focus:border-orange-bitcoin block w-full sm:text-sm border-gray-300 rounded-md ${className}`}
    />
  );
}
```

src/app/components/Onboard/Welcome/features.jsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+export default function Features(props) {
```
namxam, line 3: Destructure props so other devs know what props to give it. In this case it would be really helpful to add proptypes, add a comment, or somehow document which properties the elements of the feature array should/could have.

src/app/components/Onboard/LastStep/index.jsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import Template from "../Shared/template";
+import Button from "../../Shared/button";
+
+export default function lastStep() {
```
namxam, line 5: Should be `LastStep`

src/app/components/Onboard/Welcome/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+          </h2>
+        </div>
+        <Features features={features} />
+      </div>
+      <div className="sm:py-16 sm:px-6 lg:px-8">
+        <div className="max-w-3xl mx-auto divide-y-2 divide-gray-200 mr-0">
+          <FAQs faqs={faqs} />
+        </div>
+      </div>
+      <button
```
namxam, line 78: Is there a reason why you do not use the button component?

src/app/components/Onboard/Welcome/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+import Features from "./features";
+import {
+  GlobeAltIcon,
+  LightningBoltIcon,
+  MailIcon,
+  ArrowRightIcon,
+  ScaleIcon,
+} from "@heroicons/react/outline";
+
+const features = [
```
namxam, line 13: I would inline `features` and `faqs` variables into component definition. But not sure why. I guess as it is content and therefore belongs in the body :)

namxam (changes requested):
I am not sure, but I have seen a lot of `text-base` classes. From what I recall, those are not necessary in most cases. But I might be wrong.

src/app/components/Onboard/Shared/template.jsx
```diff
@@ -0,0 +1,19 @@
...
+  { id: "Step 1", name: "Job details", href: "#", status: "complete" },
+  { id: "Step 2", name: "Application form", href: "#", status: "current" },
+  { id: "Step 3", name: "Preview", href: "#", status: "upcoming" },
+];
```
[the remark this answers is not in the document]
johnsBeharry: I think the stepper makes sense here. the designs were made in haste during a hackathon so such details like a progress indicator was missed.

src/app/components/Onboard/Shared/steps.jsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+export default function Steps(props) {
```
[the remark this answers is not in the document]
johnsBeharry: Should style be used if we're using tailwind classes? From what I can grok that could make adding padding difficult if classname is passed down unless some logic is written to check for the existing spacing classes on the component.

Instead it may work better to pass down some defined props like "size", "color", "spacing" etc to build the list of classes. Overall though I think this level of customisation of a component with classname is not worth it, static component styes, a flexible grid system and I think that should be it.

src/app/components/Onboard/Shared/card.jsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export default function Card({ topLabel, lastLabel, heading }) {
```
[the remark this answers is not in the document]
johnsBeharry: Hey @namxam so this component is from the [wallet-ui-kit](https://www.figma.com/community/file/916680391812923706) and while it may only be used in this application once (so far) since the designs are based on the wallet-ui-kit @bumi and I spoke about being able to extract them out so other projects can use them.

src/app/components/Onboard/Welcome/features.jsx
johnsBeharry: wish we used typescript. but i'm not going to push it at this point 😅

src/app/components/Onboard/Welcome/features.jsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <dl className="mt-10 space-y-10 sm:space-y-0 sm:grid sm:grid-cols-2 sm:gap-x-8 sm:gap-y-10 lg:mt-0 lg:col-span-2">
+      {props.features.map((feature) => (
+        <div key={feature.name}>
```
johnsBeharry, line 7: like @namxam said dont think its valid html to have a div here `dl > div > dt`

src/app/components/Onboard/Welcome/index.jsx
johnsBeharry: I'd disagree on this one, more smaller components FTW. Esp, when using storybook which has been install on the project but not utilised in this PR.

src/stories/Button.stories.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import Button from "../app/components/Shared/button";
+import '../app/styles/index.css';
```
dylancom, line 3: Instead of importing the styles into every story, you can use .storybook/preview.js for global code that applies to all stories. See: https://storybook.js.org/docs/riot/configure/overview#configure-story-rendering
So it can be imported only once.

dylancom:
Every onboarding step now has it's own story file. I think it's cleaner to have one Onboarding.stories.js that bundles them.

This makes the code easier to manage and removes the need to expand every "Primary" story.
Schermafbeelding 2021-06-15 om 12 02 55

.storybook/preview.js
```diff
@@ -1,3 +1,4 @@
+import '../app/styles/index.css';
```
dylancom, line 1: The path should be '../src/app/styles/index.css';

src/stories/Button.stories.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
```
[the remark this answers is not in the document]
bumi: 👍 this is also what @dylancom did in #62

dylancom:
Maybe too much details but the loading of the top image (ConnectLND, LastStep, SetPassword), causes "layout shifting" while it is being loaded. It seems this image always has the same width / height so it could be wrapped in a fixed height div for e.g.?

bumi:
I am merging this as it grows big and becomes blocking.
we will address the issues in separate PRs. 

thanks @ShawkiS <3 🎉

bumi (approved):

PR #62, opened 2021-06-14
Add some Storybook stories

dylancom:


src/app/components/LndHub/LndHub.stories.jsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+import LndHub from ".";
```
bumi, line 3: what's actually a good convention here: As this is a form should this be called `LndHubForm`

bumi (approved):
guess this is mergable?

src/app/components/LndHub/LndHub.stories.jsx
dylancom: That would be a good convention. And maybe bundle it in a Forms folder for e.g. if there are many.

dylancom:
yes it's mergable, just some small stories.

PR #63, opened 2021-06-15
Add global state management

bumi:
This uses `zustand` to have a global state object in the background script that holds all data and also can sync the data to the extension's browser storage.
Frontend code should use the extension messaging to interact with the background script where all the state logic happens.

All logic is implemented in "actions". (which addresses the problem that too much unrelated logic was in the connectors) 
The connectors now only handle any lightning related functionality.

The global state object solves the problem that the user has to re-enter the password on every page. 

This is basically the idea: 
state


chris-aeviator:
what's your take on why you chose zustand over the  proposed https://github.com/bumi/lightning-browser-extension/pull/56 redux?

bumi:
@chris-aeviator #56 adds redux for the frontend components and does not yet communicate with the background. 
I struggled a lot with redux and then also to somehow get it work to keep the state in the background script and also store parts of it in the extension's browser storage.

In this PR I experiment to have a persistent state object in the background script and the frontend sends a message to the background script where the business logic is implemented to change and read that state. 

I used zustand because it gives me a state object and I can subscribe to changes (e.g. whenever the settings are changed those are also persisted to the browser storage). 
The state is not intended to be used in react components.

chris-aeviator:
> I used zustand because it gives me a state object and I can subscribe to changes (e.g. whenever the settings are changed those are also persisted to the browser storage

I think what would work with redux is to load the store in the background script and in the frontend and use `mapStateToProps` for state and `mapDispatchToProps` to call actions to subscribe/ plug everything together

https://react-redux.js.org/using-react-redux/connect-mapstate

bumi:
feel free to try. though I can not imagine how that could work because the background script and the UI pages are completely separate and do not share anything. So I think either communication needs to be done manually or through that webextension redux libraries (which wrap the extension messaging APIs somehow) 

But I feel like redux and all that add so much complexity and has a quite steep learning curve. 

What I like here is that it simply uses the default extension messaging API and a simple state object in the background and nothing more.

(do you know any simpler "observable object" library?)

chris-aeviator:
I'm not particularly interested in pushing redux vs. something else (though redux works very good and has an excellent community around it)… more into reducing the sources of truth ideally down to a single one.

bumi:
we must not forget this is not a typical frontend application. thus I am not sure how much redux alone helps in the end. 
UI <-> Background Script is the question. the logic and storage should be in the background script.

bumi:
This PR also does not touch the UI, it only introduces a state object in the background script which can be used from different business logic actions.

chris-aeviator:
Maybe, even though counterintuitive, the UI could ask (poll) the backend regularly for it's state and we still maintain a central state with redux?!

bumi:
yep, any UI page/component can handle its own state... the background script is then more a backend API. 

I don't think it will happen that the state changes in the background script without UI interaction... if it does we can deal with it then.

typically it would be: 
1. user adds an account
2. backend is called to add the account
3. update the UI (either by getting all accounts again from the backend or by simply using local "state")

chris-aeviator:
Ok will review and approve accordingly

src/app/components/Welcome/Welcome.jsx
```diff
@@ -44,11 +44,20 @@ class Welcome extends React.Component {
...
+        .then((unlocked) => {
+          console.log(unlocked);
+          return utils.call("accounts.add", account).then((added) => {
+            console.log(added);
+            return utils
+              .call("accounts.select", { accountId: added.accountId })
+              .then((selected) => {
+                console.log(selected);
+                alert("Test account is saved. Your password is: btc");
+              });
```
bumi, line 58: The UI always calls the backend script. So here first we `unlock` the extension (set the password) then we add an account (the password from the state is used to encrypt the config) and then we select the newly created account.

src/common/lib/crypto.js
```diff
@@ -1,10 +1,10 @@
 import { AES, enc } from "crypto-js";
 
 export function encryptData(data, password, salt) {
-  return AES.encrypt(JSON.stringify(data), password + salt).toString();
+  return AES.encrypt(JSON.stringify(data), password).toString();
```
fregante, line 4: No salt?

src/extension/background-script/index.js
```diff
@@ -1,85 +1,40 @@
...
 };
 
-// listen to calls from the content script and pass it on to the native application
+// listen to calls from the content script and calls the actions through the router
 // returns a promise to be handled in the content script
-const handleConnectorCalls = (message, sender, sendResponse) => {
+const routeCalls = (message, sender, sendResponse) => {
   // if the application does not match or if it is not a prompt we ignore the call
   if (message.application !== "Joule" || !message.prompt) {
     return Promise.resolve();
```
fregante, line 27: This does not ignore the message, it handles it. To ignore the call return undefined
```suggestion
    return;
```

If there are other `onMessage` listeners (in here or in external modules) they won't receive the message.

Another comment in the code seems to suggest you do it for security, but I might have misunderstood.

src/extension/background-script/index.js
```diff
@@ -1,85 +1,40 @@
...
   if (details.reason === "install") {
     utils.openUrl("welcome.html");
   }
 };
 
-// listen to calls from the content script and pass it on to the native application
+// listen to calls from the content script and calls the actions through the router
 // returns a promise to be handled in the content script
-const handleConnectorCalls = (message, sender, sendResponse) => {
+const routeCalls = (message, sender, sendResponse) => {
```
fregante, line 24: Unused/unnecessary `sendResponse`. This was the old, callback-based way to return a message. Now the promise does it already.

```suggestion
const routeCalls = (message, sender) => {
```

src/extension/background-script/state.js
```diff
@@ -0,0 +1,62 @@
...
+  connector: null,
+  account: null,
+  settings: {},
+  accounts: {},
+  allowances: [],
+  currentAccountId: null,
+  password: null,
+  lock: () => set({ password: null, connector: null, account: null }),
+  init: () => {
+    return browser.storage.sync
```
fregante, line 23: Sync storage can be updated from other instances of a browser so you might want to use `browser.storage.onChanged` to avoid getting the state out of sync. 

A bit rare, but this might happen if, say, I leave the browser always open at home, change something to my extension at work, and come back to it at night.

src/extension/background-script/state.js
```diff
@@ -0,0 +1,62 @@
...
+      .get(Object.keys(browserStorage))
+      .then((result) => {
+        const data = {};
+        Object.keys(browserStorage).forEach((key) => {
+          data[key] = result[key] || browserStorage[key];
+        });
+        set(data);
+      });
+  },
+  flush: () => {
```
fregante, line 33: What is this and what does it do?

fregante (commented):
These changes rely on persistent background pages, but it seems that they're persistent only in Chrome and Opera: https://github.com/bumi/lightning-browser-extension/blob/261a2510f64c6facde4109f1df915bfc5223dda2/src/manifest.json#L67

I think it should be `true` unconditionally.

src/extension/background-script/state.js
bumi: currently it is not used. was experimenting with how to actively write the object to the browser storage

src/extension/background-script/state.js
bumi: very true. should add an onchanged handler to update the state in memory. 👍

src/common/lib/crypto.js
bumi: it is too complicated to calculate a salt per installation for now. We just require a strong password.

bumi:
This PR gets too big. I am merging this, did not hear any big objections and we can change things anyway again. :)

PR #64, opened 2021-06-15
Wallet screen and overall improvements

dylancom:
PR includes:
- Wallet screen front end styling
- Improved Tailwind integration, compile Tailwind in both the app / storybook and remove unused css on production.
- Fix unlocking when pressing enter.
- Inter font (hack needed until antd lib get's fully replaced by Tailwind).
- Some folder restructuring, moved several components to "screens". Screens can then be composed by reusable components.


src/app/screens/Home/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+        </div>
+        <div className="p-5">
+          <h2 className="text-xl">Transactions</h2>
+          {this.state.loadingTransactions ? (
+            <div className="pt-4 flex justify-center">
+              <Loading />
+            </div>
+          ) : (
+            <Transactions
+              exchangeRate={this.exchangeRate}
```
bumi, line 89: later we will have the exchange rate on the individual transaction level. (the rate at the time of the transaction)

src/app/screens/Home/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+            <div className="text-xs text-gray-500">{alias || "\u00A0"}</div>
+            <div className="text-xs">₿0.0016 7930 €33.57</div>
+          </div>
+          <button
+            className="ml-auto"
+            onClick={() => {
+              return utils.openPage("options.html");
+            }}
+          >
+            <CogIcon className="h-6 w-6 text-gray-400" aria-hidden="true" />
```
bumi, line 71: wondering if the header should be a component that we can also use on the options.html and other pages?

bumi (approved):
great! looks good to me. 
We need to fill some of the dummy text with actual correct data, but this can be added separately.  

feel free to merge

static/views/lsat.html
```diff
@@ -4,6 +4,7 @@
 <head>
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=500" />
+  <link href="https://fonts.googleapis.com/css?family=Inter" rel="stylesheet">
```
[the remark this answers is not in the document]
dylancom: Got it, I replaced it by a local implementation.

src/app/screens/Home/index.jsx
dylancom: Yes, it definately should be.

PR #65, opened 2021-06-16
feature: add bitcoin design icons

dylancom:


bumi:
does this then include the whole icon set or also only the imported ones?

dylancom:
Only the imported ones (wallet / gear), the rest is ignored.

bumi (approved):

PR #66, opened 2021-06-16
UI Elements

ShawkiS:


dylancom:
When trying to run storybook I'm getting the error: `Module not found: Error: Can't resolve '../app/components/Wallet/Transactions/index' in '/Users/dylancompanjen/Apps/lightning-browser-extension/src/stories'`

dylancom:
I think we don't need the "Shared" subfolder. App (in master) is now divided into components, router and screens. Screens can then be composed by components (which are always re-usable/shared).

ShawkiS:
> I think we don't need the "Shared" subfolder. App (in master) is now divided into components, router and screens. Screens can then be composed by components (which are always re-usable/shared elements).

I think there were be a lot of shared components. the structure will be be cleaner if developers just navigate through folders

ShawkiS:
Note: some commits are not responsive yet. I'll be fixing them soon

ShawkiS:
*** Categories table

bumi:
I am actually also not very sure about a "shared" folder with the recent changed structure which has introduced also some kind of hierarchy. 

typically I am worried with folders like "shared" or "common" or "lib" that those become a "dump" for anything :) 

If you think there will be a lot of shared components that are used by other components and that the folders get to bloated we can move it then?

bumi:
for example the publisherstable I don't see as a shared component. the publishers will be a Screen/ a full page.

dylancom:
Could you update your branch with the latest master? It seems there will be now some duplicate components. For NavBar there already is AppBar for e.g. that could be extended with functionality.

dylancom:
The classNames function can be util function, so it doesn't have to be recreated for every component.

src/app/components/Shared/progressbar.jsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+export default function Progressbar({
```
dylancom, line 3: For the current styling I think we can just use the "native" progress element? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

ShawkiS: Oh I didn't know there is one. cool!

ShawkiS:
@bumi I thought we may use the publishers table in an extension screen so I though it's better to be in a shared folder

bumi:
there seems to be a syntax error? 

```
8:48:52 PM: ERR! => Failed to build the preview
8:48:56 PM: ERR! Module build failed (from ./node_modules/babel-loader/lib/index.js):
8:48:56 PM: ERR! SyntaxError: /opt/build/repo/src/app/components/Wallet/Transactions/index.jsx: Unexpected token (8:2)
8:48:56 PM: ERR!
8:48:56 PM: ERR!    6 |
8:48:56 PM: ERR!    7 |   return (
8:48:56 PM: ERR! >  8 |   );
8:48:56 PM: ERR!      |   ^
8:48:56 PM: ERR!    9 | }
8:48:56 PM: ERR!   10 |
8:48:56 PM: ERR!     at
``` 
(from netlify)

bumi:
merging this PR, then the code can be used and integrated
Thanks a lot!! 🚀

PR #67, opened 2021-06-17
implement onboarding screens

dylancom:
- Moved entry point components (Routers) into /route folder.
- Integrated the onboarding screens into the welcome page.
- Dynamically generate routes based on a given routes array (allows for easily re-ordering the steps etc.)
- The progress steps bar in the top updates itself based on the active route.
- Fixed layout "jumping/shifting" while onboarding screens are loaded. 
- As an example, the screen will create a new user after adding a password.


bumi (approved):
:shipit: 🚀 
looks great. I need to hook it up with the logic. probably as started in #63

PR #68, opened 2021-06-19
TypeScript Progressive

dylancom:
This PR allows us to use TypeScript when desired.


bumi:
that basically means to use TypeScript just use it?

dylancom:
Yes, file extension needs to change to .ts(x) though if you want to use it.

PR #71, opened 2021-06-22
catch onboard errors and faster css bundling 

dylancom:
This PR catches errors in the onboard flow, which prevent's crashes.
- Also fixes a few webpack warnings / faster css bundling.
- Removes some unused imports.


bumi (approved):
what errors did you encounter?

dylancom:
If an async back-end call would fail, it will be catched and logged inside a catch block. Letting the rest of the code execute fine.

PR #82, opened 2021-06-27
[WIP] Implement new send payment action and use pub/sub pattern for events

bumi:
This adds the pub/sub pattern for payment events to decouple functionality like the user notifications from the actual payment. 

Actions publish events (e.g. `ln.sendPayment.success`) which can be consumed by `event subscribers`.

TODO:
- [x] run PubSub subscriptions in the init function - and not automatically when loading the file


bumi:
still WIP. but merging to integrate further with frontend

PR #84, opened 2021-06-28
feat: payment summary component, enable design refactoring

dylancom:
I kept the "Save Budget" part in the ConfirmPayment screen. Not sure if it needs to be removed.


bumi:
Yeah we need a new screen for setting the allowance budget.

PR #87, opened 2021-06-28
Load real allowance data

bumi:
ToDo:

- [ ] remove search bar?
- [ ] show more data
- [ ] link to detail page
- [ ] rename `percentag` to `percentageRemaining`
- [ ] relevant badges


src/app/components/PublishersTable.jsx
```diff
@@ -1,8 +1,16 @@
...
 
 export default function PublishersTable({ publishers }) {
+  const deletePublisher = (id) => {
+    utils.call("deleteAllowance", { id });
```
bumi, line 10: @dylancom this is the delete call. how do we best handle the refresh of the publishers/or the removal of that specific publisher?

src/app/components/PublishersTable.jsx
```diff
@@ -29,30 +37,33 @@ export default function PublishersTable({ publishers }) {
...
-                    <div
-                      style={{ width: `${publisher.percentage}%` }}
-                      className="ml-auto h-2 shadow-none rounded flex flex-col text-center whitespace-nowrap text-white justify-center bg-blue-500"
-                    />
-                    <span
-                      className={`ml-2 text-sm font-semibold text-blue-500`}
-                    >
-                      {publisher.percentage}%
-                    </span>
+              <span onClick={deletePublisher(publisher.id)}>delete</span>
```
dylancom, line 46: Should be () => deletePublisher(publisher.id). Now the function is invoked directly.

src/app/components/PublishersTable.jsx
dylancom: If you move the deletePublisher function to the Publishers screen, you can directly update the state from there.
Inside the deletePublisher function you can add: setData(data.filter(publisher => publisher.id !== id)), which will refresh the publishers table.
From Publishers.tsx you need to pass down deletePublisher as a prop to PublishersTable.

This also makes sure that your component stays "dumb" and "smart" things are handled within your screens (smart components). (Clean code)

src/app/components/PublishersTable.jsx
```diff
@@ -1,8 +1,16 @@
...
 
 export default function PublishersTable({ publishers }) {
```
dylancom, line 8: Add extra prop: deletePublisher (or handleDelete for e.g. how you prefer).

src/app/components/PublishersTable.jsx
bumi: 👍 yep. thanks

bumi:
mergeing this to be able to continue with #89

PR #92, opened 2021-06-30
Load real wallet data for navbar

bumi:
- Do not show options icon in navbar if no options handler is provided
- Load account info and details into the navbar
- Remember the enable option by default


src/app/screens/Enable/index.jsx
```diff
@@ -16,7 +16,7 @@ function Enable(props) {
     setEnabled(true);
     msg.reply({
       enabled: true,
-      remember,
+      remember: true, // remember by defaul
```
dylancom, line 19: I think it's better to set it to true in const [remember, setRemember] = useState(false).
Now it will always be overridden in case you would use "setRemember" somewhere.

PR #93, opened 2021-06-30
feat: show contextual website data when making a payment

dylancom:


bumi:
nice!
let's remove `amountAlt="$28.55"` for now and we will add that again when work on fiat conversions

PR #96, opened 2021-07-01
Contextual home

dylancom:
When you visit for e.g. https://lightning-roulette.com/ and there is an allowance already set, the extension will show the publisher/payments when opening the extension.

If not, it will show the default view.

### DIFF
--- a/src/app/screens/Home/index.jsx
+++ b/src/app/screens/Home/index.jsx
@@ -35,7 +35,6 @@ class Home extends React.Component {
       .then((tabs) => {
         const url = new URL(tabs[0].url);
         utils.call("getAllowance", { host: url.host }).then((result) => {
-          console.log(result);
           if (result.enabled) {
             this.setState({ allowance: result });
           }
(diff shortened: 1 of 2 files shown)


bumi:
Whoop whoop! super awesome!

src/app/screens/Home/index.jsx
```diff
@@ -27,6 +30,17 @@ class Home extends React.Component {
   }
 
   componentDidMount() {
+    browser.tabs
+      .query({ active: true, lastFocusedWindow: true })
```
bumi, line 34: why not `currentWindow` do you know what's the difference?

dylancom: Good catch! Actually haven't tested with currentWindow instead.

dylancom: Pushed a change.
Should be currentWindow: https://developer.chrome.com/docs/extensions/reference/tabs/#get-the-current-tab

PR #106, opened 2021-07-06
Add external links to publishers and transaction pages

bumi:


src/app/components/PublisherCard/index.tsx
```diff
@@ -22,7 +23,15 @@ export default function PublisherCard({ title, image }: Props) {
         src={image}
         alt=""
       />
-      <h2 className="text-center mb-0 text-xl text-white">{title}</h2>
+      <h2 className="text-center mb-0 text-xl text-white">
```
dylancom, line 26: The icon isn't vertically centered aligned.

If you add the classes "flex justify-center items-center" to <h2>, "inline-flex" to <a> and "inline-flex" to <ExternalLinkIcon> it is centered.

bumi: thanks!

PR #108, opened 2021-07-07
Improved tables design, removed duplicate components

dylancom:


bumi:
nice! I think we should make the entries a bit smaller/reduce the height of the transaction line.
We hope there will be many  :)  (at some point we might want to group those)

dylancom:
Fixed merge conflicts and reduced the size :)

PR #110, opened 2021-07-07
Set extension width

dylancom:


bumi:
actually this causes problems with the transactions in the popup. The table there is too wide.
but I also feel like we need to show less information in the popup and make the line even smaller. what are your thoughts?

PR #135, opened 2021-07-13
Highlight publishers nav link on sub page

dylancom:


bumi:
👍 
I think we need to update the redirect here: https://github.com/bumi/lightning-browser-extension/blob/24a36df2ee88b7d7fd29e8afc751eb6fee1e9ea5/src/app/screens/Publisher.tsx#L80

dylancom:
@bumi good catch. Should be replace instead of push btw as the previous route doesn't exist anymore after deletion. (Otherwise using the native browser back button would be broken)

bumi:
ah I see. thanks!

PR #136, opened 2021-07-13
Remove publisher from details page

dylancom:
I created an "accessible" `<Menu />` component (which can be re-used around the app).
On the publisher page it will represent a hamburger button than can open a menu to "edit" and "delete" the publisher.


bumi:
this is super cool

PR #138, opened 2021-07-14
Lock account from options screen

dylancom:


bumi:
👍 I think we should call it "lock" or something like that. the user is not really logged in.

bumi:
was wondering if we need a lock button also on the popup.... what do you think @johnsBeharry? 
Should the user often lock the extension wallet? or is the default to keep it unlocked?

dylancom:
Might have a "user" button on the top right that can open a menu to "lock account", "edit account" and such? (Just like here on GitHub etc.)

bumi:
actually I think that’s a good idea.
and the link to the options page could be a “options/settings” link in that menu.

dylancom:
I created a `<UserMenu />` component which now contains the buttons "Options" and "Lock".

bumi:
that's great and can be now easily extended. 👍

PR #141, opened 2021-07-21
[WIP] LNURL pay

dylancom:
Can be tested with: https://lnurl.fiatjaf.com/

TODO:

- [ ] Additionally, a payment dialog must include:
- Domain name extracted from LNURL query string.
- A way to view the metadata sent of text/plain format.
- A text input where user can enter a comment string (max character count is equal or less than commentAllowed value)


bumi:
wondering if we should just call it `lnurl` and have all the lnurl features and options implemented there. (later also lnurl-auth #7 )

dylancom:
@bumi makes sense!

src/extension/inpage-script/index.js
```diff
@@ -1,8 +1,10 @@
 import WebLNProvider from "../ln/webln";
+import LNURLProvider from "../ln/lnurl";
 import Donation from "./donation";
 
 if (document) {
   window.webln = new WebLNProvider();
+  window.lnurl = new LNURLProvider();
```
bumi, line 7: actually I think we should stick with WebLN here and propose a new `lnul` function on the webln object. 

so if it is a `lnurl` link we call the `webln.lnurl()` function which sends the message to lnurl action. There we decode the lnurl, do the GET request and then continue depending on the `type` of the lnurl (pay, auth, etc.)

what do you think?

dylancom: The lnurl action sounds good to me. (Moving the decoding and GET away from React) I'm only not sure about if lnurl should be part of webln as they are different directions?

bumi: yes, that needs to be discussed. Though I am not sure if it is different directions. WebLN describes a way how to interact with lightning from the web/JS. 
Instead of passing a lightning invoice we are passing in a lnurl. 

(somewhat related discussion around extending webln with keysend: https://github.com/joule-labs/webln/issues/33)

src/app/screens/LNURLPay.tsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+function LNURLPay({ details, origin }: Props) {
+  const [value, setValue] = useState<string | number>(details.minSendable);
+
+  async function confirm() {
+    try {
+      // Get the invoice
+      const params = {
+        amount: value, // user specified sum in MilliSatoshi
```
bumi, line 31: we have to be careful here currently we use satoshis but we get MilliSatoshi here.

src/extension/background-script/actions/lnurl/index.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    try {
+      const response = await connector.sendPayment({
+        paymentRequest,
+      });
+      publishPaymentNotification(message, paymentRequestDetails, response);
+
+      // Once payment is fulfilled LN WALLET executes a non-null successAction
+      // LN WALLET should also store successAction data on the transaction record
+      if (successAction) {
```
bumi, line 69: wondering if we should handle the successAction in the component? 

displaying the URL (with link to open) and the decrypted message could be easier in the component? or we have to open another prompt?

dylancom: @bumi Yes indeed, was already planning to move this to the component. That way we can easily show custom modals etc.

src/app/screens/LNURLPay.tsx
```diff
@@ -50,32 +50,29 @@ function LNURLPay({ details, origin }: Props) {
...
       // LN WALLET pays the invoice, no additional user confirmation is required at this point
-      return await msg.reply({
-        confirmed: true,
+      return await utils.call("lnurlPay", {
+        message: { origin },
         paymentRequest,
-        successCallback,
+        // Once payment is fulfilled LN WALLET executes a non-null successAction
+        // LN WALLET should also store successAction data on the transaction record
+        successCallback: successAction
```
bumi, line 59: interesting, didn't even think of passing a function through those calls. good that it works.

src/app/screens/LNURLPay.tsx
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  return (
+    <div>
+      <PublisherCard title={origin.name} image={origin.icon} />
+      <div className="p-6">
+        <dl className="shadow p-4 rounded-lg mb-8">
+          <dt className="font-semibold text-gray-500">Send payment to</dt>
+          <dd className="mb-6">{origin.name}</dd>
```
bumi, line 126: here we should also display `details.url.host`

PR #143, opened 2021-07-25
Drop node-lightning/invoice package

dylancom:
https://github.com/bumi/lightning-browser-extension/issues/142

### DIFF
--- a/src/app/screens/ConfirmPayment/index.jsx
+++ b/src/app/screens/ConfirmPayment/index.jsx
@@ -14,7 +14,7 @@ class ConfirmPayment extends React.Component {
     super(props);
     this.history = createHashHistory();
     this.state = {
-      budget: (this.props.invoice?.valueSat || 0) * 10,
+      budget: this.props.invoice?.mtokens || 0,
       rememberMe: false,
     };
   }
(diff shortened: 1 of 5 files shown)


src/app/screens/ConfirmPayment/index.jsx
```diff
@@ -58,8 +58,8 @@ class ConfirmPayment extends React.Component {
         <div className="p-6">
           <div className="mb-8">
             <PaymentSummary
-              amount={this.props.invoice?.valueSat}
-              description={this.props.invoice?.desc}
+              amount={this.props.invoice?.mtokens}
```
bumi, line 61:
```suggestion
              amount={this.props.invoice?.tokens}
```

I think this should be tokens. (not m(illi)tokens)

lightning supports sub-satoshi amounts and some amouts are represented as milli-satoshis. 

Currently we have only been using satoshi. But we should probably store the smallest possible amount and thus always use milli-satoshi and convert to sats (or other denominations) in the UI?

bumi (changes requested):

src/app/screens/ConfirmPayment/index.jsx
```diff
@@ -14,7 +14,7 @@ class ConfirmPayment extends React.Component {
     super(props);
     this.history = createHashHistory();
     this.state = {
-      budget: (this.props.invoice?.valueSat || 0) * 10,
+      budget: this.props.invoice?.mtokens || 0,
```
bumi, line 17:
```suggestion
      budget: (this.props.invoice?.tokens || 0) * 10,
```
was it intentional to use mtokens here? when I initially wrote the code I just thought a budget of 10 times the invoice amount sounds good...no big reasoning behind.

dylancom: Nope, should me tokens in this case indeed!

PR #157, opened 2021-08-05
Review onboarding

bumi:
### TODO:

Allow users to select their connectors (LND, LndHub, Lnbits) and render forms accordingly


bumi:
@dylancom I think we need to remove the links on the steps. It is for now too complicated to support manually back/forth in the wizzard. (which would for example mean we need to update an account instead of adding)

dylancom:
@bumi I agree, will update the steps to work without links.

dylancom:
See: https://github.com/bumi/lightning-browser-extension/pull/160

PR #169, opened 2021-08-18
Add macaroon by dragging / dropping a file

dylancom:


src/common/lib/utils.js
```diff
@@ -45,6 +45,11 @@ const utils = {
     }
     return hex.join("");
   },
+  buf2hex: (buffer) => {
```
bumi, line 48: I think you can use the `bytesToHexString` which we have a few lines below?

dylancom: Correct, if I turn the buffer into a Uint8Array.

bumi: ah thought it was using the same, didn't look close enough.

PR #171, opened 2021-08-19
Move lnurl related functions into lib

dylancom:


src/common/lib/lnurl.ts
```diff
@@ -6,7 +6,20 @@ import { parsePaymentRequest } from "invoices";
 import { bech32Decode } from "../utils/helpers";
 
 const lnurl = {
-  async getDetails(lnurlEncoded) {
+  findLnurl(bodyOfText: string) {
+    const res =
+      /^(?:http.*[&?]lightning=|lightning:)?(lnurl1[02-9ac-hj-np-z]+)/.exec(
```
bumi, line 11: hmm. I can not read that regex, what all does it match?

dylancom: Schermafbeelding 2021-08-21 om 09 58 33

dylancom: I borrowed it from https://github.com/BlueWallet/BlueWallet/blob/bfb10a0278d23538917fb0f5ddc37278a69aaceb/class/lnurl.js as I wasn't sure which characters have to be matched after "lnurl".
There might be a breaking change coming up though: https://github.com/fiatjaf/lnurl-rfc/blob/luds/17.md

bumi: I still don't understand this regex. 

so for my understanding the LNURL is either a string starting with LNURL or a URL like string starting with lnurl?://
(or maybe a [lightning address](https://lightningaddress.com/)

PR #174, opened 2021-08-20
QR Code Scanner

dylancom:


static/views/welcome.html
```diff
@@ -9,6 +9,7 @@
 
 <body>
   <div id="welcome-root"></div>
+  <script src="/assets/js/html5-qrcode.min.js"></script>
```
bumi, line 12: do we need this AND the npm package as dependency?

src/app/screens/Onboard/ConnectLndHub/index.jsx
```diff
@@ -78,13 +80,28 @@ export default function ConnectLndHub() {
...
                 <small>
                   BlueWallet users find the Backup/Export URI in the lightning
                   wallet settings.
                 </small>
               </div>
+              <p className="text-center my-4">OR</p>
+              <QrcodeScanner
+                fps={10}
+                qrbox={250}
+                qrCodeSuccessCallback={(decodedText) => {
```
bumi, line 95: maybe we can "stop" the scanning after we successfully have done a scan and set the form data?

static/views/welcome.html
dylancom: Currently only the script tag. So it could be left out. I left the dependency for a couple of reasons:
- The package might support module in the future (that way we can import directly from within react).
- We could create a build task that copies html5-qrcode.min.js from the node_modules, instead of the currently manually included one.

bumi: 👍 adding a TODO there

PR #176, opened 2021-08-22
Add option to export LndHub backup string

bumi:


src/app/screens/Onboard/NewWallet/index.js
```diff
@@ -65,12 +79,44 @@ export default function ConnectLndHub() {
         <div className="lg:w-1/2">
           <h1 className="text-3xl font-bold">Get a new lightning wallet</h1>
           <p className="text-gray-500 mt-6"></p>
-          <div className="w-4/5">
-            <div className="mt-6">
-              <strong>Remember, not your keys, not your coins.</strong>
-              This quick setup uses a custodial service to manage your wallet.
+          {!lndHubData.login && (
```
dylancom, line 82: You can group this logical operator together with the one at line 90.
And wrap a "Fragment" around them so there is no `<div>` needed.
See: https://reactjs.org/docs/fragments.html#short-syntax

src/app/screens/Onboard/NewWallet/index.js
```diff
@@ -80,7 +126,11 @@ export default function ConnectLndHub() {
                 return false;
               }}
             />
-            <Button label="Continue" onClick={handleNext} primary />
+            <Button
+              label={lndHubData.login ? "Next" : "Create a wallet"}
+              onClick={handleNext}
```
dylancom, line 131: You can directly pass in the right func to onClick, so there is no handleNext method needed.
`onClick={lndHubData.login && lndHubData.password ? next : signup}`

dylancom (changes requested):

src/app/screens/Onboard/NewWallet/index.js
bumi: but this one is `!lndHubData.login` (when we do not have a login, yet) while the other one is `lndHubData.login` (when we got a login)

bumi: ah ja, that's nicer. thanks!

PR #177, opened 2021-08-22
Add LN faucet

bumi:
This gives users with a zero balance a small faucet to quickly get started and test the extension

### DIFF
--- a/src/app/screens/Onboard/TestConnection/index.jsx
+++ b/src/app/screens/Onboard/TestConnection/index.jsx
@@ -1,14 +1,39 @@
 import React, { useState, useEffect } from "react";
+import axios from "axios";
 import { useHistory } from "react-router-dom";
+import Modal from "react-modal";
+import CrossIcon from "@bitcoin-design/bitcoin-icons/svg/outline/cross.svg";
+import Input from "../../../components/Form/input";
 import Button from "../../../components/button";
 import Card from "../../../components/card";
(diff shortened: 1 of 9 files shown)


src/app/screens/Onboard/TestConnection/index.jsx
```diff
@@ -32,8 +94,61 @@ export default function TestConnection() {
...
+              name="uri"
+              type="text"
+              onChange={handleEmailChange}
+              required
+              placeholder="you@email.com"
+            />
+          </div>
+        </div>
+        <div className="flex justify-end p-5">
+          {faucetLoading && <Loading />}
```
dylancom, line 143: You can refactor this to: `{faucetLoading ? <Loading /> : <Button onClick={claimSats} label="Get Sats" primary />}`

src/app/screens/Onboard/TestConnection/index.jsx
```diff
@@ -49,19 +164,28 @@ export default function TestConnection() {
...
+                        className="underline"
+                        onClick={(e) => {
+                          e.preventDefault();
+                          setShowFaucet(true);
+                        }}
+                      >
+                        To get started we can send you some Sats...
                       </a>
-                    </p>
+                      {faucetURL && renderFaucet()}
```
dylancom, line 185: You're already checking if faucetURL exists at line 172, so an extra check is not necessary.

PR #181, opened 2021-08-23
Fix typo in thanks section

wbobeirne:
Just wanted to avoid the connection to Jeff Bezos. Unless that was intentional, but I'll assume it was not.

But seriously, thanks for dropping the note in the README, and really excited to see all the progress here. 100% agree with the "why not joule" section and really glad to see a more agnostic and lightweight solution in the works. I'll be hanging out in the Slack channel, feel free to ping me if there's anything you need from me!


bumi:
what...uuuupppsss...lol. wondering how that happened...🙈  
sorry, it should obviously be "**super amazing work**"!

thanks! and I am happy to hear that feedback.

PR #190, opened 2021-08-30
Experiment with lightning enhancement of existing websites

bumi:
# Lightning charge the current web! 

This inspects websites for lightning information (e.g. a special file in a GitHub repository or the podcast value data on podcast index). This information then can be used to display in the popup and allow the user to easily send a lightning payment or stream payments.

## Questions

* How do we show this information?
* How do we support payments to multiple recipients? 

### GitHub
We check GitHub repositories for a `.lndonate` file (as also proposed by [ln-donate](https://github.com/Kixunil/ln-donate)) and extract a LNURL from it. 

### Podcast Index
We extract the podcast value data provided by podcast index. 
Currently this seems to be mainly keysend payments which we do not yet support (but can easily be added)

### Metatag
We extract lightning information from a metatag (similar to the shitcoin [webmonetization](https://webmonetization.org/))

`<meta name="lightning" content="lnurlp:LNURL" />`

### Twitter
Extract a LNURL/Lighting Address from the profile bio. 
⚡️followed by an lighting address or a lnurlp://... string

Twitter makes it super hard to parse and extract data from the DOM...we need to test what works reliably

### YouTube
Extract a lnurlp:// string from the channel about page


src/app/components/Loading/index.jsx
```diff
@@ -1,27 +1,29 @@
 import React from "react";
 
-function Loading() {
+function Loading({ color = "currentColor" }) {
```
bumi, line 3: nice, did not know about `currentColor`

bumi:
Current Firefox build: https://wormhole.app/70WDW#g68DSysCZV1cpzSUlJCINQ

src/extension/background-script/index.js
```diff
@@ -14,11 +14,21 @@ setInterval(() => {
...
     return;
   }
-  if (!changeInfo.url.startsWith("http")) {
+
+  // delay execution a bit to give the website chance to potentially load async data (e.g. most data from twitter is loaded async)
+  setTimeout(() => {
+    browser.tabs.executeScript(tabId, {
+      code: "if ((document.location.protocol === 'https:' || document.location.protocol === 'http:') && window.LBE_EXTRACT_LIGHTNING_DATA) { LBE_EXTRACT_LIGHTNING_DATA(); };",
+    });
+  }, 500);
```
dylancom, line 27: This can easily become buggy with slow connections etc. Maybe better to check with an interval (for e.g. every 100ms) inside a battery if the html contains certain data and if so, cancel the interval.

dylancom (changes requested):

fiatjaf:
Can we change the `<meta>` tag to use [LUD-17](https://github.com/fiatjaf/lnurl-rfc/blob/luds/17.md) instead? i.e., `<meta name="lightning" content="lnurlp:domain.com/path" />`.

fiatjaf:
Ok, great!

Kixunil:
Awesome! Some notes from me:

* Just to be clear, I intend `.lndonate` to be independent of web VSC frontend provider so if anyone implements it for GitLab and Bitbucket I would be more than happy. Having just GitHub is still a good start of course.
* There is a way to use GitHub sponsor button by having `.github/FUNDING.yml`. I don't see a clean way to put LNURL there unless web fallback is implemented but even that will have privacy implications. Given `.lndonate` doesn't depend on GitHub I believe it's slightly better. I suggest **not** to pursue `FUNDING.yml` parsing.
* Meta tag feels like missing a bit of context. Is it for donations or a paywall? Would be nice to provide some more information into the tag so that it's distinguishable.
* We should write proper specification for the meta tag somewhere. Perhaps LUD or BIP.

fiatjaf:
Why can't you just write

```
lightningaddress: me@myself.com
```

There? It doesn't break, it's just not parsed, right, @kaloudis?

Kixunil:
@fiatjaf not sure if you're replying to me. GitHub will just show an error that FUNDING.yml is malformed and will show no fallback links inside `custom`. Try it, you'll see.

bumi:
@Kixunil 
* yes. I like the idea of an .lndonate file in a repository to be independent of the web frontend. Here in the wallet the question is how to find that file and for that we need checks for the individual web frontends - though it is easy to add more there. or do you have an idea how to make it work for other git web UIs?
* agree. even though with a .FUNDING link like this: https://github.com/bumi/lnme/blob/master/.github/FUNDING.yml - the extension here recognizes the `lightning` query parameter and opens the lnurl payment flow. But a website is required as fallback. 
* For now the metatag is for donations. A "paywall" would require more interaction/validation of the payment and then you would rather do that with WebLN. The idea was to make it simple for any website to indicate that lightning donations are accepted. Similar to the value4value podcast tag. 
* Yes, would be good to write a spec and do a RFC for that. Also an accessible website like lightningaddress would be great... so much exciting work to do :D

bumi:
I am merging this PR as the feedback has been great so far and we can improve it from there. The PR gets big and old :)

Issue #47, opened 2021-04-10
Command husky-run not found


chris-aeviator:
hey @bumi is it possible that branches I take from you include a githook that tries to find husky.sh? I cannot base any new commit ontop of https://github.com/bumi/lightning-browser-extension/pull/37 

```
❯ git commit
error Command "husky-run" not found.
```

EDIT: I actually have husky.sh in .git

bumi:
I was complaining about that in the chat also. :)
it is doing some linting in a pre-hook: https://github.com/bumi/lightning-browser-extension/blob/master/package.json#L33

I think this should not be a default and running `yarn install` should not automatically add git hooks. what do you think?

PR #191, opened 2021-08-30
Remove husky

bumi:
do we need that?

[#47]


dylancom:
@bumi if you want to remove Husky, you want to remove lint-staged (config + package) aswell

bumi:
#194

PR #202, opened 2021-09-04
Components cleanup

dylancom:
- Remove shared folder.
- Move stories into component folders
- Remove component duplicates


bumi:
@dylancom I think we somehow lost the fixed width/size of the popup in this PR?

dylancom:
@bumi which browser / screen? I'm not noticing any differences.

bumi:
image
The popup in FF. but haven't looked in detail

PR #211, opened 2021-09-07
[WIP] Send screen

dylancom:
- [x] Refactor routing to re-use navbar (in popup).
- [ ] Paste invoice.
- [ ] Scan Invoice.
- [ ] Confirm.
- [ ] Make it work from options

### DIFF
--- a/src/app/components/Menu/MenuList.tsx
+++ b/src/app/components/Menu/MenuList.tsx
@@ -22,7 +22,7 @@ function List({ position = "left", children }: Props) {
           position === "left"
             ? "left-0 origin-top-left"
             : "right-0 origin-top-right"
-        } absolute mt-2 py-1 w-56 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none`}
+        } absolute z-50 mt-2 py-1 w-56 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none`}
       >
         {children}
       </Menu.Items>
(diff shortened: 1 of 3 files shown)


bumi:
The QR code scanning does not work for me in the popup (in FF) - I guess I do not get asked if I want to allow access to the camera. 
When I fully open the popup.html in it's own tab it works. Maybe we have to open the full page for scanning?

dylancom:
It seems asking for permission from within the extension is problematic in Firefox. After granting permissions from within popup.html in it's own tab it's working fine.
It returns the following error from the addon: **"NotAllowedError : The request is not allowed by the user agent or the platform in the current context."**

dylancom:
So the scanning can work fine from within the popup, but it seems we need to ask permission from within a tab. See: https://discourse.mozilla.org/t/call-getusermedia-from-web-extension/36835/2

bumi:
Is there a way that we can keep the permission status on a page reload. 
Now if the user visits the settings page again it again shows the button to grant the permissions.

dylancom:
@bumi this is possible in Chrome, but unfortunately not in Firefox & Safari yet AFAIK.
See: https://stackoverflow.com/a/64274323

bumi:
@johnsBeharry what do you say?

bumi:
Firefox build: https://wormhole.app/R8NBl#CY7X9aKDrHoXE-LT5y7I4A
Chrome build: https://wormhole.app/bbxqD#ufpnbXlBFxPQv5RYmtTVnA

PR #216, opened 2021-09-09
Remove dev scripts from prod

dylancom:


bumi:
hmm... wondering why we don't have it in the options.html and welcome.html?

dylancom:
Found out it's not going to work anyway: https://github.com/facebook/react/issues/20453

bumi:
👍

PR #221, opened 2021-09-09
Improved (default) button design

dylancom:


bumi:
he, nice... was just before thinking about opening an issue about that.

PR #223, opened 2021-09-10
Add LNbits connector

bumi:
this allows the user to use an LNbits account.
Connection is done with the LNbits URL and the admin key


bumi:
Build for chrome: [chrome.zip](https://github.com/bumi/lightning-browser-extension/files/7149862/chrome.zip)

Issue #214, opened 2021-09-08
Display preimage on transaction list


bumi:
The user should be able to access the preimage of each payment in the transaction list. 

This should be possible by adding the preimage to: https://github.com/bumi/lightning-browser-extension/blob/master/src/app/components/TransactionsTable/index.tsx#L45

The preimage is already stored in the payments (https://github.com/bumi/lightning-browser-extension/blob/master/src/extension/background-script/db.js#L11)

lejara:
Hello!

Is it okay if i can work on this?

If yes,
how do you want it styled?

bumi:
Hey @lejara! yay, that's would be great! 

I think we can add it to the `subtitle` of the transaction or add a new similar row after that? (https://github.com/bumi/lightning-browser-extension/blob/master/src/app/components/TransactionsTable/index.tsx#L54-L56)

maybe you can try it and then @johnsBeharry can have a look and we can iterate on it? 

let me know if you need  help.

lejara:
Cool! 
I added it under the subtitle. Though I think best to add in a drop down within the transaction. Since preimages are referred as secrets, up to you all.

https://github.com/lejara/lightning-browser-extension/commit/16dd91b69365659de1b7ed3695a383cd82e7d69a



![preimage_under_subtile](https://user-images.githubusercontent.com/16671822/132703825-899f98f7-3266-4939-a78f-c9303c180053.JPG)

bumi:
yeah, that's already great and helpful. feel free to open a pull request.
How would you do the dropdown? 
Maybe we could also indicate that it is the preimage and maybe also make the font smaller? 

@johnsBeharry do you have any design thoughts?

lejara:
Just by adding a chevron...

Closed:
![dropdown_close](https://user-images.githubusercontent.com/16671822/132933679-d83679b0-39da-470a-b270-84d98bc0e30c.jpg)

Open:
![dropdown_open](https://user-images.githubusercontent.com/16671822/132933686-386a03c5-b127-421b-90b3-fb06f97f9468.jpg)

And yeah indicating its a preimage seems like a good idea! I can add it.

Other than that if your okay with everything, i will open a pull request.

bumi:
I think that's a great idea. Maybe we can move the whole details behind that?

johnsBeharry:
Welcome @lejara thanks for your interest in the project!

[...]

PR #229, opened 2021-09-13
Add preimage and disclosure to transaction table

lejara:
If there are any style changes you like, ill be happy to add

Resolves: #214

### DIFF
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -16,6 +16,7 @@ type Transaction = {
   amount: string;
   currency: string;
   value: string;
+  preimage: string;
 };
 
 export default function TransactionsTable({ transactions }: Props) {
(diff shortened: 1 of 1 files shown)


bumi:
great! I think that is perfect! thanks

src/app/components/TransactionsTable/index.tsx
```diff
@@ -43,35 +46,60 @@ export default function TransactionsTable({ transactions }: Props) {
...
-                  {tx.type && renderIcon(tx.type)}
-                </div>
-                <div className="overflow-hidden">
-                  <div className="text-gray-900 truncate">{tx.title}</div>
-                  <div className="text-sm text-gray-500 truncate">
-                    {tx.subTitle}
+          <div className="px-3 py-2">
+            <Disclosure>
+              {({ open }) => (
+                <>
```
dylancom, line 52: The "Fragment" can be left out as `<div key {tx.id} />` is already the outer div.

dylancom (changes requested):

bumi:
yay, thanks! 🎉

PR #241, opened 2021-09-29
set default memo for invoice

pseudozach:
This PR fixes an error when a service creates an invoice with no memo field, for instance via WebLN: `var payreq = await webln.makeInvoice(parseInt(amount));`.
As per WebLN spec, memo field is optional but backends like lnbits require a memo.


bumi:
thanks for the debugging and the PR!

I think one ToDo that we have here is to open a prompt on the `webln.makeInvoice` calls and ask the user for confirmation and then also for a memo. 
And then we need some input validations to make sure we always call the methods on the connected nodes with the correct data.

Issue #217, opened 2021-09-09
Refactor sendPayment to close the prompt only after the payment is sent


bumi:
Currently we close the prompt directly after the user confirms the payment. This is because of the architecture of the prompt.

The sendPayment action waits for data from the prompt (waits for a promise to resolve) and depending on that data (confirmed or not) it continues with the payment. 
We might want to refactor this that the sendPayment action only handles the payment if an allowance is available. Otherwise it just opens the confirmation prompt and the confirmation prompt is then in charge to send the payment and then close the prompt.

We already do this for the lnurl-pay flow.

PR #253, opened 2021-10-05
refactor: close prompt after succesful payment

dylancom:
This fixes: #217

For now I called the action weblnPay.

### DIFF
--- a/src/extension/background-script/actions/ln/index.js
+++ b/src/extension/background-script/actions/ln/index.js
@@ -1,6 +1,6 @@
 import PubSub from "pubsub-js";
 import state from "../../state";
-import sendPayment from "./sendPayment";
+import sendPayment, { weblnPay } from "./sendPayment";
 import signMessage from "./signMessage";
 import getInfo from "./getInfo";
 import makeInvoice from "./makeInvoice";
@@ -36,6 +36,7 @@ export {
   getBalance,
   getTransactions,
   sendPayment,
+  weblnPay,
   signMessage,
   makeInvoice,
   verifyMessage,
--- a/src/extension/background-script/actions/ln/sendPayment.js
+++ b/src/extension/background-script/actions/ln/sendPayment.js
@@ -24,7 +24,7 @@ const sendPayment = async (message, sender) => {
   ) {
     return sendPaymentWithAllowance(message, paymentRequestDetails, allowance);
   } else {
-    return sendPaymentWithPrompt(message, paymentRequestDetails);
+    return payWithPrompt(message);
   }
 };
 
@@ -42,28 +42,31 @@ async function sendPaymentWithAllowance(
   return response;
 }
 
-async function sendPaymentWithPrompt(message, paymentRequestDetails) {
+async function payWithPrompt(message) {
+  await utils.openPrompt(message);
+}
+
+export async function weblnPay(message) {
+  const { paymentRequest } = message.args;
   const connector = state.getState().getConnector();
-  let promptResponse;
-  let paymentResponse;
+  const paymentRequestDetails = parsePaymentRequest({
+    request: paymentRequest,
+  });
+
   try {
-    promptResponse = await utils.openPrompt(message);
-    if (promptResponse && promptResponse.data.confirmed) {
-      paymentResponse = await connector.sendPayment({
-        paymentRequest: message.args.paymentRequest,
-      });
-      publishPaymentNotification(
-        message,
-        paymentRequestDetails,
-        paymentResponse
-      );
-      return paymentResponse;
-    }
+    const response = await connector.sendPayment({
+      paymentRequest,
+    });
+    publishPaymentNotification(
+      message.args.message,
+      paymentRequestDetails,
+      response
+    );
+
+    return response;
   } catch (e) {
-    console.log("Payment cancelled", e);
-    return { error: e.message };
+    console.log(e.message);
   }
-  return { error: "Failed" };
 }
 
 export function publishPaymentNotification(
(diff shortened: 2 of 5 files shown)


bumi:
> What about when a user has set an allowance. Should the extension show some loading state in a different way, or is it up to the publisher?

I would say it is up to the publisher but maybe we can indicate that with some icon state (color, or a different character then)

src/extension/background-script/router.js
```diff
@@ -30,6 +30,7 @@ const routes = {
   lnurl,
   lnurlPay,
   sendPayment: ln.sendPayment,
+  weblnPay: ln.weblnPay,
```
bumi, line 33: maybe that's a good time not to use those namespaces that we actually support? 

`webln.sendPayment` (is what the content script calls) vs. `ln.sendPayment` is what we internally call to actually perform the payment.  - though it might not be good to have to files with the same name? (but different folder?)

right now I would reverse the names: 
weblnsendPayment for what the content script calls (sendPayment right now) and then `sendPayment` to do the actual payment (your weblnPay) - I see anything webln close to the website...

src/extension/background-script/actions/ln/index.js
```diff
@@ -1,6 +1,6 @@
 import PubSub from "pubsub-js";
 import state from "../../state";
-import sendPayment from "./sendPayment";
+import sendPayment, { weblnPay } from "./sendPayment";
```
bumi, line 3: I'd move that in a different file. Then we keep the convention of one action per file? - does that make sense?

src/extension/background-script/actions/ln/sendPayment.js
```diff
@@ -42,28 +42,31 @@ async function sendPaymentWithAllowance(
...
+    const response = await connector.sendPayment({
+      paymentRequest,
+    });
+    publishPaymentNotification(
+      message.args.message,
+      paymentRequestDetails,
+      response
+    );
+
+    return response;
```
bumi, line 66: I think this new flow is better, isn't it? 
better than expecting a respone from the openPrompt and then continuing based on that response

what do you think?

src/extension/background-script/actions/ln/sendPaymentOrPrompt.js
```diff
@@ -0,0 +1,49 @@
...
+  const response = await connector.sendPayment({
+    paymentRequest: message.args.paymentRequest,
+  });
+  utils.publishPaymentNotification(message, paymentRequestDetails, response);
+
+  return response;
+}
+
+async function payWithPrompt(message) {
+  await utils.openPrompt({ ...message, type: "sendPayment" });
```
bumi, line 46: I think this is a problem here because now we change the response of the `sendPaymentOrPrompt` call. `sendPaymentWithAllowance` and `payWithPrompt` have different responses. 

the call is initiated from the website with something like the following:

```js
webln.sendPayment(invoice).then(response => { 
  if (verifyPayment(response)) {
    // do something.
  }
}).catch(e => { 
  // show fallback
});
```

so we have to make sure that either way the `sendPayment` promise resolves with the same object.

which might be a problem? hmm..

bumi (changes requested):

bumi (commented):
What do you think about moving the new `sendPaymentOrPrompt.js` to a new `actions/webln` folder? 
Then the `ln` folder contains all actions that directly talk to the lightning connector and these more "highlevel" actions get their own folder?

bumi:
> Something is still broken with the events. The payments do not get persisted.
this is because of:
https://github.com/getAlby/lightning-browser-extension/pull/253/files#r727028434

dylancom:
> What do you think about moving the new `sendPaymentOrPrompt.js` to a new `actions/webln` folder? Then the `ln` folder contains all actions that directly talk to the lightning connector and these more "highlevel" actions get their own folder?

I think it's a good idea to make this distinction.

src/common/lib/utils.js
```diff
@@ -61,6 +62,17 @@ const utils = {
   stringToUint8Array: (str) => {
     return Uint8Array.from(str, (x) => x.charCodeAt(0));
   },
+  publishPaymentNotification: (message, paymentRequestDetails, response) => {
```
bumi, line 65: I am not so happy with this being in `utils` this is very payment related and we need to make sure utils does not become too big with random functions. :D 

(but only for the future I think)

src/extension/background-script/actions/webln/sendPaymentOrPrompt.js
```diff
@@ -0,0 +1,48 @@
...
+  ) {
+    return sendPaymentWithAllowance(message, paymentRequestDetails, allowance);
+  } else {
+    return payWithPrompt(message, paymentRequestDetails);
+  }
+};
+
+async function sendPaymentWithAllowance(message, paymentRequestDetails) {
+  const response = sendPayment(message);
+  utils.publishPaymentNotification(message, paymentRequestDetails, response);
```
bumi, line 30: It think this should also go into `sendPayment` - we always want to do that when a paument is sent and we have all the data there, don't we?

src/extension/background-script/actions/webln/sendPaymentOrPrompt.js
```diff
@@ -0,0 +1,50 @@
...
+    return payWithPrompt(message);
+  }
+};
+
+async function sendPaymentWithAllowance(message) {
+  try {
+    const response = await sendPayment(message);
+    return response;
+  } catch (e) {
+    return { error: e.message };
```
bumi, line 33:
```suggestion
    console.error(e);
    return { error: e.message };
```

PR #255, opened 2021-10-05
feat: edit or delete allowance from popup

### LINKED ISSUE #112: Disable allowance/publisher from popup
We show the publisher details in the popup if the user is on the publisher's website. 
Currently the user can not change the allowance or remove/disable the publisher. 

As a user I want to be able to "disable" a publisher and update the allowance details.

dylancom:
Adds #112

### DIFF
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -1,0 +1,119 @@
+import React, { useState } from "react";
+import Modal from "react-modal";
+import { CogIcon } from "@heroicons/react/solid";
+import CrossIcon from "@bitcoin-design/bitcoin-icons/svg/outline/cross.svg";
+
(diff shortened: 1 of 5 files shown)


src/app/components/AllowanceMenu/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  function closeModal() {
+    setIsOpen(false);
+  }
+
+  async function updateAllowance() {
+    await utils.call("updateAllowance", {
+      id: parseInt(allowance.id),
+      totalBudget: budget,
+    });
+    onEdit && onEdit();
```
bumi, line 39: naming nitpick: we do an `updatAllowance` but call `onEdit`

bumi (approved):
this is great! it is good to be able to directly edit it in the popup. 
gives also the user the feeling to be in control I think.

PR #259, opened 2021-10-06
Limit PR builds

bumi:
This runs the builds on master branches and if the PR has a certain label.
PR runs now also get access to the secrets and full repo access (event: pull_request_target).
To limit this use we now require PRs to have a certain label (which can only be added from maintainers)


bumi:
test, please ignore

bumi:
test, please ignore

PR #263, opened 2021-10-07
Improved popup design

dylancom:
Fixed width, smallers texts, better alignment, backgrounds etc.

### DIFF
--- a/src/app/screens/Home/index.jsx
+++ b/src/app/screens/Home/index.jsx
@@ -151,7 +151,7 @@ class Home extends React.Component {
                     <a
                       target="_blank"
                       title={payment.name}
-                      href={`options.html#/publishers`}
+                      href={`options.html#/publishers/${allowance.id}`}
                       rel="noreferrer"
                     >
                       {payment.name}
@@ -163,7 +163,7 @@ class Home extends React.Component {
                     <a
                       target="_blank"
                       title={payment.name}
-                      href={`options.html#/publishers`}
+                      href={`options.html#/publishers/${allowance.id}`}
                       rel="noreferrer"
                     >
                       {payment.host}


src/app/router/Popup/styles.scss
```diff
@@ -1,7 +1,5 @@
 #popup {
-  min-width: 350px;
-  max-width: 540px;
-  min-height: 420px; /* Needed to expand the user menu. */
+  min-height: 420px; /* Needed to account for the user menu. */
```
bumi, line 2: do we need a max-height?

src/app/components/TransactionsTable/index.tsx
```diff
@@ -24,46 +24,42 @@ type Transaction = {
...
-                          </div>
-                        </Disclosure.Panel>
+                <>
+                  <div key={tx.id} className="flex">
+                    <div className="flex items-center flex-shrink-0 mr-3">
+                      {tx.type && renderIcon(tx.type)}
+                    </div>
+                    <div className="overflow-hidden mr-3">
+                      <div className="text-sm font-medium text-gray-900 truncate">
+                        {tx.title}
```
bumi, line 58: where do we set the link for this? can we link directly to the publishers detail page? (right now we link to the publishers list.

but I guess in the parent component where we use the TransactionTable we do not know the link of the individual payment and thus only have one link for all..?

dylancom: We can link to: options.html/publishers/${id}

bumi: ah true, we can do that here? https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/index.jsx#L154 correct?

dylancom: I updated the links for an allowance. When there is no allowance (just payments), I'm not sure yet how we can get the publisher id.

Issue #264, opened 2021-10-07
Back button missing for the LNbits connector


bumi:
On the onboarding screens the "Back" button is missing on the lnbits connector screen.

<img width="533" alt="image" src="https://user-images.githubusercontent.com/318/136401530-df384f63-1459-4620-862e-924070f496f0.png">

PR #268, opened 2021-10-11
fix: missing back button

dylancom:
Issue: #264

### DIFF
--- a/src/app/screens/Onboard/ConnectLnbits/index.jsx
+++ b/src/app/screens/Onboard/ConnectLnbits/index.jsx
@@ -89,8 +89,16 @@ export default function ConnectLnbits() {
               </div>
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


src/app/screens/Onboard/ConnectLnbits/index.jsx
```diff
@@ -89,8 +89,16 @@ export default function ConnectLnbits() {
...
+          <div className="mt-8 flex space-x-4">
+            <Button
+              label="Back"
+              onClick={(e) => {
+                e.preventDefault();
+                history.goBack();
+                return false;
+              }}
+            />
+            <Button type="submit" label="Continue" primary />
```
bumi, line 101: :) I think we have to unify the use of the word "Continue" vs. "Next"

PR #280, opened 2021-10-13
feat: tipping directly from popup

dylancom:
I replaced Hash Router by Memory Router which has more capabilities such as letting you pass data to a route.


bumi:
@MoritzKa can you test this one with the package linked above?

dylancom:
@bumi Done

Issue #315, opened 2021-10-23
Redesign add new account screen


bumi:
* work with wallet/node logos?

<img width="1084" alt="CleanShot 2021-10-23 at 16 46 34@2x" src="https://user-images.githubusercontent.com/318/138561166-bcd0992e-8981-42a4-880b-175462e10451.png">

secondl1ght:
I can add the logos to this screen if this is needed?

secondl1ght:
Created PR #349 for this

PR #349, opened 2021-11-03
Add logos to new account screen

secondl1ght:
Added logos for each node/wallet option when setting up a new account #315.

Screenshot: 
image

I am open to any feedback and suggestions if anyone wants to change anything i've done here! :)


dylancom:
Cool, looks a lot better already.
Would say that the logo's of LNDHub & LNbits should be a bit more in balance with the other logo's (too big currently).
For the logo's I'd say we use SVG instead of .png?

bumi:
👍 for SVGs also adding @johnsBeharry for a UI review

secondl1ght:
Ok I have made them all the same size and converted from PNG to SVG. I found an online converter tool to do this, however it looks like some of the colors changed a bit when doing that. If anyone knows a better way to convert the images just let me know. Please let me know what you guys think of this iteration:

Screenshot: 
image

secondl1ght:
Here's another version where the LND logo has a blue background: 
image

dylancom:
I would prefer the logo's to appear without a background.

bumi:
> converted from PNG to SVG.

do you know if these projects provide SVG version of the logo? those would have better quality and can easily be scaled then.

secondl1ght:
I'll look into finding the official SVGs for each project tonight. I agree that would be much better.

secondl1ght:
So far I haven't had any luck finding SVGs for all the projects. I found one in the lnbits repo but did not see one in LNDhub or LND. I asked in the LND Slack group if there was an SVG icon for developers but did not get a response. I tried contacting the BlueWallet guys over e-mail and did not hear back yet either. 

So for now we can either use PNGs with the correct colors or use the SVGs with the slight color variations from converting them. Or just wait to merge this PR until we can track down the official SVGs.

secondl1ght:
So it looks like not every project has an official SVG logo. I have changed this PR back to using the PNGs for now. My vote would be to use those rather than convert the logos and cause some distortions.

secondl1ght:
Here's another option for everyone to consider, I changed to the Blue Wallet logo as suggested by @MoritzKa and I also removed the forced rounding and let each logo have its own shape.

image

This one probably looks the best IMO but if we can agree on feedback then maybe we can merge. 

I know @dylancom would like ones without background but with my limited design skills I haven't been able to make PNGs without backgrounds while also making them all appear the same size. Because a 200x200 image with no background will look smaller than one with a background. So if anyone else wants to send me some of their own images I can use those and post screenshots.

MoritzKa:
I wonder how to create a logo without background out of the LNbits logo for instance. The latest update is a great improvement for the extension. Let's ask for feedback in the community call and decide.

secondl1ght:
We didn't bring up this PR in today's meeting but is there anything anyone else wants me to do for this one?

MoritzKa:
It is a great improvement.
What do you think, @johnsBeharry ?

secondl1ght:
Ok @johnsBeharry was nice enough to provide some uniform images that we have been looking for. So now they are all the same size and look good, here's what the updated version looks like:

image

He also gave an alternative option for the Alby logo:

image

secondl1ght:
Johns also gave us a yellow version which I personally like better but we can take a vote on it:

image

MoritzKa:
I would also choose the yellow background for the logo. But the extension design is not final anyway. 
So let's go with the full body Alby logo (not just head) on a white background.

secondl1ght:
Here's what that version looks like:

image

bumi:
I like it! @johnsBeharry can you review

secondl1ght:
Oh sorry I see everyone wants the white background, I just need Johns to put the white version back on dropbox and then I can make the final changes.

bumi:
ok, we need to merge the conflict and then get decide and get this in!

bumi:
so it is very likely that we get the Galoy connector in and we should directly plan that we will have an icon for it

secondl1ght:
OK I have merged the conflicts, yep we can for sure add a Galoy logo as well. Maybe johns has one that is formatted the same as the others? Also the current version of this PR is using the yellow, full-body Alby logo just so everyone knows, but it can be changed still.

bumi (approved):

Issue #382, opened 2021-11-15
Bug: NaN displaying in GUI


secondl1ght:
When you set an allowance for a website and backspace any numbers in the input field, the input will then display NaN and you will not be able to continue entering numbers into the field. 

![image](https://user-images.githubusercontent.com/85003930/141712295-5b3480d5-9905-4a3e-a5e6-f85bca140529.png)

secondl1ght:
I put in a PR that fixes this #392

secondl1ght:
Closed with #392

PR #385, opened 2021-11-15
Add section recommending Polar

secondl1ght:
Added section to README recommending Polar for lightning network testing while working on Alby, for those who may not know about it.


bumi:
What is needed to be done to get Alby connected with Polar? is there any special setup needed because of the certificate?

secondl1ght:
I was able to just run Polar and then connect to Alby via the LND node option by copying the REST Host into Address and Admin Macaroon.

secondl1ght:
Not sure if this was supposed to be approved and merged but it actually got removed in PR #382 because I messed up my master branch fork, sorry I should have just left it in there maybe. Maybe a maintainer can just add this to the README if it was meant to be merged.

bumi:
let's create a new PR and in that PR we also build a small publisher website that can be used for testing locally. 

Alby -> connects to Alice
Website -> connects to Bob

then it is easy to test Alby locally

secondl1ght:
Can you explain a bit more what you want here? I am not sure I understand everything.

bumi:
something like that basically: https://github.com/bumi/402-server-js (but just in a single HTML)

if you use polar and want to fully test Alby you also need a publisher website connected to polar.

secondl1ght:
Isn't there a test like that on https://getalby.com/demo or is that different than what you were thinking?

bumi:
the demo would not work because that one is connected to the mainnet and with polar a local development network is used. So to do a test like that we also need an website that uses the local development network.

secondl1ght:
Oh right ok got it, I think I know what you are looking for now. Not entirely sure how to implement it, but I can give it a try.

PR #386, opened 2021-11-15
Add optional debug mode

bumi:
This allows to easily access important objects from the background script console.
e.g.
debugAlby.state.getState().accounts gives you the list of accounts
debugAlby.connectors gives ja access to the connector classes

this helps with quickly testing certain changes in a REPL.

### DIFF
--- a/src/extension/background-script/index.js
+++ b/src/extension/background-script/index.js
@@ -5,6 +5,7 @@ import utils from "../../common/lib/utils";
 import { router } from "./router";
 import state from "./state";
 import db from "./db";
+import connectors from "./connectors";
 
 import * as events from "./events";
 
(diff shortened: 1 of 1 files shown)


dylancom (commented):
Debug mode seems only available after clicking on reset on: chrome-extension://dekoapbghfekfgdpjcplbdfpnmidpenn/welcome.html

Shouldn't it be directly available when calling yarn run dev:{browser}?

bumi:
yes, in the console of the background script. 
can't think of a reason why that reset button is related.

dylancom:
> yes, in the console of the background script. can't think of a reason why that reset button is related.

https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/setup/reset.js#L5

bumi:
ah sure. yes, true.. we need that setting to enable the debug flag. forgot about that. 
we are missing a UI for this. (but I guess we can add that later)

dylancom (approved):

PR #387, opened 2021-11-17
Faster Webpack building in development

dylancom:
Currently starting webpack in dev is slow and waiting for changes aswell.


bumi (approved):
Works for me great! and I like the yarn.lock diff :D thanks!

PR #396, opened 2021-11-18
Show LNURL-pay responses in the app instead of a notification

### LINKED ISSUE #342: LNURL response notification is not shown in chrome
In Firefox the LNURL response message (e.g. thanks for the payment) is shown as notification.
In chrome I only see the payment success notification with the pre-image. 

Maybe this is a bug, maybe this is only how my chrome notifications are configured?

dylancom:
Fixes #342


src/app/screens/LNURLPay.tsx
```diff
@@ -110,7 +110,9 @@ function LNURLPay(props: Props) {
         }
       }
 
-      window.close();
+      if (shouldCloseWindow) {
+        window.close();
```
bumi, line 114: if we move the window.close() into the switch case, then we do not need to set a new variable.
and it is not much difference between writing: `shouldCloseWindow = true/false` and `window.close();`

what do you think?

bumi (commented):
we also could do a simple `alert` :) but I guess this is much nicer (haven't tested it yet)

src/app/screens/LNURLPay.tsx
dylancom: What if the code doesn't pass the `(successAction && !payment.payment_error)` condition? Currently we'd close the window then too.

bumi: 👍

bumi:
two thoughts: 

while we are at that: shall we also convert the alert in the `url` tag to display such a message with a link (that opens in a new tab)? 

and we might need a `close` link? - when I tested it I had the question how to close/get back to the home screen. ?

dylancom:
I added a close button, which indeed was missing.
For the alert, I'm not sure to open a new tab as the spec speaks about a popup.

bumi:
> For the alert, I'm not sure to open a new tab as the spec speaks about a popup.

I think opening a tab is the easiest (as we do right now) and the ones I have seen so far have been full web pages.

src/app/screens/LNURLPay.tsx
```diff
@@ -212,42 +205,96 @@ function LNURLPay(props: Props) {
     return elements;
   }
 
+  function renderSuccessAction() {
+    if (successAction) {
```
bumi, line 209: maybe `if (!successAction) { return; }` to prevent the nesting?

bumi (approved):

src/app/screens/LNURLPay.tsx
dylancom: Will remove the check and make TS assume the value is present.
As renderSuccessAction is only called when successAction exists.

Issue #381, opened 2021-11-15
Should we match 'Websites' and 'Allowances' naming


secondl1ght:
Currently there is the 'Websites' button on the NavBar and on the dropdown menu there is 'Allowances'. Both buttons take you to the same page so I think they should be named the same. Do people agree with this?

bumi:
@johnsBeharry is this part of some copy writing brainstorming? :)

@MoritzKa ack for using "Websites" there, too?

MoritzKa:
I slightly prefer the "Allowances" in the nav bar more. For me as a user "websites" could be anything. And the main use case for clicking on it (in the pop up) would be to change or check my existing allowances. 
On the other side a user doesn't have an allowance for every website. Also, if I am on the Alby fullscreen then the use case might be different. I would rather like to have an overview of websites that I visited. Therefore "websites" would fit better. 
Maybe in the future we have two pages: active allowances and just websites (or a filter).

secondl1ght:
These are good points @MoritzKa, I thought it would be easy to pick one or the other but you make a good case for both here haha.

Could even use something like 'Connections' or 'Permissions' as well...but I do think Websites is the most obvious for now.

secondl1ght:
PR #397 if we agree on the naming for this.

secondl1ght:
Closed with #397

PR #397, opened 2021-11-18
Update allowances button

secondl1ght:
As we talked about in issue #381 I updated the Allowances button to be called Websites so that it matches the same button in the NavBar which takes you to the page called Websites.


johnsBeharry (approved):
LGTM

bumi (approved):

PR #416, opened 2021-11-25
Upgrade to ESLint 8, don't depend on CRA, introduce new JSX transforms

dylancom:
- Removes depending on the Create React App linting setup. (which requires old eslint / eslint babel deps)
- Introducing the New JSX Transform https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports)
--> No more React imports, which is the preferred style going into the future.
- Improved linting.
- Updates all dev dependencies without errors.


src/app/components/Form/Select/Select.stories.tsx
```diff
@@ -1,9 +1,13 @@
...
-
 import Select from ".";
 
 export const Primary = () => (
-  <Select name="user" value="" onChange={() => {}}>
+  <Select
+    name="user"
+    value=""
+    onChange={() => {
+      console.log("changed");
```
bumi, line 8: do we need this log?

bumi (approved):

bumi (approved):
nice, thanks for doing this!

PR #425, opened 2021-11-26
Fix some linting warnings

bumi:
This fixes some linting warnings of unused variables.

Though I actually like those because it shows the developer what can be used in the function - even if it unused right now.

### DIFF
--- a/src/extension/background-script/actions/allowances/list.js
+++ b/src/extension/background-script/actions/allowances/list.js
@@ -1,6 +1,6 @@
 import db from "../../db";
 
-const list = async (message, sender) => {
+const list = async () => {
   // TODO add filter and ordering?
   let allowances = await db.allowances
     .toCollection()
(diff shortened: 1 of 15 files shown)


dylancom:
@bumi if you don't agree with certain rules, we can add our own rules to estlintrc. (Rules that override the rules we extend)

bumi:
I think we should go with the most widely used and accepted standard. 
what are your thoughts on those?

dylancom:
In general I prefer to go for the most standard rules. But if there's a good reason for an exception I'd say we can add that.
We could add for e.g.: `"@typescript-eslint/no-unused-vars":["error", { "args": "none" }]` to our rules so unused vars still lead to warnings except for function arguments.

bumi:
and what is your opinion?

dylancom:
For our use-case we can add the exception.

bumi:
so what do you think, what should we enforce?

dylancom:
This can be closed in favor of https://github.com/getAlby/lightning-browser-extension/pull/481

PR #429, opened 2021-11-29
Support default memo

dylancom:
https://webln.dev/#/api/make-invoice describes using defaultMemo

### DIFF
--- a/src/extension/background-script/actions/ln/makeInvoice.js
+++ b/src/extension/background-script/actions/ln/makeInvoice.js
@@ -3,7 +3,7 @@ import state from "../../state";
 
 const makeInvoice = async (message, sender) => {
   if (message.args.memo === undefined) {
-    message.args.memo = "Alby invoice memo";
+    message.args.memo = memo.args.defaultMemo || "Alby invoice memo";
   }
   PubSub.publish(`ln.makeInvoice.start`, message);
 
--- a/src/extension/background-script/actions/ln/makeInvoice.js
+++ b/src/extension/background-script/actions/ln/makeInvoice.js
@@ -5,7 +5,7 @@ const makeInvoice = async (message, sender) => {
   PubSub.publish(`ln.makeInvoice.start`, message);
 
   const amount = parseInt(message.args.amount);
-  const memo = memo.args.memo || memo.args.defaultMemo || "Alby invoice memo";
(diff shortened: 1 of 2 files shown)


src/extension/background-script/actions/ln/makeInvoice.js
```diff
@@ -3,7 +3,7 @@ import state from "../../state";
 
 const makeInvoice = async (message, sender) => {
   if (message.args.memo === undefined) {
-    message.args.memo = "Alby invoice memo";
+    message.args.memo = memo.args.defaultMemo || "Alby invoice memo";
```
bumi, line 6: maybe instead of the if and the line 11 we do something like: 

```
const memo = message.args.memo || message.args.defaultMemo || "Alby invoice";
```

bumi (changes requested):

src/extension/background-script/actions/ln/makeInvoice.js
```diff
@@ -2,13 +2,11 @@ import PubSub from "pubsub-js";
 import state from "../../state";
 
 const makeInvoice = async (message, sender) => {
-  if (message.args.memo === undefined) {
-    message.args.memo = "Alby invoice memo";
-  }
+  const memo = memo.args.memo || memo.args.defaultMemo || "Alby invoice memo";
+  message.args.memo = memo;
```
bumi, line 6:
```suggestion
```
I think we do not even need that. we only use `amount` and `memo` in line 12 and ff. 
or am I missing the message usage somewhere?

bumi (approved):

src/extension/background-script/actions/ln/makeInvoice.js
```diff
@@ -2,13 +2,10 @@ import PubSub from "pubsub-js";
...
 
 const makeInvoice = async (message, sender) => {
-  if (message.args.memo === undefined) {
-    message.args.memo = "Alby invoice memo";
-  }
   PubSub.publish(`ln.makeInvoice.start`, message);
 
   const amount = parseInt(message.args.amount);
-  const memo = message.args.memo;
+  const memo = memo.args.memo || memo.args.defaultMemo || "Alby invoice memo";
```
bumi, line 8:
```suggestion
  const memo = message.args.memo || message.args.defaultMemo || "Alby invoice";
```

bumi (approved):

Issue #283, opened 2021-10-15
Convert connectors to TypeScript


bumi:
The connector classes should be the first parts of the app that we should fully convert to TypeScript and create type definitions.
This makes sure that the functions are called with the correct parameters and makes it easier for connector developers to create new connectors. 
Also we need to write specs for those that can be used to test new connectors.

secondl1ght:
Could you point me to where the connector classes are?

bumi:
https://github.com/getAlby/lightning-browser-extension/tree/master/src/extension/background-script/connectors

PR #433, opened 2021-11-29
Typed connectors

dylancom:
Solves #283


src/extension/background-script/connectors/lnbits.ts
```diff
@@ -136,13 +148,19 @@ class LnBits extends Base {
     });
   }
 
-  async request(method, path, apiKey, args, defaultValues) {
+  async request(
+    method: string,
+    path: string,
+    apiKey: string | undefined,
```
bumi, line 154: can this one really be undefined?

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -0,0 +1,71 @@
...
+    balance: number;
+  };
+}
+
+export type SendPaymentResponse =
+  | {
+      data: {
+        preimage: string;
+        paymentHash: string;
+        route: { total_amt: number; total_fees: number };
```
bumi, line 34: maybe it is a good time to not have this nested in the `route` option? 

originally I tried to follow lnd's response as much as possible, but I am not sure if that makes sense for us

bumi (approved):

Issue #213, opened 2021-09-08
Add support for dynamic invoice creation


bumi:
WebLN specifies invoice creation for fixed values and dynamic values: https://webln.dev/#/api/make-invoice
currently we only support fixed values. 

For dynamic values we need to prompt the user for an amount.

PR #436, opened 2021-12-01
Dynamic invoice

dylancom:
This will prompt the user for an amount when needed.
Can be tested at: https://webln.dev/#/api/make-invoice
Solves: #213


src/app/router/Prompt/Prompt.jsx
```diff
@@ -69,11 +65,19 @@ class Prompt extends Component {
                   />
                 }
               />
+              <Route
+                path="invoice"
+                element={
+                  <MakeInvoice
+                    invoice={this.state.args.invoice}
```
bumi, line 72: maybe we should call this invoiceAttributes or something like that to make clear what is in there? (in this case no bolt11 payment request/invoice)

src/extension/background-script/actions/ln/makeInvoice.js
```diff
@@ -1,14 +1,29 @@
...
   }
   PubSub.publish(`ln.makeInvoice.start`, message);
 
-  const amount = parseInt(message.args.amount);
+  let amount;
   const memo = message.args.memo;
+  if (message.args.amount) {
+    amount = parseInt(message.args.amount);
+  } else {
+    const response = await utils.openPrompt({
```
bumi, line 16: I think we should refactor this pattern as it had been problematic in the past (e.g. impossible to have a loading indicator) 

I think we should do that similarly to what we are doing here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/ConfirmPayment/index.jsx#L22

bumi (commented):
what do you think?

src/extension/background-script/actions/ln/makeInvoice.js
dylancom: Hmm the https://webln.dev/#/api/make-invoice fails if the makeInvoice is not awaiting a response. Not sure how to integrate this pattern here.

bumi: we have to have the response, but here we did it like this: 
prompt -> does another action call and then returns that response back

https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/ConfirmPayment/index.jsx#L29-L34

dylancom: Not sure why I didn't got this working yesterday, but working now :)

Issue #306, opened 2021-10-23
Add option to disable/enable enhancements


bumi:
It should be possible to disable the website tipping enhancements (for Twitter, YouTube, etc.)

PR #438, opened 2021-12-01
Website enhancements toggle

dylancom:
Can be toggled from settings.
Solves: #306


src/app/screens/Settings.tsx
```diff
@@ -1,36 +1,109 @@
...
 function Settings() {
+  const [loading, setLoading] = useState(true);
+  const [settings, setSettings] = useState<{ websiteEnhancements: boolean }>({
+    websiteEnhancements: false,
+  });
   const [cameraPermissionsGranted, setCameraPermissionsGranted] =
     useState(false);
 
+  function getSettings() {
+    browser.storage.sync
```
bumi, line 36: I think this should go through an `action` and the state. what do you think? 

in the end it is a `state.setState({})`

bumi (changes requested):

src/app/screens/Settings.tsx
dylancom: Agree, will update :)

src/extension/content-script/batteries/index.ts
```diff
@@ -8,6 +9,9 @@ import YouTubeVideo from "./YouTubeVideo";
 const enhancements = [Twitter, YouTubeVideo, Monetization];
 
 async function extractLightningData() {
+  const { settings } = await utils.call("getSettings");
+  if (!settings.websiteEnhancements) return;
```
bumi, line 13: just a thought for the future maybe: 
```
const { enabled }  = await utils.call("isFeatureEnabled", {feature: "websiteEnhancements"} );
```

bumi (approved):
utACK

(still need to test it)

PR #449, opened 2021-12-06
Improve connector types with individual configs

dylancom:
Use individual configs

### DIFF
--- a/src/extension/background-script/connectors/lnbits.ts
+++ b/src/extension/background-script/connectors/lnbits.ts
@@ -5,7 +5,6 @@ import Base from "./base";
 import utils from "../../../common/lib/utils";
 import HashKeySigner from "../../../common/utils/signer";
 import Connector, {
-  LnBitsConfig,
   SendPaymentArgs,
   SendPaymentResponse,
   GetInfoResponse,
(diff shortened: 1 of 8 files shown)


src/extension/background-script/connectors/lnbits.ts
```diff
@@ -17,7 +17,19 @@ import Connector, {
...
+interface Config {
+  adminkey: string;
+  url: string;
+}
+
 class LnBits extends Base implements Connector {
+  config: Config;
+
+  constructor(config: Config) {
+    super();
```
bumi, line 29: should this always be called? even if Base does not implement it?
I am even thinking if we should completely get rid of `Base` right now nothing is left. 
an `init()` and `unload()` might be needed for future connections (like #444 ) but then those can also just be added to any connector individually.

src/extension/background-script/connectors/lnd.ts
```diff
@@ -14,7 +14,19 @@ import Connector, {
   VerifyMessageResponse,
 } from "./connector.interface";
 
+interface Config {
+  macaroon?: string;
```
bumi, line 18: I guess the macaroon then is mandatory.

src/extension/background-script/connectors/lndhub.ts
```diff
@@ -19,13 +19,25 @@ import Connector, {
 } from "./connector.interface";
 import { AxiosRequestConfig, Method } from "axios";
 
+interface Config {
+  login?: string;
+  password?: string;
+  url?: string;
```
bumi, line 25: all those are mandatory now

bumi (approved):
good Base class... I had spent so much time with it... byebye

PR #450, opened 2021-12-06
Feedback button

secondl1ght:
I added a feedback button to the popup menu and to the navbar. I also created a feedback component which currently just says 'Thank you for providing feedback to Alby!' but can be updated in the future to have an embedded feedback form. This was my first time using React Router and I just went off the current code as an example, I have been meaning to take a course on it so if I did anything wrong there just let me know. Both buttons open a new tab to: https://feedback.getalby.com

Screenshots:
image

image

Check it out @MoritzKa !

### DIFF
--- a/src/app/components/UserMenu/index.tsx
+++ b/src/app/components/UserMenu/index.tsx
@@ -78,7 +78,7 @@ export default function UserMenu() {
         </Menu.ItemButton>
         <Menu.ItemButton
           onClick={() => {
-            window.open("https://feedback.getalby.com");
+            utils.openUrl("https://feedback.getalby.com");
           }}
         >
           <QuestionIcon className="h-5 w-5 mr-2 text-gray-500" />
--- a/src/app/components/UserMenu/index.tsx
+++ b/src/app/components/UserMenu/index.tsx
@@ -6,6 +6,7 @@ import {
   SendIcon,
   TransactionsIcon,
   ReceiveIcon,
+  QuestionIcon,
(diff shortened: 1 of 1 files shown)


src/app/components/UserMenu/index.tsx
```diff
@@ -75,6 +76,14 @@ export default function UserMenu() {
           <GearIcon className="h-5 w-5 mr-2 text-gray-500" />
           Settings
         </Menu.ItemButton>
+        <Menu.ItemButton
+          onClick={() => {
+            window.open("https://feedback.getalby.com");
```
bumi, line 81: you can use `utils.openUrl` here: https://github.com/getAlby/lightning-browser-extension/blob/00c03bf895a295b90813de1b68d73f05959ea4f0/src/common/lib/utils.ts#L90

bumi (changes requested):
I would say let's merge it with the link only in the menu (and not on the options screen) 
when we know what exactly we show on that Feedback component we can add it again.

secondl1ght:
Sounds good, should I update this PR with only the changes you want for now?

bumi:
yeah, let's do that, then we ship this one

secondl1ght:
Ok sounds good, I have updated this PR so that only the feedback button in the popup menu is included currently.

I stashed my other edits in a new branch so we can pull them back in later if we want.

bumi (approved):
👍 perfect.

PR #452, opened 2021-12-08
Capitalize Unlock button

secondl1ght:
All of the other buttons within the app have capitalization, currently the unlock button is lowercase. I think it looks more official and inline with the rest of the styles if it is also capitalized.

Before:

image

After:

image


bumi (approved):
thanks for looking at those details!

PR #460, opened 2021-12-10
Proposal: updated onboarding screens

bumi:


tailwind.config.js
```diff
@@ -11,6 +13,7 @@ module.exports = {
       },
       fontFamily: {
         sans: ["Inter var", ...defaultTheme.fontFamily.sans],
+        serif: ["Catamaran", ...defaultTheme.fontFamily.sans],
```
bumi, line 16: is this changing the font? I would be hesitant to do that right now in an iterative improvement.

MoritzKa: Catamaran was the font used in the Figma document.

src/app/screens/Unlock/index.jsx
```diff
@@ -61,9 +65,18 @@ function Unlock() {
           primary
           disabled={password === ""}
         />
+
+        <div className="flex justify-center space-x-1 mt-5">
+          <span className="text-text-gray-bitcoin ">Need help? Contact </span>
+          <TextButton
```
dylancom, line 71: I don't see the need for a new TextButton component. This can be just a simple <a> link with a mailto: in the href.

src/app/screens/Unlock/index.jsx
```diff
@@ -61,9 +65,18 @@ function Unlock() {
...
+            label="Alby Support"
+            primary
+            className="font-serif text-base"
+          />
+        </div>
       </form>
       {error && (
```
dylancom, line 78: In the new design I would place the error directly below the input field and give it a red color. (Feels out of place now below support.)

See for e.g. t
Schermafbeelding 2021-12-16 om 11 14 08
he setting a password screen:

dylancom:
The new design of the unlock screen feels great!

dylancom (changes requested):
.h-screen and .min-h-full are being used to stretch the background but those result in unnecessary scrollbars. We can have a continuous background just like the options screen, by removing those classes and applying the "bg-gray-bitcoin" directly on the body.

This can be done in: static/views/welcome.html (see static/views/options.html for a reference)

dylancom:
Avoid adding extra colors to the tailwind setup when unnecessary. Tailwind already has lot's or color classes integrated. So for a gray bg color, we can use bg-gray-50 or for e.g. bg-gray-100. For text gray colors aswel.

bumi:
@umarrg we should try to split each screen or change into a small PR. that makes it easier to review and then faster to merge and iterate on it.

umarrg:
Ok

Issue #310, opened 2021-10-23
Limit amount of transactions show in the popup


bumi:
Currently we just list all transactions. Probably we should only list the last n transactions.

bumi:
@secondl1ght do you maybe want to look into this? 
I feel like as more people start using Alby this becomes more important. otherwise the list becomes super long and useless.

secondl1ght:
I can look into this for sure, I am a bit busy over the next few days so if it needs to be done right away then maybe someone else should do it otherwise i'd be happy to give it a try!

secondl1ght:
PR #461 in

PR #461, opened 2021-12-11
Limit tx history in popup to most recent 10

secondl1ght:
Related to issue #310 I limited the number of transactions displaying in the popup to the most recent 10. I also updated the header to say 'Recent Transactions' so it is more clear that it does not include all of them. I just picked the number 10 but we can change it if everyone thinks that is too short, maybe 21 would be better?

image

### DIFF
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -36,7 +36,7 @@ export default function TransactionsTable({ transactions }: Props) {
   return (
     <div className="shadow overflow-hidden rounded-lg">
       <div className="bg-white divide-y divide-gray-200">
-        {transactions.map((tx) => (
+        {transactions.slice(0, 10).map((tx) => (
           <div key={tx.id} className="px-3 py-2">
             <Disclosure>
               {({ open }) => (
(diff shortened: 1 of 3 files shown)


bumi:
thanks for the PR. 

can you check if we can limit the amount of payments in the action? Then this applies to everywhere we might load the payments. 
We use [dexie](https://dexie.org/) there, it should have some offset or some standard pagination functions.

secondl1ght:
What do you mean by 'in the action', do you mind linking to the code the are referring to?

bumi:
I think this is where we load the payments: 

https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/allowances/getById.js#L20-L24

and:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/payments/all.js#L6

If we limit the payments in the view we still load all payments and return them to the view.
We could add some pagination there then we do not need to take care of it in every view.

bumi:
if you want we can also pair on this one.

secondl1ght:
Sorry I've been super busy and haven't had a chance to dive into this yet. I had a quick look though and it looks like it involves some backend stuff, I am interested in learning that. I wouldn't turn down an opportunity to pair with you if you're offering. I could probably figure it out by myself but would take longer. I'll message you in Slack about setting up a time! Thanks :)

secondl1ght:
Looks good @bumi thanks for pairing on this!

bumi (approved):

Issue #446, opened 2021-12-05
Add link to payment location/website


bumi:
We store the location URL from where the payment was triggered but this is currently not shown in the transaction list. 

If the location is present we should show it and link to it. 

https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/TransactionsTable/index.tsx#L39

Payment data in the DB: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/db.ts#L18

PR #463, opened 2021-12-13
Add payment location link in tx table

secondl1ght:
Requested in issue #446, I am not sure if this is exactly what you wanted @bumi so please review and let me know. Thanks!

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -74,6 +74,7 @@ export type Transaction = {
   badges: IBadge[];
   totalAmount: string;
   description: string;
+  location: string;
 };
 
 export interface Allowance {
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,7 +64,7 @@ export interface IBadge {
 export type Transaction = {
   type: string;
   id: string;
-  title: any;
+  title: string | React.ReactNode;
   subTitle: string;
   date: string;
   amount: string;
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,7 +64,7 @@ export interface IBadge {
 export type Transaction = {
   type: string;
   id: string;
-  title: string;
+  title: any;
   subTitle: string;
   date: string;
   amount: string;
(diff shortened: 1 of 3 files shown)


secondl1ght:
I could remove the link in the popup if we want the Website page to be the main way to link to the payment location (maybe we don't need it in the popup). 

Also I did not test a whole bunch of transactions in a production wallet to view and see how the styling looks for different locations. We should probably do that and check to make sure everything looks good because I did not add any additional styling to this element.

bumi:
Didn't you now remove the link again? 

what I meant was that when you click on a transaction from within the popup we open the options page with the publisher (this is what happens currently)

and on the publisher page when you click on the location or on the title of the payment we open the tx.location in a new tab.

secondl1ght:
Ohhh ok sorry, so you DO want to have the location linked in the popup, but you want me to also remove the title link in the popup? I personally don't really have a strong opinion either way, I wonder if any users like having the title link or even know it is an option currently...

bumi:
I think we have two cases where we show transactions: in the popup and on the full options page (publisher show).
When you click on the transaction in the popup we open the options page (publisher show). - I would leave that like this and only add the location link under the preimage. 

on the publisher show page we can link the full transaction entry (when the user clicks on the the transaction) to the location. 
and also add the location under the preimage. 

does that make sense?

secondl1ght:
@bumi yes I understand now, sorry for the confusion! Please have a look again, I think it is the way you want now. Also just a note that I had to change the title property Type to 'any' in order for the checks to pass. I tried changing it to Element and 'string' or Element but that didn't work. I am not too familiar with TypeScript so maybe you want to make a small edit to fix the type? I can also do some more research on it as well.

src/types.ts
```diff
@@ -64,7 +64,7 @@ export interface IBadge {
 export type Transaction = {
   type: string;
   id: string;
-  title: string;
+  title: any;
```
bumi, line 67: @dylancom do you have an idea what's the best type for this?

bumi (approved):

dylancom:
@bumi string | React.ReactNode

src/types.ts
bumi, line 67:
```suggestion
  title: string | React.ReactNode;
```

@secondl1ght can you try this one?

secondl1ght:
Yep it works, thanks @dylancom!

dylancom:
@secondl1ght you're welcome!

Issue #289, opened 2021-10-18
Poll for invoice status and give success message once paid


bumi:
Currently we only show the invoice QR code and allow the invoice to be copied. 
We should also poll and check if the invoice is paid and show a success message once paid.

secondl1ght:
What do you mean by poll? If you were able to give me an overview of how you want this implemented I could try working on it.

PR #465, opened 2021-12-13
feat: check if invoice is paid

dylancom:
Should solve #289

### DIFF
--- a/src/app/screens/Receive.tsx
+++ b/src/app/screens/Receive.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { useNavigate } from "react-router-dom";
 import { CaretLeftIcon } from "@bitcoin-design/bitcoin-icons-react/filled";
 import { CopyIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
@@ -20,7 +20,8 @@ function Receive() {
     expiration: "",
   });
   const [loading, setLoading] = useState(false);
-  const [invoice, setInvoice] = useState<{ paymentRequest: string }>();
+  const [invoice, setInvoice] =
+    useState<{ paymentRequest: string; rHash: string }>();
   const [copyLabel, setCopyLabel] = useState("Copy");
 
   function handleChange(
@@ -31,6 +32,17 @@ function Receive() {
       [event.target.name]: event.target.value.trim(),
     });
   }
+
+  useEffect(() => {
+    if (invoice) {
+      // utils
+      //   .call("checkPayment", { paymentHash: invoice.rHash })
+      //   .then(({ paid }) => {
+      //     console.log("paid", paid);
+      //   });
+      console.log(invoice);
+    }
+  }, [invoice]);
 
   async function createInvoice() {
     try {
--- a/src/app/screens/Receive.tsx
+++ b/src/app/screens/Receive.tsx
@@ -3,6 +3,7 @@ import { useNavigate } from "react-router-dom";
 import { CaretLeftIcon } from "@bitcoin-design/bitcoin-icons-react/filled";
 import { CopyIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
 import QRCode from "react-qr-code";
+import Confetti from "react-confetti";
 
 import utils from "../../common/lib/utils";
 import { poll } from "../../common/utils/helpers";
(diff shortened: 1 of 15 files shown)


src/app/screens/Receive.tsx
```diff
@@ -32,6 +34,19 @@ function Receive() {
     });
   }
 
+  useEffect(() => {
+    if (invoice) {
+      const pollForPayment = poll({
+        fn: () => utils.call("checkPayment", { paymentHash: invoice.rHash }),
+        validate: (payment) => payment.paid,
+        interval: 5000,
+        maxAttempts: 10,
```
bumi, line 43: I think we should poll a little longer but decrease the interval a bit (maybe 3 seconds? - 5 seconds feels quiet long?)

src/extension/background-script/connectors/lnd.ts
```diff
@@ -63,6 +65,17 @@ class Lnd implements Connector {
     });
   }
 
+  async checkPayment(args: CheckPaymentArgs): Promise<CheckPaymentResponse> {
```
bumi, line 68: to be more generally used ideally we return more invoice data and maybe even call the function `getInvoice`? 
sadly it seems lndhub/lnbits to not really return much useful invoice data.

bumi:
I heard some reports that it people had been waiting for a payment and had been confused. good that you fix it with this PR!

RicardoCapricciosa:
How exactly can I test this? I installed the Firefox extension from above and sent some sats to may mobile wallet. I've received this notification from firefox 
### image

The text is a bit confusing, because I sent the 500 sats FROM Alby, not TO Alby

RicardoCapricciosa:
> How exactly can I test this? I installed the Firefox extension from above and sent some sats to my mobile wallet. I've received this notification from firefox
> ### image
> 
> The text is a bit confusing, because I sent the 500 sats FROM Alby, not TO Alby

Ok, I just tried again with the released version, which behaved the same. So the above assumption wasn't correct :-)

RicardoCapricciosa:
ok, after looking at #289 I got it
image
image

So it works. But what I find a bit confusing is, that once the Alby window is closed the invoice is gone. At first I had a couple of failed attempts to pay the invoice, but that was a problem on my end (zeus, tor, etc.). But I could not try to pay with another wallet because the invoice was gone although not yet paid. I had to create a new invoice.

src/app/screens/Receive.tsx
```diff
@@ -75,6 +101,21 @@ function Receive() {
...
+              <span>waiting for payment...</span>
+            </div>
+          )}
+          {!paid && !pollingForPayment && (
+            <Button
+              onClick={() => checkPayment(invoice.rHash)}
+              label="Check payment status"
+            />
+          )}
+          {paid && <p>Payment received!</p>}
```
bumi, line 117: I think we should then also hide the QR code or maybe have some kind of ✅ overlay. 
or maybe @johnsBeharry confetti?

bumi: is this too much of a dependency?

bumi:
@RicardoCapricciosa nice, thanks for testing. those are some good points. 

can you create an issue for those two topics? I think we do not have issues for those yet. 
The reload is caused because whenever the extension is opened the popup is reloaded and we do not store state across those reloads - which I think for now is fine, it is too complex for now to change that. 

The confusion with "To Alby" is more important.

src/app/screens/Receive.tsx
dylancom: Agree, yeah I like the "fun-factor" of that and think it's a small lib to add :)

bumi: yeah, we need more fun factors! :)

RicardoCapricciosa: > Agree, yeah I like the "fun-factor" of that and think it's a small lib to add :)

I also agree, confetti are nice, at least in the beginning. Once I use the app five times a day the confetti animation might get a bit old, but for now I think it's great!

bumi:
looks nice. 
Though I still think we should replace or overlay the QR code to make it visual that it is paid. 
the `Payment received!` message can easily be missed I think.

dylancom:
Updated the success screen. Could improve it in the future with for e.g. a modal sliding up from the bottom. Something like: 
image

RicardoCapricciosa:
> @RicardoCapricciosa nice, thanks for testing. those are some good points.
> 
> can you create an issue for those two topics? I think we do not have issues for those yet. The reload is caused because whenever the extension is opened the popup is reloaded and we do not store state across those reloads - which I think for now is fine, it is too complex for now to change that.
> 
> The confusion with "To Alby" is more important.

- notification confusion text: https://github.com/getAlby/lightning-browser-extension/issues/473#issue-1084187974
- Alby state storing: https://github.com/getAlby/lightning-browser-extension/issues/474#issue-1084190298

bumi:
perfect thanks @RicardoCapricciosa

bumi (approved):

PR #472, opened 2021-12-19
feat: Citadel connection

AaronDewes:
This implements connection to Citadel. It doesn't work if 2FA is enabled currently, but otherwise works with Citadel 0.5.16.


bumi:
whoot! how cool is that?! 
nice! I really like what you are dong in the sdk! 👍

bumi:
We are currently working on a native companion app that allows to proxy requests through Tor. Tor will have a few downsides over clearnet requests. 
How do you see that in relation to citadel? You do requests within the local network (or within some zeronet or similar), correct?

AaronDewes:
> How do you see that in relation to citadel? You do requests within the local network (or within some zeronet or similar), correct?

We use the local network currently, but Tor can easily be added (+ automatically getting the onion URL, so you do the initial connection with your password, but use Tor for future connections).

AaronDewes:
grafik

Here's a beta build (Different ID so I could get it signed by Mozilla): https://github.com/AaronDewes/lightning-browser-extension/releases/download/beta-1/alby_with_citadel_connection_suport_beta-1.2.1-fx.xpi

bumi:
That's great! I am wondering how to best tell the user that that connection only works in the home network. Or do you think this is not a problem? 

I am happy to merge this and get it in one of the next releases.

AaronDewes:
> Or do you think this is not a problem?

I think most users should know that, they enter the URL of their node there and they should know it only works on the local network.

AaronDewes:
When this is merged, don't consider Citadel as a blocker for some feature, if you need anything currently missing from the SDK, I can implement it.

AaronDewes:
If you want to add Umbrel connection which connects to LND directly, but has login with the Umbrel password, it should work like this (not tested, neither with Umbrel or Citadel, but should work with both):

```TypeScript
type lndConnectData = {
        restTor,
        restLocal,
        grpcTor,
        grpcLocal
};
const password = "YOURPASSWORD";

const data = await fetch("http://umbrel.local/manager-api/v1/account/login", {
    body: JSON.stringify({password}),
    method: "POST",
});
const text = await data.text();
let jwt;
try {
    jwt = JSON.parse(text).jwt;
} catch {
    throw new Error(text);
}

const lndData = await fetch("http://umbrel.local/manager-api/v1/system/lndconnect-urls", {
    headers: {
        Authorization: `JWT ${jwt}`,
    }
});

const finalData: lndConnectData = await lndData.json();
```

Every element of finalData is a lndconnect string. You can parse such strings using https://www.npmjs.com/package/lndconnect or https://www.npmjs.com/package/@runcitadel/lndconnect (My TypeScript fork). Both utilize certain node features though, so they may not work properly in the browser, but maybe the code from them can help you.

bumi:
Thanks for that. The UX of this flow is super cool! 

How do you handle the self-signed certificate? because by default we there have problems in the browser.

AaronDewes:
> How do you handle the self-signed certificate? because by default we there have problems in the browser.

Citadel currently doesn't use HTTPS, we use HTTP. A future update will add SSL by requiring the user to create and install the custom certificate on their computer.

AaronDewes:
As communication only happens on the local network, it's not such a huge security issue, but we'll add HTTPS soon anyway.

bumi:
so the SDK proxies requests through HTTP to the local LND? (meaning LND is not directly accessed, correct?)

AaronDewes:
> so the SDK proxies requests through HTTP to the local LND? (meaning LND is not directly accessed, correct?)

Yeah, it goes to the Citadel API. This means that the only thing needed for login is the user's password, nothing else. It will also ontinue to work when Citadel allows users to switch to c-lightning and when we have multi-user support.

AaronDewes:
The last commit isn't well tested yet, it should work, but I completely changed the way the SDK is bundled and that could've broken things.

AaronDewes:
Tested & published at https://github.com/AaronDewes/lightning-browser-extension/releases/download/beta-2/alby_with_citadel_connection_suport_beta-1.2.2-fx.xpi

AaronDewes:
It works (except native connector, that is untested) now, so I think it's ready to be merged.

AaronDewes:
Conflicts are fixed now.. Plz merge for the next release

yarn.lock
```diff
@@ -16526,6 +16533,11 @@ unbox-primitive@^1.0.1:
     has-symbols "^1.0.2"
     which-boxed-primitive "^1.0.2"
 
+undici@^4.14.1:
```
bumi, line 16536: I am always worried about additional dependencies, but good choice that it does not have more dependencies! :)

bumi (approved):

yarn.lock
AaronDewes: It won't actually be bundled with Alby and only used in Node.js environments. Once Node 18 is out, we'll soon drop the undici dependency completely because node will have the fetch API natively.

bumi:
Thanks for adding this! great to see that you could also adopt the Tor support. looking forward to learn how that works for users.

PR #481, opened 2021-12-23
Fix all warnings

dylancom:
Most warnings are due to using "any".
Using "any" defeats the purpose of using typescript. Instead when not possible otherwise, use "unknown". Which has stricter rules than "any".

Introduces a few new concepts such as: "Generics", which makes it possible to make for e.g. a call to utils.call() typed. (By providing a type between < >.
This works the same in libs such as axios --> `axios.get<ExpectedResponseType>("https://......")`
https://www.typescriptlang.org/docs/handbook/2/generics.html


src/app/context/AuthContext.tsx
```diff
@@ -35,27 +36,31 @@ export function AuthProvider({ children }: { children: React.ReactNode }) {
...
+      .getStatus()
       .then((response) => {
         if (!response.configured) {
           utils.openPage("welcome.html");
           window.close();
         } else if (response.unlocked) {
-          setAccount(response.currentAccountId);
-          getAccountInfo(response.currentAccountId);
+          setAccount({ id: response.currentAccountId });
+          api.getAccountInfo().then((response) => {
```
bumi, line 59: we need to introduce caching, I think we do this account info call twice now.
(independent of this PR)

bumi (approved):

PR #485, opened 2021-12-27
Add support for a connector to a native companion app

bumi:
Using the [native messaging](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging) API we can connect to a native app running on the host computer. This allows us to do Tor connections from that native app and thus make it easier for Alby to connect to nodes behind Tor. 

Because it takes a while until Tor is started we need to keep the app running and a "connection" to the native app open. 

### ToDo: 

- [x] write proper TypeScript
- [x] DRY up native connectors
- [x] dynamically choose the connector for LND: `nativelnd` vs. `lnd` depending on the provided address
- [x] dynamically choose the connector for LNDHub: `nativelndhub` vs. `lnhub` depending on the provided URL
- [x] can we somehow validate that the companion app is installed? (try to connect and if it does not work show an error)
- [ ] add explanation on how to install the native companion app

#### Choosing the connector

We should check the entered address and if it is an .onion address we should choose the `nativelnd`/`nativelndhub` connector and also check if the native companion app is available. If not we should provide an error message with a download link of the installer.

For LND we should also use the `nativelnd` connector for clearnet IP addresses and also ask the user for a certificate.


bumi:
So I think for choosing the connector we should check the address and if the address contains .onion or is https:// + an IP address (no domain) then we should show a message that the Alby native app needs to be installed. With a link to the download depending on the operating system.

PR #490, opened 2021-12-27
feat: websites where you can use alby

umarrg:


bumi:
thanks a lot for the PR. 
I still need to review it in detail, but I am wondering if we can somehow make it simple. 
It's quite a lot of code for showing those boxes. Maybe we could simple list the boxes... so we have one component and one array we iterate over?

bumi:
I wanted to follow up on this. do you think we can make this change a bit simpler/smaller?

umarrg:
Ok

bumi:
thanks for your work on this. I think it is great when we get this one in soon. 

But I still think we should simplify this and only show a nice list of websites without the need of adding so much new components and colors etc. 
just a "stipid" list. do you see a problem with that?

umarrg:
Ok

bumi:
@umarrg shall we finalize this? what do we need?

umarrg:
Pls let me know if there's anything else

escapedcat:
It's a quite a big PR and hard to comprehend for an outsider like me.

Would be super helpful if this PR had a description.
Like what has been changed and why. Even better with screenshots before/after.
Maybe there's a related issue? If yes would be great to link it.
This would help people to understand the intention of this PR.

Maybe these things have been discussed somewhere than it would be great to add that discussion into the description.

Thanks for you efforts! ❤️

escapedcat (approved):

bumi:
Currently this loads quite some images from twitter and other sources. I think we should avoid external sources and upload those images to our storage. Then we can be sure that those images not simply change or disappear and also users don't access external resources by default (but only trusted alby resources) 

@MoritzKa can you upload those images to our DO storage. 
@kiwiidb I think the DO storage does not have a custom domain. we should setup something like `static.getalby.com` or `assets.getalby.com` for this. can you configure that? (or does this needs to be done directly as DNS records?)

bumi (changes requested):
Currently this loads quite some images from twitter and other sources. I think we should avoid external sources and upload those images to our storage. Then we can be sure that those images not simply change or disappear and also users don't access external resources by default (but only trusted alby resources)

@MoritzKa can you upload those images to our DO storage.
@kiwiidb I think the DO storage does not have a custom domain. we should setup something like static.getalby.com or assets.getalby.com for this. can you configure that? (or does this needs to be done directly as DNS records?)

MoritzKa:
I can do that. Is there any specific image size required or can I go with the size of the image that we use right now?
I just need a place to upload them.

bumi:
> I can do that. Is there any specific image size required or can I go with the size of the image that we use right now?
I just need a place to upload them.

They should all have the same format and size with a reasonable resolution. also check the file-size that it is not more than a low 100s of KB, metadata should be removed for PNGs. but probably SVG versions are the best. 
@dylancom do you have suggestions for that, maybe. - I'd have to do some research.

You still have that DO storage that you used for makers.bolt.fun? let's upload it there to a new folder.

bumi:
we should use the static.getalby.com domain instead of the digital ocean domain directly. 
@kiwiidb what do I need to do to set it up?

bumi:
@MoritzKa what about?
* makers.bolt.fun
* geyser.fund

MoritzKa:
Yes. I will add them once we figured out where to store the logos. 
Let's get it into the next release. We can always add more websites later on.

bumi:
no, the URL must be an alby URL, before that I will not merge it. 
logos will be stored on DO as you uploaded them. 

just the URL will change.

src/app/screens/Publishers/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    fetchData();
+  }, []);
+
+  return (
+    <Container>
+      <h2 className="mt-12 mb-2 text-2xl font-bold dark:text-white">
+        Your ⚡️ Websites
+      </h2>
+      <p className="mb-6 text-gray-500">
+        Websites where you have used Alby before
```
bumi, line 53:
```suggestion
        Some websites where you have used Alby before
```

should we indicate here that it is only a selection?

src/app/screens/Publishers/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </h2>
+      <p className="mb-6 text-gray-500">Websites where you can use Alby</p>
+      <div className="mb-12">
+        {websites.map(({ title, items }) => (
+          <div className="mb-6" key={title}>
+            <h4 className="mb-4 text-xl font-bold">{title}</h4>
+            <div className="grid gap-4 md:grid-cols-4">
+              {items.map(({ title, subtitle, logo, url }) => (
+                <a key={url} href={url} target="_blank" rel="noreferrer">
+                  <div className="bg-white shadow-md flex space-x-3 h-44 px-4 rounded-lg pt-7 hover:bg-gray-50 cursor-pointer w-full">
```
bumi, line 74: is `h-32` or something maybe enough? I feel like the boxes are pretty big. 
Every description seems to fit except the lightning store one - but there we can shorten the description.

bumi (commented):
@MoritzKa we now have `https://cdn.getalby-assets.com/` as domain.

src/app/screens/Publishers/index.tsx
dylancom: I'm not too fond of "Some websites".

Issue #148, opened 2021-07-28
Display auto login option for publishers


bumi:
The publisher list and detail page and popup should show of auto login with lnurl-auth is enabled or not.

Similar to updating the budget it should also be possible to enable/disable the auto login from the publishers edit form.

<img width="1171" alt="image" src="https://user-images.githubusercontent.com/318/142949835-795aeee3-ceb0-4b0c-a21e-42b841f6a709.png">
<img width="726" alt="image" src="https://user-images.githubusercontent.com/318/142949892-c313b669-8c9d-4b68-a2f1-fad8587c7210.png">

secondl1ght:
I could work on the frontend for this if someone wants to do the backend?

bumi:
yeah, let's do it. 
I think the `lnurlAuth` attribute is already part of the allowance data.

I can then update the action that it stores the update if you add it to the form.

PR #495, opened 2021-12-28
Add lnurl-auth input checkbox to allowance menu

secondl1ght:
Linked to #148, I have started this PR but it still needs work. Me and @bumi are working on this together. I have added the input checkbox element to the allowances menus. I have just done this initial work so far and haven't looked at hooking it up yet. It will be something like ```<input type="checkbox" id="lnurl-auth" name="lnurl-auth" checked={lnurlAuth ? checked : ""}/>```. And we will have to add an onclick to change the ```lnurlAuth``` which I think is a ```boolean```. Thanks for helping with this bumi!

Screenshot:

image

image

### DIFF
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -14,7 +14,7 @@ type Props = {
   allowance: {
     id: string;
     totalBudget: number;
-		lnurlAuth: boolean;
+    lnurlAuth: boolean;
   };
   onEdit?: () => void;
   onDelete?: () => void;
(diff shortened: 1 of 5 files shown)


dylancom:
@secondl1ght We have a <Checkbox /> component in the Form folder that you can use.

secondl1ght:
Awesome thanks, I've updated this to use the nice checkbox component with Alby styles.

bumi:
wanna try to add the logic to save that value? 
you can actually just hook into how the saving of the budget is currently done.

secondl1ght:
Ok I gave it a try, I am still getting a type error but I just pushed it anyway so that you could see it and I can fix that later if I am on the right track. Just let me know if what I did here is correct or not, I tried to follow what was already set up for the budget. But it is likely that there is still some work to be done here (maybe I am close?). Thanks!

bumi:
I think @secondl1ght you could pick this up again and try if you now get it working

escapedcat:
Closing in favour of #1539

PR #496, opened 2021-12-28
Load account info after unlocking

bumi:
this asynchronously loads the balance and account info after unlocking the accounts.

This fixes a bug that the balance is not shown directly after unlocking the popup.


bumi:
@dylancom why is this line 57 needed: https://github.com/getAlby/lightning-browser-extension/pull/496/files#diff-746a7eac0422151ff7dbb6d284688b1e54131a065e8953ae8df7c9c7f2839f62R57

dylancom:
@bumi that was added so if you switch account, you'll see the skeleton loading animation when alias & balance aren't present. Could handle it differently if desired.

PR #500, opened 2021-12-30
Add dark mode to test connection page

secondl1ght:
Discovered this page needs dark mode styling - had not seen it before.


bumi:
So yeah, this is the problem I have with such changes. we will miss things and break things.
it looks simple first but it just adds so much complexity and code that we have to maintain. 

Can you please use every part of the extension and review it again. Every page for every connection etc.

PR #434, opened 2021-11-30
Add connector to Galoy wallets

bumi:
PR by @vindard 
This allows Alby to connect to [Galoy](https://galoy.io/) wallets.
(original branch: https://github.com/vindard/lightning-browser-extension/tree/feature/galoy-connector)

PR #505, opened 2022-01-01
Add commits to PR #434 branch (Galoy connector)

vindard:
### Description

I added some improvements to the error handling based on some testing and I implemented the `checkPayment` interface for Galoy connector.

Question: How do you all currently circle back to updating the status on pending payments? I added a conditional branch for when our `sendPayment` status comes back as `pending` but I believe my change still puts the pending transaction into the txn list in Alby.

Also, where does `checkPayment` currently get used?


src/app/screens/Onboard/ConnectGaloy/index.tsx
```diff
@@ -51,20 +51,22 @@ export default function ConnectGaloy() {
     };
 
     try {
-      const { data } = await axios.post<any>(url, query, {
```
vindard, line 54: I had to remove these `<any>` types because otherwise it wouldn't compile

src/extension/background-script/connectors/galoy.ts
```diff
@@ -108,36 +110,157 @@ class Galoy implements Connector {
...
+          return {
+            error: "Invoice was already paid.",
+          };
+        case "FAILURE":
+          return {
+            error: "Payment failed.",
+          };
+        case "PENDING":
+          return {
+            data: {
```
vindard, line 139: Not sure what the implications of returning a valid object on pending payment are here

src/extension/background-script/connectors/galoy.ts
```diff
@@ -108,36 +110,157 @@ class Galoy implements Connector {
...
   async checkPayment(args: CheckPaymentArgs): Promise<CheckPaymentResponse> {
-    return {
-      data: {
-        paid: false,
-        preimage: "",
-      },
+    const TRANSACTIONS_PER_PAGE = 20;
+
+    const query = {
+      query: `
```
vindard, line 161: We probably need an easier way to query single transactions than this 🤕  (cc @samerbuna @dolcalmi @daviroo)

bumi:
`checkPayment` is used on the receive screen.  it periodically checks if the payment was received to show an success screen.

vindard:
Converted back to draft because I'd like to add a helper method for all that graphql error handling

vindard:
@bumi this should be ready to merge in the open #434 branch now.

I improved the error handling and brought in `master` to resolve merge conflicts. The commit diff looks like a lot here but I've arranged the commits so that it'll resolve itself once we merge this into the PR #434 branch (final diff looks [like this](https://github.com/getAlby/lightning-browser-extension/compare/master...vindard:feature/galoy-connector))

PR #507, opened 2022-01-05
add eclair connector

fiatjaf:


kiwiidb (changes requested):
Would it be possible to also add the connector to the `options/ChooseConnector/index.jsx` screen? Right now you only added it to the "onboarding" ChooseConnector screen.

bumi:
yeah, we sadly got some slight code duplication here. it is kinda the same but not 100%. (needs refactoring at some point)

dylancom:
Just refactored this in: #516 (So we only have 1 < ChooseConnector /> component and define the connector routes in 1 place).

secondl1ght:
Testing: I have tested this and was able to successfully connect an eclair node to Alby. Nice work @kiwiidb 👍

fiatjaf:
Thank you for finishing this. I was going to get to it someday.

bumi:
thanks again @fiatjaf for adding this. will go online in the next release

bumi (approved):

PR #516, opened 2022-01-11
refactor: shared connector routes (DRY)

dylancom:
Currently it's a bit of a struggle to add new connectors. We have two ChooseConnector screens and have to define the routes in the Welcome and Options Router.

This PR makes this smarter and makes use of a shared connectorRoutes array and re-uses 1 ChooseConnector compontent.


bumi:
YOLO merging and testing in master. :)

PR #519, opened 2022-01-14
feat: locally cache accounts

dylancom:


bumi:
thinking out loud and wondering if we can somehow abstract that in the `api`. ?

calling `api.cached.getAccountInfo()` then we have all the caching logic in the same place, like the general api calls.

bumi:
hmm. adding it to the the `api` would probably require a change in the function signatures. and would not work with a simple Promise... just because we would first resolve with the cached data, but then how will the data be updated in the component.

bumi:
we could use a callback which then is called twice. once with the cached data, and once with the updated data?
```
api.cached.getAccountInfo(callback)
```

dylancom:
I would prefer to keep the caching and fetching of fresh data separate (than to include it in getAccountInfo and provide a callback) but definitely agree on abstracting it into an api.
So I'm thinking more like we would have a `api.cache.get("accounts")` and `api.cache.set("accounts", data}` or: `api.cache.getAccounts` / `api.cache.getAccount(id)`

bumi:
why would you like to keep it separate? I was thinking that `getAccountInfo()` knows best how to cache its data. 
and wherever we use it we only need to know if we must get fresh data or if it can be cached (and refreshed)

dylancom:
I kind of liked the idea to keep a separate api for our own caching mechanisms / keep getAccountInfo really simple. And for code readibility you can easily see we first show stale data, then fresh data and update our cache with it.

But I also get your point of view that getAccountInfo knows best how to cache it's data.

bumi:
I'd also keep a separate API - I would not add it to `getAccountInfo` itself. but provide a function that uses `getAccountInfo` and does the caching. 

right now we would need to replicate that code whenever we want to use a cached version of `getAccountInfo`.

Regarding readability: I agree. isn't this a bit of a naming problem? if the calls are descriptive enough it becomes obvious. 


so I'd just move the storage setter/getter code to a "lower" level

bumi:
if we make the `callback` optional then the API could be exactly the same returning a promise and we could resolve whatever we find first?

maybe too much, but came to my mind when looking at your latest PR.

PR #521, opened 2022-01-14
Add bitcast peertube instance and load lightning data from support fields 

bumi:


bmac21M:
Thank you for the update and keep up the great work @bumi !

PR #524, opened 2022-01-16
Update regex to allow for whitespace before ln address

### LINKED ISSUE #520: In the enhancements support a whitespace
currently we use the regex to find lightning information: 
`text.match(/(⚡️:?|lightning:|lnurl:)(\S+@\S+)/i)`

we should allow one single whitespace between the key (`(⚡️:?|lightning:|lnurl:)`) and the lightning address.

users tend to add a whitespace there.

secondl1ght:
Closes #520, this was my first time trying regex so I am not 100% sure that this is correct. If it is wrong please let me know and I will go back and learn more about regex.


bumi:
here are some tools to inspect/debug regex: 
https://www.debuggex.com/
https://rubular.com/


I think I'd do something like: `(⚡️:?|lightning:|lnurl:)\s?(\S+@\S+)` to not have the whitespace part of the match group and only have the lightning address in there. 

and I think there are some more places where this is used. can you look for them also

bumi:
and don't forget to look for other occurrences of the regex

PR #530, opened 2022-01-18
feat: add lnurl-withdraw support

dylancom:
Tested with Kollider: https://pro.kollider.xyz/ & https://stacker.news/

### DIFF
--- a/src/app/screens/LNURLWithdraw.tsx
+++ b/src/app/screens/LNURLWithdraw.tsx
@@ -22,7 +22,7 @@ function LNURLWithdraw(props: Props) {
   const [origin] = useState(props.origin || getOriginData());
   const { minWithdrawable, maxWithdrawable } = props.details;
   const [valueMSat, setValueMSat] = useState<number | undefined>(
-    minWithdrawable
+    maxWithdrawable
   );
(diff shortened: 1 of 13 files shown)


kiwiidb:
Is this PR still necessary? I can already withdraw from Kollider and Stacker News, or is that using webLN?

dylancom:
At Kollider, from the wallet page it's using WebLN. Try to withdraw in the bottom left @ https://pro.kollider.xyz/

src/app/screens/LNURLWithdraw.tsx
```diff
@@ -0,0 +1,147 @@
+import { useState, MouseEvent } from "react";
+import axios from "axios";
+
+import { LNURLWithdrawServiceResponse } from "../../types";
```
escapedcat, line 4: This is only used as a type, right? Can `import type ...` be used?

dylancom: Correct!

bitkarrot:
Any ETA on when this pull request will get merged? I'd like to make use of the feature. Thanks much in advance! Thanks @dylancom @escapedcat

bumi (approved):
utACK

Issue #529, opened 2022-01-18
Re-think or replace slider


kiwiidb:
I don’t find the slider for choosing the amount very useful. It's not a great tool for numerical input and it's wildly inaccurate.
For example, in an LN address payload there is a minimum and a maximum field, but these can differ greatly. If I touch the slider on my own twitter profile even slightly, the amount jumps from 10 sats to >10000 sats immediately, which is not something most people would want. What I would find more useful is some buttons with sane pre-filled amounts.

secondl1ght:
Do we have an idea on what we would like the pre-filled amounts to be for the buttons?

secondl1ght:
Making a note that in the future we want to add click and hold functionality to add the amounts for the buttons, for now we will just have a single click function to set the input.

secondl1ght:
Should I close this now that the PR is merged and open a new issue to add the click and hold function?

dylancom:
Yes, this can be closed. We can add that as a "nice-to-have" or something I think, so we can currently play with this and get some ideas to improve on it.

secondl1ght:
Closed by #539

PR #539, opened 2022-01-20
Replace slider with sat amount buttons when sending

secondl1ght:
As requested in #529, I have removed the slider on the sending sats screen. I replaced it with 4 buttons with pre-determined satoshi amounts for each: 10, 100, 1000 and 10000. For the functionality I made it so that when you click each button it adds the satoshi amount to the input, I thought this may be useful rather than only allowing for the number to be set once.

What does everyone think about this?

Screenshot:
image

I think there is another screen I will have to adjust but I thought I would get feedback on this first and if everyone likes it then I can update the other one as well.

### DIFF
--- a/src/app/screens/LNURLPay.tsx
+++ b/src/app/screens/LNURLPay.tsx
@@ -47,7 +47,7 @@ function LNURLPay(props: Props) {
   const [successAction, setSuccessAction] = useState<
     LNURLPaymentSuccessAction | undefined
   >();
-
+  const [clear, setClear] = useState<boolean>(false);
   useEffect(() => {
     if (searchParams) {
       // lnurl was passed as querystring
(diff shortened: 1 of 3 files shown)


dylancom:
Schermafbeelding 2022-01-20 om 09 05 19
Some inspiration from bolt.fun

bumi:
that's actually great. I like it!! 
best to keep it in one line probably

src/app/screens/LNURLPay.tsx
```diff
@@ -47,7 +47,7 @@ function LNURLPay(props: Props) {
   const [successAction, setSuccessAction] = useState<
     LNURLPaymentSuccessAction | undefined
   >();
-
+  const [clear, setClear] = useState<boolean>(false);
```
bumi, line 50: is this needed? 
if we make sure that `valueMSat` is never undefined but always a integer then we
can simple always do `setValueMSat(valueMSat + 10000);` I guess?

secondl1ght: The reason I added this was because currently `valueMSat` starts off with a value of the minimum sendable. So for me when I was testing there was already a value of 1 in the input. This made it so that when I clicked the buttons instead of going up in even numbers like 10 or 1000 it would do 11, 1001 etc. So I thought it would be cleaner to clear what is the value at the start so that when you use the buttons it reacts the way you would expect. Does this make sense?

secondl1ght:
What do you guys think of this?

secondl1ght:
I was also thinking we could make these buttons light up on click/active to a lightning yellow color so when you click them it looks like there is lightning striking? ⚡⚡

dylancom:
I agree with Bumi to keep it on 1 line. No need for 2 rows.

secondl1ght:
Ok sounds good, I didn't think I'd be able to fit everything on 1 row but it looks like we can:
image

dylancom:
Then I would remove the gap on the  left of the first button and the right of the last button so it aligns nicely with the amount field (just like the bolt.fun example)

secondl1ght:
How's this?
image

dylancom:
Much better! Bonus points if the horizontal spacing between the buttons is the same as the vertical spacing between the amount field and the buttons (i'd say).

secondl1ght:
Woot!
image

secondl1ght:
Do you know where the other slider is in the app? I see it in the code but I'm not sure how to navigate to it on Alby...

secondl1ght:
Latest screenshot:
image

secondl1ght:
Ok I have updated the functionality to only set the input one time, I also updated the values of the buttons.
image

secondl1ght:
> instead of a lessPadding prop we could opt for a noPadding prop, and let flex stretch the buttons to available space.

Here's how it looks with no padding, I think it looks better with padding personally:
image

Or do I have something wrong with flex and it is not filling the width properly?

dylancom:
You should remove the extra div's around the buttons. Add the fullWidth props to the buttons and remove "flex-wrap" from the parent wrapper.

secondl1ght:
Oh right, I had those extra divs in there earlier to apply styling but wasn't using them anymore, good call. Here is how it looks now:
image

To add X margin should I create another prop on the Button component called marginX?

secondl1ght:
Yeah! I forgot about that class, I have to remember all these useful tricks.

secondl1ght:
Ok I have also updated the slider on the make invoice screen as well:
image

For this one I did add the validation for the min/max because that is part of the invoice.

I think this PR is now ready for final review and then merging if all looks good.

dylancom:
Schermafbeelding 2022-01-24 om 18 16 18

Tigerix:
Suggestion for the amount buttons:
Clicking the 100 Sats button 3 times in a row, should result in 300 Sats!
Clicking it 10 times, should result in 100 Sats again.

PR #541, opened 2022-01-20
feature/lud 18 email support

AustinKelsay:


src/app/screens/LNURLPay.tsx
```diff
@@ -73,16 +74,34 @@ function LNURLPay(props: Props) {
...
+        setUserEmail(response.userEmail);
+      }
     });
   }, []);
 
+  const getPayerData = (details: LNURLPayServiceResponse) => {
+    if (
+      userName &&
+      userEmail &&
+      userName.length &&
```
escapedcat, line 87: Does i.e. `userName?.length` get rid of the first two entries?  
Same question for situations further down.

AustinKelsay: Yeah good point I suppose just having userName.length and userEmail.length would accomplish the same check

AustinKelsay:
Hey sorry @escapedcat was just a bit confused with that syntax initially but I see what you're saying now. Just updated the variable checks with your suggested changes and retested!

bumi (approved):

PR #542, opened 2022-01-20
refactor: improved state.ts

dylancom:
Note: this doesn't import the whole lodash lib, just the actual imported utility functions.

merge: deep merges objects.
pick: creates a subset from an object.


bumi:
ok, good to know it is not loading the whole lodash.. would be a bit much ;)
we need to keep dependencies to a minimum

PR #549, opened 2022-01-24
feat: update balance info after payment

dylancom:
Currently after sending a payment with webln or lnurl pay, the balance isn't directly updated.

Current:
Scenario: Tipping on Twitter -> Balance stays the same until reopening the popup.
Scenario: Deposit sats on Kollider -> Prompt appears to make a payment, after payment, the prompt closes. When opening the popup, it first shows the old cached balance and then starts to fetch the new balance.

Desired:
Scenario: Tipping on Twitter -> Balance directly updates after a successful payment.
Scenario: Deposit sats on Kollider -> Prompt appears to make a payment, after payment, the prompt closes. When opening the popup, it directly shows the new balance.

### DIFF
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -30,7 +30,7 @@ function AccountMenu() {
     await utils.call("selectAccount", {
       id: accountId,
     });
-    auth.getAccountInfo(accountId);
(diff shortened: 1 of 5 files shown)


bumi:
this is great! once the conflict is fixed we release this! :shipit: 🚀

bumi (approved):

PR #550, opened 2022-01-24
refactor: optionally ask for native messaging permission

### LINKED ISSUE #532: Ask for native messaging permission on runtime
We need the native messaging permission only when the companion app is used. 
We could define this as an optional permission and request that permission on runtime when it is needed:

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/optional_permissions
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request

dylancom:
Fixes #532

### DIFF
--- a/src/app/screens/Onboard/ConnectLnd/index.tsx
+++ b/src/app/screens/Onboard/ConnectLnd/index.tsx
@@ -53,7 +53,15 @@ export default function ConnectLnd() {
       // TODO: for native connectors we currently skip the validation because it is too slow (booting up Tor etc.)
       if (account.connector === "nativelnd") {
         validation = { valid: true, error: "" };
-        browser.permissions.request({ permissions: ["nativeMessaging"] });
+        const permissionGranted = await browser.permissions.request({
+          permissions: ["nativeMessaging"],
+        });
+        if (!permissionGranted) {
+          validation = {
+            valid: false,
+            error: "Native permissions are required to connect through Tor.",
+          };
+        }
       } else {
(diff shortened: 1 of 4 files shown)


src/app/screens/Onboard/ConnectLnd/index.tsx
```diff
@@ -52,6 +53,7 @@ export default function ConnectLnd() {
       // TODO: for native connectors we currently skip the validation because it is too slow (booting up Tor etc.)
       if (account.connector === "nativelnd") {
         validation = { valid: true, error: "" };
+        browser.permissions.request({ permissions: ["nativeMessaging"] });
```
bumi, line 56: this returns a promise with true/false if the permissions were granted or not. 
I think we should fail if the user does not provide the permissions.

bumi (changes requested):
nice. thanks!

src/app/screens/Onboard/ConnectLnd/index.tsx
bumi: maybe something:
```js
browser.permissions.request({ permissions: ["nativeMessaging"] }).then(hasPermissionGranted => { 
  if (hasPermissionGranted) { 
    validation = { valid: true, error: "" };
  } else {
    validation = { valid: false, error: "Native permissions are required to connect through Tor." };
  }
});
```

bumi (approved):

PR #555, opened 2022-01-24
feat: handle invalid invoices

### LINKED ISSUE #547: Handle invalid lightning addresses
Currently when I enter an invalid lightning address into the Send input, Alby will crash with no indication of what happened. Maybe we should have a message appear saying something like 'invalid lightning address'?

dylancom:
Fixes #547

### DIFF
--- a/src/app/screens/ConfirmPayment/index.jsx
+++ b/src/app/screens/ConfirmPayment/index.jsx
@@ -1,7 +1,6 @@
 import { useState, useRef } from "react";
 import { Transition } from "@headlessui/react";
 import { useSearchParams, useNavigate } from "react-router-dom";
-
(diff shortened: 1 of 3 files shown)


bumi:
@secondl1ght can you test this?

secondl1ght:
What is the expected behavior for this? Right now when I enter 'secondl1ght@getalby.comm' I do not get an error message, but I think we are supposed to get an error for this right?

secondl1ght:
When I enter 'asdfkjhsdf' it does give an error though which is good.

dylancom:
Now it also gives an error message on: secondl1ght@getalby.comm

secondl1ght:
Ok I tested this and it's working now.

PR #558, opened 2022-01-24
Change grid to allow 5 connectors per row

secondl1ght:
@bumi wanted to have all on one row instead of having 1 connector by itself. 

image

### DIFF
--- a/src/app/screens/Onboard/ChooseConnector/index.tsx
+++ b/src/app/screens/Onboard/ChooseConnector/index.tsx
@@ -56,7 +56,7 @@ export default function ChooseConnector({ title, description }: Props) {
             </p>
           )}
         </div>
-        <div className="grid grid-cols-4 gap-4">
+        <div className="grid grid-cols-5 gap-5">
           {connectors.map(({ to, title, description, logo }) => (
             <LinkButton
               key={to}


bumi:
whoot. you're fast. 
@johnsBeharry are you OK with that. and then we think what do do once we have Galoy in?

johnsBeharry:
We won't have Galoy, as that's just the tech from what I understand. So it will be BitcoinBeach, and I think [Bitcoin Jungle](https://bitcoinjungle.app). @vindard is that correct? OR do those two apps share a database?

Either way I reckon that list will continue to grow, and I'd opt we go for a more App Store style card. Similar to what Auwal did

Screenshot 2022-01-25 at 1 07 12 AM

bumi:
Check out the #434 branch. It is a Galoy connector.

I like the app store style cards...those can hold enough information.

bumi:
but are you OK with this PR for this release?

johnsBeharry (approved):
LGTM

vindard:
> So it will be BitcoinBeach, and I think [Bitcoin Jungle](https://bitcoinjungle.app)

I guess tbd because the only thing that separates these two is the api link we point at.

secondl1ght:
Should we create a new issue to track the connector re-design?

bumi:
:+1: please do.

PR #560, opened 2022-01-25
refactor: use html input correctly

dylancom:
Inputs are currently using mixed types of number / undefined.
We should handle everything as a string (during input handling) and convert to number in the appropriate places (api's)

Results:
No more clumsy onChange handlers / falling back on undefined when empty.

### DIFF
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -20,10 +20,10 @@ type Props = {
 
 function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
   const [modalIsOpen, setIsOpen] = useState(false);
-  const [budget, setBudget] = useState<number | undefined>(0);
+  const [budget, setBudget] = useState("0");
 
   function openModal() {
-    setBudget(allowance.totalBudget);
+    setBudget(allowance.totalBudget.toString());
     setIsOpen(true);
   }
 
@@ -34,7 +34,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
   async function updateAllowance() {
     await utils.call("updateAllowance", {
       id: parseInt(allowance.id),
-      totalBudget: budget,
+      totalBudget: parseInt(budget),
     });
     onEdit && onEdit();
     closeModal();
@@ -90,13 +90,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
               placeholder="sat"
               value={budget}
               type="number"
-              onChange={(event) => {
-                setBudget(
-                  !isNaN(event.target.valueAsNumber)
-                    ? event.target.valueAsNumber
-                    : undefined
-                );
-              }}
+              onChange={(e) => setBudget(e.target.value)}
             />
           </div>
         </div>
@@ -105,7 +99,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
             onClick={updateAllowance}
             label="Save"
             primary
-            disabled={budget === undefined}
+            disabled={budget === ""}
           />
         </div>
       </Modal>
--- a/src/app/screens/LNURLPay.tsx
+++ b/src/app/screens/LNURLPay.tsx
@@ -40,11 +40,11 @@ function LNURLPay(props: Props) {
         JSON.parse(searchParams.get("origin") as string)) ||
       getOriginData()
   );
-  const [valueMSat, setValueMSat] = useState<number | undefined>(
-    details?.minSendable
+  const [valueSat, setValueSat] = useState(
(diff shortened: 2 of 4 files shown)


src/app/screens/MakeInvoice.tsx
```diff
@@ -32,32 +32,16 @@ function MakeInvoice({ invoiceAttributes, origin }: Props) {
   );
   const [error, setError] = useState("");
 
-  function handleValueChange(e: React.ChangeEvent<HTMLInputElement>) {
+  function handleValueChange(amount: string) {
```
bumi, line 35: in this one we call `setValue` now with a string instrad of a number. is this fine?

src/app/screens/LNURLPay.tsx
```diff
@@ -58,7 +58,7 @@ function LNURLPay(props: Props) {
         lnurl.getDetails(lnurlString).then((lnurlDetails) => {
           if (lnurlDetails.tag === "payRequest") {
             setDetails(lnurlDetails);
-            setValueMSat(lnurlDetails.minSendable);
+            setValueSat((lnurlDetails.minSendable / 1000).toString());
```
bumi, line 61: is `lnurlDetails.minSendable` always defined? what if the response does not include it?

src/app/screens/LNURLPay.tsx
```diff
@@ -40,11 +40,11 @@ function LNURLPay(props: Props) {
         JSON.parse(searchParams.get("origin") as string)) ||
       getOriginData()
   );
-  const [valueMSat, setValueMSat] = useState<number | undefined>(
-    details?.minSendable
+  const [valueSat, setValueSat] = useState(
+    (details?.minSendable && (details?.minSendable / 1000).toString()) || ""
```
bumi, line 44: also here isn't `details?.minSendable` something from the JSON response. what if it is a string?

src/app/screens/MakeInvoice.tsx
dylancom: e.target.value is always a string, no matter what you pass into an input.

dylancom: Yes, utils.call("makeInvoice"), which will use value as an argument, can handle both strings / numbers.

src/app/screens/LNURLPay.tsx
dylancom: Currently in our interface we haven't defined them as optional, as the spec doesn't speak about this as far as I'm aware? https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md

src/app/screens/LNURLPay.tsx
dylancom: According to the spec it should be a number?

bumi: it should but it is still an API response and people to strange things there :)

src/app/screens/LNURLPay.tsx
bumi: ok.thx

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -34,7 +34,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
   async function updateAllowance() {
     await utils.call("updateAllowance", {
       id: parseInt(allowance.id),
-      totalBudget: budget,
+      totalBudget: parseInt(budget),
```
escapedcat, line 37: Is it still a thing to add the radix? I remember not adding radix used to cause issues in certain situations:  
https://eslint.org/docs/rules/radix

src/app/screens/LNURLPay.tsx
```diff
@@ -40,11 +40,11 @@ function LNURLPay(props: Props) {
         JSON.parse(searchParams.get("origin") as string)) ||
       getOriginData()
   );
-  const [valueMSat, setValueMSat] = useState<number | undefined>(
-    details?.minSendable
+  const [valueSat, setValueSat] = useState(
+    (details?.minSendable && (+details?.minSendable / 1000).toString()) || ""
```
escapedcat, line 44: Here you're using `+` instead of `parseInt`. Would one way work everywhere?

src/app/components/AllowanceMenu/index.tsx
dylancom: Our ECMAScript target version is ES6 so this shouldn't be a problem.

src/app/screens/LNURLPay.tsx
dylancom: Here we treat it as raw input and just coerce to number (if it was a string), leaving decimals intact. parseInt always expects a string which in this case could be a number already.

Issue #477, opened 2021-12-21
Tipping feature for Vimeo videos 


bumi:
Similar to the tipping feature on twitter we should also support tipping on Vimeo.

vimeo seems to have a `window.vimeo.clip_page_config` object that holds the needed data. (But we might need to wait a bit until this variable is available for us. )

PR #563, opened 2022-01-26
add vimeo tips

pseudozach:
add tips on vimeo: here's demo video: https://vimeo.com/670055676

also fixed a strange bug that was adding these characters while parsing lightningaddress to lnurlp url `https://lntxbot.com/.well-known/lnurlp/%EF%B8%8Fpseudozach`. this was causing lnurlp tips on youtube/vimeo to fail silently.

### DIFF
--- a/src/common/lib/lnurl.ts
+++ b/src/common/lib/lnurl.ts
@@ -13,7 +13,10 @@ const fromInternetIdentifier = (address: string) => {
       /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
     )
   ) {
-    const [name, host] = address.split("@");
+    let [name, host] = address.split("@");
+    // remove invisible characters %EF%B8%8F
+    name = name.replace(/[^ -~]+/g, "");
+    host = host.replace(/[^ -~]+/g, "");
     return `https://${host}/.well-known/lnurlp/${name}`;
   }
   return null;
--- a/src/extension/content-script/batteries/VimeoVideo.ts
+++ b/src/extension/content-script/batteries/VimeoVideo.ts
@@ -4,9 +4,8 @@ import setLightningData from "../setLightningData";
 const urlMatcher = /^https:\/\/vimeo.com\/.*/;
 
 const battery = (): void => {
-  document.getElementsByClassName("first")[0].innerHTML;
-  const videoDescription = document.querySelector(".first");
-  const channelLink = document.querySelector(".js-user_link");
+  const videoDescription = document.querySelector("[data-description-content]");
+  const channelLink = document.querySelector(".clip_info-subline--watch");
   if (!videoDescription || !channelLink) {
     return;
   }
--- a/src/extension/content-script/batteries/VimeoVideo.ts
+++ b/src/extension/content-script/batteries/VimeoVideo.ts
@@ -1,0 +1,48 @@
+import getOriginData from "../originData";
+import setLightningData from "../setLightningData";
+
+const urlMatcher = /^https:\/\/vimeo.com\/.*/;
+
+const battery = (): void => {
+  document.getElementsByClassName("first")[0].innerHTML;
+  const videoDescription = document.querySelector(".first");
+  const channelLink = document.querySelector(".js-user_link");
+  if (!videoDescription || !channelLink) {
+    return;
+  }
+  const text = videoDescription.textContent || "";
+  let match;
+  let recipient;
+  // check for an lnurl
+  if ((match = text.match(/(lnurlp:)(\S+)/i))) {
+    recipient = match[2];
+  }
+  // if there is no lnurl we check for a zap emoji with a lightning address
+  // we check for the @-sign to try to limit the possibility to match some invalid text (e.g. random emoji usage)
+  else if ((match = text.match(/(⚡:?|lightning:|lnurl:)\s?(\S+@\S+)/i))) {
+    recipient = match[2];
+  } else {
+    return;
+  }
+
+  const name = channelLink.textContent || "";
+  const imageUrl =
+    document.querySelector<HTMLImageElement>(".clip_info-subline--watch img")
+      ?.src || "";
+  setLightningData([
+    {
+      method: "lnurl",
+      recipient,
+      ...getOriginData(),
+      name,
+      icon: imageUrl,
+    },
+  ]);
+};
+
+const VimeoVideo = {
+  urlMatcher,
+  battery,
+};
+
+export default VimeoVideo;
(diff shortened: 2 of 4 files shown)


pseudozach:
hopefully fixes: https://github.com/getAlby/lightning-browser-extension/issues/477

src/extension/content-script/batteries/VimeoVideo.ts
```diff
@@ -0,0 +1,48 @@
+import getOriginData from "../originData";
+import setLightningData from "../setLightningData";
+
+const urlMatcher = /^https:\/\/vimeo.com\/.*/;
+
+const battery = (): void => {
+  document.getElementsByClassName("first")[0].innerHTML;
+  const videoDescription = document.querySelector(".first");
```
bumi, line 8: it seems here `.first` is not enough?
image

src/extension/content-script/batteries/VimeoVideo.ts
```diff
@@ -0,0 +1,48 @@
...
+  const channelLink = document.querySelector(".js-user_link");
```
bumi, line 9: I think this query matches also the ones from the "similar videos" on the side? 
Is it save that this is the link to the current channel?

maybe we can additionally add `.clip_info-subline--watch`
or additionally even check that `data-fatal-attraction` attribute on the link? 

we just need to be save that it is the name of the channel :)

src/common/lib/lnurl.ts
```diff
@@ -13,7 +13,10 @@ const fromInternetIdentifier = (address: string) => {
       /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
     )
   ) {
-    const [name, host] = address.split("@");
+    let [name, host] = address.split("@");
+    // remove invisible characters %EF%B8%8F
+    name = name.replace(/[^ -~]+/g, "");
+    host = host.replace(/[^ -~]+/g, "");
```
bumi, line 19: that's a great addition! we have already seen the problem of invisible characters. 
do you think we can even be more restrictive there?

bumi (changes requested):
Thanks a lot for taking this on! 

how was it for you to create this? when thinking of the architecture we tried to make it easy for people to create such extractors without knowing much of the extension code.
would love to hear your feedback.

pseudozach:
> Thanks a lot for taking this on!
> 
> how was it for you to create this? when thinking of the architecture we tried to make it easy for people to create such extractors without knowing much of the extension code. would love to hear your feedback.

it's great, just the way I like it. I just copied youtubevideo battery and modified it to work with vimeo. Definitely beginner friendly!

src/extension/content-script/batteries/VimeoVideo.ts
pseudozach: I'll look into this but I wasn't sure what channelLink is used for. I assumed it's just to identify that we're on a video page. So I was thinking just the confirmation that it exists would be enough.

src/common/lib/lnurl.ts
pseudozach: this worked for my error `%EF%B8%8F` not sure what other invisible characters could cause such issues. do you know a better regex? I'm terrible with those.

src/extension/content-script/batteries/VimeoVideo.ts
pseudozach: I've changed to `.clip_info-subline--watch` I agree this makes more sense. I don't think `data-fatal-attraction` helps though.

src/extension/content-script/batteries/VimeoVideo.ts
pseudozach: updated. not sure about etiquette, should I resolve the conversation myself or wait until you review?

src/extension/content-script/batteries/VimeoVideo.ts
bumi: 👍 
(the name is extracted from that link and that name is shown in Alby then)

src/extension/content-script/batteries/VimeoVideo.ts
```diff
@@ -0,0 +1,47 @@
+import getOriginData from "../originData";
+import setLightningData from "../setLightningData";
+
+const urlMatcher = /^https:\/\/vimeo.com\/.*/;
```
bumi, line 4: I think we can limit this a bit down.
seems all vimeo video pages end with an ID.
https://vimeo.com/584985260 or https://vimeo.com/channels/bestoftheyear2021/584985260

maybe: `^https:\/\/vimeo.com\/.*\d{4,}\/?\z`

that it must end at least with a 4 digits number?

bumi (commented):
Looking great. 
I think we should adjust the regex and then we need to do a bit more testing but then we can release this! :shipit:

src/extension/content-script/batteries/VimeoVideo.ts
pseudozach: I couldn't get the one you have working I think due to the `\z` at the end.
but made this one:
^https:\/\/vimeo.com\/.*(\d+)(?:|\/\?)
link: https://regexr.com/6e6au
yours does seem simpler and better if I can understand the `\z`

bumi: maybe it should be `$` instead of `\z` - I just thought to check that it ends with a number

pseudozach: that did it because with `\z` eslinter was complaining.

bumi: ah ja. great. 
that regexr.com site is nice. didn't know it.

bumi:
I found one more issues. when an account is "pro" then a `pro` badge is in that channelLink span. 
e.g. here: https://vimeo.com/658711759

secondl1ght:
Not sure why but I am testing this and it is not working for me...

pseudozach:
> I found one more issues. when an account is "pro" then a `pro` badge is in that channelLink span. e.g. here: https://vimeo.com/658711759

good point `.clip_info-subline--watch .js-user_link` works for both I'll change it to this.

bumi:
@secondl1ght what means "not working". 
can you explain what you did? which video did you test? what was the URL? what did you see? what did you expect?

secondl1ght:
I went to https://vimeo.com/670055676 and when I opened up Alby there was no prompt to send any sats to the LN address displayed on the page.. did I miss a step?

bumi:
@secondl1ght hmm. works for me. are you using the latest version of this branch?

secondl1ght:
yep.. :\

kiwiidb:
This worked for me, I've sent 1000 sats to @pseudozach for his nice video.

kiwiidb:
Only (unrelated to PR) feedback that I have is that it takes a bit (2-3 s) for the second screen to show up, and there is no loading indicator in the mean time.

Schermafbeelding 2022-02-04 om 11 41 40
Schermafbeelding 2022-02-04 om 11 41 48

bumi:
thanks for the review!!
I think a simple loading indicator was added in: https://github.com/getAlby/lightning-browser-extension/pull/575 
and I am not sure why the lntxbot sometimes seems slow?

Issue #226, opened 2021-09-13
Setup CI action to run linter and tests


bumi:

escapedcat:
Done with #564  
Can be closed

PR #564, opened 2022-01-27
chore: add code-quality github action #226

escapedcat:
Hoping to solve #226 with this


bumi:
thanks a lot!

.github/workflows/codequality.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  build:
```
bumi, line 11: this is the name of the job, isn't it? 
should we call something like `ci`?

PR #565, opened 2022-01-27
Improve LNURL-Pay notification

dylancom:
Current Scenario 1:
- On the send screen, I'm pasting in: bumi@getalby.com
The details screen shows: _Send payment to: getalby.com_
Paying a few sats shows a notification: _Paid x sats to Alby_

Desired Scenario 1:
- On the send screen, I'm pasting in: bumi@getalby.com
The details screen shows: _Send payment to: bumi@getalby.com_
Paying a few sats shows a notification: _Paid x sats to bumi@getalby.com_

What's currently also weird is that "Send payment to:" is using `details.domain` while the notification uses: `metadata.name`, I think they should use the same info.

### DIFF
--- a/src/app/screens/LNURLPay.tsx
+++ b/src/app/screens/LNURLPay.tsx
@@ -117,7 +117,12 @@ function LNURLPay(props: Props) {
       const payment = await utils.call(
         "lnurlPay",
         { paymentRequest },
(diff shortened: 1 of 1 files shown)


bumi:
utACK

Issue #569, opened 2022-01-29
Dark mode updates


bumi:
Collecting some dark mode todos here
<img width="366" alt="image" src="https://user-images.githubusercontent.com/318/151663275-a1c0a358-8073-454c-8835-8b8524905d9c.png">

secondl1ght:
Fixed with #571, let me know if you find any more!

![image](https://user-images.githubusercontent.com/85003930/151718776-19d1ce5d-4588-4348-bd5e-2e3ecf669bc5.png)

escapedcat:
Fixed according to comment

PR #571, opened 2022-01-30
Fix macaroon text in dark mode

secondl1ght:
Reported in issue #569


secondl1ght:
I just found another dark mode fix so I included it in this PR rather than making a new one.

PR #581, opened 2022-02-03
Proposal: improved lnurlpay screen

dylancom:
I think the PublisherCard is unnecessary in the LNURLPay screen and takes too much space.
(Currently you'd have to scroll to see the confirm button).
My preference goes to solely using it for contextual info on the Home screen.

Before:
Schermafbeelding 2022-02-03 om 21 34 13

After:
Schermafbeelding 2022-02-03 om 21 12 42

### DIFF
--- a/src/app/components/PublisherCard/index.tsx
+++ b/src/app/components/PublisherCard/index.tsx
@@ -55,7 +55,7 @@ export default function PublisherCard({
             )}
           </h2>
           {description && (
-            <p className="text-white opacity-75 truncate">{description}</p>
+            <p className="text-white opacity-60 truncate">{description}</p>
           )}
         </div>
       </div>
--- a/src/app/components/PublisherCard/index.tsx
+++ b/src/app/components/PublisherCard/index.tsx
@@ -6,46 +6,60 @@ const DEFAULT_IMAGE =
 
 export type Props = {
   title: string;
+  description?: string;
   image: string;
   url?: string;
   children?: React.ReactNode;
 };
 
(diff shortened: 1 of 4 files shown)


bumi:
I actually like the color and that it picks something up from the website. it gives a clear reference to the website and who/what you send satoshis to. 

The size is maybe a problem. maybe we simply can make it a bit smaller? and only use half the size? e.g. having the icon on the left and text on the right in one line?

bumi:
and the publisher card when I manually enter an invoice/lnurl/... somehow does not make much sense: 
image

MoritzKa:
"The size is maybe a problem. maybe we simply can make it a bit smaller? and only use half the size? e.g. having the icon on the left and text on the right in one line?" +1

dylancom:
Thanks for the feedback. I like the icon left + text right option.
See:
Schermafbeelding 2022-02-04 om 12 56 22

Schermafbeelding 2022-02-04 om 12 56 39

bumi:
this is a great improvement!! looks and feels great

bumi:
though I am not sure about the `description` on all pages I tested the description is very truncated and too short to make sense. maybe we can add a tiptool (title tag?) that the full description can be read?

dylancom:
Or we go for a slightly smaller font size and allow 2 lines?

dylancom:
For now I added title tags. Let me know what you think.

bumi:
👍 
btw. I think we need to soon think about how we can handle multiple recipients. This would be needed to support the value4value podcast index spec.

bumi (approved):

PR #586, opened 2022-02-06
Use new Alby hosted wallets

bumi:
With this change we create Alby accounts and LndHub compatible wallets.

This also means that the `NewWallet` component no longer works with default lndhub instances.
Do you think we should still support those?

### DIFF
--- a/src/app/screens/Onboard/NewWallet/index.tsx
+++ b/src/app/screens/Onboard/NewWallet/index.tsx
@@ -35,16 +35,13 @@ export default function NewWallet() {
     })
       .then((res) => {
         res.json().then((data) => {
-          if (
-            data.lndhub &&
-            data.lndhub.login &&
-            data.lndhub.password &&
-            data.lndhub.url
-          ) {
+          if (data.lndhub?.login && data.lndhub?.password && data.lndhub?.url) {
             setLndHubData(data.lndhub);
           } else {
             console.error(data);
-            alert("Failed to create a new wallet. Please contact support.");
+            alert(
+              "Failed to create a new wallet. Please try again and contact support."
+            );
           }
         });
       })
@@ -145,7 +142,7 @@ export default function NewWallet() {
 
               <div className="mt-6">
                 <label
-                  htmlFor="password"
+                  htmlFor="email"
                   className="block font-medium text-gray-700 dark:text-gray-400"
                 >
                   Email Address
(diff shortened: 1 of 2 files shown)


src/app/screens/Onboard/NewWallet/index.tsx
```diff
@@ -25,15 +28,20 @@ export default function NewWallet() {
...
       headers,
-      body: JSON.stringify({ partnerid: "bluewallet", accounttype: "common" }),
+      body: JSON.stringify({ email }),
     })
       .then((res) => {
         res.json().then((data) => {
-          if (data.login && data.password) {
-            setLndHubData(data);
+          if (
+            data.lndhub &&
```
dylancom, line 39: [SUGGESTION] could be simplified to: `data.lndhub?.login && data.lndhub?.password && data.lndhub?.url`

src/app/screens/Onboard/NewWallet/index.tsx
```diff
@@ -130,10 +138,27 @@ export default function NewWallet() {
...
+                <strong>We host a lightning wallet for you!</strong>
                 <br />
-                This quick setup uses a custodial service <br /> to manage your
-                wallet.
+                ...but remember: not your keys, not your coins.
+              </div>
+
+              <div className="mt-6">
+                <label
+                  htmlFor="password"
```
dylancom, line 148: Should be "email".

dylancom (changes requested):

src/app/screens/Onboard/NewWallet/index.tsx
escapedcat: To support label-click/functionality do you also need the `id` on the input?

dylancom: @escapedcat correct!

src/app/screens/Onboard/NewWallet/index.tsx
```diff
@@ -130,10 +138,27 @@ export default function NewWallet() {
...
+                  className="block font-medium text-gray-700 dark:text-gray-400"
+                >
+                  Email Address
+                </label>
+                <div className="mt-1">
+                  <Input
+                    name="email"
+                    type="email"
```
dylancom, line 156: add: id="email".
Thanks to @escapedcat

src/app/screens/Onboard/NewWallet/index.tsx
bumi: uuuupppsss.. failed at prorgamming rule no.1: never c&p 
thanks

dylancom (approved):

PR #589, opened 2022-02-07
feat: associate labels with its inputs (A11Y)

dylancom:
Associating a `<label>` with an `<input>` element offers some major advantages:

- The label text is not only visually associated with its corresponding text input; it is programmatically associated with it too. This means that, for example, a screen reader will read out the label when the user is focused on the form input, making it easier for an assistive technology user to understand what data should be entered.
- When a user clicks or touches/taps a label, the browser passes the focus to its associated input (the resulting event is also raised for the input). That increased hit area for focusing the input provides an advantage to anyone trying to activate it — including those using a touch-screen device.

Thanks to @escapedcat I did a double-check on this and noticed we were still missing id attributes in a lot of places.

### DIFF
--- a/src/app/screens/Onboard/ConnectEclair/index.tsx
+++ b/src/app/screens/Onboard/ConnectEclair/index.tsx
@@ -80,6 +80,7 @@ export default function ConnectLnbits() {
               <div className="mt-1">
                 <Input
                   name="password"
+                  id="password"
                   type="text"
                   required
                   onChange={handleChange}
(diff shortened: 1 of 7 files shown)


bumi (approved):

escapedcat:
I other projects we use a i.e. `InputField` component which combines `label+input(+error-message, info, etc.)` for the usual input-cases. Not sure if it's worth (yet) for this projects. Just an idea.

PR #590, opened 2022-02-07
TextField component (label + input combined)

dylancom:
- Combines `<label>` and `<input>` for easier code maintenance.
- Removes CurrencyInput which was basically a double component.


src/app/screens/ConfirmPayment/index.tsx
```diff
@@ -150,17 +150,12 @@ function ConfirmPayment(props: Props) {
...
                 </p>
                 <div>
-                  <label
-                    htmlFor="budget"
-                    className="mb-1 block text-sm font-medium text-gray-700 dark:text-gray-400"
-                  >
-                    Budget
-                  </label>
-                  <CurrencyInput
+                  <TextField
```
escapedcat, line 153: Component is called `TextField` but here it's used as a number field. Would `InputField` be a better name? Might be confusing for people thinking this should/will only be used for text or confuse it with a TextArea?

escapedcat (commented):
🚀  That was fast!

src/app/screens/ConfirmPayment/index.tsx
dylancom: Wasn't quite sure about the naming tbh, so I took inspiration from: https://mui.com/components/text-fields/

dylancom: Might want to call it just "<InputWithLabel />" at this point.

dylancom: In Material Design it's als called a Textfield, whether you enter number / string.
https://material.io/components/text-fields

escapedcat: 👍  Alright, I get that.  
Might be good to keep that in mind for future component naming. And document it i.e. "We stick to material-ui naming even though we don't use that framework" or something like that.

dylancom: Not sure we need that for all components but when naming becomes hard I like to take inspiration from proven ui libs.
But open to any name that be better :)

escapedcat: No all good, the question was just popping up when I looked through it. Don't mind me :)

escapedcat: No all good, the question was just popping up when I looked through it. Don't mind me :)

src/app/components/Form/TextField.tsx
```diff
@@ -0,0 +1,47 @@
...
+  max,
+}: React.InputHTMLAttributes<HTMLInputElement> & { label: string }) => (
+  <>
+    <label
+      htmlFor={id}
+      className="block font-medium text-gray-700 dark:text-white"
+    >
+      {label}
+    </label>
+    <div className="mt-1">
```
bumi, line 26: Is this HTML structure generally usable for us or is this matching for the ones in this PR?

bumi (approved):
utACK

src/app/components/Form/TextField.tsx
dylancom: Generally indeed. Makes our current code easier and future code :)

PR #594, opened 2022-02-09
Allowance empty screen design

dylancom:
Just like the DefaultView, the AllowanceView should show some info when there are no transactions yet.

Before:
Schermafbeelding 2022-02-09 om 09 59 09

After:
Schermafbeelding 2022-02-09 om 10 05 14

### DIFF
--- a/src/app/screens/Home/index.jsx
+++ b/src/app/screens/Home/index.jsx
@@ -142,7 +142,7 @@ function Home() {
             />
           ) : (
             <p className="text-gray-500 dark:text-gray-400">
-              No transactions yet.
+              No transactions on <strong>{allowance.name}</strong> yet.
             </p>
           )}
         </div>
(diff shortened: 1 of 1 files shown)


src/app/screens/Home/index.jsx
```diff
@@ -138,6 +140,10 @@ function Home() {
                 ),
               }))}
             />
+          ) : (
+            <p className="text-gray-500 dark:text-gray-400">
+              No transactions yet.
```
bumi, line 145: this is also website specific, isn't it? 
maybe we say No transactions on xxx yet?

bumi (changes requested):

src/app/screens/Home/index.jsx
dylancom: Correct, nice suggestion.
Schermafbeelding 2022-02-09 om 12 06 40

PR #599, opened 2022-02-10
feat: delete accounts

dylancom:
Adds an account page to the options and allows you to delete accounts.
Schermafbeelding 2022-02-14 om 15 37 20

### DIFF
--- a/src/app/screens/Accounts.tsx
+++ b/src/app/screens/Accounts.tsx
@@ -1,0 +1,106 @@
+import { useState, useEffect } from "react";
+import { WalletIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
+
+import type { Accounts } from "../../types";
+import api from "../../common/lib/api";
+import { useAuth } from "../context/AuthContext";
+
+import Container from "../components/Container";
+
+function AccountsScreen() {
+  const [accounts, setAccounts] = useState<Accounts>({});
+  const auth = useAuth();
+
+  function getAccounts() {
+    api.getAccounts().then((response) => {
+      setAccounts(response);
+    });
+  }
+
+  async function selectAccount(accountId: string) {
+    auth.setAccountId(accountId);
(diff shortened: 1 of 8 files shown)


secondl1ght:
Do we also want to add the ability to rename the accounts in this PR or make then a separate one?

dylancom:
To keep things tiny / easily reviewable I think we can split that up

src/app/screens/Accounts.tsx
```diff
@@ -0,0 +1,126 @@
...
+                          {account.name}
+                        </div>
+                        <div className="text-sm text-gray-500">
+                          {account.connector}
+                        </div>
+                      </div>
+                    </div>
+                  </td>
+                  <td className="px-6 py-4 whitespace-nowrap">
+                    <Badge
```
bumi, line 102: How about we remove this centered column?
I don't really think that connector badge provides much value (initially I put the connector there because in the design there was a badge and I did not have anything else 🙈😱)

src/app/screens/Accounts.tsx
```diff
@@ -0,0 +1,126 @@
...
+    if (window.confirm(`Are you sure you want to delete account: ${name}?`)) {
+      let nextAccountId;
+      let accountIds = Object.keys(accounts);
+      if (auth.account?.id === id && accountIds.length > 1) {
+        nextAccountId = accountIds.filter((accountId) => accountId !== id)[0];
+      }
+
+      await Promise.all([
+        utils.call("deleteAccount", { id }),
+        removeAccountFromCache(id), // Clear cached account info (alias/balance).
```
bumi, line 40: ahhh.. I hate cache invalidation... :D

can we do this in the api? and use `api.deleteAccount()` instead of `utils.call("deleteAccount"...)?

bumi (commented):
I actually think we should remove the centered badge and then merge this PR. 
It already brings great value and an often requested functionality.

PR #605, opened 2022-02-16
Feature/reddit-tipping

AustinKelsay:
Implemented Reddit tipping similarly to how Twitter is done and tested it on multiple reddit profiles

### DIFF
--- a/src/extension/content-script/batteries/Reddit.ts
+++ b/src/extension/content-script/batteries/Reddit.ts
@@ -3,70 +3,43 @@ import setLightningData from "../setLightningData";
 
 const urlMatcher = /^https:\/\/www.reddit\.com\/user\/(\w+).*/;
 
-// can we detect the user description from the description metatag?
-function isOnProfilePage() {
-  return (
-    document.querySelector<HTMLMetaElement>(
-      'head > meta[name="description"]'
-    ) != null
-  );
-}
-
-function getUserData() {
-  if (isOnProfilePage()) {
-    const element = document.querySelector<HTMLMetaElement>(
-      'head > meta[name="description"]'
-    );
-    const imageUrl = document.querySelector<HTMLImageElement>(
-      `img[alt="User avatar"]`
-    )?.src;
-    if (element && imageUrl) {
-      return {
-        element,
-        imageUrl,
-        name: document.title,
-      };
-    }
-  }
-  return null;
-}
-
 function battery(): void {
-  let userData;
-  if ((userData = getUserData())) {
-    let match;
-    let recipient;
-    // attempt to extract lnurlp: from the description text
-    if ((match = (userData.element.content || "").match(/lnurlp:(\S+)/i))) {
-      recipient = match[1];
-    } else {
-      // else let's look for an ⚡ emoji in the userDescription
-      if (userData.element.content && userData.element.content.includes("⚡")) {
-        const descriptionText = userData.element.content.split(" ");
-        // If we detect the empji we can iterate through the text seperated by whitespace to try and find a match
-        for (const text of descriptionText) {
-          if ((match = (text || "").match(/(\S+@\S+)/))) {
-            recipient = match[1];
-          }
-        }
-      }
+  const descriptionElement = document.querySelector<HTMLMetaElement>(
+    'head > meta[name="description"]'
+  );
+  const imageUrl = document.querySelector<HTMLImageElement>(
+    `img[alt="User avatar"]`
+  )?.src;
 
-      // if we still did not find anything ignore it.
-      if (!recipient) {
-        return;
-      }
-
-      setLightningData([
-        {
-          method: "lnurl",
-          recipient,
-          ...getOriginData(),
-          icon: userData.imageUrl,
-          name: userData.name,
-        },
-      ]);
-    }
+  if (!descriptionElement || !imageUrl) {
+    return;
   }
+  let match;
+  let recipient;
+  // attempt to extract lnurlp: from the description text
+  if ((match = (descriptionElement.content || "").match(/lnurlp:(\S+)/i))) {
+    recipient = match[1];
+  } else if (
+    (match = descriptionElement.content.match(
+      /(⚡:?|lightning:|lnurl:)\s?(\S+@\S+)/i
+    ))
+  ) {
+    recipient = match[2];
+  }
+  // if we still did not find anything ignore it.
+  if (!recipient) {
+    return;
+  }
+
+  setLightningData([
+    {
+      method: "lnurl",
+      recipient,
+      ...getOriginData(),
+      icon: imageUrl,
+      name: document.title,
+    },
+  ]);
 }
 
 const reddit = {
--- a/src/extension/content-script/batteries/Reddit.ts
+++ b/src/extension/content-script/batteries/Reddit.ts
@@ -1,22 +1,22 @@
 import getOriginData from "../originData";
 import setLightningData from "../setLightningData";
 
-declare global {
-  interface Window {
-    LBE_REDDIT_MUTATION_OBSERVER: MutationObserver;
-  }
-}
-
 const urlMatcher = /^https:\/\/www.reddit\.com\/user\/(\w+).*/;
 
-// can we get the user description from the primary column that holds the profile information?
+// can we detect the user description from the description metatag?
 function isOnProfilePage() {
-  return document.querySelector('[class="bVfceI5F_twrnRcVO1328"]') != null;
+  return (
+    document.querySelector<HTMLMetaElement>(
+      'head > meta[name="description"]'
+    ) != null
+  );
 }
 
 function getUserData() {
   if (isOnProfilePage()) {
-    const element = document.querySelector('[class="bVfceI5F_twrnRcVO1328"]');
+    const element = document.querySelector<HTMLMetaElement>(
+      'head > meta[name="description"]'
+    );
     const imageUrl = document.querySelector<HTMLImageElement>(
       `img[alt="User avatar"]`
     )?.src;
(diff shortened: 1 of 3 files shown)


src/extension/content-script/batteries/Reddit.ts
```diff
@@ -0,0 +1,100 @@
...
+  interface Window {
+    LBE_REDDIT_MUTATION_OBSERVER: MutationObserver;
+  }
+}
+
+const urlMatcher = /^https:\/\/www.reddit\.com\/user\/(\w+).*/;
+
+// can we get the user description from the primary column that holds the profile information?
+function isOnProfilePage() {
+  return document.querySelector('[class="bVfceI5F_twrnRcVO1328"]') != null;
```
escapedcat, line 14: Nice PR! We had a look yesterday at this and didn't trust those classes. Can we be sure these won't change? They do look dynamic. That's why we were worried a bit.  
What do you think?

escapedcat, line 14: Nice PR! We had a look yesterday at this and didn't trust those classes. Can we be sure these won't change? They do look dynamic. That's why we were worried a bit.  
What do you think?

escapedcat, line 14: Nice PR! We had a look yesterday at this and didn't trust those classes. Can we be sure these won't change? They do look dynamic. That's why we were worried a bit.  
What do you think?

src/extension/content-script/batteries/Reddit.ts
```diff
@@ -0,0 +1,100 @@
...
+      ) {
+        recipient = match[1];
+      } else {
+        // else let's look for an ⚡ emoji in the userDescription
+        if (
+          userData.element.textContent &&
+          userData.element.textContent.includes("⚡")
+        ) {
+          const descriptionText = userData.element.textContent.split(" ");
+          console.log({ descriptionText });
```
escapedcat, line 55: Is this here on purpose? If so, how about switching it to `.info`?

escapedcat, line 55: Is this here on purpose? If so, how about switching it to `.info`?

src/extension/content-script/batteries/Reddit.ts
```diff
@@ -0,0 +1,100 @@
...
+    window.LBE_REDDIT_MUTATION_OBSERVER = new MutationObserver(
+      redditDOMChanged
+    );
+  }
+  window.LBE_REDDIT_MUTATION_OBSERVER.observe(document, {
+    childList: true,
+    subtree: true,
+  });
+  // On slow connections the observer is added after the DOM is fully loaded.
+  // Therefore the callback twitterDOMChanged needs to also be called manually.
```
escapedcat, line 92: Copy&paste type: `twitterDOMChanged`

escapedcat, line 92: Copy&paste type: `twitterDOMChanged`

escapedcat (commented):
Nice one!

src/extension/content-script/batteries/Reddit.ts
escapedcat, line 14: Arg, not sure if you got the first comment. Had some network issues.   
We also had a look at this yesterday and were a bit worried that these classnames might change because they look dynamic and this might be a way of avoiding others to scrape reddit. What do you think? Would there be a nicer way/saver to find the wanted divs?

AustinKelsay: Yeah I definitely thought the same thing when I was first looking at their classnames but surprisingly this class is universal for all of the user profile descriptions (I checked multiple profiles)

I couldn't find any other way to programmatically grab this element

escapedcat: Interesting. Good to know! I guess it would be ok to start with this and adjust if something changes.

AustinKelsay: My thoughts exactly, this works fine for now but it would be ideal to find a more robust method for grabbing this element over the long term.

bumi:
whoot whoot! nice, thanks for tackling this! looking good. I think it is great to bring this to reddit! 🚀 

I got a few comments and thoughts: 
* on twitter all content is loaded asynchronously and also the website navigation only happens in JS. This is the reason why this MutationObserver is used. That observer allows to execute some code and check the DOM changes. I am actually not sure if we need the same approach on Reddit. It seems on reddit the page is fully reloaded and we should be able to just look into the DOM. Did you check this? 
* it also seems there is a meta tag in the page that contains the user profile description 
image maybe this can even be used to extract the lightning information? So we would not need to check for the CSS class?

AustinKelsay:
> 
Hey thanks!

- Yeah I assumed at first because of the classnames everything was being generated dynamically so I thought it would be good to use the same MutationObserver but if it's not needed I can take it out an retest
- I didn't think to check the meta tags but that seems like a much cleaner way of grabbing the user description. Should I use [https://github.com/BetaHuhn/metadata-scraper] to do this?

bumi:
Yeah, the MutationObserver is great. but if it works without it it would reduce complexity which I think is good. 

to check for the meta tag I would do that dircttly (like [here](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/content-script/batteries/Monetization.ts#L7-L12) 

something like:
```ts
const descriptionTag = document.querySelector<HTMLMetaElement>(
    'head > meta[name="description"]'
);
descriptionTag.content...
```

src/extension/content-script/batteries/Reddit.ts
```diff
@@ -0,0 +1,76 @@
...
+function isOnProfilePage() {
+  return (
+    document.querySelector<HTMLMetaElement>(
+      'head > meta[name="description"]'
+    ) != null
+  );
+}
+
+function getUserData() {
+  if (isOnProfilePage()) {
```
bumi, line 16: I am wondering if we need to check this on reddit. 
On twitter we for example need to somehow differentiate between `/home` `/notifications` and `/your-username`. but on reddit it seems we can do that directly from the URL `reddit.com/user/...` I guess if our regex matches then we can be sure to be on a user profile.

src/extension/content-script/batteries/Reddit.ts
```diff
@@ -0,0 +1,76 @@
...
+        name: document.title,
+      };
+    }
+  }
+  return null;
+}
+
+function battery(): void {
+  let userData;
+  if ((userData = getUserData())) {
```
bumi, line 36: probably we could even inline the two functions? 

```
const descriptionElement = document.querySelector<HTMLMetaElement>(
  'head > meta[name="description"]'
);
const imageUrl = document.querySelector<HTMLImageElement>(
  `img[alt="User avatar"]`
)?.src;

if (!descriptionElement) {
  return;
}
```

and then we check the `descriptionElement.content`

src/extension/content-script/batteries/Reddit.ts
```diff
@@ -0,0 +1,76 @@
...
+    let match;
+    let recipient;
+    // attempt to extract lnurlp: from the description text
+    if ((match = (userData.element.content || "").match(/lnurlp:(\S+)/i))) {
+      recipient = match[1];
+    } else {
+      // else let's look for an ⚡ emoji in the userDescription
+      if (userData.element.content && userData.element.content.includes("⚡")) {
+        const descriptionText = userData.element.content.split(" ");
```
bumi, line 45: Twitter replaces the emoji with an image tag which is the reason why we can not use a regex. 
Reddit seems a bit nicer there and we should be able to use the regex.

so we should be able to do something like here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/content-script/batteries/YouTubeVideo.ts#L20-L29

```js
const text = descriptionElement.content;
// check for an lnurl
if ((match = text.match(/(lnurlp:)(\S+)/i))) {
  recipient = match[2];
}
// if there is no lnurl we check for a zap emoji with a lightning address
// we check for the @-sign to try to limit the possibility to match some invalid text (e.g. random emoji usage)
else if ((match = text.match(/(⚡:?|lightning:|lnurl:)\s?(\S+@\S+)/i))) {
  recipient = match[2];
} else {
  return;
}
```

bumi (changes requested):
nice, looks already simpler. 
got some more thoughts in the comments.

AustinKelsay:
> nice, looks already simpler.
> got some more thoughts in the comments.

Great thanks for the suggestions I implemented each one of these and then retested, let me know if anything else looks out of place.

src/extension/content-script/batteries/Reddit.ts
```diff
@@ -0,0 +1,49 @@
...
+import setLightningData from "../setLightningData";
+
+const urlMatcher = /^https:\/\/www.reddit\.com\/user\/(\w+).*/;
+
+function battery(): void {
+  const descriptionElement = document.querySelector<HTMLMetaElement>(
+    'head > meta[name="description"]'
+  );
+  const imageUrl = document.querySelector<HTMLImageElement>(
+    `img[alt="User avatar"]`
```
bumi, line 11: it seems this does not match on all pages. e.g. https://www.reddit.com/user/bitdex/
but maybe we do not need this actually. It seems there is also a meta tag for the image: 
```
document.querySelector('head > meta[property="og:image"]')
```

and I think this one should be already extracted through the `getOriginData()`. so maybe we we can even just rely on that. 
Can you test that? Or alternatively explicitly use the `og:image` here. (`document.querySelector('head > meta[property="og:image"]')?.content`)

bumi (changes requested):
I found one more thing. 

but then it looks good: 
image

AustinKelsay:
> I found one more thing.
> 
> but then it looks good: image

Ahh good catch! I just updated and retested with a few more accounts now

bumi (approved):

bumi:
:shipit: yay!

PR #613, opened 2022-02-19
feat: add integration test and fix bolt11 decoding in jest

dylancom:
I polyfilled UIntArray8 so we can directly decode invoices and don't need to mock it anymore.


src/app/screens/ConfirmPayment/index.test.tsx
```diff
@@ -39,4 +38,17 @@ describe("ConfirmPayment", () => {
...
   });
+
+  test("toggles set budget, displays input with a default budget", async () => {
+    render(
+      <MemoryRouter>
+        <ConfirmPayment {...props} />
+      </MemoryRouter>
+    );
+    fireEvent.click(screen.getByText("Remember and set a budget"));
+    await waitFor(() => screen.getByLabelText("Budget"));
```
escapedcat, line 49: Would [findby](https://testing-library.com/docs/dom-testing-library/api-async/#findby-queries) work here instead of waitFor+getBy?

package.json
```diff
@@ -128,6 +130,7 @@
     "jest-webextension-mock": "^3.7.19",
     "lint-staged": "^12.3.2",
     "mini-css-extract-plugin": "^2.5.3",
+    "patch-package": "^6.4.7",
```
escapedcat, line 133: Nice! We use thsi as well

patches/jest-environment-jsdom+27.5.1.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/jest-environment-jsdom/build/index.js b/node_modules/jest-environment-jsdom/build/index.js
+index 77176ba..1644673 100644
+--- a/node_modules/jest-environment-jsdom/build/index.js
++++ b/node_modules/jest-environment-jsdom/build/index.js
+@@ -102,6 +102,7 @@ class JSDOMEnvironment {
+     (0, _jestUtil().installCommonGlobals)(global, config.globals); // TODO: remove this ASAP, but it currently causes tests to run really slow
+ 
+     global.Buffer = Buffer; // Report uncaught errors.
++    global.Uint8Array = Uint8Array;
```
escapedcat, line 9: How did you find this solution? 🙌

dylancom: I found this line in the node test env: https://github.com/facebook/jest/blob/1b402ab76b8bd0f8c5c8512345e2fd7aca8d83ed/packages/jest-environment-node/src/index.ts#L46

In most issues people advice to define the node test env, which in our case doesn't work.

bumi (approved):
utACK, looks good to me.
@escapedcat what do you think?

escapedcat:
looksgreattome!

Issue #598, opened 2022-02-09
Make dark/light mode configurable


bumi:
Currently we use the system's dark/light settings. Can this also be a custom Alby setting? (mode: light/dark/system?) 

https://feedback.getalby.com/feature-requests-/posts/i-would-like-no-dark-mode-option-on-alby

bumi:
@secondl1ght you want to check this out?

secondl1ght:
Sure I can check this out

secondl1ght:
Just wanted to give an update on this, I have it setup on the front-end and have starting working on hooking it up. Just need to figure out how to get it working but I should have time soon to look into this. Here is how the settings look, there are 3 options: Light, Dark and System for the theme.

![image](https://user-images.githubusercontent.com/85003930/154831615-8be1d6fe-54eb-468a-8095-179a4ca25c77.png)

bumi:
Super cool. Feel free to open a draft PR then people already can follow the development and maybe can help with hints maybe.

secondl1ght:
That would be great! I'll do that now.

PR #618, opened 2022-02-20
Add theme selection options to settings

secondl1ght:
As requested in issue #598 and in our feedback board I have added an option to the settings screen that allows for selecting if the user wants dark, light or system settings for the extension theme. Currently we use the system settings only for choosing between dark/light mode.

This is still a work in progress, I have a type error that I need to solve and I also still need to figure out how to actually get this config working. But I have the front-end and I started working on the back-end, if anyone wants to review what I have done so far and provide any feedback that would be good! 

I am doing more research on how to implement both functionalities. Tailwind makes it easy to use either the system settings or a selectable theme. But it is a bit more complicated when trying to allow for both functionalities, especially with React.

Here's how it looks:

image


secondl1ght:
UPDATE: This PR had major merge conflicts to be resolved because I submitted it so long ago, wasn't able to resolve on GitHub so I had to do locally. I submitted a new PR with the same changes here but that has now been merged with master: #618

Issue #607, opened 2022-02-16
Show tipping options even if the publisher card is shown


bumi:
When a website is connected (a publisher/allowance is stored) we currently show only the publisher card and do not show the lightning tipping options. Ideally we mage to show both.

This can be reproduced on getalby.com (login with Alby on [getalby.com/login](https://getalby.com/login) and go to the [homepage](https://getalby.com) which contains lightning details.
 
https://feedback.getalby.com/feature-requests-/posts/the-https-getalby-com-site-has-no-option-to-donate-via-ln-

secondl1ght:
I'm not able to reproduce this - when I login to the Alby site and then go back to the homepage and open the extension, there is a 'Send Satoshis' button that I can click to send tips.

bumi:
Is getalby.com on the websites list / do you have an allowance saved?

secondl1ght:
Oh no it isn't, I wasn't prompted to add as a publisher, how can I do this?

bumi:
so I guess you did not login with Alby on getalby.com
You have to play around and dig into that yourself. Try to understand what is happening. That is the most work when debugging and fixing bugs. But it's easy just click around a bit. This also helps you to understand how things work in Alby.

secondl1ght:
That's weird because I was taken to the logged in screen to I assumed that meant I was logged in.. :\

![image](https://user-images.githubusercontent.com/85003930/154852577-bd339eca-97e7-4734-bb4e-19c3d4631a90.png)

secondl1ght:
Ok I put in PR #619 to resolve this!

bumi:
@secondl1ght good find! great!! I left a comment there...we should try to not copy&paste the code...somehow we need to refactor this to keep it DRY.

PR #619, opened 2022-02-20
Add tips button for when meta tag exists and also connected to site

secondl1ght:
As discussed in issue #607 and reported in our feedback board, the tips button was not displaying if a website had the meta tag but you were also connected to them as a publisher, with or without an allowance.

I have added the tips button to this screen so that you can still tip publishers even if you are connected to their website:

image

### DIFF
--- a/src/app/screens/Home/index.jsx
+++ b/src/app/screens/Home/index.jsx
@@ -72,7 +72,36 @@ function Home() {
   function renderAllowanceView() {
     return (
       <>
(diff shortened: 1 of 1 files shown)


bumi:
We should not copy & paste the code. This would mean we need to change the code in multiple places which easily leads to errors. 

so I think we have to somehow find a way to render the lndata

secondl1ght:
Ok sounds good, i'll look into making this a component.

bumi:
closing in favor of #706

PR #622, opened 2022-02-21
refactor: use typescript for all components

dylancom:
Refactored the left-over jsx components as part of fully transitioning to TypeScript.


dylancom:
@pavanjoshi914 no not yet. In another PR I'm working now on migration all code in the /screens folder. (which still has some jsx left).

Then:
- /app/router
- /extension

are not completely over on TS yet.

bumi (approved):
utACK
should merge these asap to prevent conflicts...
but we need to test it well.

PR #626, opened 2022-02-22
feat: improve accounts ux

dylancom:


bumi:
(currently we do not update the accounts list in the dropdown when we update the accounts.)

src/app/screens/Accounts.tsx
```diff
@@ -52,46 +60,47 @@ function AccountsScreen() {
...
+        <table className="min-w-full">
+          <tbody className="divide-y divide-gray-200">
             {Object.keys(accounts).map((accountId) => {
               const account = accounts[accountId];
               return (
                 <tr key={accountId}>
                   <td className="px-6 py-4 whitespace-nowrap">
                     <div className="flex items-center">
-                      <WalletIcon className="w-8 h-8 -ml-1 text-gray-500" />
+                      <div className="w-12 h-12 flex justify-center items-center rounded-full bg-orange-bitcoin-50">
```
bumi, line 80: actually I like the one without the orange color currently better... 🤔

PR #630, opened 2022-02-23
feat: accounts context (updating accounts troughout the whole app)

dylancom:
Create an Accounts context so accounts state can be used across multiple components easily.
For e.g. when an account get's deleted, it should resemble this in the Accounts list on the accounts screen but also in the AccountsMenu.


bumi:
@dylancom can you rebase this one? 
You're probably faster adding this now in the new typed components.

escapedcat (approved):

bumi (approved):
yay, thanks.

Issue #635, opened 2022-02-25
Switch account Ids to UUIDs


bumi:
**Describe the bug**
Currently we use numbers as account IDs. Those are currently simply generated by counting the accounts and +1. 
Before being able to delete an account this was fine. Now since we can delete accounts this is a problem because it can cause "duplicates"

**To Reproduce**
1. Make sure to have two accounts: (ID 1 and ID 2)
2. Delete the first account (ID 1) (ID 2 is still there)
3. Try to add a new account

Error: This will try to create a new account with ID accounts.lenght+1 == 2 - and thus overwrite the existing account

**Idea**

We should switch to UUIDs when creating new accounts. 
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/accounts/add.js#L14

dylancom:
This has been integrated.

PR #627, opened 2022-02-22
chore: add integration tests using Playwright and Puppeteer

MiguelMedeiros:
Fix #556.

---

### New Dev Dependencies
This PR implements integration tests using:
- [Jest](https://jestjs.io/) 
- [Puppeteer](https://github.com/puppeteer/puppeteer)
- [Puppeteer Testing Library](https://testing-library.com/docs/)
- [Complete Randomer](https://www.npmjs.com/package/complete-randomer)

---

### Integrations tests
I've already added the following tests:
- [x] follow the welcome screen and connect using LNBits
- [x] follow the welcome screen and connect using BlueWallet
- [x] follow the welcome screen and create an Alby wallet

There are still more connections to the welcome screen, and other screens like spend and receive sats, connect and disconnect websites, lock and remove accounts. 

But we can do that in other PRs, the purpose of this PR is to start testing the extension so that the screens could have uniform paths and don't break due to any future change.

---

### How to test

Run this command to run all the tests.
```bash
yarn run test
```

---

## Preview

https://user-images.githubusercontent.com/5798170/155232637-eec1800d-f1a3-45aa-9363-c766a6bf7180.mp4

PR #636, opened 2022-02-25
fix(addAccount): use uuid ad account-id #635

escapedcat:
- switch add.js to TS
- add types

If possible would like to add a test for this when #627 is merged

### DIFF
--- a/src/extension/background-script/actions/accounts/add.ts
+++ b/src/extension/background-script/actions/accounts/add.ts
@@ -1,9 +1,20 @@
 import type { Runtime } from "webextension-polyfill";
 import { encryptData } from "../../../../common/lib/crypto";
-import type { Message } from "../../../../types";
+import type { OriginData, Account } from "../../../../types";
 import state from "../../state";
 
-const add = async (message: Message, _sender: Runtime.MessageSender) => {
+interface AddAccountMessage {
+  args: Account;
+  origin: OriginData;
+  application?: string;
+  prompt?: boolean;
+  type?: string;
+}
+
+const add = async (
+  message: AddAccountMessage,
+  _sender: Runtime.MessageSender
+) => {
   const newAccount = message.args;
   const accounts = state.getState().accounts;
 
--- a/src/extension/background-script/actions/accounts/add.ts
+++ b/src/extension/background-script/actions/accounts/add.ts
@@ -1,7 +1,9 @@
+import type { Runtime } from "webextension-polyfill";
 import { encryptData } from "../../../../common/lib/crypto";
+import type { Message } from "../../../../types";
 import state from "../../state";
 
-const add = async (message, sender) => {
+const add = async (message: Message, _sender: Runtime.MessageSender) => {
   const newAccount = message.args;
   const accounts = state.getState().accounts;
 
@@ -11,9 +13,9 @@ const add = async (message, sender) => {
   const password = state.getState().password;
   const currentAccountId = state.getState().currentAccountId;
 
-  const accountId = Object.keys(accounts).length + 1;
-  console.log(`Created account ${accountId}`);
+  if (!password) return { error: "Password is missing" };
 
+  const accountId = crypto.randomUUID();
   newAccount.config = encryptData(newAccount.config, password);
   accounts[accountId] = newAccount;
 
--- a/src/extension/background-script/actions/accounts/add.ts
+++ b/src/extension/background-script/actions/accounts/add.ts
@@ -3,6 +3,8 @@ import { encryptData } from "../../../../common/lib/crypto";
 import type { OriginData, Account } from "../../../../types";
 import state from "../../state";
 
+// @TODO: https://github.com/getAlby/lightning-browser-extension/issues/652
+// align Message-Types
 interface AddAccountMessage {
   args: Account;
   origin: OriginData;
--- a/src/types.ts
+++ b/src/types.ts
@@ -51,6 +51,9 @@ export interface Battery extends OriginData {
 export interface Message {
   args: Record<string, unknown>;
   origin: OriginData;
+  application?: string;
+  prompt?: boolean;
+  type?: string;
 }
 
 interface LNURLChannelServiceResponse {
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,8 +1,10 @@
 import connectors from "./extension/background-script/connectors";
 
+export type ConnectorType = keyof typeof connectors;
+
 export interface Account {
-  connector: keyof typeof connectors;
-  config: string;
+  connector: ConnectorType;
+  config: string | Record<string, unknown>;
   name: string;
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -2,6 +2,9 @@ import connectors from "./extension/background-script/connectors";
 
 export type ConnectorType = keyof typeof connectors;
 
+// @TODO: https://github.com/getAlby/lightning-browser-extension/issues/652
+// align Message-Types
+// Where is this used? Do we still need this if 652 is solved?
 export interface Account {
   connector: ConnectorType;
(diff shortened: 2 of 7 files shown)


src/types.ts
```diff
@@ -51,6 +51,9 @@ export interface Battery extends OriginData {
 export interface Message {
   args: Record<string, unknown>;
```
escapedcat, line 52: In the `add` case args looks like i.e.:

```js
config: "U2FsdGVkX1+jfieW172FcOgLK0KBKSLNly6zTQ...EPZjw0FRUsr1Y5HfcCyPTMMOZikwcCXm9Hsb5grS"
connector: "lnd"
name: "LND"
```

I assume in other cases this different.  
@dylancom you have an idea how to handle this?

src/extension/background-script/actions/accounts/add.ts
```diff
@@ -11,10 +13,11 @@ const add = async (message, sender) => {
   const password = state.getState().password;
   const currentAccountId = state.getState().currentAccountId;
 
-  const accountId = Object.keys(accounts).length + 1;
-  console.log(`Created account ${accountId}`);
+  if (!password) return;
```
bumi, line 16: I think we should return some error here. Though this should not happen and the error might not be handled well in the UI, we should still indicate the error by returning a error object.

src/types.ts
bumi: hmm. currently it always is `connector` (indicating which connector class to use), `name` (human readable name that we can use in the frontend), and `config` (the actual connector config; encrypted as string)

escapedcat: Ah ok, that helps, will update

escapedcat: - This is expecting a `.message` as well
  https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/webln/signMessageOrPrompt.ts#L8
- This also has `prompt` and `origin`
  https://github.com/getAlby/lightning-browser-extension/blob/master/src/common/lib/utils.ts#L17-L22

Maybe I'm reading it wrong but I would assume this would need to be aligned?

bumi: ah you mean the `args` - sorry somehow I was thinking about the `account` object. 

not sure what's the best typescript way for general/unknown arguments. 
in the `accounts` it needs to be come a `Account` type.

bumi: this needs to get the correct Account type, doesn't it? https://github.com/getAlby/lightning-browser-extension/pull/636/files#diff-0029b68659877433cec643410ec71d719a7cf163141ca11e7b0ced59716530caL5

escapedcat: - Yup, it's here:  
  https://github.com/getAlby/lightning-browser-extension/blob/master/src/types.ts#L51-L54  
- Could add missing props and use this I guess:  
  https://github.com/getAlby/lightning-browser-extension/blob/master/src/types.ts#L51-L54  

Just not sure if this would be correct in the overall app.

src/extension/background-script/actions/accounts/add.ts
```diff
@@ -11,9 +13,11 @@ const add = async (message, sender) => {
   const password = state.getState().password;
   const currentAccountId = state.getState().currentAccountId;
 
-  const accountId = Object.keys(accounts).length + 1;
-  console.log(`Created account ${accountId}`);
+  if (!password) throw new Error(
```
bumi, line 16: I think we need to return something like `{ error: "Password is missing" }`; 
Because of the messaging required from the "frontend" to the backrgound script we can not throw an error I think. (this might be handled on a different level, but I am not sure)

src/types.ts
bumi: 🤔 

```diff
diff --git a/src/extension/background-script/actions/accounts/add.ts b/src/extension/background-script/actions/accounts/add.ts
index bb02736..4853212 100644
--- a/src/extension/background-script/actions/accounts/add.ts
+++ b/src/extension/background-script/actions/accounts/add.ts
@@ -1,10 +1,9 @@
 import type { Runtime } from "webextension-polyfill";
 import { encryptData } from "../../../../common/lib/crypto";
-import type { Message } from "../../../../types";
+import type { Message, Account, ConnectorType } from "../../../../types";
 import state from "../../state";

 const add = async (message: Message, _sender: Runtime.MessageSender) => {
-  const newAccount = message.args;
   const accounts = state.getState().accounts;

   // TODO: add validations
@@ -18,8 +17,11 @@ const add = async (message: Message, _sender: Runtime.MessageSender) => {
   );

   const accountId = crypto.randomUUID();
-  newAccount.config = encryptData(newAccount.config, password);
-  accounts[accountId] = newAccount;
+  accounts[accountId] = {
+    name: message.args.name as string,
+    connector: message.args.connector as ConnectorType,
+    config:  encryptData(message.args.config, password),
+  };

   state.setState({ accounts });

diff --git a/src/types.ts b/src/types.ts
index ef403a2..f97497e 100644
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,7 +1,9 @@
 import connectors from "./extension/background-script/connectors";

+export type ConnectorType = keyof typeof connectors;
+
 export interface Account {
-  connector: keyof typeof connectors;
+  connector: ConnectorType;
   config: string;
   name: string;
 }
```

dylancom: I think the best way for general/unknown arguments is to use Generics: https://www.typescriptlang.org/docs/handbook/2/generics.html

We also use this for e.g. with `utils.call<Accounts>("getAccounts")` where we write what type we expect within `<>`. (libs like Axios do it also like this)

bumi: ahhh. I see. so we define the `args` as generics here?

dylancom: Had a closer look at the code and come to the following conclusion:
First option: [narrowing-the-unknown-type](https://mariusschulz.com/blog/the-unknown-type-in-typescript#narrowing-the-unknown-type)
Second option: [using-type-assertions-with-unknown](https://mariusschulz.com/blog/the-unknown-type-in-typescript#using-type-assertions-with-unknown)

src/extension/background-script/actions/accounts/add.ts
escapedcat: Updated it.

Looking at i.e. this, both is being used:  
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/router.js#L60
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/router.js#L71

Some other places in `background-script` is using both as well.  
@dylancom do you know when which is being used or should be used or how it's propagated in a way?

src/types.ts
escapedcat: Sorry, just noticed that my links above both pointed to the same lines... fixed it.  

> Had a closer look at the code and come to the following conclusion

I wonder if it's possible to know the exact type. The app knows what `Message` will be in each situation, no? If yes, could we define all needed types (and/or variations)?

dylancom: Yes, we could also "assume" that we pass in the correct MessageType and have an "AddAccountMessage" or something, if that's safe enough.

escapedcat: THIS! 🎉   
Will come up with a suggestion for this PR.

bumi:
I've pushed some TypeScript change here. What is missing? 
Would be good to get this in because then we can make a new release

src/types.ts
escapedcat: Thanks for tackling this @bumi .  
For further cleanup I created #652 and mentioned in related places

src/extension/background-script/actions/accounts/add.ts
escapedcat: Can be discussed later on

bumi (approved):
thanks for the adding the TODO! indeed something we need to tackle in this TypeScripting process :)

PR #637, opened 2022-02-26
Split galoy into instance-specific configurations

vindard:
## Description

This PR is to abstract away some of the Galoy-specific details in favor of instance-specific ones.

### Connect Page
Screenshot from 2022-02-26 12-41-44

### Accounts View
Screenshot from 2022-02-26 12-39-08


bumi:
ok, great! thanks. so instance specific connectors it is! looks good to me.

PR #642, opened 2022-02-28
refactor: LNURLPay & MakeInvoice should use the new accessible TextField

dylancom:
Currently labels aren't clickable in those screens and styles are inconsistent.
This makes also the rendering more clear which became too abstract within methods.

### DIFF
--- a/src/app/screens/LNURLPay.tsx
+++ b/src/app/screens/LNURLPay.tsx
@@ -1,4 +1,4 @@
-import { Fragment, useState, useEffect, MouseEvent } from "react";
+import React, { useState, useEffect, MouseEvent } from "react";
 import axios from "axios";
 import { useNavigate, useSearchParams } from "react-router-dom";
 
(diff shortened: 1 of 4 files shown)


src/app/screens/LNURLPay.tsx
```diff
@@ -28,6 +28,14 @@ type Props = {
   origin?: Origin;
 };
 
+const Dt = ({ children }: { children: React.ReactNode }) => (
```
escapedcat, line 31: nit: We do this with `Button` as well, but I feel like naming these as the HTML-tag might be somewhat confusing.  
No need to tackle it here and now. Just a thought.

escapedcat: Might just be a react-coding-style-thing I'm not used to though.

dylancom: It's what I'm used to by known UI libs for e.g.: https://mui.com/components/buttons/
But debatable for sure :)

escapedcat: No worries, thanks for the link. Might be worth looking at some libs.

bumi (approved):
utACK

PR #648, opened 2022-03-01
feat(accounts): add basic rename action #647

### LINKED ISSUE #647: Feature: Add ability to rename account
Splitting up  #404 

Under "account screens" you can see the design:  
https://www.figma.com/file/xEPJFGFdc3HD0JS1ax2IpH/Extension-(MVP)?node-id=1%3A22152

image
image

Screenshots might be outdated. Please double check with the designs in the provided figma link.

escapedcat:
pairing with @dylancom 

fixes #647

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
 # #### This all good standard ###
 
-.nova
-
 # Logs
 logs
 *.log
@@ -133,6 +131,7 @@ dist/
 
 # Editor directories and files
 .idea
+.nova
 .vscode
 *.suo
 *.ntvs*
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,11 @@ dist/
 .awcache
 
 .pnp.*
+
+# Editor directories and files
+.idea
+.vscode
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -87,6 +87,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
             <TextField
               id="budget"
               label="Budget"
+              autoFocus
               placeholder="sat"
               value={budget}
               type="number"
--- a/src/app/screens/Accounts/index.test.tsx
+++ b/src/app/screens/Accounts/index.test.tsx
@@ -1,0 +1,17 @@
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter } from "react-router-dom";
+import Accounts from "./index";
+import { AccountsProvider } from "../../context/AccountsContext";
+
+describe("Accounts", () => {
+  test("render", async () => {
+    render(
+      <AccountsProvider>
+        <MemoryRouter>
+          <Accounts />
+        </MemoryRouter>
+      </AccountsProvider>
+    );
+    expect(await screen.findByText("Accounts")).toBeInTheDocument();
+  });
+});
--- a/src/app/screens/Accounts/index.tsx
+++ b/src/app/screens/Accounts/index.tsx
@@ -8,14 +8,14 @@ import type { FormEvent } from "react";
 import { useState } from "react";
 import Modal from "react-modal";
 import { useNavigate } from "react-router-dom";
-import api from "../../common/lib/api";
-import utils from "../../common/lib/utils";
-import Button from "../components/Button";
-import Container from "../components/Container";
-import TextField from "../components/Form/TextField";
-import Menu from "../components/Menu";
-import { useAccounts } from "../context/AccountsContext";
-import { useAuth } from "../context/AuthContext";
+import api from "../../../common/lib/api";
+import utils from "../../../common/lib/utils";
+import Button from "../../components/Button";
+import Container from "../../components/Container";
+import TextField from "../../components/Form/TextField";
+import Menu from "../../components/Menu";
+import { useAccounts } from "../../context/AccountsContext";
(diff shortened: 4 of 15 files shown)


.gitignore
```diff
@@ -130,3 +130,11 @@ dist/
 .awcache
 
 .pnp.*
+
```
escapedcat, line 133: Unless we have good reasons we don't need to save individual code-editor configs I believe.  
@bumi you added `.nova` further abov. What kind of file is that? Just asking out of cuirosity.

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -87,6 +87,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
             <TextField
               id="budget"
               label="Budget"
+              autoFocus
```
escapedcat, line 90: Added these to inputs in modals to improve UX.  
Turns out it doesn't seem to work. WIll double check and create an issue if needed.

src/app/screens/Accounts/index.test.tsx
```diff
@@ -0,0 +1,17 @@
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter } from "react-router-dom";
+import Accounts from "./index";
+import { AccountsProvider } from "../../context/AccountsContext";
+
+describe("Accounts", () => {
+  test("render", async () => {
+    render(
+      <AccountsProvider>
```
escapedcat, line 9: I think for now this is ok but we should imrpove this:
- Provide a render functions with AccountsProvider and MemoryRouter which can be imported by other tests
- AccountsProvider should provide mocked accounts for further tests

@dylancom hapyy for input/feedback here. I could work on this because I feel like atm I'm blocking releases if I work on high-prio issues.

src/app/screens/Accounts/index.tsx
```diff
@@ -87,8 +109,65 @@ function AccountsScreen() {
...
+                        <h2 className="text-2xl font-bold dark:text-white">
+                          Edit account
+                        </h2>
+                        <button onClick={closeModal}>
+                          <CrossIcon className="w-6 h-6 dark:text-white" />
+                        </button>
+                      </div>
+
+                      <form
+                        onSubmit={(e: FormEvent) => {
```
escapedcat, line 142: Moved action to onSubmit to provide "on Enter" submit

src/custom.d.ts
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+declare type FixMe = any
```
escapedcat, line 2: We use this in other projects to mark "unwanted" `any`s. I just swtiched `route.js` to TS and it needs to be improved with exact types.

tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "target": "es6",                                /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', 'ES2021', or 'ESNEXT'. */
     "module": "commonjs",                           /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
     // "lib": [],                                   /* Specify library files to be included in the compilation. */
-    // "allowJs": true,                             /* Allow javascript files to be compiled. */
+    "allowJs": true,                                /* Allow javascript files to be compiled. */
```
escapedcat, line 10: Adding this to avoid TS errors for JS imports

.gitignore
bumi: I had tried https://nova.app/ editor. .nova I think is their editor config file.

escapedcat: Ah thanks, tried to figure out what it was but too many "nova"s out there.  Will add it to the editor section.

src/app/screens/Accounts/index.tsx
```diff
@@ -1,22 +1,44 @@
...
   const navigate = useNavigate();
 
+  const [modalIsOpen, setIsOpen] = useState(false);
+  const [newAccountName, setNewAccountName] = useState("");
+
+  function closeModal() {
+    setIsOpen(false);
+  }
+
+  async function updateAccountName(accountId: string) {
```
bumi, line 32: it seems the accountId is not passed properly here. 
In my test it was always the same.

I have multiple accounts but the ID here was always the ID of the last account in the list

bumi (changes requested):

src/app/screens/Accounts/index.tsx
escapedcat: Issue is that currently all modals wil be rendered, butyou only see the one on top, which is the last one.  
Will fix.

src/app/screens/Accounts/index.tsx
```diff
@@ -97,6 +130,50 @@ function AccountsScreen() {
...
+            <h2 className="text-2xl font-bold dark:text-white">Edit account</h2>
+            <button onClick={closeModal}>
+              <CrossIcon className="w-6 h-6 dark:text-white" />
+            </button>
+          </div>
+
+          <form
+            onSubmit={(e: FormEvent) => {
+              e.preventDefault();
+              updateAccountName(currentAccountId);
```
bumi, line 152: as we already pass in the `currentAccountId` we could also pass in the `newAccountName`? then this method uses consistently data from the arguments.

bumi (approved):

src/app/screens/Accounts/index.tsx
escapedcat: Good point, done here: #705

Issue #612, opened 2022-02-19
Enhancement: Integrate commitlint with proper commit conventions 


pavanjoshi914:
Alby doesn't have any proper commit conventions that everyone should follow

we can integrate any available linter so that everyone follows the same commit conventions, this will help in
a.  keeping commit history clean for the codebase, 
b. do regressions and analysis of previous commits easy in future. or while using git bisect
c. Commit messages could also be used to determine the next version number according to [semver](https://semver.org/). 

Proposed plugins:
1. https://github.com/conventional-changelog/commitlint 
2.  https://github.com/conventional-commits/conventionalcommits.org (personally I love this)

coding conventions of conventional commits

https://www.conventionalcommits.org/en/v1.0.0/

Once mentors decide which would be best for us, I can start working on it

pavanjoshi914:
@escapedcat

escapedcat:
Some background info: https://slides.com/marionebl/the-perks-of-committing-with-conventions

dylancom:
Let's add in the industry standard: https://github.com/conventional-changelog/commitlint

PR #653, opened 2022-03-05
chore(commitlint): set husky executable #612

escapedcat:
avoid warning: hook was ignored because it's not set as executable  

```sh
hint: The '.husky/commit-msg' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

See: https://newbedev.com/why-is-my-git-pre-commit-hook-not-executable-by-default/

@pavanjoshi914 fyi


pavanjoshi914:
Interesting, I was not aware of this, thanks!. 
I never faced such a warning yet 🤔 

git-hooks documentation::
```Hooks are programs you can place in a hooks directory to trigger actions at certain points in git’s execution. Hooks that don’t have the executable bit set are ignored.```

PR #654, opened 2022-03-05
chore: always require native messaging permission

bumi:
On chrome based browsers we can not request permissions on they fly.
Requesting the native messaging permission requires a restart
This would kick the user out of the onboarding process.
To work around this we now always require the native messaging permission.


bumi:
Merging, but let me know if somebody has comments.

PR #657, opened 2022-03-06
Micro-improving the UX

altoidkind:
For a better view

PR #668, opened 2022-03-08
Removing the shadow in the icon

altoidkind:


escapedcat:
Thanks!  
Please describe what your PR is solving? Can existing issue be referenced?  
Maybe adding a screenshot helps?  
Would be great to understand you PR better to get this moving faster.

src/app/screens/Receive.tsx
```diff
@@ -232,16 +232,6 @@ function Receive() {
                   }
                 />
               </div>
-              <a
```
escapedcat, line 235: How is this related to removing a shadow?

bumi: I think that's a left over from: https://github.com/getAlby/lightning-browser-extension/pull/657

PR #672, opened 2022-03-09
feat: capitalize QR code data

### LINKED ISSUE #665: Use capitilized QR code data
Currently receiving invoices on Alby use lower-cased QR code data. Capitalizing this data makes the QR code smaller and easier to read by lower end devices. 

https://github.com/btcpayserver/btcpayserver/issues/2110 
https://bitcoin.design/guide/payments/request/#qr-code

pavanjoshi914:
#### Link this PR to an issue
Fixes #665 

#### Type of change
- [X ] New feature (non-breaking change which adds functionality)

capitalize QR code data 

1. scanning QR code after making the changes 
2.  trying to see the payment data(using send section) using a capitalized form of payment request scanned from QR code

#### Checklist:


pavanjoshi914:
@bumi

escapedcat:
Would it be possible to test this with a "simple" jest test?

pavanjoshi914:
> Would it be possible to test this with a "simple" jest test?

I haven't used the jest framework before, I will try to do that.  

I think adding a test just because we used the built-in javascript function to uppercase the string might not be much required!

adding a test for the whole class to check its functionality might be fine and can be done separately
WDYT?

pavanjoshi914:
also, I cant see ```.test.tsx ``` test files for most of the components! this might be a separate issue to write unit tests for Alby

escapedcat:
> I think adding a test just because we used the built-in javascript function to uppercase the string might not be much required!
> 
> adding a test for the whole class to check its functionality might be fine and can be done separately WDYT?

> also, I cant see `.test.tsx ` test files for most of the components! this might be a separate issue to write unit tests for Alby

Alright, yeah, not having enough tests in general is the main issue here, that's why I'm trying to push for it.  
Even testing this small PR will add a general render test for the component and that's better than nothing.  

[This is a very basic example](https://github.com/getAlby/lightning-browser-extension/blob/master/src%2Fapp%2Fcomponents%2FBalanceCard%2Findex.test.tsx  ).
Maybe that could could be a start for adding a test for this component later on. Adding tests could be a good entry for contributors as well.

It's ok to not add a test it in this PR.

bumi:
Great! can you do the same in `NewWallet` screen? there we show a QR code to connect the wallet with BlueWallet

pavanjoshi914:
> 

what about the URL which is shown in the text field?  shall we capitalize that too, to have consistency between URL given by qr code and text field

bumi:
no, I think we should leave the input field as it is. I think most URLs are displayed lowercase.

pavanjoshi914:
> no, I think we should leave the input field as it is. I think most URLs are displayed lowercase.

got it! did required changes kindly review and merge

src/app/screens/Onboard/NewWallet/index.tsx
```diff
@@ -126,7 +126,7 @@ export default function NewWallet() {
                 </div>
                 <div className="float-right">
                   <QRCode
-                    value={`lndhub://${lndHubData.login}:${lndHubData.password}@${lndHubData.url}/`}
+                    value={`lndhub://${lndHubData.login}:${lndHubData.password}@${lndHubData.url}/`.toUpperCase()}
```
bumi, line 129: actually. sorry... I am stupid and did not think when I thought about this QR code. 
We must noch change that one as it contains login data which would be altered and would no longer work. 
so we should leave it as it was.

bumi (changes requested):

src/app/screens/Onboard/NewWallet/index.tsx
pavanjoshi914: no worries did the required changes! kindly review and merge

PR #683, opened 2022-03-12
feat(connectors): add raspiblitz

### LINKED ISSUE #673: Add RaspiBlitz connector
In order to make it easier for RaspiBlitz user we want to show them within Alby how to connect their node and where they find the connection settings. 
This could be done in a special RaspiBlitz connector UI with special instructions that then adds a LND connection.

secondl1ght:
#### Link this PR to an issue
Fixes #673 

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

Begin to add the RaspiBlitz connector, I repurposed the LND one for this because the backend should be the same. I still need to add the RaspiBlitz specific steps to this so if anyone knows those can they please post them here.

image

I added a LND node successfully.

#### Checklist:

- [x] I have made corresponding changes to the documentation (If Any)


secondl1ght:
Note: This PR is not ready to be merged, we need to add the Raspi specific details first.

escapedcat:
> Note: This PR is not ready to be merged

Wanna switch it to `draft` till it's ready to indicate it?

secondl1ght:
I don't see the button to do that..?

escapedcat:
image

Upper right, or when you open a PR you can select it here:  
image

secondl1ght:
Ok I have added the instructions for this and it is now ready for review, @MoritzKa do you want to verify everything looks correct?

image

MoritzKa:
Thanks. I´ve two points to report
1. adjust instruction to ...":80**8**0 at the end"
2. I could not test the connection. The build has different ID than the companion app is configured for. 
image

bumi:
@MoritzKa you can change the ID in that JSON file.

MoritzKa:
Changed the ID. Connection works fine!

bumi:
We don't need the macaroon drag&drop here - we have specific instructions how users get the macaroon hex above. So we should remove that.

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+            <br />
+            Paste the URL in the input below.
+          </p>
+          <div className="w-4/5">
+            <div className="mt-6">
+              <TextField
+                id="url"
+                label="REST API host and port"
+                placeholder="https://your-node:8080"
+                onChange={handleChange}
```
bumi, line 151: can you sync that with the latest textfield from lnd? 
there is an additional `pattern` validation. 

(that's why code needs to be DRY :D)

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+              <div>
+                <TextField
+                  id="macaroon"
+                  label="Macaroon (HEX format)"
+                  value={formData.macaroon}
+                  onChange={handleChange}
+                  required
+                />
+              </div>
+              <p className="text-center my-4 dark:text-white">OR</p>
```
bumi, line 173: this whole `OR` we don't need as we instruct users to copy the macaroon HEX.

secondl1ght:
Ok I have made all the changes, have a look:

image

MoritzKa:
ACK

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+            You need your node URL and a macaroon with read and send permissions
+            (e.g. admin.macaroon).
+            <br />
+            <br />
+            <b>SSH</b> into your <b>RaspiBlitz</b>.<br />
+            Run the command <b>sudo cat /mnt/hdd/tor/lndrest8080/hostname</b>.
+            <br />
+            Copy your <b>.onion</b> address.
+            <br />
+            Add <b>https://</b> at the beginning and <b>:8080</b> at the end.
```
bumi, line 96: we actually could easily do that dynamically. 
always add https:// and have a second port field.
on change we then save `https://${url}:${port}` as `url`.

the user does not need to manually add.

bumi (changes requested):

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
secondl1ght: Great idea bumi!

secondl1ght:
I updated the code so that https:// and port:8080 get added automatically to keep it as simple as possible for the user.

bumi:
now you should test it and check what you broke.

secondl1ght:
Bugs fixed and tested - can now add credentials for both inputs.

MoritzKa:
ACK. Tested the connection successfully.Thanks for your work!

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
```diff
@@ -16,10 +16,17 @@ export default function ConnectRaspiBlitz() {
   const [loading, setLoading] = useState(false);
 
   function handleChange(event: React.ChangeEvent<HTMLInputElement>) {
-    setFormData({
-      ...formData,
-      [event.target.name]: `https://${event.target.value.trim()}:8080`,
-    });
+    if (event.target.name === "url") {
```
bumi, line 19: I'd say whenever you have a function that consists of one big`if` that's an indication that the function should be split up. (generally it's good to avoid `if` as much as possible).
This makes the function smaller, doing less and thus also easier to test. 

so in this case write a `handleUrl` or something that only handles the URL and use a different function for the macaroon. 
This makes it also less likely to silently break if somebody potentially changes the input name (from url to something eles). - your code would fail silently and it would not be noticed.

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
```diff
@@ -16,10 +16,17 @@ export default function ConnectRaspiBlitz() {
...
+      setFormData({
+        ...formData,
+        [event.target.name]: `https://${event.target.value.trim()}:8080`,
```
bumi, line 22: maybe we also want to have the port in an input? can the user change the port in raspiblitz? - if not then it's fine.

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
```diff
@@ -101,8 +108,8 @@ export default function ConnectRaspiBlitz() {
                 id="url"
                 label="REST API host"
                 placeholder="your-node-url"
-                pattern="https://.+"
-                title="https://your-node-url:8080"
+                pattern="^[A-Za-z0-9]*\.onion$"
```
bumi, line 109: is it always the .onion address? can we be sure about that?

bumi (changes requested):
great, got some more comments

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
MoritzKa: no we cannot be sure. Here it works only for nodes with a onion address.
What type of address does the user need if he wants to connect a public node?

bumi: the public IP address.
but then let's just not add this .onion validation and leave it open.

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
secondl1ght: @MoritzKa do you know if the user can change the port?

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
secondl1ght: Ok but now we also need to update the instructions - currently they are instructing the user to use the .onion address. Can we add the other instructions to get the non-tor connection info?

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
secondl1ght: Sounds good, I thought about making 2 functions for this but wasn't sure if that was the right thing to do. Will update now!

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
secondl1ght: Ok I have removed the validation but I don't think any users will assume they can use clearnet with the current instructions because it says 'paste the .onion address below'. But i'll post a screenshot of the latest instructions so you can take a look.

secondl1ght:
I have made the requested changes, just waiting to hear back about the Port question. Here is an updated screenshot for reference:

image

secondl1ght:
@MoritzKa let me know that users can change the port number, so I will remove the fixed 8080 port.

secondl1ght:
Updated screenshot:

image

Please let me know what you think!

MoritzKa:
but there are users that have a public IP and did not enable Tor. Right now they think that they cannot use Alby.
Why not writing this:
If Tor is not activated and you have a public IP address for your RaspiBlitz 
Run the command lncli getinfo and copy the address in the input below.

bumi:
because it is not enough and not working out of the box.

MoritzKa:
I see. Thanks for the clarification. That means we would provide anyway dedicated support for this type of user.
Let's not make it more complicated and learn how many non Tor RaspiBlitz are actually out there until we provide a proper onboarding for them. 
Let's get this connector out.

bumi:
Generally there are many ways how people can configure their nodes. So all these connectors just try to make it easier but can not catch all the different cases. 
We can link to guides and try to have as many guides as possible for the different setups.

bumi:
one more thing: while we are there. 

you can check if the url starts with `http` and if NOT add `https` by default, but if it starts with `http` then we use it like that. 
(and we also don't need the note in the instructions of not adding http)

bumi:
Sadly there have been some refactorings (now making these screens much easier) but that means this PR has conflicts.
@secondl1ght can you update it? best check how the other connector screens now work.

secondl1ght:
Ok I refactored the code so that it now checks for `http` and adds `https` if the user does not enter that themselves.

All that we need to do now is resolve the merge conflicts - but for some reason GitHub won't let me resolve them. The 'resolve conflicts' button is disabled for me...

src/app/screens/Onboard/ConnectRaspiBlitz/index.tsx
```diff
@@ -16,10 +16,17 @@ export default function ConnectRaspiBlitz() {
   const [loading, setLoading] = useState(false);
 
   function handleUrl(event: React.ChangeEvent<HTMLInputElement>) {
-    setFormData({
-      ...formData,
-      [event.target.name]: `https://${event.target.value.trim()}`,
-    });
+    if (event.target.value.substring(0, 4) === "http") {
```
bumi, line 19: I'd do something like this. It is not good to duplicate the function calls here, just for adding the `https://` - just put it in a variable and update that if needed. 

Everytime you duplicate some code ask if that is really needed or if it can somehow be refactored. 
Duplication easily leads to errors. e.g. when you want to update this code here and miss one of the setFormData calls... 

```js
let url = event.target.value.trim();
if (event.target.value.substring(0, 4) !== "http") {
  url = `https://${url}`;
}
...
```
this should work, doesn't it?

secondl1ght: Yep I just tested it and it works, I wasn't sure if it would but it does. I didn't think to create a variable in this way to DRY this function. Thanks for the suggestion!

secondl1ght:
Ok so resolving the merge conflicts was a bit of a hassle! GitHub wouldn't let me do it on the website because it said the conflicts were too complex. So I did it locally, I ended up having to create a new branch on this repo and merging this PR onto master on that new branch. I then tried to push those changes back to this PR but it looks like something happened during that so this PR is now not correct. Easiest solution was to to create a new PR from the new branch and I did that with #777. We can leave this PR open to track the conversation history but we should merge #777 and then close this one when it gets merged. Hopefully what I explained here makes sense.

PR #694, opened 2022-03-14
refactor: improve connector errors

dylancom:
#### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

Instead of resolving with error messages we should throw errors. This currently leads to weird things such as **SendPaymentResponse** being two responses.


bumi:
ahh, I think we have to be careful there... due to that messaging between the content script and the background script we can not have real error objects there. We need to pass a object..
This means the error somewhere must be transformed into an object with an `error` key... not sure if this happens somewhere?

bumi (approved):
good convention: 

make only the actions return the correct response. so an object with either a `data` key or `error` key. then no other parts of the app need to know about that. 

👍

PR #699, opened 2022-03-16
Feature/keysend

### LINKED ISSUE #671: Keysend
Keysend is a way of sending spontaneous payments over Lightning, using just a public key and an amount as input.
Currently, the main use-case are "Value4Value" Podcasts, enable by podcasts apps like [Breez](https://medium.com/breez-technology/podcasts-on-breez-streaming-sats-for-streaming-ideas-d9361ae8a627), [Fountain](https://www.fountain.fm/) or [Sphinx](https://sphinx.chat/). An index of Value4Value podcasts can be found on [Podcastindex](https://podcastindex.org).
However, there is no web-based value4value podcast player.

Integrating keysend into Alby and into the webLN standard would allow for easy web-based value4value podcast players.
The keysend integration can be added to the following connectors:
- Alby wallet: [lndhub.go](https://github.com/getAlby/lndhub.go/blob/main/main.go#L137) now supports keysend, this is already live in the default Alby wallet in the extension.
- LND connector: for an example on how to do this, check out [Blixt wallet](https://github.com/hsjoberg/blixt-wallet/blob/9fcc56a7dc25237bc14b85e6490adb9e044c009c/src/lndmobile/index.ts#L251-L270).
- Eclair connector: [API docs](https://acinq.github.io/eclair/#sendtonode).
- LNBits connector: I don't think this is supported in the LNBits API, so this would not be possible (correct me if I'm wrong).
- Legacy LNDhub: not possible, but we probably would add the same endpoint as lndhub.go anyway, in case users want to connect to their own lndhub.go.

We would also need to extend the webLN spec to include support for keysend payments, so websites can integrate it.

kiwiidb:
#### Link this PR to an issue
Fixes #671 
#### Type of change

- [x] New feature (non-breaking change which adds functionality)

- Added a new keysend screen that is mostly copied over from the LNurlPay screen. This screen is triggered either by pasting in a pubkey in the send box (by checking for length and starting characters), or by a `webln.keysend` call.
- Added more fields to the `sendPaymentArgs` to allow for keysend and offer payments (later).
- Implements the keysend method for the `LND` and `LNDhub` connectors.
- Adds a webLN keysend function.
Schermafbeelding 2022-03-16 om 08 55 52
- webLN tested on https://amboss.space - check out the keysend billboard. Tested with lndhub.go backend.
- by pasting in random pubkeys and doing payments. Get a pubkey from lnd-2 on rtl.regtest.getalby.com - connect to alby-simnet-lnd-1 in the extension - make payment. Check status on RTL.

Schermafbeelding 2022-03-16 om 09 05 57

I don't have any tests and also the code could use some refactoring since I have copy-pasted quite a bit. I think it would be more opportune if someone else picks this up.

#### Checklist:

- [ ] I have made corresponding changes to the documentation (If Any)

### DIFF
--- a/src/extension/background-script/actions/webln/keysendOrPrompt.ts
+++ b/src/extension/background-script/actions/webln/keysendOrPrompt.ts
@@ -1,0 +1,24 @@
+import { Message } from "../../../../types";
+import {
+  checkAllowance,
+  sendPaymentWithAllowance,
+  payWithPrompt,
+} from "./sendPaymentOrPrompt";
+
+const keysendOrPrompt = async (message: Message) => {
+  const destination = message.args.destination;
+  const amount = message.args.amount;
+  if (typeof destination !== "string" || typeof amount !== "string") {
+    return {
+      error: "Destination or amount missing.",
+    };
+  }
+
+  if (await checkAllowance(message.origin.host, parseInt(amount))) {
+    return sendPaymentWithAllowance(message, true);
+  } else {
+    return payWithPrompt(message, "keysend");
+  }
+};
+
+export default keysendOrPrompt;
--- a/src/extension/background-script/connectors/connector.interface.ts
+++ b/src/extension/background-script/connectors/connector.interface.ts
@@ -43,6 +43,10 @@ export type SendPaymentResponse =
 
 export interface SendPaymentArgs {
   paymentRequest: string;
+  offer: string;
+  pubkey: string;
+  amount: number;
+  memo: string;
 }
 
 export interface CheckPaymentArgs {
@@ -88,6 +92,7 @@ export default interface Connector {
   getBalance(): Promise<GetBalanceResponse>;
   makeInvoice(args: MakeInvoiceArgs): Promise<MakeInvoiceResponse>;
   sendPayment(args: SendPaymentArgs): Promise<SendPaymentResponse>;
+  sendPaymentKeySend(args: SendPaymentArgs): Promise<SendPaymentResponse>;
   checkPayment(args: CheckPaymentArgs): Promise<CheckPaymentResponse>;
   signMessage(args: SignMessageArgs): Promise<SignMessageResponse>;
   verifyMessage(args: VerifyMessageArgs): Promise<VerifyMessageResponse>;
--- a/src/extension/background-script/connectors/lnd.ts
+++ b/src/extension/background-script/connectors/lnd.ts
@@ -81,7 +81,11 @@ class Lnd implements Connector {
       };
     });
   }
-
+  async sendPaymentKeySend(
+    args: SendPaymentArgs
+  ): Promise<SendPaymentResponse> {
+    throw new Error("not supported");
+  }
   async checkPayment(args: CheckPaymentArgs): Promise<CheckPaymentResponse> {
     const data = await this.request<{ settled: boolean }>(
       "GET",
--- a/src/extension/background-script/connectors/lnd.ts
+++ b/src/extension/background-script/connectors/lnd.ts
@@ -1,7 +1,7 @@
 import Base64 from "crypto-js/enc-base64";
-import Hex from "crypto-js/enc-hex";
 import UTF8 from "crypto-js/enc-utf8";
-import sha256 from "crypto-js/sha256";
+import CryptoJS from "crypto-js";
+import _ from "lodash";
 import utils from "../../../common/lib/utils";
 import Connector, {
   SendPaymentArgs,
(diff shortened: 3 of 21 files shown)


src/extension/background-script/connectors/lnd.ts
```diff
@@ -81,7 +83,58 @@ class Lnd implements Connector {
...
+      payment_route: { total_amt: number; total_fees: number };
+      payment_error?: string;
+    }>(
+      "POST",
+      "/v1/channels/transactions",
+      {
+        dest: dest_pubkey_base64,
+        amt: args.amount,
+        payment_hash: hash,
+        dest_custom_records: {
```
bumi, line 119: can we allow passing in of this custom records? that developers can decide what records to set? 

e.g. I think the podcasting uses a special one: https://github.com/lightning/blips/blob/master/blip-0010.md

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -88,6 +92,7 @@ export default interface Connector {
   getBalance(): Promise<GetBalanceResponse>;
   makeInvoice(args: MakeInvoiceArgs): Promise<MakeInvoiceResponse>;
   sendPayment(args: SendPaymentArgs): Promise<SendPaymentResponse>;
+  sendPaymentKeySend(args: SendPaymentArgs): Promise<SendPaymentResponse>;
```
bumi, line 95: I'd call this `keysend`? why add `sendPayment`?

src/extension/background-script/connectors/lnd.ts
```diff
@@ -81,7 +83,58 @@ class Lnd implements Connector {
       };
     });
   }
+  async sendPaymentKeySend(
+    args: SendPaymentArgs
```
bumi, line 87: those should be KeysendArgs - it's different to what sendpayment accepts

src/extension/background-script/connectors/lnd.ts
```diff
@@ -1,5 +1,7 @@
 import Base64 from "crypto-js/enc-base64";
 import UTF8 from "crypto-js/enc-utf8";
+import CryptoJS from "crypto-js";
```
bumi, line 3: as we use import the specific libs above, we should do that here, too.
```
import sha256 from "crypto-js/sha256";
```

src/extension/background-script/connectors/lnd.ts
```diff
@@ -81,7 +83,58 @@ class Lnd implements Connector {
...
+    const dest_pubkey_hex = args.pubkey;
+    const dest_pubkey_base64 = Buffer.from(dest_pubkey_hex, "hex").toString(
+      "base64"
+    );
+    console.log(dest_pubkey_base64);
+
+    for (let i = 0; i < 32; i++) {
+      const randomNumber = _.random(0, 255);
+      byteArray[i] = randomNumber;
+    }
```
bumi, line 101: we should move this into a utils function: `utils.randomHex(32)` or something. 
@dylancom do you know a nicer way on how to get a random hex string?

kiwiidb:
I've fixed @bumi 's comments.

However the code currently does not work, I'm not sure on how to pass in a map all the way to the network call. I used a TypeScript `Map`, but this won't serialize into json. How to best do this?

Also, would we only allow this to be called by webln, or would we also allow people just pasting in a pubkey? Would we allow custom records in the latter case? Or just a comment (`TLV_WHATSAT_MESSAGE`)?

dylancom:
> I've fixed @bumi 's comments.
> 
> However the code currently does not work, I'm not sure on how to pass in a map all the way to the network call. I used a TypeScript `Map`, but this won't serialize into json. How to best do this?
> 
> Also, would we only allow this to be called by webln, or would we also allow people just pasting in a pubkey? Would we allow custom records in the latter case? Or just a comment (`TLV_WHATSAT_MESSAGE`)?

I would suggest an object of shape: Record<string, string>.

kiwiidb:
- Left out the option of a manual comment.
- Keysend can now only be triggered from webLN:
```
await webln.keysend({destination: "025c1d5d1b4c983cc6350fc2d756fbb59b4dc365e45e87f8e3afe07e24013e8220", amount: "100", customRecords: {"test123":"test"}})
```

src/extension/background-script/actions/webln/keysendOrPrompt.ts
```diff
@@ -0,0 +1,24 @@
...
+  const destination = message.args.destination;
+  const amount = message.args.amount;
+  if (typeof destination !== "string" || typeof amount !== "string") {
+    return {
+      error: "Destination or amount missing.",
+    };
+  }
+
+  if (await checkAllowance(message.origin.host, parseInt(amount))) {
+    return sendPaymentWithAllowance(message, true);
```
bumi, line 18: I am not a fan of magic boolean flags. When I read this code here it is unclear what this `true` means. 
looking into `sendPaymentWithAllowance` this function basically now does two things based on the boolean flag. 
I would split this up and use here a new `keySendWithAllowance` method. it is more descriptive and we don't need to add some `if` in those methods sendPayment methods

bumi, line 18: I am not a fan of magic boolean flags. When I read this code here it is unclear what this `true` means. 
looking into `sendPaymentWithAllowance` this function basically now does two things based on the boolean flag. 
I would split this up and use here a new `keySendWithAllowance` method. it is more descriptive and we don't need to add some `if` in those methods sendPayment methods

kiwiidb:
I've added some commits that handle the encoding of the custom records. The custom record values passed in through webln should now be plain strings (these strings can be serialized json). Depending on the connector used, they are either encoded as base64 (LND connector) or as hex (LNDhub connector).

PR #707, opened 2022-03-18
feat: add native connector for lnbits

bumi:
This allows LnBits connections through Tor

#### Type of change

-  New feature (non-breaking change which adds functionality)

#### Checklist:

- [ ] I have made corresponding changes to the documentation (If Any)


src/extension/background-script/connectors/nativelnbits.ts
```diff
@@ -0,0 +1,86 @@
...
+  id: string;
+  method: string;
+  url: string;
+  body?: string;
+  params?: Record<string, unknown>;
+  headers: Record<string, string>;
+};
+
+export default class NativeLndHub extends NativeConnector {
+  _nativeRequest(postMessage: PostMessage) {
```
bumi, line 18: @dylancom so far we haven't had this `_nativeRequest` function in the Native class. why not? can we move it there it is the same in `nativelndhub` I think

dylancom: I think because nativelnd.ts isn't integrating it?

PR #708, opened 2022-03-18
Some keysend suggestions

bumi:
This mainly now uses two different screens: 
one for the prompt (which is initiated from webln)  (`ConfirmKeysend`)
and one for manual keysend payments (`Keysend`)

on the manual screen users can enter amounts whereas the webln call must provide an amount which is not changeable by the user. 

I think this clear separation makes the code a bit easier to understand and we should do this also for the current `ConfirmPayment` screen and split it up.


src/extension/background-script/connectors/lnd.ts
```diff
@@ -85,24 +84,16 @@ class Lnd implements Connector {
     });
   }
   async keySend(args: KeysendArgs): Promise<SendPaymentResponse> {
-    //See: https://gist.github.com/dellagustin/c3793308b75b6b0faf134e64db7dc915
```
kiwiidb, line 88: Why not leave this comment in? I added it to make clear where I got this code sample from.

kiwiidb (approved):

src/extension/background-script/connectors/lnd.ts
bumi: ups. of course.

PR #709, opened 2022-03-18
fix: react modals should get focus after opening

dylancom:
Modals triggered from a `<Menu />` currently don't receive focus.
This is because @headless-ui/menu restores focus after closing a menu, to the button that opened it.
There is currently no way to change this behavior so a hack is needed.

More info + added comment: https://github.com/tailwindlabs/headlessui/issues/259

Thanks @escapedcat for noticing it.

### DIFF
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -24,9 +24,12 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
 
   function openModal() {
     setBudget(allowance.totalBudget.toString());
-    // HACK!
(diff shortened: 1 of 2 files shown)


src/app/components/AllowanceMenu/index.tsx
```diff
@@ -24,7 +24,12 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
 
   function openModal() {
     setBudget(allowance.totalBudget.toString());
-    setIsOpen(true);
+    // HACK!
```
escapedcat, line 27: Wanna mention the issue in these places?

escapedcat (approved):

PR #710, opened 2022-03-19
refactor: use normal list instead of checked list in pull request template

### LINKED ISSUE #702: Lists appearing as tasks in pull request templates
list items present in pull request templates appearing as tasks in pull request overview

1. Use  double space in ```-[  ]``` brackets so that only checked items appear as a task in the PR overview, and unchecked items remain still without appearing in the PR overview

2. if we don't want checked items not appearing in pr overview as well,  Instead of using lists which can be checked, use numbered list and contributors can decide what to keep and what not to keep depending on the pr created.

**Additional context**
before 
image

after :
only checked items appear in the PR overview
image

for second  solution, no task status will appear in pr overview

pavanjoshi914:
#### Link this PR to an issue
Fixes #702 

1. remove example of linking of the issue (so that PR which is not linked with any issue, doesn't get linked to the issue number shown in the example)
2. use a normal list instead of a checked list
3. for the checklist section still using a checklist to enforce certain tasks
4. for checklist section using double spaces inside square brackets so that only checked  item appear in pull request overview (

#### Checklist:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,12 @@
 #### Link this PR to an issue
-Fixes #ISSUE-NUMBER (eg. Fixes #147)
+Fixes #ISSUE-NUMBER
 
-#### Type of change
+#### Type of change (Remove other not matching type)
 
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- [ ] Documentation update
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- Documentation update
 
 #### Describe the changes you have made in this PR -
 
@@ -16,12 +16,8 @@ Fixes #ISSUE-NUMBER (eg. Fixes #147)
 
 Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
 
-- [ ] Test A
-- [ ] Test B
-
-
 #### Checklist:
 
 - [ ] My code follows the style guidelines of this project and performed a self-review of my own code
 - [ ] New and existing tests pass locally with my changes
-- [ ] I have made corresponding changes to the documentation (If Any)
+- [ ] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)


pavanjoshi914:
@escapedcat @bumi

.github/pull_request_template.md
```diff
@@ -16,12 +16,12 @@ Fixes #ISSUE-NUMBER (eg. Fixes #147)
 
 Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
 
-- [ ] Test A
-- [ ] Test B
+- Test A
```
escapedcat, line 19: We could just remove these I think. People can list however they tested their PR

.github/pull_request_template.md
```diff
@@ -16,12 +16,12 @@ Fixes #ISSUE-NUMBER (eg. Fixes #147)
...
+- Test B
 
 
 #### Checklist:
 
-- [ ] My code follows the style guidelines of this project and performed a self-review of my own code
-- [ ] New and existing tests pass locally with my changes
-- [ ] I have made corresponding changes to the documentation (If Any)
+- [  ] My code follows the style guidelines of this project and performed a self-review of my own code
```
escapedcat, line 25: I'm worried people might get these wrong and will not check/uncheck/correct these.  
For the checklist I would be ok to leave these as tasks. We can use these to remind people to check these.
@bumi what do you think?

pavanjoshi914: yeah, the problem is with the last one! documentation not necessary for every pr other two is a must. so if we keep as it is we will always have ```2 out of 3 tasks done

escapedcat: Hm, maybe we can change it to something like "I checked if I need to made corresponding changes to the documentation (and made these if needed)" so it can be checked also if no changes need to be done? What do you think?

bumi: yep, good idea maybe we can indicate that comments/docs should be checked updated or written?

bumi: and maybe we should add something non code releated:

```
- [ ] I have made a somersault
- [ ] I've been excellent today
```

:D

escapedcat: - [ ] I like 🐢🐢🐢

pavanjoshi914: - [X] My code follows the style guidelines of this project and performed a self-review of my own code
> I checked if I need to made corresponding changes to the documentation (and made these if needed)

pavanjoshi914: what do you think

escapedcat: I would prefer a checkbox to keep the same style.  Everything else might be confusing for people.

pavanjoshi914: > I would prefer a checkbox to keep the same style. Everything else might be confusing for people.

yeah true! so single spaces in brackets with the last statement changed!

- [X] I checked if I need to make corresponding changes to the documentation (and made these if needed)

pavanjoshi914: also if all three are now must what if we keep them already checked! will it be effective?

.github/pull_request_template.md
pavanjoshi914: removed!

.github/pull_request_template.md
pavanjoshi914: did the suggested changes!

pavanjoshi914:
@escapedcat  kindly review again

escapedcat (approved):

pavanjoshi914:
@bumi  kindly review and merge

bumi (approved):

Issue #600, opened 2022-02-10
Internationalization / Translations


bumi:
In order to make it easier for people we should localize the app and provide translations. 
Opening this issue to track how we can approach this.

escapedcat:
[raspiblitz-web](https://github.com/cstenglein/raspiblitz-web) ist running well with a [react-i18next](https://github.com/i18next/react-i18next) and [weblate](https://weblate.org/en/) setup:  
https://github.com/cstenglein/raspiblitz-web/tree/master/src/i18n

escapedcat:
@pavanjoshi914 What setup would you suggest? You had a look at the raspiblitz-web project?

pavanjoshi914:
> @pavanjoshi914 What setup would you suggest? You had a look at the raspiblitz-web project?

yeah, I just want to ask, if it is feasible to work on this at this moment. If yes, I will propose a detailed plan about, which setup we can integrate, locale management, translation key management , i18n rules for naming of keys etc

escapedcat:
> yeah, I just want to ask, if it is feasible to work on this at this moment. If yes, I will propose a detailed plan about, which setup we can integrate, locale management, translation key management , i18n rules for naming of keys etc

Sounds great! I assume because this issue exists you're good to go ahead and propose a draft-setup for further discussions. Would that be ok?

pavanjoshi914:
> > yeah, I just want to ask, if it is feasible to work on this at this moment. If yes, I will propose a detailed plan about, which setup we can integrate, locale management, translation key management , i18n rules for naming of keys etc
> 
> Sounds great! I assume because this issue exists you're good to go ahead and propose a draft-setup for further discussions. Would that be ok?

okay thanks!

bumi:
I did not look into any of the options, yet. But you know it better anyway...
Suggestions and PRs on this are for sure welcome! thanks.

pavanjoshi914:
sorry for the late! I had my semester exams so didn't have the bandwidth to work on this.
@bumi @escapedcat  here is the proposed i18n architecture for Alby
https://gist.github.com/pavanjoshi914/a65af473209541c3a5cfecbac0f6e70f

after a few revisions, this will also act as developer documentation for i18n in Alby! I will start implementation from now onwards. suggest any changes (if Any)

escapedcat:
Setup was merged 🎉

PR #711, opened 2022-03-19
feat(i18n): initial setup

pavanjoshi914:
Solves #600 

#### Type of change

 New feature (non-breaking change which adds functionality)

setup i18next react-i18next browser-language-detector
create locale switcher
styling for both light and dark mode
translate heading
enable JSON imports in config
language persistence throughout the component
language persistence between different sessions
browser language detections
format JSON translation file using built-in eslint JSON formatted  and run automated using lintstaged

image
image
image
image

## Updates
locale switcher shifted under settings
image
locale switcher on Intro shown only under development mode
store locale in extension's storage as well

#### Checklist:


pavanjoshi914:
@bumi  @escapedcat initial setup for i18n

package.json
```diff
@@ -34,6 +34,7 @@
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "prepare": "husky install"
+
```
escapedcat, line 37: This empty line can be removed

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
```diff
@@ -0,0 +1,25 @@
+import { useState, ChangeEvent } from "react";
```
escapedcat, line 1: Separate these into `import ...` and `import type ...`

src/i18n/locales/en.json
```diff
@@ -0,0 +1,5 @@
+{
```
escapedcat, line 1: Can these json-file be formatted?

src/app/components/Navbar/Navbar.tsx
```diff
@@ -39,6 +40,7 @@ export default function Navbar({ title, subtitle, children }: Props) {
         </div>
       )}
       <div className="md:w-4/12 lg:w-3/12 flex justify-end items-center">
+        <LocaleSwitcher />
```
escapedcat, line 43: Not sure if we need this in the navbar or if it would be ok to have this i.e. under settings

pavanjoshi914:
if you have gone through the documentation! either we can use [nesting of keys](https://gist.github.com/pavanjoshi914/a65af473209541c3a5cfecbac0f6e70f#using-nesting-of-keys)  or [react i18next namespaces](https://gist.github.com/pavanjoshi914/a65af473209541c3a5cfecbac0f6e70f#alternate-for-nesting-of-keys)

currently, I  am using namespace. if we wish to use nesting of keys as a key management strategy we can do that as well

escapedcat (commented):
Really cool, thanks! Some minor comments.  
@bumi do we want use [Weblate](https://weblate.org/) so others can help with further translations?

src/app/components/Navbar/Navbar.tsx
pavanjoshi914: this will be a case of dropdown under dropdown! mostly locale switchers shall be handy so I thought navbar shall be the best place! 
keeping it in settings and then it might happen that user never knows that we support multiple languages

bitcoinuser: I prefer on settings too.

escapedcat: Sure, I understand. Just thinking about this regarding things being stuffed in the navbar and if this is really needed or not. If the user is already using the browser in a certain language and alby is using that as a default then chances are high the user doesn't need to adjust the language all the time. Just wanted to give this a thought.  
@bumi @MoritzKa what do you think?

bumi: was thinking the same. I guess the user does not change the language often. I'd prefer to save the user's language in the settings. If we can default to the browser's language then this should be fine, isn't it?

package.json
pavanjoshi914: done

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
pavanjoshi914: done

src/i18n/locales/en.json
pavanjoshi914: yes its possible eslint provides built in json formatter. done

src/app/components/Navbar/Navbar.tsx
pavanjoshi914: okay so moving locale switcher under settings

package.json
```diff
@@ -33,11 +33,10 @@
     "test:coverage": "jest --coverage --collectCoverageFrom='src/**/*.(js|ts|jsx|tsx)'",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "prepare": "husky install"
-    
+    "prepare": "husky install"  
   },
   "lint-staged": {
-    "src/**/*.{js,jsx,ts,tsx}": [
+    "src/**/*.{js,jsx,ts,tsx,json}": [
```
escapedcat, line 39: 🙌

src/app/components/Navbar/Navbar.tsx
pavanjoshi914: I moved the locale switcher under settings! shall we preserve the locale along with other settings?

pavanjoshi914: the problem I am getting is I am not able to do ```saveSettings``` from the intro page. where locale switcher is present as well.

pavanjoshi914: as we are using a browser language detector! so this will handle the case for new users! 

for existing users locale is persisted by default which is changed from language switcher, so do we really need to store locale in local browser storage?

escapedcat: I think we can go with not storing it for now and adjust later if needed. Shouldn't block this.

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
```diff
@@ -0,0 +1,26 @@
+import { useState } from "react";
+import type { ChangeEvent } from "react";
+import i18n from "i18next";
+// import i18n (needs to be bundled ;))
```
escapedcat, line 4: I don't understand this comment

escapedcat (approved):
Awesome job @pavanjoshi914 !

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
pavanjoshi914: we use a single JSON file per locale, but when react i18next bundles everything into a single object {en: {} , hi:{}} something like that. so that comment is pointing something like this.

pavanjoshi914:
Squashed all the commits into a single commit. @bumi  PR is ready to be merged!

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
escapedcat: Ah ok, do we need that comment though? Does it explain anything the import doesn't?

pavanjoshi914: not much but standard documentation keeps this so didn't change that. if you wish I can remove that for sure

pavanjoshi914:
> Really cool, thanks! Some minor comments. @bumi do we want to use [Weblate](https://weblate.org/) so others can help with further translations?

regarding weblate! currently, we can just focus on completing i18n with a single alternate language. setup for continuous localization can be done once we are finished with this
https://gist.github.com/pavanjoshi914/a65af473209541c3a5cfecbac0f6e70f#continuous-localization

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
escapedcat: Ah I see, it's from here:  
https://react.i18next.com/latest/using-with-hooks  
I think it's good to have it in the docs, but it doesn't add any value here.

pavanjoshi914: removed that

escapedcat:
Applied for the "Libre hosting" plan on weblate already

bumi (changes requested):
I think we should save the selected language in the settings storage. Otherwise it will get lost after the app reloads. 

You can see how to save the settings there: https://github.com/getAlby/lightning-browser-extension/blob/4a8d6336acdbdd2a0848426d404cf79a71f7404b/src/app/screens/Settings.tsx#L42

pavanjoshi914:
> I think we should save the selected language in the settings storage. Otherwise it will get lost after the app reloads.
> 
> You can see how to save the settings there:
> 
> https://github.com/getAlby/lightning-browser-extension/blob/4a8d6336acdbdd2a0848426d404cf79a71f7404b/src/app/screens/Settings.tsx#L42

sure so we will have incoming locale from three places
1. browser
2. language switcher
3. settings

Case 1. storing browser detected language or fallback language when language is not available 
when a user installs the app we let the browser detector, detect the language and use that language. either browser detected language will be used and if not available fallback language ```:en ``` will be used

```resolvedLanguage``` can handle this case, we exported the final resolved language under default settings

```
export const DEFAULT_SETTINGS = {
  websiteEnhancements: true,
  userName: "",
  locale: i18n.resolvedLanguage,
};
```

Case 2:  setting language changed via locale switcher

1. locale switcher under settings
passing saveSetting function as props and executed in ```LocaleSwitcher.tsx```

```
<LocaleSwitcher saveSettings={saveSetting} />
```
```
 if (dropdownLang !== newLanguage) {
      setDropdownLang(newLanguage);
      i18n.changeLanguage(newLanguage);
      props.saveSettings({ locale: newLanguage });
    }
```

2. locale switcher under intro section

I can't pass a function as props for this locale switcher. one idea I think is refactor and make function accessible


Before I knew just the basics of react before, I am still learning :)  what can be the best possible approach for handling case 2 @bumi  @escapedcat  any ideas?

**Comment by bumi** [comment]
Can you elaborate on `2.`? Why doesn't it work the same way as in the settings on the welcome page?

**Comment by pavanjoshi914** [comment]
> Can you elaborate on `2.`? Why doesn't it work the same way as in the settings on the welcome page?

I made a  silly mistake! I thought ```saveSetting``` is the function which is providing save functionality and I was trying this to pass as props in localeSwitcher component. it was just a wrapper for ```const response = await api.setSetting(setting);	``` 
```
async function saveSetting(
    setting: Record<string, string | number | boolean>
  ) {
    const response = await api.setSetting(setting);
    setSettings(response);
  }
  ```
  
  did required changes @bumi @escapedcat kindly review

escapedcat:
> I made a silly mistake

image

src/app/router/Welcome/Welcome.tsx
```diff
@@ -73,10 +76,10 @@ function App() {
   return (
     <div>
       {process.env.NODE_ENV === "development" && <DevMenu />}
-
+      <LocaleSwitcher />
       <div className="relative max-w-7xl mx-auto px-4 sm:px-6 lg:px-8">
         <div className="text-center font-serif font-medium text-2xl pt-7 pb-3 dark:text-white">
-          <p>The power of lightning in your browser</p>
+          <p>{t("heading")}</p>
```
bumi, line 82: what is the convention for naming these keys? 
i18n automatically nests these keys under the component names? (do we need to document this for translators?)

bumi (commented):
From the code I don't fully understand where we load the language from the settings. 
Can you point me to that? 

Also how do we roll this out? Should we only show the language switcher in dev mode until we have a decent amount of the app translated?

pavanjoshi914:
> if you have gone through the documentation! either we can use [nesting of keys](https://gist.github.com/pavanjoshi914/a65af473209541c3a5cfecbac0f6e70f#using-nesting-of-keys) or [react i18next namespaces](https://gist.github.com/pavanjoshi914/a65af473209541c3a5cfecbac0f6e70f#alternate-for-nesting-of-keys)
> 
> currently, I am using namespace. if we wish to use nesting of keys as a key management strategy we can do that as well

...

bumi:
sorry, my bad. forgot...sure. really good, let's add that to the wiki? and add a link in the i18config in code to that wiki page?

src/app/router/Welcome/Welcome.tsx
escapedcat: I think it's a combination with `useTranslation(["welcome"])` in line 53.  
This is a short way of writing `welcome.heading`.

bumi:
(ups and I don't even hit the right button this morning, sorry. :) )

pavanjoshi914:
> sorry, my bad. forgot...sure. really good, let's add that to the wiki? and add a link in the i18config in code to that wiki page?

sure , but before that i have to do few revisions to make it more formal.

escapedcat:
> Also how do we roll this out? Should we only show the language switcher in dev mode until we have a decent amount of the app translated?

Good point. We can also add `alpha` to the settings-label, already show it and open an issue to move all texts step by step to translation-strings.

pavanjoshi914:
> From the code I don't fully understand where we load the language from the settings. Can you point me to that?
> 
> Also how do we roll this out? Should we only show the language switcher in dev mode until we have a decent amount of the app translated?


> > I made a silly mistake
> 
> image
haha nice one @escapedcat 😁

AaronDewes:
@bumi If you're looking for a translation platform, I recommend hosted.weblate.org, they are free for open source projects if you apply. I'm using it for Citadel and it works great

pavanjoshi914:
so any special use case do you have in mind @bumi? which supports storing locale in the custom objects along with other settings and loading locale from settings?
if not then we shall not interrupt the normal flow of the plugin and do something as shown below! WDYT

```
 lng: localStorage.getItem(I18N_LANGUAGE) || "en"
```

pavanjoshi914:
> @bumi If you're looking for a translation platform, I recommend hosted.weblate.org, they are free for open source projects if you apply. I'm using it for Citadel and it works great

thanks for the suggestion @AaronDewes ! yes we are already planned to use weblate for continuous localization

bumi:
@pavanjoshi914 hmm, I mean we now save the settings, so we should also use that. 
Also I am not sure how good it is to use the localstorage in the extension. for that extensions have some extension storage (which we use to save the settings)

@AaronDewes nice, thanks. good to hear you're using and recommending it.

pavanjoshi914:
> 

sure doing that then! @escapedcat  WDYT

escapedcat:
I think if there are no issues we should go with what the plugin does. At least for now.  
Would be great to get this in and start translating.

bumi:
I think I am missing something...
IF we use the default locale detector WHY do we then have the setting? that setting now has no effect.

And I think in the extension context it is recommended to use the [extension storage](https://developer.chrome.com/docs/extensions/reference/storage/) I would also prefer not to store things in different places.

bumi:
As discussed in the call: 
* we leave it as is and make a follow up PR that loads the locale from the our settings ([custom detector example code](https://github.com/i18next/i18next-browser-languageDetector/issues/140)) (@pavanjoshi914 can you create an issue for that?
* we show the language switcher on the welcome/intro screens ONLY in dev mode (or remove it for now) - because we do not have any translations, yet.
* we add something like "work in progress - help us translate Alby in your language" note to the locale switcher setting

pavanjoshi914:
@bumi  @escapedcat

pavanjoshi914:
> As discussed in the call:
> 
> * we leave it as is and make a follow up PR that loads the locale from the our settings ([custom detector example code](https://github.com/i18next/i18next-browser-languageDetector/issues/140)) (@pavanjoshi914 can you create an issue for that?
> * we show the language switcher on the welcome/intro screens ONLY in dev mode (or remove it for now) - because we do not have any translations, yet.
> * we add something like "work in progress - help us translate Alby in your language" note to the locale switcher setting



> As discussed in the call:
> 
> * we leave it as is and make a follow up PR that loads the locale from the our settings ([custom detector example code](https://github.com/i18next/i18next-browser-languageDetector/issues/140)) (@pavanjoshi914 can you create an issue for that?
> * we show the language switcher on the welcome/intro screens ONLY in dev mode (or remove it for now) - because we do not have any translations, yet.
> * we add something like "work in progress - help us translate Alby in your language" note to the locale switcher setting

https://github.com/getAlby/lightning-browser-extension/issues/743

bumi:
@dylancom can you have a quick look and merge it when approved?

bumi (approved):
utACK

pavanjoshi914:
any updates here?

dylancom:
@pavanjoshi914 Going to test it now!

dylancom:
The vertical alignment of the LanguageSwitcher seems off.
We already have a <Select /> component by the way that you can use from components/form
Schermafbeelding 2022-03-29 om 16 22 19

dylancom (changes requested):

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
```diff
@@ -0,0 +1,27 @@
+import { useState } from "react";
+import type { ChangeEvent } from "react";
+import i18n from "i18next";
+import "../../../i18n/i18nConfig";
+import api from "../../../common/lib/api";
+
+export default function LocaleSwitcher() {
+  const [dropdownLang, setDropdownLang] = useState(i18n.language || "en");
```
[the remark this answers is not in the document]
pavanjoshi914: done

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
```diff
@@ -0,0 +1,27 @@
+import { useState } from "react";
+import type { ChangeEvent } from "react";
+import i18n from "i18next";
+import "../../../i18n/i18nConfig";
```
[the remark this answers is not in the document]
pavanjoshi914: done

src/app/router/Welcome/Welcome.tsx
```diff
@@ -72,11 +75,15 @@ function App() {
 
   return (
     <div>
-      {process.env.NODE_ENV === "development" && <DevMenu />}
-
+      {process.env.NODE_ENV === "development" && (
+        <Fragment>
```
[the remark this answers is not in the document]
pavanjoshi914: done

pavanjoshi914:
@dylancom

escapedcat:
> > The vertical alignment of the LanguageSwitcher seems off. It also doesn't show the pointer cursor while hovering over it. We already have a component by the way that you can use from components/form
> 
> not sure what I have to do here! any screenshot which can help to understand where is the problem?

I think he meant you can use this component instead of a normal select:  
https://github.com/getAlby/lightning-browser-extension/blob/test/select-account-after-account-delete/src/app/components/form/Select/index.tsx  

You can find an example how to use it here:
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/form/Select/Select.stories.tsx
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Receive.tsx/#L199-L214
  Funnily it looks like this is commented out so we do not actually use this component anywhere atm. But still it would be good to use it to make sure future selects are aligned

pavanjoshi914:
> > > The vertical alignment of the LanguageSwitcher seems off. It also doesn't show the pointer cursor while hovering over it. We already have a component by the way that you can use from components/form
> > 
> > 
> > not sure what I have to do here! any screenshot which can help to understand where is the problem?
> 
> I think he meant you can use this component instead of a normal select: [https://github.com/getAlby/lightning-browser-extension/blob/test/select-account-after-account-delete/src/app/components/form/Select/index.tsx](https://github.com/getAlby/lightning-browser-extension/blob/test/select-account-after-account-delete/src/app/components/form/Select/index.tsx?rgh-link-date=2022-04-01T08%3A15%3A21Z)
> 
> You can find an example how to use it here:
> 
> * [https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/form/Select/Select.stories.tsx](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/form/Select/Select.stories.tsx?rgh-link-date=2022-04-01T08%3A15%3A21Z)
> * [https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Receive.tsx/#L199-L214](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Receive.tsx/?rgh-link-date=2022-04-01T08%3A15%3A21Z#L199-L214)
>   Funnily it looks like this is commented out so we do not actually use this component anywhere atm. But still it would be good to use it to make sure future selects are aligned

already using! but i refactored by default class present with component!

pavanjoshi914:
> already using! but i refactored by default class present with component!

the only important class was ```h-14``` in that default class (```w-full h-14 border-gray-300 rounded-md shadow-sm text-gray-700 bg-white```), added that too! not sure if this is what Dylan willing to have

escapedcat:
> already using! but i refactored by default class present with component!

👍  Ah sorry, didn't see.

Might be good having default styles on a select which can be extended or overwritten, but we can take care of that later on.

pavanjoshi914:
> > already using! but i refactored by default class present with component!
> 
> 👍 Ah sorry, didn't see.
> 
> Might be good having default styles on a select which can be extended or overwritten, but we can take care of that later on.

yup but the class has ```w-full``` which we don't want in case of language switcher also tried overwriting it! nothing happened

dylancom:
@pavanjoshi914 To give the <Select /> a certain width, you can just wrap it in a div with a specified width...
Our form inputs also work like that.

pavanjoshi914:
> 

w-full covers entire screen
image

pavanjoshi914:
> @pavanjoshi914 What is the problem with the w-full class? You let the width of the component to be decided by the parent that it's wrapped in. It might need some updated styling if it's too high now, as it has been styled a while ago and not used in prod yet. But I would keep default styles instead of passing in custom classes, and let a parent element decide how it's aligned, what it's width is etc.

so I can wrap the select element in div tag to overwrite few of the styles.

dylancom:
w-full means width: 100%; which means take 100% width of the parent container. So if your parent container is 400px wide, w-full means a width of 400px...

dylancom:
If your screen width is smaller than 1300px in the Welcome screen, then you can't click the language switcher. (elements overlapping)

dylancom (approved):

pavanjoshi914:
@bumi kindly review and merge

PR #723, opened 2022-03-22
feat: AccountMenu in prompts

### LINKED ISSUE #675: People are unsure which account is used to login
**Describe the bug**
In the login (or payment) prompt it is currently not clear which account is used. This can be confusing when users have two or more accounts.

**Expected behavior**
Maybe we can name the current selected account in the prompts. And/or allow changing the account from within the prompt.

**Screenshots (If Any)**
image

dylancom:
#### Link this PR to an issue
Fixes #675

You can now see from the prompt which account is active and also switch your account.
Previously this was unclear.

We now have a more flexible `<Navbar />` that will show an "advanced" version by default. Meaning it will include a user menu and show advanced `<AccountMenu />` options. For the prompt screen I added the improved `<Navbar />` and disabled the advanced prop so it will just show a simplified account menu, that allows you to switch.

Schermafbeelding 2022-03-22 om 09 28 32

### DIFF
--- a/src/app/components/Navbar/Navbar.tsx
+++ b/src/app/components/Navbar/Navbar.tsx
@@ -7,14 +7,16 @@ import UserMenu from "../UserMenu";
 type Props = {
   title: string;
   subtitle: string;
-  advanced?: boolean;
+  showAccountMenuOptions?: boolean;
+  showUserMenu?: boolean;
   children?: React.ReactNode;
 };
 
 export default function Navbar({
   title,
   subtitle,
-  advanced = true,
+  showAccountMenuOptions = true,
+  showUserMenu = true,
   children,
 }: Props) {
   return (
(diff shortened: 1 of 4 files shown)


src/app/components/Navbar/Navbar.tsx
```diff
@@ -30,17 +36,19 @@ export default function Navbar({ title, subtitle, children }: Props) {
...
       </div>
       {children && (
         <div>
           <nav className="flex space-x-8">{children}</nav>
         </div>
       )}
-      <div className="md:w-4/12 lg:w-3/12 flex justify-end items-center">
-        <UserMenu />
-      </div>
+      {advanced && (
```
escapedcat, line 47: Bit confused by the `advance` naming.  
Above `advance` displays the account-switcher/new-account. Here `advance` is also used to show the user-menu.  
Maybe name it like that? `showUserMenu` and `showAccoundSwitcher` or something like that?  
`advance` can be too many things I'm worried.

dylancom: Good suggestion, I agree that advanced can be too vague.

escapedcat (approved):
Test for `AccountMenu` would be nice.

dylancom:
@escapedcat added tests

PR #725, opened 2022-03-22
fix: prevent switching accounts while in progress

dylancom:

You should not be able to quickly jump between accounts, while it was still fetching a previous account.
This currently leads to bugs, showing the wrong balance for e.g. after fastly switching between different accounts.


bumi (approved):
super good. especially for the Tor connectors

PR #730, opened 2022-03-23
feat: move screen code into reusable components

dylancom:

Currently many screens duplicate the same kind of code. This makes it more difficult to make code changes and results in maintenance errors such as "dark mode missing in several parts".
Componetizing this code results in just needing 1 place to edit code.

### DIFF
--- a/src/app/components/ConfirmOrCancel/index.tsx
+++ b/src/app/components/ConfirmOrCancel/index.tsx
@@ -1,6 +1,6 @@
 import { MouseEventHandler } from "react";
 
-import Button from "../../Button";
+import Button from "../Button";
 
 type Props = {
   disabled?: boolean;
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -7,7 +7,7 @@ import utils from "../../../common/lib/utils";
 
 import Button from "../Button";
 import Menu from "../Menu";
-import TextField from "../Form/TextField";
+import TextField from "../form/TextField";
 
 type Props = {
   allowance: {
--- a/src/app/components/layout/ConfirmOrCancel/index.tsx
+++ b/src/app/components/layout/ConfirmOrCancel/index.tsx
@@ -1,0 +1,48 @@
+import { MouseEventHandler } from "react";
+
+import Button from "../../Button";
+
+type Props = {
+  disabled?: boolean;
+  loading?: boolean;
+  label?: string;
+  onConfirm: MouseEventHandler;
+  onCancel: MouseEventHandler;
+};
+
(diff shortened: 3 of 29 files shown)


bumi (approved):
utACK. this is great @escapedcat can you have a look and then we quickly merge this.

src/app/components/ConfirmOrCancel/index.tsx
```diff
@@ -0,0 +1,48 @@
+import { MouseEventHandler } from "react";
```
escapedcat, line 1: nit: This is called `ConfirmOrCancel` which sounds very general but it looks very specific. Maybe reflect this in the naming? Eventually we might need a general `ConfirmOrCancel`?

src/app/components/Setting/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+function Setting({ title, subtitle, right }: Props) {
+  return (
+    <div className="py-4 flex justify-between items-center">
+      <div>
+        <span className="text-gray-700 dark:text-white font-medium">
+          {title}
+        </span>
+        <p className="text-gray-400 text-sm">{subtitle}</p>
+      </div>
+      {right}
```
escapedcat, line 16: Eventually design will be changed and this isn't `right` anymore. Is this a i.e. `settingComponent`?

src/app/components/ConfirmOrCancel/index.tsx
dylancom: Hmm unsure how to name it then.

escapedcat: This is only being used in `ConfirmSignMessage` and comes with prepared texts matching that usage, yes? What about `ConfirmOrCancelSignMessage`?

dylancom: This is used in 8 screens.

escapedcat: Ah, didn't see. Not aware of the context.   
I'm hestitating because it has this specific "Only connect with sites you trust.". Can it be named in relation to that?  
Could look into the screens and come up with something more matching tommrow.

dylancom: We might need to change the message or make it changable by a prop at some point. This sentence is used everywhere also when I want to send you a few sats for e.g.
So I'd say it's already a general component but we might need to improve on the sentence at some point.

escapedcat: Ahhh, ok, well then... yolo

escapedcat (approved):

bumi:
:shipit:

PR #736, opened 2022-03-24
chore: remove unused code #721

### LINKED ISSUE #721: Replace sha.js library with crypto-js SHA265 library
Currently we use sha.js and crypto-js/SHA265 for hashing. We should replace all sha.js usages with the crypto-js hashing lib.

escapedcat:
Fixes #721

- Chore

Also resorted imports


bumi (approved):
damn didn't fully got rid of sha.js. :)

PR #745, opened 2022-03-25
fix(delete account): update currentAccountId only if needed

bumi:

In the delete action we now only update the currentAccountId state if the current account was deleted.

#### Checklist:

Would be nice if somebody could help me write a spec for this


escapedcat:
> Would be nice if somebody could help me write a spec for this  

I was thinking if it might be better to test the background-code with intergration/e2e-tests instead of real unit-tests. With that we would test the whole interactivity for i.e. "delete account". The underlying code can change without the need of updating uni-tests.
So as soon as #627 is ready we can cover this code here as well.
What do you think @dylancom @bumi ?

bumi:
I like both. often unit tests are easier to write, faster and the units are reused in many parts of the app.

escapedcat:
Test via #747

bumi:
thanks for the help with the tests @escapedcat

jest.config.ts
```diff
@@ -3,6 +3,7 @@ module.exports = {
   transform: {
     "^.+\\.[t|j]sx?$": "babel-jest",
   },
+  transformIgnorePatterns: ["node_modules/(?!(@runcitadel))/"],
```
bumi, line 6: do we still need this?

bumi (commented):
mergable? @escapedcat

jest.config.ts
escapedcat: I believe so, yes

bumi: wasn't there a change in that package? 
(and I don't like such special cases :) )

escapedcat: Not sure, will check

escapedcat: My understanding is that runcitadel's dist code is coming as ESM (using import/export) which can't be handled by Jest. Therefor we need to transform the citadel code eventough normally we do not compile any code from `node_modules`.  
@dylancom what do you think about this?

bumi: ah I see. thanks for the explanation. 

but then we can merge this PR?

escapedcat: I think so, yes

PR #747, opened 2022-03-26
Test/select account after account delete

escapedcat:
This is fun on multiple levels:

- ✅ `origin` for delete-message only is `internal: true`, so the `OriginData`-type doesn't match
   Should we create a special `OriginData`-type for internal messages? Or use the same but handle it differently?  
   What do you think @dylancom ?  
   It's a good side-effect that the unit-tests can help re-organizing the types.
- ✅ [Citadel/NativeCitadel](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/citadel.ts#L1)-connectors break the Jest setup
  Seems to be a mix of ESM/TS/babel. Couldn't solve it fast enough and moved on. But should be solved.

Might talk to @AaronDewes regarding handling the Citadel import. Was wondering why it's using `browser` instead of `dist`. But changing that didn't help either. Steps of trying trying to solve this can be found in `jest.config.ts` and `.babelrc`.

Error is:

```sh
    SyntaxError: Unexpected token 'export'

    > 1 | import Citadel from "@runcitadel/sdk/browser/index.js";
        | ^
      2 |
      3 | enum Invoice_InvoiceState {
      4 |   OPEN = 0,
```

Fixed by transforming `@runcitadel` with Jest and switching to `babel.config.json` accoring to:
- https://babeljs.io/docs/en/configuration#whats-your-use-case
- https://stackoverflow.com/questions/61781271/jest-wont-transform-the-module-syntaxerror-cannot-use-import-statement-outsi#comment120522535_61785012


src/types.ts
```diff
@@ -36,15 +36,16 @@ export interface MetaData {
 }
 
 export interface OriginData {
-  location: string;
```
escapedcat, line 39: Maybe add `OriginDataInternal` and below do this?:

```js
origin: OriginData | OriginDataInternal
```

escapedcat: Talked to @dylancom , will add something like this.

src/extension/background-script/actions/accounts/__tests__/delete.test.ts
```diff
@@ -0,0 +1,50 @@
...
+const message: Message = {
+  application: "LBE",
+  args: { id: "8b7f1dc6-ab87-4c6c-bca5-19fa8632731e" },
+  origin: { internal: true },
+  prompt: true,
+  type: "deleteAccount",
+};
+
+describe("delete account", () => {
+  test("if current account is being deleted first existing account will be set as current", async () => {
```
bumi, line 40: I tried to add a second test to test the opposite case. 

But it seems the spy on the setState is somewhat persistent and in my new test it also knows about the two first calls? 

can you help me with that test?

```
  ● delete account › if other account is being deleted the current account is not updated

    expect(jest.fn()).not.toHaveBeenCalledWith(...expected)

    Expected: not {"currentAccountId": "d892e2d7-ac72-49b7-94c2-9fa024c5c1d3"}
    Received
           2:     {"currentAccountId": "d892e2d7-ac72-49b7-94c2-9fa024c5c1d3"}

    Number of calls: 3
```

escapedcat: Yeah, will do, I think we need to reset the mock/spy.

bumi (commented):
Thanks for the help with the mocks!

src/extension/background-script/connectors/citadel.ts
```diff
@@ -1,4 +1,4 @@
-import Citadel from "@runcitadel/sdk/browser/index.js";
+import Citadel from "@runcitadel/sdk";
```
escapedcat, line 1: @AaronDewes this should be prefered and working, right?

jest.config.ts
```diff
@@ -3,6 +3,7 @@ module.exports = {
   transform: {
     "^.+\\.[t|j]sx?$": "babel-jest",
   },
+  transformIgnorePatterns: ["node_modules/(?!(@runcitadel))/"],
```
bumi, line 6: what is this doing?

bumi (approved):

jest.config.ts
escapedcat: From my understanding @runcitadel is using ESM aka i.e `import` instead of i.e. `required`. Alby is also using this in its code. For Jest to handle these files Babel is being used to transpile this into something Jest can handle (same for the tsx-files).

Per default `node_modules` are not being transpiled because it can be expected that node-modules are shipping already transpiled dist code. 
[@runcitadel is using ESM](https://github.com/getAlby/lightning-browser-extension/pull/769#issuecomment-1086654095) also in the dist-files. Therefor we need to include `@runcitadel` into transpiling. 

Does this make sense? @dylancom what do you think?

bumi: when I hear babel 🏃‍♂️😱😵‍💫 

:D

PR #749, opened 2022-03-26
fix: pre-defined amount buttons text wrapping

### LINKED ISSUE #727: The pre-defined amounts have a line break on windows
**Describe the bug**
A minor UI bug: The pre-defined amounts have a line break on windows.

**To Reproduce**
1. Visit to https://twitter.com/dergigi
2. Open Alby
3. Click "send satoshis" 

**Information About Alby**
- Alby Version:  1.7.0
- Alby installed through the Firefox browser store
- Wallet connected with Alby 

**Screenshots (If Any)**
image

**Device Information [optional]:**
 - OS: Windows
 - Browser Firefox

rdfriedl:
#### Link this PR to an issue
Fixes #727 

 - Added a `small` prop to `Button` which changes the padding from `px-7 py-2` to `px-4 py-1`
 - Changed `SatButtons` component to use `grid` and `grid-rows-2 lg:grid-rows-1` instead of `flex` for layout

 - Before (using chrome on linux) 
image

 - After (using chrome on linux) 
image

I tested this locally on Firefox and Brave (chrome) and I was able to reproduce and fix the issue.
I was not able to test this on windows since I do no have access to a windows pc.

#### Checklist:

### DIFF
--- a/src/app/components/Button/Button.stories.tsx
+++ b/src/app/components/Button/Button.stories.tsx
@@ -14,12 +14,6 @@ export const Disabled = () => (
     <Button label="Button" primary loading disabled />
   </div>
 );
-export const Size = () => (
-  <div className="space-x-3">
-    <Button label="Normal" />
-    <Button label="Compact" compact />
-  </div>
-);
 
 export default {
   title: "Components/Buttons/Button",
--- a/src/app/components/Button/Button.stories.tsx
+++ b/src/app/components/Button/Button.stories.tsx
@@ -14,6 +14,12 @@ export const Disabled = () => (
     <Button label="Button" primary loading disabled />
   </div>
 );
+export const Size = () => (
+  <div className="space-x-3">
+    <Button label="Normal" />
+    <Button label="Compact" compact />
+  </div>
+);
 
 export default {
   title: "Components/Buttons/Button",
--- a/src/app/components/Button/index.tsx
+++ b/src/app/components/Button/index.tsx
@@ -3,7 +3,7 @@ import Loading from "../Loading";
 
 export type Props = React.ButtonHTMLAttributes<HTMLButtonElement> & {
   fullWidth?: boolean;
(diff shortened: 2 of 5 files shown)


dylancom:
Although this fixes the wrapping, the design still looks off (no horizontal padding in the buttons). I think we just need a button variant where the text is smaller.

escapedcat:
Other option could be to arrange them 2x2.

dylancom:
Thanks a lot for your contribution :). The only downside now with 2 rows is that it pushes the submit button below "the fold". We discussed today in our community call that we'd prefer to have it always directly visible in our screens.

rdfriedl:
I removed the 2x2 grid and with the smaller text it dose not wrap anymore
image
the confirm button is visible now, but unfortunately the horizontal padding is gone :(

dylancom:
How about we just drop the "sat" wording (keep the current styling) and place the satoshi symbol in front of the amount?
https://bitcoinicons.com/

Similar to most fiat amount options:
image

rdfriedl:
that works a lot better :) 
image

dylancom (approved):

dylancom:
On makers.bolt.fun 3 buttons are used, with a variable width depending on the content.
They also use less horizontal spacing between the buttons.
@bumi what do you say compared to the current change we have now?

Schermafbeelding 2022-03-30 om 10 28 24

bumi:
tho only thought I have is we should use 3 buttons and put it horizontally next to the input field. 
1M Sats (currently nearly 475USD) in the input take a bit more space than 1 button. This would mean less horizontal scrolling 

but to me it's fine. it's much better now with @rdfriedl's PR. what do you think?

dylancom:
I think this works great for now. I would not put those buttons next to the input field.

bumi:
Thanks again for the contribution! your change will be in the next release :shipit:

PR #752, opened 2022-03-28
feat(send): support BIP21 URI with BOLT 11 invoice QRCode

krsjoseph:
#### Link this PR to an issue
Fixes #ISSUE-NUMBER

support BIP21 URI with BOLT 11 invoice QRCode

Not been fully as I've been having problems scanning QR codes on my laptop.

#### Checklist:

### DIFF
--- a/src/app/screens/Send.tsx
+++ b/src/app/screens/Send.tsx
@@ -46,6 +46,18 @@ function Send() {
     }
   }
 
+  function checkInvoice(data: string) {
+    const reqExp = /lightning=([^&|\b]+)/i;
+
+    const invoice = data.match(reqExp);
+
+    if (invoice) {
+      setInvoice(invoice[1]);
+    } else {
+      setInvoice(data);
+    }
+  }
+
   if (qrIsOpen) {
     return (
       <div>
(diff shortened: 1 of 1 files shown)


krsjoseph:
@bumi here's the PR from our paring session

src/app/screens/Send.tsx
```diff
@@ -63,7 +75,7 @@ function Send() {
             qrbox={200}
             qrCodeSuccessCallback={(decodedText) => {
               if (invoice !== decodedText) {
-                setInvoice(decodedText);
+                checkInvoice(decodedText);
```
kiwiidb, line 78: I would name this function `checkBip21AndSetInvoice`, so it's more clear what this actually does.

kiwiidb (changes requested):

src/app/screens/Send.tsx
bumi: I would even do something like: 

`setInvoice(extractInvoiceFrom(decodedText))`

this way the `extractInvoiceFrom` does not need to know anything about context where it is used and just returns the invoice string if it found one. 
Potentially this makes it easier to extract and reuse somewhere else and also easier to test.

bumi (approved):
thanks for the PR! 
I have suggested a small refactoring, but up to you.

src/app/screens/Send.tsx
krsjoseph: Thanks @bumi! I'll make these changes

bumi: (not sure if it's a good idea, was just a thought. :D )
thnks

escapedcat (approved):

Issue #631, opened 2022-02-23
Make lnurl-auth with Lnbits compatible the their implementation


bumi:
Currently we create a custom key used for lnurl-auth with lnbits. 
Since lnbits supports lnurl-auth we should make sure to be comptabile. This then results in the same accounts when the user logs in with Alby or with Lnbits.

code in lnbits:
https://github.com/lnbits/lnbits-legend/blob/dbab18175935a65fa3a41a4869257953f5ec38cb/lnbits/core/models.py#L41

kirkawolff:
No matter how hard I try, I can't figure out what lnbits is putting in for Wallet.id and how that's defined. I believe this compatibility would only be important for lnd, eclair and lnbits wallets, and not for lndhub wallets.

bumi:
I think this can only be compatible if the user connect Alby to lnbits - lnbits connected to LND or something else is unrelated there. 
The wallet ID must be something lnbits specific. 
If you create a lnbits wallet (e.g. on https://legend.lnbits.com/) you see in "API Info" on the right a "Wallet ID" entry. isn't that what is used there?

kiwiidb:
What should be the signing key then exactly?

bumi:
<img width="754" alt="image" src="https://user-images.githubusercontent.com/318/160807668-f1f7fd95-64af-4b32-92b3-8cf16240d591.png">

Turns out I am wrong. the code I linked in this issue is only a place holder so far. They don't have full lnurl-auth support. 

So I think we should use a similar schema as lndhub (`lnbits://${this.config.adminkey}`) to create the key used for signing.

PR #756, opened 2022-03-29
feat: add opt-out of new signmessage

### LINKED ISSUE #629: Make lnurl-auth with lndhub compatible to BlueWallet
BlueWallet is adding lnurl-auth support. \o/ BUT they are not using the same secret to create the key as we do. 
We also use an [Alby specific key](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/lndhub.ts#L164) (which was a mistake initially made)
To be compatible with BlueWallet, meaning that lnurl-auth with the same lndhub account in Alby and BlueWallet has the same result we have to update the way we create the key and signature. 

Updating the code should be easy, but we need to think about a way to not break it for existing users. Existing users would need a way to login using the old mode. 
Ideally this is a setting on the "website". 
something like: 
```
if website.useOldLnurlAuthMode {
  // use old key generation / signing method
} else { // this is the default then
  // use new key generation and signing method compatible with BlueWalet
}
```

Code in BlueWallet: 
https://github.com/BlueWallet/BlueWallet/pull/4493/files#diff-8aaf0739679ead0a7c5e8cd819fad19bd972ccc1e8ef4dd641af75b7dd87c82cR672
-> https://github.com/BlueWallet/BlueWallet/pull/4493/files#diff-ec4c16453004f9b974f5659081ec5eafd3f4f5ae6a009fbc1bbe953a2de8b9f0R298-R306

used secret: https://github.com/BlueWallet/BlueWallet/blob/6f87a73a52ba648efa3cf33202bf965432b9054a/class/wallets/lightning-custodian-wallet.js#L113

kiwiidb:
#### Link this PR to an issue
Fixes #629

- Breaking change (fix or feature that would cause existing functionality to not work as expected)
Users can opt-out of the new login method by checking a setting.

- A new "private key" format for LNDhub connectors
- Hash the message that we need to sign in both LNBits and LNDhub connectors.
- Add a toggle in the Settings so people can sign in the old way and recover their old accounts.

Schermafbeelding 2022-03-30 om 10 33 41

- Check out the master branch of the extension and go to https://lnurl.fiatjaf.com
- Make sure that you have a LNDhub wallet as your current wallet in Alby.
- Click on LNURL Auth QR code
- Make a note of the `key` value. This is your identity on this site.
- Now, check out this PR, reload the extension, refresh the page and do this again. Your `key` should be different.
- Toggle the setting to use the legacy LNURL auth. Your `key` should be the same again.

To test that it is compatible with Bluewallet, use Bluewallet connected to an Alby account to login to https://lnurl.fiatjaf.com. Make a not of the public key. This should be the same key as the one that you get by logging in with Alby, if you use the new login and are connected to the same account. 

Alternatively, go to a site where you have logged in with Alby like lnmarkets.com or stacker.news. If you log in with the new method, you should have a fresh, empty account. If you toggle the legacy LNURL auth setting, you should have your old account back.

TODO: new key format for LNBits? @bumi ?


bumi:
this is for LNbits: https://github.com/getAlby/lightning-browser-extension/issues/631#issuecomment-1082897771

we should just change the format on how we create the root signing key to follow a the same format as lndhub does

src/app/screens/Settings.tsx
```diff
@@ -56,6 +57,21 @@ function Settings() {
           )}
         </Setting>
 
+        <Setting
+          title="Legacy LNURL Auth"
```
bumi, line 61: I'd call this maybe "Legacy messaging signing for LNDhub and LNbits" 
because this changes message signing (and not only/specifically LNURL-Auth).  

Ideally we also only show this IF the user has an lndhub or lnbits account? there is an `getAccounts` action that returns all accounts (config is encrypted). We could check in that array for a connector === 'lndhub' || connector === 'lnbits' ? what do you think?

src/app/screens/Settings.tsx
```diff
@@ -56,6 +57,21 @@ function Settings() {
...
+          subtitle="Login with the deprecated LNURL Auth method"
```
bumi, line 62: maybe more details for the user: 

We have made compatibility changes to the signing with LNDhub and LNbits accounts. If you used LNURL-auth logins with those accounts you can enable the deprecated signing method. (not recommended)

src/extension/background-script/connectors/lnbits.ts
```diff
@@ -121,12 +122,18 @@ class LnBits implements Connector {
     if (!args.message) {
       return Promise.reject(new Error("Invalid message"));
     }
-    const message = utils.stringToUint8Array(args.message);
+    let message: string | Uint8Array;
+    message = sha256(args.message).toString(Hex);
     // create a signing key from the lnbits URL and the adminkey
     const keyHex = sha256(
       `LBE-LNBITS-${this.config.url}-${this.config.adminkey}`
```
bumi, line 129: https://github.com/getAlby/lightning-browser-extension/issues/631#issuecomment-1082897771

src/app/screens/Settings.tsx
kiwiidb: Ok this would be a good idea, but I think it would be even more useful if we somehow get the time when an account was added, and don't show this setting if that is after a certain time?

bumi: nice idea. but we don't have the time or what are you're ideas? 

I would have left it in for maybe ~2months and then we remove the deprecated option. I think that's OK at this stage.

kiwiidb: That's also possible, but then we need to make sure that users now that they need to empty their old accounts.

kiwiidb: Schermafbeelding 2022-03-30 om 14 43 11

I would just leave it at this then to be honest. People will only visit this page when they have this problem with their legacy accounts anyway.

kiwiidb: Is this text ok, @bumi ?

bumi: to base it on time we would need to store more information, don't we?

bumi (approved):

PR #757, opened 2022-03-29
feat: new wallet account creation with lightning address

bumi:
#### Describe the changes you have made in this PR

This PR changes the signup for a new alby account. 
We make it clear that the user gets an account on getalby.com and we require an email and a password. 
The user now can also directly create a lightning address during the signup. 

image

image

#### Checklist:

### DIFF
--- a/src/app/screens/connectors/NewWallet/index.tsx
+++ b/src/app/screens/connectors/NewWallet/index.tsx
@@ -177,9 +177,10 @@ export default function NewWallet() {
             />
           </div>
           <div className="mt-6">
-            <p className="text-gray-500 dark:text-gray-400">
+            <p className="mb-2 text-gray-500 dark:text-gray-400">
               Your Alby account also comes with an optional{" "}
               <a
+                className="underline"
                 href="https://lightningaddress.com/"
                 target="_blank"
                 rel="noreferrer"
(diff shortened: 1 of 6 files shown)


src/app/screens/connectors/NewWallet/index.tsx
```diff
@@ -136,21 +146,66 @@ export default function NewWallet() {
...
+                learn more
+              </a>
+              )
+            </p>
+            <TextField
+              id="lnAddress"
+              label="Choose your Lightning Address (optional)"
+              type="text"
+              onChange={(e) => {
+                setLnAddress(e.target.value.trim().split("@")[0]); // in case somebody enters a full address we simple remove the domain
```
escapedcat, line 205: Because an Alby address will always have the `@getAlby.com` domain? Not too familiar with this yet.  
If yes, can a placeholder for the input help to avoid confusion and set expectations?
If an address is already taken it will lead to "Failed to create a new wallet" further up, yes?

bumi: because people might enter @getalby.com or something. 

and yes, something as we have on the web app would be good, but I did not know how :) 
image

if the address is taken, then yes, it would fail with that alert (including the errors as JSON) - there are no good error messages so far.

PR #762, opened 2022-03-31
feat(notification): icon and payment fees

### LINKED ISSUE #205: Show fees on payment confirmation
As a user I want to know what are the fees of the payment and the total amount of the payment.

reneaaron:
#### Link this PR to an issue
Fixes #205 

image
image

- [x] [Successful payment](https://webln.twentyuno.net/) 
- [x] [Login](https://webln.twentyuno.net/login)
- [ ] Payment failed
- [ ] Login failed

👆 Is there a way to test the failed notifications, too? 🤔

#### Checklist:

### DIFF
--- a/src/extension/background-script/events/notifications.js
+++ b/src/extension/background-script/events/notifications.js
@@ -6,11 +6,18 @@ const paymentSuccessNotification = (message, data) => {
   const recipient = data.origin.name;
   const paymentResponse = data.response;
   const route = paymentResponse.data.route;
-  const { total_amt } = route;
+  const { total_amt, total_fees } = route;
+  const paymentAmount = total_amt - total_fees;
+
+  function formatAmount(amount) {
+    return `${amount} sat${amount != 1 ? "s" : ""}`;
+  }
 
   return utils.notify({
-    title: `Paid ${total_amt} sat to ${recipient}`,
-    message: `pre image: ${paymentResponse.data.preimage}`,
+    title: `✅ Successfully paid ${formatAmount(
+      paymentAmount
+    )} to »${recipient}«`,
+    message: `Payment fee: ${formatAmount(total_fees)}`,
   });
 };
 
@@ -24,21 +31,21 @@ const paymentFailedNotification = (message, data) => {
     error = data.response.data.payment_error;
   }
   return utils.notify({
-    title: `Payment failed`,
+    title: `⚠️ Payment failed`,
     message: `Error: ${error}`,
   });
 };
 
 const lnurlAuthSuccessNotification = (message, data) => {
   return utils.notify({
-    title: `Login to ${data.origin.name}`,
+    title: `✅ Login to ${data.origin.name}`,
     message: `Successfully logged into ${data.lnurlDetails.domain}`,
   });
 };
 
 const lnurlAuthFailedNotification = (message, data) => {
   return utils.notify({
-    title: `Login failed`,
+    title: `⚠️ Login failed`,
     message: `${data.error}`,
   });
 };
--- a/src/common/lib/utils.ts
+++ b/src/common/lib/utils.ts
@@ -29,7 +29,7 @@ const utils = {
   notify: (options: { title: string; message: string }) => {
     const notification: browser.Notifications.CreateNotificationOptions = {
       type: "basic",
-      iconUrl: "assets/icons/satsymbol-48.png",
+      iconUrl: "assets/icons/alby_icon_yellow_48x48.png",
       ...options,
     };
 


src/extension/background-script/events/notifications.js
```diff
@@ -6,11 +6,18 @@ const paymentSuccessNotification = (message, data) => {
   const recipient = data.origin.name;
   const paymentResponse = data.response;
   const route = paymentResponse.data.route;
-  const { total_amt } = route;
+  const { total_amt, total_fees } = route;
+  const paymentAmount = total_amt - total_fees;
+
+  function formatAmount(amount) {
```
reneaaron, line 12: This possibly opens a whole new discussion. 😅 What is the plural of 1 sat? 
 - 2 sat
 - 2 sats
 
 To me sats feel a little bit more natural but I have no strong opinion on that.

src/extension/background-script/events/notifications.js
```diff
@@ -6,11 +6,18 @@ const paymentSuccessNotification = (message, data) => {
...
+    return `${amount} sat${amount != 1 ? "s" : ""}`;
+  }
 
   return utils.notify({
-    title: `Paid ${total_amt} sat to ${recipient}`,
-    message: `pre image: ${paymentResponse.data.preimage}`,
+    title: `✅ Successfully paid ${formatAmount(
+      paymentAmount
+    )} to »${recipient}«`,
```
reneaaron, line 19: The `recipient` is more like the `initiator` of the payment, right? We don't know who the user actually just paid 🤔

src/extension/background-script/events/notifications.js
bumi: hehe. yes. is it sat or sats that's the question. maybe we should do a twitter poll?

src/extension/background-script/events/notifications.js
```diff
@@ -6,11 +6,18 @@ const paymentSuccessNotification = (message, data) => {
...
+    message: `Payment fee: ${formatAmount(total_fees)}`,
```
bumi, line 20:
```suggestion
    message: `fee: ${formatAmount(total_fees)}`,
```

making it shorter?

bumi (approved):

src/extension/background-script/events/notifications.js
bumi: I mean it could be "paid X ON foo.com" - if that's better?

src/extension/background-script/events/notifications.js
reneaaron: While drafting on the poll I found a very promising definition. There are basically 2 types of usages:

 - Unit (an amount of money, i.e. 99 satoshi, cent, etc) "please pay this invoice of 100 satoshi"
 - Objects (coins, i.e. 99 satoshis, cents, etc) "stacking sats"
 
 Since I feel we use it more like a unit I would vote for against using `sats`. 🤔

src/extension/background-script/events/notifications.js
reneaaron: Generally I like ON very much because that describes exactly what actually happened. 

But that wouldn't work for LightningAddress payments directly in Alby. There the success notification now says `Successfully paid X to example@example.com` which is a good thing IMO. We would need to come up with one phrase that fits both use cases or be able to distinguish the 2 cases:

 - You have successfully paid X on example.com (calls through webln)
  - You have successfully paid X to lnaddress@example.com (payments initiated by alby batteries, send tab, etc)

src/extension/background-script/events/notifications.js
bumi: this is what twitter says: https://twitter.com/getAlby/status/1509884716517928963

reneaaron: That's a clear result. Let's keep the `s`! 😀

src/extension/background-script/events/notifications.js
reneaaron: After having a look at how this is handled internally I think this is a whole new topic that goes way beyond this notification: 

Currently Alby does not differentiate the `initiator` or `payee`. If you for instance pay an invoice the UI & notifications indicate that you are currently paying Alby. (as this is probably the default?)

I'd create a separate issue for that and discuss it in there and leave the notification text as it is now.

bumi: > If you for instance pay an invoice the UI & notifications indicate that you are currently paying Alby. (as this is probably the default?)

this was because of pragmatic reasons to reuse what was there. 

+1 for a new issue to look into this as a whole. thx.

reneaaron: I've created a follow-up issue and tried to summarize my findings:
https://github.com/getAlby/lightning-browser-extension/issues/770

Please let me know if I missed anything.

bumi (approved):
utACK

PR #763, opened 2022-03-31
feat(lnurl-pay): handle fetch invoice errors

### LINKED ISSUE #761: Handle LNURL pay callback errors
**Describe the bug**
Currently we do not handle errors of the LNURL-pay callback

**To Reproduce**
1. pay an LNURL-pay with a certain minSendable
2. try to send a payment below the minimum
3. the error response from the callback is not handled

**Expected behavior**
Show an error alert to the user

related code: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/LNURLPay.tsx#L127

reneaaron:
#### Link this PR to an issue
Fixes #761 

Error handling for fetching the invoice via lnurl-pay.

image

 - [x] Try to pay to a Lightning Address with below minimum allowed amount

#### Checklist:

### DIFF
--- a/src/app/screens/LNURLPay.tsx
+++ b/src/app/screens/LNURLPay.tsx
@@ -3,6 +3,7 @@ import axios from "axios";
 import { useNavigate, useSearchParams } from "react-router-dom";
 
 import {
+  LNURLPaymentInfoError,
   LNURLPaymentInfo,
   LNURLPaymentSuccessAction,
   LNURLPayServiceResponse,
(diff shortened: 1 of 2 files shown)


src/app/screens/LNURLPay.tsx
```diff
@@ -106,13 +106,20 @@ function LNURLPay(props: Props) {
...
+        const request = await axios.get<LNURLPaymentInfo>(details.callback, {
           params,
-        }
-      );
-      const { pr: paymentRequest } = paymentInfo;
+        });
+        paymentInfo = request.data;
+        paymentRequest = paymentInfo.pr;
+      } catch (e) {
+        alert("Payment aborted: Could not fetch invoice.");
```
bumi, line 120: The LNURL server can also return an "error" in the response. 
Right now we do not handle this one either. 

```
{"status":"ERROR", "reason":"error details..."}
```

I guess checking for the status "error" and then doing an alert(reason) would be good?

bumi (changes requested):
thanks, let's directly also add a check for an lnurl error response

src/app/screens/LNURLPay.tsx
reneaaron: Thanks for the hint, I just added the reason to the alert displayed to the user (if available). 

image

bumi: can you check the spec. I think that the service can also respond with a HTTP 200 BUT have an error field.

src/app/screens/LNURLPay.tsx
```diff
@@ -106,13 +107,36 @@ function LNURLPay(props: Props) {
...
+        response = await axios.get<LNURLPaymentInfo | LNURLPaymentInfoError>(
+          details.callback,
+          {
+            params,
+            // https://github.com/fiatjaf/lnurl-rfc/blob/luds/01.md#http-status-codes-and-content-type
+            validateStatus: () => true,
+          }
+        );
+
+        const isSuccessResponse = function (
```
reneaaron, line 123: Do you know the reason the spec decided against HTTP status codes? Doesn't that make the implementation / tooling more complicated for all services consuming those APIs? 

If you can think of a more elegant way to solve that, please let me know!

bumi: nice, implementation. 

No idea why status codes should not be respected. honestly I am still OK with respecting them. If a server for example returns a 500 as a valid response, then imo the server is so broken, that I do not respect the response :) 

I think in other places we do assume a success response. 
but this PR looks small and clean to me.  👍

bumi (approved):

PR #764, opened 2022-03-31
feat: upgrade to react 18

dylancom:

Upgrade to react 18.
Although we didn't support IE anyway, I like to mention that the framework React that we use dropped IE support as of version 18.


escapedcat (approved):

bumi:
wait, no support for IE6 anymore?! 😱

Issue #748, opened 2022-03-26
Return pubkey to webln getInfo call


staab:
I am looking at using webln to build a social website which uses signing to verify messages, and shares a user's public key as an account's unique identifier. It looks like Alby doesn't fulfill the entire webln spec, since when I call:

```
provider.getInfo().then(info => console.log(info))
```

I get

```
{
    "node": {
        "alias": "Alby"
    }
}
```

I'd love to get a node.pubkey back per [the spec](https://webln.dev/#/api/get-info). I can sort of work around this by asking the user to sign a constant message to get a derived key, e.g.,

```
provider.signMessage('Account ID generation').then(x => console.log(x))
```

But that brings up a prompt, which is a weird user experience.

reneaaron:
I noticed the same behavior on https://webln.twentyuno.net/info for @getalby.com acounts: 

![image](https://user-images.githubusercontent.com/100827540/161124258-e93930e1-b022-43c6-9f4a-f84aebe229e3.png)

I think we can add the missing fields for LNDHub, but I think there are a few cases where connectors do not provide this information. (Galoy, LNBits) I'll have a look into it.

bumi:
sorry for my late reply here. 
What do you try to do exactly? 

Please note that when using a custodial connector (Alby connected to LndHub for example) then the node pub_key is the key of the shared node. (the same for many users)

staab:
Interesting — I am trying to put together sort of a "sign in with wallet" feature for a social-sign-in type of thing. But if public keys are not unique, it seems I won't be able to use Alby to support this use-case. But I'm a little confused, if the node id is shared, how do payments get routed to the right user?

staab:
No, just dipping my toes in here, though I am hoping for a client-side-only solution. I'll take a look at LNURL-auth!

escapedcat:
Can this be closed then?

staab:
Yep, thanks for your help!

bumi:
as an alternative to lnurl-auth you can also use the `signMessage` call. typically those authentication options work with requesting a signature from the user. The user then proofs to have the private key to a public key. 
but for web authentications I think lnurl-auth is the current way to go. It generates a key for each website and thus preserves the user's privacy.

Jared-Dahlke:
@escapedcat or @reneaaron ,
I'm using alby with bitcoin-connect-react. I am trying to get the pubkey from my connected alby account so that i can do a keysend to it later, but looks like that is not working.
[...]

PR #765, opened 2022-03-31
feat(connector): added further fields to getinfo

reneaaron:
#### Link this PR to an issue
#748 

 - [x] Go to https://webln.twentyuno.net/info with your Alby / LNDHub account

#### Checklist:


escapedcat:
What's the status here? 😬

bumi:
WebLN back then did not think about custodial setups and in fact WebLN probably did not think much about what that getInfo information is useful for. (Why return the color there?)

I did not return the pub key there because it is is the shared pubkey... and not the user's pubkey. 

but let's merge it. 

@reneaaron can you fix the merge conflict?

PR #767, opened 2022-04-01
feat(tipping): github profile support

### LINKED ISSUE #766: Tipping support for GitHub profiles
When I add a ⚡Address to my GitHub profile, I want Alby to recognize it.

**Are you working on this? 👍**

reneaaron:
#### Link this PR to an issue
Fixes #766 

Currently we only support Github profiles. We could widen support to repositories (so you have the option to donate on i.e. reneaaron/webln-demo at the cost of an extra request, too. 

 - Add a lightning address to your GitHub profile

#### Checklist:


reneaaron:
@bumi Support for GitHub profiles is here! 💯 

Do you think we should support donation on user repositories for the price of an extra request? Since currently we try to avoid that I initially skipped that, but I am happy to add it. 

Regarding docs update:
I've updated the table [here](https://github.com/getAlby/website/issues/78), do we need to add info about this battery to other places as well?

src/extension/content-script/batteries/index.ts
```diff
@@ -20,6 +20,9 @@ const enhancements = [
   Peertube,
   VimeoVideo,
   Medium,
+  GitHubProfile,
+
+  // Monetization must likely be always the last one
```
escapedcat, line 25: Wanna add a short reason to this?

escapedcat (approved):

bumi (commented):
as the URL matches for both repos and profiles. Maybe we can directly also read the data from the readme? 

so this one would work for both, user profiles and repos.

bumi:
I would also delete the existing one: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/content-script/batteries/GitHubRepo.ts - it used the lndonate idea of a file. but requires an additional HTTP request and imo it is good to put the info in the readme.

for a second I also thought we could additionally load the info from the `.github/FUNDING` file (check for a link that has `a lighting=...` parameter. 
(see https://github.com/bumi/lnme/blob/master/.github/FUNDING.yml

reneaaron:
> for a second I also thought we could additionally load the info from the .github/FUNDING file (check for a link that has a lighting=... parameter.

I didn't know about that. That is awesome! 💯 

> as the URL matches for both repos and profiles. Maybe we can directly also read the data from the readme?

That should definitely be possible, too. 

So we would have the following options
 - GitHub Profile via⚡ in short bio or README
 - GitHub Repo via .github/FUNDING.yml
 - GitHub Repo via ⚡in README

bumi:
> For supporting the FUNDING.yml I'd probably create another issue as I think we need to put some more thought into that. (how to identify lightning addresses, lnurlp, etc)

+1 - and it already works when the user clicks on the sponsor link. Alby will open with such lightning query params.

src/extension/content-script/batteries/GitHub.ts
```diff
@@ -0,0 +1,79 @@
...
+function handleProfilePage() {
+  const shortBioElement =
+    document.querySelector<HTMLHeadingElement>("[data-bio-text]");
+
+  // This is not a GitHub profile
+  if (!shortBioElement) return;
+
+  const address = parseElement("[data-bio-text]") || parseElement("article");
+
+  if (address) {
```
bumi, line 35: I kinda prefer the guards like in line 31 - if there is no else branch

bumi (approved):

src/extension/content-script/batteries/GitHub.ts
```diff
@@ -0,0 +1,79 @@
...
+        recipient: address,
+        ...getOriginData(),
+        description:
+          document.querySelector<HTMLHeadingElement>("[data-bio-text]")
+            ?.innerText ?? "",
+        name:
+          document.querySelector<HTMLHeadingElement>('[itemprop="name"]')
+            ?.innerText ?? "",
+        icon:
+          document.querySelector<HTMLImageElement>(`img.avatar-user`)?.src ??
```
bumi, line 48: we need to be more specific here. 

e.g. (`.Layout-sidebar .h-card img.avatar-user`)

the current one matches also other avatars on the page (e.g. the current logged in user)

bumi: this goes also for the other selectors.

bumi (commented):
@reneaaron I was too fast with merging... :/ 
do you have time to update the CSS selectors and make those more specific?

reneaaron:
@bumi Sure, i'll have a look at it!

src/extension/content-script/batteries/GitHub.ts
reneaaron: I just created a PR that should fix those issues 👍 Sorry for missing that! 😑

PR #769, opened 2022-04-01
fix: update citadel sdk for core lightning support

AaronDewes:

This updates the Citadel SDK and makes a minor change that is required to support Core Lightning (previously called c-lightning) on Citadel in the future.

I manually ensured everything was still working correctly when connected to my local Citadel.

#### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@babel/runtime": "^7.15.4",
     "@bitcoin-design/bitcoin-icons-react": "^0.1.9",
     "@headlessui/react": "^1.5.0",
-    "@runcitadel/sdk": "^0.3.3",
+    "@runcitadel/sdk": "^0.3.8",
     "@tailwindcss/forms": "^0.5.0",
     "assert": "^2.0.0",
     "axios": "^0.26.1",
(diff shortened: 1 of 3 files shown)


bumi:
utACK. 

in your the sdk code the `MiddlewareLND` now works for both lnd and cln? 

(btw. slightly related thread: https://twitter.com/EricSirion/status/1509906940767481859 )

bumi (approved):

escapedcat:
@AaronDewes would you mind having a look at this?: https://github.com/getAlby/lightning-browser-extension/pull/747   
For some reason the jest setup has issues with the citadel-import. Maybe you have an idea what needs to be changed. If not no worries.

AaronDewes:
>  in your the sdk code the MiddlewareLND now works for both lnd and cln?

Yes, only the name is a leftover.

AaronDewes:
> @AaronDewes would you mind having a look at this?: #747 For some reason the jest setup has issues with the citadel-import. Maybe you have an idea what needs to be changed. If not no worries.

It seems to be because we use ESM in @runcitadel/sdk, but Alby doesn't use it otherwise

bumi:
@escapedcat #747 is independent of this, correct? 
so we can merge this one?

escapedcat:
> @escapedcat https://github.com/getAlby/lightning-browser-extension/pull/747 is independent of this, correct?
so we can merge this one?

Yup, all good here

PR #776, opened 2022-04-05
feat: add prompt for webln makeInvoice calls

### LINKED ISSUE #459: User prompt for makeInvoice call
The user should be able to confirm the first makeInvoice call similar to the confirmPayment call and the [signMessage call](https://github.com/getAlby/lightning-browser-extension/issues/242)

By default we should remember that decision and only ask the user once but provide a checkbox for this.

bumi:
#### Link this PR to an issue
Fixes #459

So far we did not prompt the user to create invoices.
With this change we now always show a prompt and ask the user for confirmation.
In the future we should add some "allowance" like system for invoices, too. Or allow the user to "remember" the decision for that website

image
image

do a `webln.makeInvoice()` call

#### Checklist:


dylancom:
I think it would be better UX wise to just show text instead of an input when amount or memo is disabled.
(Visually it still makes you think that you could enter something)

bumi:
@dylancom updated. can you check again?

dylancom:
Route not found: webln/makeInvoice

bumi:
merged to release :)

Issue #781, opened 2022-04-09
Use i18n for the whole app


escapedcat:
### Is your feature request related to a problem?

#711 added support for i18n-translation.  

As a first example [this headline was translated](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Welcome/Welcome.tsx/#L88
).  
Translations (for i.e. English) will be maintained [in this file.](https://github.com/getAlby/lightning-browser-extension/blob/master/src/i18n/locales/en.json)

### Describe the solution you'd like

All other strings in the app need to be replaced by i18n-placeholders and moved to the `en.json`.  
It should be somewhat possible to guess where a certain string is being used by looking at a structured json-file.

Could be good to look at other projects using the same plugin, i.e. 
- https://github.com/cstenglein/raspiblitz-web/blob/master/src/i18n/langs/en.json
- ...happy for further examples

This can be done in multiple steps. If you are interested to give this a try maybe start with a PR for one section first and get feedback first before moving on.

### Are you working on this?

No

pavanjoshi914:
thanks willing to push translation pr's as soon  as possible

escapedcat:
Via Weblate after approving the OSS libre hosting plan:  

> Don’t forget to mention you use Weblate in the README and the website of your project! You can use the lovely stuff from the Share menu of your Weblate project for it.

Let's support them with mentions.

thebrandonlucas:
Hello, I've started working on this [here](https://github.com/getAlby/lightning-browser-extension/pull/883). This just includes the "welcome" pages since I wanted to receive feedback on that before doing more. Thanks!

thebrandonlucas:
> It should be somewhat possible to guess where a certain string is being used by looking at a structured json-file.

@escapedcat Should we be optimizing for guessing where strings are in the code or guessing where they are in the UI? For example, in `en.json` in `welcome.choose_connector.alby`, I could add a key called `new_wallet` which contains all the strings for the Alby hosted wallet page, because the enclosing folder for the code with those strings is called `NewWallet`, or I could just name it something like `page` because it's the corresponding "page" under the `choose_connector.alby` key in the UI.

escapedcat:
Good question. I think most projects I've seen so far are using a mix.  
On one hand you have certain "views" (i.e. a special headlien will only show up on one page) an on the other hand we have more general things like "forms" (i.e. a "sats" input mitht show up in different forms).

Does this help?

[...]

PR #786, opened 2022-04-10
feat: theme selector

### LINKED ISSUE #598: Make dark/light mode configurable

secondl1ght:
#### Link this PR to an issue
Fixes #598 
Closes #618 

Added a theme selector to the Settings screen which allows the user to choose dark, light or system theme for Alby. This settings is stored and used on all screens for Alby, I also added a transition animation for when you change the selection on the Settings page.

image

Tested all the 3 different theme options on all the screens within Alby. Also ran `console.log` tests to make sure the Settings were persistently updated.

#### Checklist:


src/app/context/AuthContext.tsx
```diff
@@ -27,6 +27,24 @@ interface AuthContextType {
 const AuthContext = createContext({} as AuthContextType);
 
 export function AuthProvider({ children }: { children: React.ReactNode }) {
+  //theme
+  useEffect(() => {
+    api.getSettings().then((response) => {
```
escapedcat, line 32: More like a general question: Does Alby prefer `.then()` or `async/await` or just doesn't matter?

src/app/context/AuthContext.tsx
```diff
@@ -27,6 +27,24 @@ interface AuthContextType {
...
+      const settings = response;
```
escapedcat, line 33: Cab it be worth typing this?

src/app/context/AuthContext.tsx
escapedcat, line 32: Does it make sense to move this into a i.e. util function because it looks like the same code is being used in 3 places?

src/app/screens/Settings.tsx
```diff
@@ -140,6 +157,26 @@ function Settings() {
...
+              <Select
+                name="theme"
+                value={settings.theme}
+                onChange={(ev) => {
+                  saveSetting({
+                    theme: ev.target.value,
+                  });
+                }}
+              >
+                <option value="dark">Dark</option>
```
escapedcat, line 173: Now that we have translaitons in should each ne PR already be forced to use translation-placeholders instead of strings?

src/app/context/AuthContext.tsx
secondl1ght: `.then()` was already being used for this function previously so that's why I used it, but it looks like both then and `await` are used in the codebase. Not sure which one is preferred though.

src/app/context/AuthContext.tsx
secondl1ght: You mean typing the response of the `getSettings` call?

src/app/context/AuthContext.tsx
secondl1ght: I thought about what I could do to make this more DRY, but didn't know you could turn a `useEffect` hook into a function. It is possible?

src/app/screens/Settings.tsx
secondl1ght: Ok I can update this with the translations.

secondl1ght: I just noticed none of the other items on this page have the translations yet it looks like, should we put in a separate PR to update the entire app with translations?

src/app/context/AuthContext.tsx
escapedcat: Yeah, just noticed this. Might be good to think about a preferred way.  
@dylancom what do you think?

Definitely not blocking this.

escapedcat: Not sure, I assumed you can use a function in that place.

src/app/context/AuthContext.tsx
escapedcat: Yes, would that work?

src/app/screens/Settings.tsx
escapedcat: Created #781 for that.  
But it might be good to just start with this and do it step by step already.

src/app/context/AuthContext.tsx
secondl1ght: I think it would work but the variables that get sent from the API call are already typed so it might be redundant?

src/app/context/AuthContext.tsx
secondl1ght: Ok I have never done that before but i'll give it a try and see if it works.

secondl1ght: Looks like I was able to get it working, see latest commit.

src/app/utils/index.ts
```diff
@@ -1,3 +1,25 @@
...
+
 export function classNames(...classes: (string | boolean)[]) {
   return classes.filter(Boolean).join(" ");
 }
+
+export function useThemeEffect(dependencies: undefined | string) {
+  useEffect(() => {
+    api.getSettings().then((response) => {
+      const settings = response;
+      if (settings.theme === "system") {
```
bumi, line 12: can you add a comment here what this is doing?

src/app/router/Welcome/Welcome.tsx
```diff
@@ -53,6 +53,12 @@ function App() {
   const { t } = useTranslation(["welcome"]);
   const location = useLocation();
   const routesElement = useRoutes(routes);
+  //theme
+  useEffect(() => {
+    if (window.matchMedia("(prefers-color-scheme: dark)").matches) {
```
bumi, line 58: why is this here but not on the other router/* ?

bumi (changes requested):
Where do those `transition-colors` need to be added?

src/app/utils/index.ts
secondl1ght: Ok I added some comments to explain how the function works, please let me know if this is what you wanted.

src/app/router/Welcome/Welcome.tsx
secondl1ght: This screen needed its own `useEffect` function because I could not use the one I wrote in `utils` for the other screens. The reason is that for the other screens I am checking the `Settings` for the theme selection. But on the Welcome screen the settings have not yet been initiated by the initial user on boarding. So in this case we only need to check what the users system theme settings are and not use the API call that doesn't exist yet. After the user sets up their first wallet they then have access to the Settings screen and can adjust the theme from there, which is why all of the screens after the initial on boarding check the Settings API.

secondl1ght: Does this make sense?

secondl1ght:
> Where do those `transition-colors` need to be added?

I added this so that there was a smoother transition effect when the user switches between dark/light themes. It is not necessary though, I can remove it if you want?

src/app/router/Welcome/Welcome.tsx
bumi: hmm. yes makes sense. 
but does it hurt doing the settings call? When there is no special setting then it just uses the default system theme, doesn't it? 

having less code is always better - less things that break :)

secondl1ght: I think I tired using the same function for this screen but it wasn't working but I will t

> > I added this so that there was a smoother transition effect when the user switches between dark/light themes. It is not necessary though, I can remove it if you want?
> 
> I don't know. I just fear that this will be forgotten when developing new screens.

Yeah it's not a big deal, most people wouldn't notice it to be honest. I will remove.

secondl1ght:
> > I added this so that there was a smoother transition effect when the user switches between dark/light themes. It is not necessary though, I can remove it if you want?
> 
> I don't know. I just fear that this will be forgotten when developing new screens.

Yeah it's not a big deal, most people wouldn't notice it to be honest. I will remove.

src/app/router/Welcome/Welcome.tsx
secondl1ght: I think I tried this originally but it was not working - I will try again.

secondl1ght: So yes you are right about defaulting to system theme but for some reason when I first launch Alby this code does not work. There is an error in the devtools console, but when I refresh the page then it starts working. So I think in production it may also not work when you first launch Alby. Maybe it is because you cant use the API call right when Alby opens for the first time? Maybe @dylancom would know more but personally I think having one unique `useEffect` call here to get it working is not that much of an issue.

This is the error it gives:
`Uncaught (in promise) Error: Could not establish connection. Receiving end does not exist.
    at wrappedSendMessageCallback (browser-polyfill.js:1187:1)`

secondl1ght: I think it means that `api.getSettings` does not exist yet when you first launch the extension.

bumi: maybe we first need to debug and understand that error before drawing conclusions.

secondl1ght: Ok I did some more debugging and was able to add a new parameter to the function to check if it is on the initial Welcome page. If it is then it will run the code without the API call, all the other pages will use the API call like before.

bumi: I really think we should first understand that error and then see how we can fix it instead of hacking around it. 
why is doing that call to load the settings a problem? I don't know the setup on this right now but I am wondering why the different pages/screens need to be treated differently. 

though it is great that it works, magic parameters make the code harder to read. people who don't have the background or in a few months don't know what the `true/false` means when reading `useThemeEffect(noDependencies, false);`

src/app/utils/index.ts
```diff
@@ -1,3 +1,32 @@
+import { useEffect } from "react";
+import api from "../../common/lib/api";
+
 export function classNames(...classes: (string | boolean)[]) {
   return classes.filter(Boolean).join(" ");
 }
+
+export function useThemeEffect(dependencies: undefined | string) {
```
dylancom, line 8: We should move this into a "hooks" folder.

dylancom (changes requested):

src/app/context/AuthContext.tsx
dylancom: I think in general async/await is preferred in terms or readability and to avoid callback hell.
It's currently a bit of a mix of both in the codebase.

dylancom:
Seems to work fine on my end (no error)

src/app/router/Welcome/Welcome.tsx
dylancom: The error was caused by onMessage listeners being registered after opening welcome.html. So I made sure that welcome is now opened after initialization has finished.

secondl1ght:
Ok thanks a lot to @dylancom for refactoring this and figuring out that hard to solve bug!! 🙌🏼 

I believe this PR is ready to be merged now.

PR #791, opened 2022-04-12
fix: more specific selectors for github battery

reneaaron:
Followup from [here](https://github.com/getAlby/lightning-browser-extension/pull/767#pullrequestreview-938252982).

Fixes some issues with the GitHub battery.

### DIFF
--- a/src/extension/content-script/batteries/GitHub.ts
+++ b/src/extension/content-script/batteries/GitHub.ts
@@ -1,7 +1,7 @@
 import getOriginData from "../originData";
 import setLightningData from "../setLightningData";
 
-const urlMatcher = /^https:\/\/github.com\/\S+$/;
+const urlMatcher = /^https:\/\/github.com\/([^/]+)(\/([^/]+))?$/;
 
 const battery = (): void => {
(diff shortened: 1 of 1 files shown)


src/extension/content-script/batteries/GitHub.ts
```diff
@@ -1,7 +1,7 @@
 import getOriginData from "../originData";
 import setLightningData from "../setLightningData";
 
-const urlMatcher = /^https:\/\/github.com\/([^/]+)(\/([^/]+))?$/;
+const urlMatcher = /^https:\/\/github.com\/\S+$/;
```
bumi, line 4: why did you change this? I liked the regex before because it is a bit more specific. 
on repo files or PRs (like this`https://github.com/getAlby/lightning-browser-extension/pull/791/files`) for example we would not even need to try to find a lightning address.

reneaaron: My intention was to include the profile tabs for the battery as well (repositories, projects, etc) but I agree that this is way too open. I'll change it back.

bumi (approved):

PR #796, opened 2022-04-16
fix: add JWT bypass screen for when Galoy requestAuthCode fails

vindard:
## Description

We currently have an issue where on some Galoy instances we require users to use the 'captcha' endpoint for request phone codes to login (see [tweet](https://twitter.com/slavakurilyak/status/1513985145329618948)). This requires an extra geetest step to complete the captcha which is not straightforward to implement for other folks using our API elsewhere.

We are currently in the process of revamping out authentication process to use Kratos from the Ory stack instead (see more: https://www.ory.sh/) but in the meantime we wanted to at least expose a potential bypass for more advanced users.

### Solution

This PR implements an additional login screen that can take a JWT directly for instances where the captcha endpoint is required. Right now this only affects the Bitcoin Beach instance, and the Bitcoin Jungle instance still works with the current phone code method.

image

Advanced users can login via our web wallet and then pull a JWT from the request `authorization` header for the graphql calls to the API.

This isn't really ideal, but it's a temporary workaround until we get the new login system in place and can clean up this flow again here.

### DIFF
--- a/src/app/screens/connectors/ConnectGaloy/index.tsx
+++ b/src/app/screens/connectors/ConnectGaloy/index.tsx
@@ -43,6 +43,10 @@ export default function ConnectGaloy(props: Props) {
   const [smsCodeRequested, setSmsCodeRequested] = useState<
     boolean | undefined
   >();
+  const [jwt, setJwt] = useState<string | undefined>();
+  const [acceptJwtDirectly, setAcceptJwtDirectly] = useState<
+    boolean | undefined
+  >();
 
   function handlePhoneNumberChange(event: React.ChangeEvent<HTMLInputElement>) {
     setPhoneNumber(event.target.value.trim());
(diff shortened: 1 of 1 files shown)


vindard:
@bumi @MoritzKa @dylancom any thoughts on this workaround?

bumi:
I guess we have to do it like this. what's the Galoy plan with this? for an API this is generally a problem, isn't it?

bumi:
thanks for fixing it, I make sure we got it in the next release.

NicolaLS:
> This requires an extra geetest step to complete the captcha which is not straightforward to implement for other folks using our API elsewhere.

Hi I stumbled upon this problem setting up alby w bbw and wanted to submit a PR but I guess it is not as easy as I thought (using the captchaCreateChallange & captchaRequestAuthCode endpoint). Could you elaborate what you mean by "(...) other folks using our API elsewhere" and what is the status on the Kratos thing ? haven't found any issues searching for that.

PR #798, opened 2022-04-17
feat: add ~ as a root import alias

dylancom:

Added `~` as an alias to our root folder. This makes imports and moving files easier (get's rid of relative imports using ../../../../../)

Added `@screens` & `@components` aliases.

**Before**
`import type { Accounts } from "../../../types";`

**After**
`import type { Accounts } from "~/types";`


escapedcat:
https://stackoverflow.com/a/68772457/1667461  
Does this help regarding the aliases for storybook?

dylancom:
@escapedcat I found an easier way where I don't have to redefine the aliases for storybook.

.storybook/main.js
```diff
@@ -31,6 +32,7 @@ module.exports = {
       ],
       include: path.resolve(__dirname, "../"),
     });
+    config.resolve.plugins = [new TsconfigPathsPlugin()];
```
escapedcat, line 35: 👏

escapedcat (approved):

PR #806, opened 2022-04-20
fix: prevent duplicate onboarding / configuration

### LINKED ISSUE #180: Prevent onboarding process from being executed multiple times
Currently we do not check if a user already has setup an account or a password when accessing the welcome page. 
To prevent a user from inconsistent config data (e.g. by setting different passwords) we should prevent accessing the welcome page if a user is already onboarded.

stefanwuest:
#### Link this PR to an issue
Fixes #180 

So far, users were able to navigate to `/welcome.html` and complete the onboarding process again, even after the setup was completed. This change redirects the user to `/options.html` when the onboarding was already completed. I also changed redirecting the current tab, instead of creating a new one and closing the current one, which should be a less "flashy" experience for the user.

I tested on Chrome.

#### Checklist:


bumi:
great, thanks for the PR.
looks good to me. using the redirect would mean that the user "looses" the currently open tab, doesn't it?

stefanwuest:
> great, thanks for the PR. looks good to me. using the redirect would mean that the user "looses" the currently open tab, doesn't it?

yes, you are right, this actually just makes sense for the case where the onboarding was already completed. will adapt.

bumi (approved):
utACK

thanks!

PR #825, opened 2022-04-26
Blocklist feature

### LINKED ISSUE #258: block webln provider request on a certain site
It should be possible to block a domain (and remove from block list later) because otherwise it might pop up every single time and user has to close the popup every time.

pseudozach:
#### Link this PR to an issue
Fixes #258

Added a new feature blocklist as explained in https://github.com/getAlby/lightning-browser-extension/issues/258 which allows repeated popups on sites that support webLN but user chooses not to allow the extension.

image

Currently in progress and *not* working as intended.
- Added new db `blocklist`
- Sites get added to blocklist if user clicks `don't ask for this site again` button on popup
- Unable to stop popup -> need help here
- Unable to show blocklisted site in `publishers` list because publishers are only listed for connected sites. -> need help here

#### Checklist:


bitcoinuser:
Feedback:

I think when asking (Do you want to grant them access?) I think instead of button confirm, should have two buttons (Yes) and (No). If the user choose No, will appear the message: Not ask for this site again? Yes or no.

pseudozach:
> Feedback:
> 
> I think when asking (Do you want to grant them access?) I think instead of button confirm, should have two buttons (Yes) and (No). If the user choose No, will appear the message: Not ask for this site again? Yes or no.

IMO there should be 2 buttons (yes/no) and `don't ask again` as another smaller option but the underlined warning is indeed confusing.

src/app/screens/Enable/index.tsx
```diff
@@ -30,6 +31,20 @@ function Enable(props: Props) {
     event.preventDefault();
   }
 
+  async function block(event: React.MouseEvent<HTMLAnchorElement>) {
+    msg.error(
+      "User added site to blocklist domain, host " +
```
escapedcat, line 36: Can [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) be used?

src/app/components/PublishersTable/index.tsx
```diff
@@ -95,6 +96,11 @@ export default function PublishersTable({
                   </div>
                 )}
               </td>
+              <td className="px-6 py-6 whitespace-nowrap text-right">
+                {publisher.blocked && (
+                  <div className="ml-40">Site blocklisted!</div>
```
escapedcat, line 101: We try to move i18n step by step, would be nice if you could move texts here already: https://github.com/getAlby/lightning-browser-extension/blob/master/src/i18n/locales/en.json

src/app/screens/Publishers.tsx
```diff
@@ -20,9 +20,14 @@ function Publishers() {
       const response = await utils.call<{
         allowances: Allowance[];
       }>("listAllowances");
+      const response2 = await utils.call<{
+        blocklist: Blocklist[];
+      }>("listBlocklist");
       const allowances = response.allowances.map((allowance) => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        let retobj: any = allowance;
```
escapedcat, line 28: Would be great to avoid `any` if possible.

src/extension/background-script/actions/blocklist/add.js
```diff
@@ -0,0 +1,20 @@
+import db from "../../db";
```
escapedcat, line 1: We are in the process of switching everything to TS. We should avoid adding ne JS files by now.

src/extension/content-script/index.js
```diff
@@ -18,6 +18,7 @@ const weblnCalls = [
 import extractLightningData from "./batteries";
 
 if (shouldInject()) {
+  console.log("content-script shouldInject");
```
escapedcat, line 21: Is left from this debugging? If yes, can it be removed.

src/extension/content-script/shouldInject.js
```diff
@@ -39,3 +46,17 @@ function documentElementCheck() {
...
+  try {
+    const currentUrl = window.location.pathname;
+    console.log("shouldinject currentUrl, ", currentUrl);
+    const isBlocked = msg.request("getBlocklist", {
+      domain: currentUrl,
+      host: currentUrl,
+    });
+    console.log("2shouldinject isBlocked? ", isBlocked);
+  } catch (e) {
+    if (e instanceof Error) console.log(e.message);
```
escapedcat, line 60: If this stays it could be switched to `console.error(...)`

src/app/components/PublishersTable/index.tsx
pseudozach: sure, I can do that but I'll need some sample code to see how to pull the text from the i18n file and show to user.

src/extension/content-script/shouldInject.js
pseudozach: the reason for these is that I couldn't block the popup (stop the webln injection) based on the blocklist.

src/extension/background-script/actions/blocklist/add.js
pseudozach: I don't understand what's needed here. I followed the same approach as many other files in codebase for db. when functionality is finalized, I can adapt it to the latest convention.

src/app/components/PublishersTable/index.tsx
escapedcat: Does this help?  
https://github.com/MiguelMedeiros/lightning-browser-extension/blob/test/integration-tests/src/app/router/Welcome/Welcome.tsx/#L87

src/extension/background-script/actions/blocklist/add.js
escapedcat: Rename the file(s) from `.js` to `.ts` and see what happens. If you run into issues, let us know.

src/extension/content-script/shouldInject.js
escapedcat: For these consoles? Then it's debugging code which will be removed later? ok

src/app/screens/Publishers.tsx
pseudozach: please check my attempt at fixing this, I added a Publisher type.

src/app/screens/Publishers.tsx
```diff
@@ -20,9 +20,13 @@ function Publishers() {
       const response = await utils.call<{
         allowances: Allowance[];
       }>("listAllowances");
+      const response2 = await utils.call<{
```
escapedcat, line 23: The responses could be renamed to reflect the usage

src/extension/background-script/actions/blocklist/add.js
pseudozach: done. please check.

src/app/components/PublishersTable/index.tsx
pseudozach: yes and no. I see how it can be used but my branch doesn't even have this package `react-i18next`, looks like I'll need to merge all of these changes first.

escapedcat: With a little bit of luck rebase should work ;)

bumi:
super cool! thanks for the PR.
I got issues with my computer right now will help here once it’s fixed.

bumi:
I've merged master and pushed a commit that does the blocklist check in the content script and only injects the in page script if the host is not on the blocklist. 

This seems to work. 

I have to think about the storage of the websites. 
There was the plan to allow the user to define permissions of what calls websites can make. (this will be relevant once we extend the exposed functionality)
So maybe these permissions can be combined with the blocklist? 
Or maybe we should make the blocklist an attribute on the "allowances" (which should be renamed to "websites" or something in the future)?

But I hope this already helps and unblocks you?

pseudozach:
This is great. it works. Thanks! It does what's needed. Once you click `dont ask again for this site` it doesn't popup on refresh.

I agree we need a different/better place to put the blocklist but I don't know where. Is there any preference? Let me know what's needed to merge this to main.

bumi:
I think the main thing that we need to do to get this in: UI/UX - how can the user remove the page from the blacklist? can we even show a small notice on the prompt of a page is on the blacklist and remove it from there. (meaning when the user is on a blacklisted page, clicks on the Alby icon and instead of the publisher screen there we show a note that this page is blacklisted and the option to remove it from the blacklist) 

maybe @dvoroneca has some ideas?

src/extension/background-script/actions/blocklist/add.ts
```diff
@@ -0,0 +1,23 @@
+import db from "../../db";
+
+const add = async (
+  message: { args: { host: string; name: string; imageURL: string } },
+  sender: unknown
```
escapedcat, line 5: Not completetely sure but we use this in other places, does it work here as well?  
https://github.com/getAlby/lightning-browser-extension/blob/master/src/common/lib/utils.ts/#L137
@dylancom what do you think?

src/extension/background-script/actions/blocklist/list.ts
```diff
@@ -0,0 +1,23 @@
+import { Blocklist } from "../../../../types";
+import db from "../../db";
+
+const list = async (message: string, sender: unknown) => {
+  // TODO add filter and ordering?
```
escapedcat, line 5: Is this a `TODO` for this PR or later on? If later on, maybe create an issue and link it here

pseudozach:
Thanks @bumi for the idea. I added a `unblock` button that shows up only when a site is in blocklist. 
Here's a demo video of what it looks like: https://www.youtube.com/watch?v=rRlyAQLs6Yc

Let me know if this is ok or any updates are needed. It definitely accomplishes what I had in mind.

escapedcat:
No need to worry about the E2E tests. It's an external issue.

src/extension/background-script/actions/blocklist/add.ts
escapedcat: Action/message handling has been updated. Please have a look at i.e.: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/accounts/add.ts#L6  

All `blocklist` actions should follow this example. Ideally [with Jest tests](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/accounts/__tests__/add.test.ts) as well but could be done later.

src/extension/background-script/actions/blocklist/add.ts
```diff
@@ -0,0 +1,23 @@
...
+) => {
+  const host = message.args.host;
```
escapedcat, line 7: This should work:

```js
const {host, name, imageURL } = message.args
```

src/extension/background-script/actions/blocklist/add.ts
```diff
@@ -0,0 +1,23 @@
...
+  const name = message.args.name;
+  const imageURL = message.args.imageURL;
+
+  await db.blocklist.add({
+    host: host,
```
escapedcat, line 12: This should work:

```js
    host,
    name,
    imageURL,
    isBlocked: true,
```

src/extension/background-script/actions/blocklist/delete.ts
```diff
@@ -0,0 +1,22 @@
...
+    .where("host")
+    .equalsIgnoreCase(host)
+    .first();
+
+  if (blocklist?.id) {
+    await db.blocklist.delete(blocklist.id);
+    await db.saveToStorage();
+    return { data: true };
+  } else {
+    return { data: false };
```
escapedcat, line 18: Should we add a error message here? See i.e.: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/accounts/delete.ts#L35

src/extension/background-script/actions/blocklist/list.ts
escapedcat: I think this was juts copy&paste. Sorting is done in the code below this. I think the comment can be removed.

escapedcat (changes requested):

src/app/screens/Enable/index.tsx
```diff
@@ -30,6 +31,19 @@ function Enable(props: Props) {
     event.preventDefault();
   }
 
+  async function block(event: React.MouseEvent<HTMLAnchorElement>) {
+    msg.error(
+      `User added site to blocklist domain, host
+        ${props.origin.domain},
+        ${props.origin.host}`
+    );
+    await utils.call("addBlocklist", {
```
bumi, line 40: @escapedcat at some point I wanted to namespace these calls. 
e.g. `utils.call("blocklist/add")` to have this grouped in the background-script. 
I did not use it for simplicity. but now that we get more and more calls we maybe slowly could think about this again.

src/app/screens/Home/index.tsx
```diff
@@ -39,17 +42,34 @@ function Home() {
...
     }
   }
 
+  async function unblock() {
+    try {
+      if (currentUrl?.host)
+        await utils.call("deleteBlocklist", {
+          host: currentUrl.host,
+        });
+      window.location.reload(); // refresh home page
```
bumi, line 67: why do we need this actually? isn't it enough to set `blocklist` to null?

src/app/screens/Publishers/index.tsx
```diff
@@ -31,7 +35,14 @@ function Publishers() {
             },
           };
         }
-        return allowance;
+        if (
+          blocklistResponse.blocklist.find(
+            (item) => item.host === allowance.host
+          )
+        ) {
+          retobj.blocked = true;
```
bumi, line 43: for simplicity reasons I think we could remove this here. 

A user can only add a page to the blocklist on the first `enable` call. In that case no `allowance` is saved. 
And because we show an unblock button directly on the popup home screen I think this is enough.

what do you think? 


in the future I also think of a permission system with very granular permissions per host. 
Then I think we have to rethink where/how we show this anyway.

src/common/lib/api.ts
```diff
@@ -96,6 +96,8 @@ export const deleteAccount = (id: string) =>
   ]);
 export const unlock = (password: string) =>
   utils.call<UnlockRes>("unlock", { password });
+export const getBlocklist = (host: string) =>
+  utils.call<{ blocked: boolean }>("getBlocklist", { host });
```
bumi, line 100: what's the best practice here? should this be a named type? (or whatever this is called?) @escapedcat

src/extension/background-script/actions/blocklist/add.ts
bumi: @pseudozach can you help with this and have a look?

src/extension/background-script/actions/blocklist/delete.ts
bumi: this is the case when there is no blocklist entry, correct?

can't we then simply also return true?

src/app/screens/Home/index.tsx
pseudozach: this is needed so the extension UI will reset and `unblock` button will disappear.

bumi: but we have this check here: https://github.com/getAlby/lightning-browser-extension/pull/825/files#diff-041371f0bc0d8e5acd8d60f55e0b217ee4320404c44f5ebe9a3d61db95753a45R254

isn't then `setBlocklist(null)` enough? the `blocklist` is falsy and that button should disappear.

src/app/screens/Publishers/index.tsx
pseudozach: remove what exactly? I don't think this can be removed - this is where we are looping through the blocklist and finding blocked host. 
agree that this will need to be rewritten at some point.

bumi: yes, but when is a host blocked? isn't it normally the case that a host is blocked OR it is in the allowance list? 
when a host gets blocked it is NOT added to the allowance list. 

or what am I missing?

src/app/screens/Home/index.tsx
pseudozach: I've tested again and indeed it doesn't automatically disappear. reload is needed, so I'm leaving this in.

src/app/screens/Publishers/index.tsx
pseudozach: I think you're right. this may be left over from when I was thinking that blocklist would go into a list in publishers screen. But we're not doing that at the moment. Removed.

src/extension/background-script/actions/blocklist/delete.ts
pseudozach: agreed. removed data:false, this will always return true. We can revisit if you think there should be a check.

pseudozach:
@bumi @escapedcat I think all change requests should be covered. can you please check if anything else pending to merge this when you get a chance?

src/common/lib/api.ts
escapedcat: Yeah, better define a type which can be reused and also can be used in unit-tests.  
See how it's done just above with `UnlockRes`

src/app/components/PublishersTable/index.tsx
escapedcat: Is latest master already merged/rebased @pseudozach ? If there any issues I can support with that.

src/app/screens/Enable/index.tsx
```diff
@@ -71,6 +85,14 @@ function Enable(props: Props) {
         </p>
 
         <ConfirmOrCancel label="Enable" onConfirm={enable} onCancel={reject} />
+
+        <a
+          className="underline mt-8 text-sm text-gray-500 dark:text-gray-400"
+          href="#"
+          onClick={block}
+        >
+          Don{`'`}t ask for this site again
```
escapedcat, line 94: This seems a bit odd. Maybe just write "Do not"? Or use one of eslints suggestions, i.e. `&apos;`

src/app/screens/Home/index.tsx
```diff
@@ -231,6 +251,21 @@ function Home() {
           />
         </div>
 
+        {blocklist && (
+          <div className="mb-2 items-center py-3 dark:text-white">
+            <p className="py-1">This site is currently in blocklist.</p>
+            <Button
+              fullWidth
+              // icon={<ReceiveIcon className="w-6 h-6" />}
```
escapedcat, line 259: What's this?

src/extension/background-script/actions/blocklist/delete.ts
```diff
@@ -0,0 +1,20 @@
+import db from "../../db";
+
+const deleteBlocklist = async (
+  message: { args: { host: string } },
+  sender: unknown
```
escapedcat, line 5: If the sender isn't being used it can be removed here and for other occurrences.

src/extension/background-script/actions/blocklist/delete.ts
```diff
@@ -0,0 +1,20 @@
+import db from "../../db";
+
+const deleteBlocklist = async (
+  message: { args: { host: string } },
```
escapedcat, line 4: For all these messages types should be used, like we do here:
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/accounts/delete.ts#L4
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/types.ts#L87

src/app/screens/Home/index.tsx
```diff
@@ -18,12 +18,15 @@ import { useNavigate } from "react-router-dom";
...
 import browser from "webextension-polyfill";
 import api from "~/common/lib/api";
+import utils from "~/common/lib/utils";
 import type { Allowance, Battery, Transaction } from "~/types";
 
 dayjs.extend(relativeTime);
 
 function Home() {
   const [allowance, setAllowance] = useState<Allowance | null>(null);
+  const [blocklist, setBlocklist] = useState<boolean | null>(null);
```
escapedcat, line 28: Can we call this `isBlocked`? That indicates it's a boolean and that this website is blocked.  
"if blocklist than show button" is a bit unclear.

src/app/screens/Home/index.tsx
escapedcat: `setBlocklist(false);` seems to work fine or am I missing something?

escapedcat (changes requested):

src/app/screens/Home/index.tsx
bumi: works fine for me, too. I did `setBlocklist(null)` - as `null` is the default value and also specified in the type.

src/app/components/PublishersTable/index.tsx
pseudozach: this is added

src/app/screens/Home/index.tsx
```diff
@@ -18,12 +18,15 @@ import { useNavigate } from "react-router-dom";
...
 import browser from "webextension-polyfill";
 import api from "~/common/lib/api";
+import utils from "~/common/lib/utils";
 import type { Allowance, Battery, Transaction } from "~/types";
 
 dayjs.extend(relativeTime);
 
 function Home() {
   const [allowance, setAllowance] = useState<Allowance | null>(null);
+  const [isBlocked, setIsBlocked] = useState<boolean | null>(null);
```
bumi, line 28:
```suggestion
  const [isBlocked, setIsBlocked] = useState<boolean>(false);
```

I think now we can always use a boolean and default to `false`

src/extension/background-script/actions/blocklist/get.ts
```diff
@@ -0,0 +1,21 @@
+import { MessageBlocklistDelete } from "~/types";
+
+import db from "../../db";
+
+const get = async (
+  message: MessageBlocklistDelete
```
escapedcat, line 6: Wrong type? This would be noticed with unit tests later on :)

pseudozach: I didn't know if I should create a new `MessageBlocklistGet` type that would be same as `MessageBlocklistDelete`. is that the best practice? I felt like it would be repetitive.

escapedcat: The `action` would be different.  
We need this anyways if we add tests later because. See `account`-types. Leaving this would be confusing.  

We would also need a type here later: https://github.com/getAlby/lightning-browser-extension/pull/825/files#diff-fd1daf979191de8eaf96a501e8ef14d98ca457e51adc21c9fc1ed4110271c997R4
For now it's ok to leave it out.

If things get too repetitive later we can optimize the types. I believe we need to create some sort of overview and testing first and optimize based on that.

pseudozach: good point on `action`. done.

src/extension/background-script/db.ts
```diff
@@ -43,25 +52,32 @@ class DB extends Dexie {
...
+    const blocklistArray = await this.blocklist.toArray();
     await browser.storage.local.set({
       allowances: allowanceArray,
       payments: paymentsArray,
+      blocklist: blocklistArray,
     });
     return true;
   }
 
+  // TODO: DRY up the duplication here.
```
escapedcat, line 76: For who is this comment? We usually create an issue for TODOs and link it here. Otherwise it's hard to forget about it and might never be done.

bumi: it was me. wasn't sure if it hurts enough to create an issue 🙈

escapedcat: If it doesn't hurt enough we do not need it? :P

escapedcat:
All [the new actions](https://github.com/pseudozach/lightning-browser-extension/tree/blocklist/src/extension/background-script/actions/blocklist) need unit tests like we already [have for accounts](https://github.com/pseudozach/lightning-browser-extension/tree/blocklist/src/extension/background-script/actions/accounts/__tests__). It's up to you if you want to add these with later PRs or not. We can do it as well. We can also do it together. Whatever works for you. 
Let me know what you think, thanks.

src/app/screens/Enable/index.tsx
escapedcat: That's a good idea! Created: https://github.com/getAlby/lightning-browser-extension/issues/1065

dvoroneca:
I've reviewed the design of the button. It's ok for now, however, there are a few things that need to be considered here:
 _How users can manage their list of blocked applications? 
 How to access this list?_

And from the vendor site: 
_Is there enough information about setting up a pop-up in a way that it doesn't load up on every interaction? Do they know how to fix if the **Cancel** button doesn't work as it should?_

escapedcat (approved):

PR #826, opened 2022-04-26
feat: tipping support for mastodon

### LINKED ISSUE #476: Tipping feature for Mastodon instances
Similar to the tipping feature on twitter we should also support tipping on mastodon 
Mastodon supports profile links which should make it easy to get the lightning address from there. 

e.g. https://bitcoinhackers.org/@openoms

This can be similarly implemented as the other "batteries" - like the "monetization" for example.

reneaaron:
#### Link this PR to an issue
Fixes #476 

Tipping support for Mastodon instances. 
 - bitcoinhackers.org
 - kosmos.social

✅ Tipped on bitcoinhackers.com

#### Checklist:


reneaaron:
@bumi I just added tipping support for Mastodon instances. There are a few questions left that I wanted to clarify before submitting the MR for review: 

1) Mastodon instances
I've added bitcoinhackers.org and kosmos.social that were mentioned in the original issue already, are there any other instances you'd want to support? 

2) Where to search for
Currently we only look for lightning addresses / LNURLP in the bio. In the original issue there was also this idea of supporting the original profile email. However I don't know how we want to know weather an email is a lightning address at this point? 🤔

bumi:
re 1: 
we should add mastodon.social
and let's ask on twitter?

re 2. we could also use custom profile metadata: 
image
I'd say the bio and a custom lightning meta data profile entry?

src/extension/content-script/batteries/Mastodon.ts
```diff
@@ -0,0 +1,65 @@
...
+      name:
+        document
+          .querySelector(".public-account-header__tabs__name > h1")
+          ?.childNodes[0]?.nodeValue?.trim() ??
+        document.querySelector(".account__header__tabs__name > h1 > span")
+          ?.textContent ??
+        "",
+      description:
+        document.querySelector(
+          ".public-account-bio, .account__header__bio .account__header__content"
```
bumi, line 47: I think this is to broad. when a user has profile metadata then this is also in the description.

reneaaron:
> re 1: we should add mastodon.social and let's ask on twitter?

Good idea, let's do a poll again! 

> re 2. we could also use custom profile metadata: image I'd say the bio and a custom lightning meta data profile entry?

Sure, let's include that as well. 👍

bumi:
Here is a list of instances: https://joinmastodon.org/communities

bumi:
maybe we can also make a PR to officially add it to Mastodon: https://github.com/mastodon/mastodon ?

bumi:
the twitter poll: https://twitter.com/getAlby/status/1520899495835635712

MoritzKa:
@reneaaron could you send me a screenshot of a profile on mastodon with a Lightning Address? I want to update the designs on getalby.com/value4value

raucao:
> maybe we can also make a PR to officially add it to Mastodon: https://github.com/mastodon/mastodon ?

I think that would generally be preferable, since there are so many instances that you can't really have a whitelist for it.

I was thinking about this in the past, and I think the cleanest way would be to add a new Mastodon user preference for adding a Lightning Address to a user profile, which could then be fetched by any other user and software (Mastodon is just one out of many ActivityPub applications). And the user's client could then implement the payment flow via the Lightning Address and using WebLN, and whatever integration is best for native mobile apps on the various mobile operating systems.

To improve things for users even further, fediverse server providers can then also go ahead and offer sats inboxes for their local users, and pre-configure local Lightning Addresses for them (same as the normal user address). This is what we had planned to do with kosmos.social, once we have the user base there ported to our LDAP server, where we already offer Lightning accounts to all kosmos.org users.

src/extension/content-script/batteries/Mastodon.ts
```diff
@@ -0,0 +1,65 @@
+import getOriginData from "../originData";
+import setLightningData from "../setLightningData";
+
+const urlMatcher =
+  /^https:\/\/(bitcoinhackers\.org|kosmos\.social)\/(web\/)?@\S+/;
```
stevenroose, line 5: Is there no way to generalize this?

bumi: it could check the actual HTML code, but we wanted to avoid this initially. 
Do you think it's OK to do that?

silverpill:
@bumi

>re 2. we could also use custom profile metadata:

There's an existing convention for cryptocurrency donation addresses: metadata fields with labels like $BTC, $ETH etc. I've never seen anything like $LN though

bumi:
> There's an existing convention for cryptocurrency donation addresses: metadata fields with labels like $BTC, $ETH etc. I've never seen anything like $LN though
@silverpill ahh, good to know. should we use $BTC? or $LN?

silverpill:
I don't know, which one do you think Alby users would prefer?
It's just a convention and most fedi servers display such fields as ordinary fields. One exception is [Soapbox](https://gitlab.com/soapbox-pub/soapbox-fe) which displays them as donation widgets. I think it will recongnize both $BTC and $LN but it won't show an icon for $LN because it uses icons from [cryptocurrency-icons](https://github.com/spothq/cryptocurrency-icons/blob/master/manifest.json) package.

reneaaron:
We have removed support for 3rd party integrations in the meantime 😞

PR #827, opened 2022-04-26
feat: explanatory videos for account setup

### LINKED ISSUE #686: Add demo videos to plug and play node connectors
@bumi had the idea to add demo videos to the node connectors like myNode, Umbrel, RaspiBlitz etc.

Here is a mockup:

image

reneaaron:
#### Link this PR to an issue
Fixes #686

This PR allows us to add an optional video for every connector and adds the first video for the Umbrel connector. If none is provided we still have the fallback to the sats symbol (like it was before). 

image

Tested on Firefox.

#### Checklist:

### DIFF
--- a/src/app/components/ConnectorForm/index.tsx
+++ b/src/app/components/ConnectorForm/index.tsx
@@ -11,6 +11,7 @@ type Props = {
   submitDisabled?: boolean;
   onSubmit: FormEventHandler;
   children: React.ReactNode;
+  video?: string;
 };
 
 function ConnectorForm({
@@ -21,6 +22,7 @@ function ConnectorForm({
   submitDisabled = false,
   onSubmit,
   children,
+  video,
 }: Props) {
   const navigate = useNavigate();
 
@@ -42,7 +44,28 @@ function ConnectorForm({
         </div>
         <div className="mt-16 lg:mt-0 lg:w-1/2">
           <div className="lg:flex h-full justify-center items-center">
-            <img src="assets/icons/satsymbol.svg" alt="sat" className="w-64" />
+            {video ? (
+              <div
+                className="flex-1 relative h-0"
+                style={{ paddingBottom: "56.25%" }}
+              >
+                <video
+                  className="absolute t-0 l-0 w-full h-full"
+                  width="1920"
+                  height="1080"
+                  controls
+                >
+                  <source src={video} type="video/mp4" />
+                  Your browser does not support the video tag.
+                </video>
+              </div>
+            ) : (
+              <img
+                src="assets/icons/satsymbol.svg"
+                alt="sat"
+                className="w-64"
+              />
+            )}
           </div>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


src/app/screens/connectors/ConnectUmbrel/index.tsx
```diff
@@ -103,6 +103,7 @@ export default function ConnectUmbrel() {
       submitLoading={loading}
       submitDisabled={formData.url === "" || formData.macaroon === ""}
       onSubmit={handleSubmit}
+      video="https://cdn.getalby-assets.com/Connector-guides/in_extension_guide_umbrel.mp4"
```
bumi, line 106: I always like everything lowercase `[a-z0-9-_]` (this will never cause errors because of the case or some special characters)

src/app/components/ConnectorForm/index.tsx
```diff
@@ -42,7 +44,28 @@ function ConnectorForm({
...
+                style={{ paddingBottom: "56.25%" }}
+              >
+                <video
+                  className="absolute t-0 l-0 w-full h-full"
+                  width="1920"
+                  height="1080"
+                  controls
+                >
+                  <source src={video} type="video/mp4" />
+                  Your browser does not support the video tag.
```
bumi, line 59: nit: I don't think we need the fallback? The browsers should support it anyway, or it's blank? 

but it does not matter also.

src/app/components/ConnectorForm/index.tsx
```diff
@@ -42,7 +44,28 @@ function ConnectorForm({
...
-            <img src="assets/icons/satsymbol.svg" alt="sat" className="w-64" />
+            {video ? (
+              <div
+                className="flex-1 relative h-0"
+                style={{ paddingBottom: "56.25%" }}
+              >
+                <video
+                  className="absolute t-0 l-0 w-full h-full"
+                  width="1920"
+                  height="1080"
```
bumi, line 55: Is it good to have fixed sizes here? what if the user's screen is different?

bumi (commented):
nice, this is great.

src/app/screens/connectors/ConnectUmbrel/index.tsx
reneaaron: @MoritzKa Can you please rename the folder to `connector-guides`?

src/app/components/ConnectorForm/index.tsx
reneaaron: You are right, browsers that are supported by Alby should definitely support that. I'll remove it.

src/app/components/ConnectorForm/index.tsx
reneaaron: The width and height are overridden by CSS rules anyway to maintain the aspect ratio, I think we can safely remove them.

reneaaron: Regarding different screen sizes: there is a responsive solution in this CSS layout so the video always retains its original aspect ratio. (padding-bottom: 56.25% --> 16:9)

bumi (approved):

PR #846, opened 2022-05-03
feat: reorganize settings page

### LINKED ISSUE #844: Improve description for additional userdata being sent via LNURL
image

I just stumbled upon some new settings and was surprised Alby asks me to fill in personal data like name & email to be sent along with LNURL payments. 👀 

 - Why would you send an name / email along with a payment? 
 - Is this info attached to every payment? 

After learning about [LUD-18](https://github.com/fiatjaf/lnurl-rfc/blob/luds/18.md) I understand the use-case, but I think we should let users know that this information is not shared without the explicit consent for every payment and that they are in control of who they share this data with.

reneaaron:
#### Link this PR to an issue
Fixes #844 

#### Checklist:

### DIFF
--- a/src/app/screens/Settings.tsx
+++ b/src/app/screens/Settings.tsx
@@ -77,46 +77,6 @@ function Settings() {
           )}
         </Setting>
 
-        <Setting
-          title="User Display Name"
-          subtitle="For sending along with LNURL payments when supported"
-        >
-          {!loading && (
-            <div className="w-64">
-              <Input
-                placeholder="Enter your username"
-                type="text"
-                value={settings.userName}
-                onChange={(ev) => {
-                  saveSetting({
-                    userName: ev.target.value,
-                  });
-                }}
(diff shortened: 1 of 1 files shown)


src/app/screens/Settings.tsx
```diff
@@ -165,6 +125,48 @@ function Settings() {
           )}
         </Setting>
       </div>
+      <h2 className="mt-12 text-2xl font-bold dark:text-white">
+        Personal data
+      </h2>
+      <div className="mb-6 text-gray-700 dark:text-gray-300 text-sm">
+        Payees can request for additional data to be sent with a payment. This
```
bumi, line 132: reading this again my thought was if "request" is too hard, and if "optional" should be mentioned.
Payees can optionally request additional data to be sent with a payment.

I don't remember what we said so far :)

bumi (approved):

PR #848, opened 2022-05-03
style: make sat buttons text a bit smaller

bumi:
The text does not fit in one line on windows

Problem: 
image

needs to be tested on windows. 

open the LNURL pay screen by sending to "hello@getalby.com"

### DIFF
--- a/src/app/components/SatButtons/index.tsx
+++ b/src/app/components/SatButtons/index.tsx
@@ -7,7 +7,7 @@ type Props = {
 
 function SatButtons({ onClick }: Props) {
   return (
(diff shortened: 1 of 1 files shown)


dylancom:
If we want smaller sat buttons, I think we should make a "small" variation of the button instead.
(Rather than creating side-effects by settings a smaller font-size on the parent)

bumi:
+1 this was more a quick hack, not knowing what I do to give Amy something to make the videos not look that broken :)

dylancom:
Is this fixed now by #850 ?

bumi:
I think so.

PR #853, opened 2022-05-04
Dispatch JS events when webln is ready and a payment for a lightning link succeeded

bumi:


src/extension/inpage-script/index.js
```diff
@@ -61,8 +69,12 @@ if (document) {
...
         .sendPayment(paymentRequest)
         .then((r) => {
-          console.log(r);
-          //alert(JSON.stringify(r));
+          const responseEvent = new CustomEvent("lightning:success", {
+            paymentRequest: paymentRequest,
+            response: r,
+          });
+          document.dispatchEvent(responseEvent);
+          link.dispatchEvent(responseEvent);
```
bumi, line 77: @escapedcat @dylancom is it better to dispatch that event on the link or on the document?

escapedcat: Where is this being used?Would need to understand where this is being used. Looks like currently everything is attached to `window`. What the purpose/usage here?

bumi: if you do a `<a href="lightning:...">..</a>` this event would allow you to get notified when a payment happened. 
`link/document.addEventListener('lightning:success', ...)`

dylancom: You dispatch it on the element, link in this case. And you can catch it on the document. This way you can also listen for which element initiated it (event.target.tagName)

bumi: does that mean it is enough to dispatch it on the link element? (and it still bubbles up to the document?)

dylancom: If you add: `bubbles: true` to the CustomEvent options, you can catch it on document.

dylancom:
If you rebase, the lint error should be gone.

src/extension/inpage-script/index.js
```diff
@@ -55,16 +62,35 @@ if (document) {
...
       if (lnurl) {
-        return window.webln.lnurl(lnurl).catch((e) => {
-          console.log(e);
-          alert(`Error: ${e.message}`);
-        });
+        // TODO: dispatch success event
+        return window.webln
+          .lnurl(lnurl)
+          .catch((e) => {
+            console.log(e);
```
escapedcat, line 69: `console.error(e)`

src/extension/inpage-script/index.js
```diff
@@ -5,6 +5,9 @@ import WebLNProvider from "../ln/webln";
 if (document) {
   window.webln = new WebLNProvider();
 
+  const readyEvent = new Event("webln:ready");
+  document.dispatchEvent(readyEvent);
```
bumi, line 9: @reneaaron what do you think about this, is this helpful?

reneaaron: Personally I never needed such an event before. 🤔 Do you have any specific use-cases in mind where one would need that event? 

Since app devs are not supposed to trigger any payments without a user interacting (I think it would be strange to trigger `webln.enable()` or any other call on page load, right?) I wonder where one would use this event?

bumi: I think the main use case would probably be to know if webln is available. A website wants to check if the user supports webln calls or if a fallback must be used or an error must be shown. 

`window.webln` is loaded on `document_end` - when the DOM is loaded. But checking for it on DOMContentLoaded might be still too early.

I just thought to fire such an event does not hurt.

reneaaron: Okay, so the app would check for WebLN capabilities right on page load to setup payment components or something. I would have never needed that before but I agree that offering this event doesn't hurt. 👍

bumi: I also don't care about this one that much.
my use case was more for the `lightning:success` event. that allows to have a normal lightning: protocol link but still get informed if a payment was done (if a user has alby)

reneaaron: Yeah, that makes sense. That could potentially allow developers to integrate a very low-impact solution for their existing apps. All they'd have to do would be to add `lightning:` links to their markup without going the full WebLN way of doing things. (and still supporting mobile wallets)

I am not too sure about the naming, though. 
 - `lightning` vs. `webln` vs. `webbtc`: Could this be part of the spec some day? I just don't want to confuse people with different names
 - Maybe it would make sense to be a bit more specific (i.e. `lightning:paymentSuccess`) in case there are other events that might become available in the future?  🤔

escapedcat: If you do a call let me know. I think I have quite some things to learn.

reneaaron: I just noticed a bug on https://webln.twentyuno.net where the detection of WebLN is run on `load` and users are redirected to an onboarding-page if they don't have WebLN. 

Sometimes it seems the `window.load` event is triggered before the webln provider is injected into the page. I think an event like `webln:ready` could help in such a case 👍  (you'd still need to add a timout for a redirect since non WebLN browsers won't trigger this event). 

I guess there is no way to inject the WebLN provider before the `load` event is fired, right?

reneaaron:
> so you're OK with this for now?

Yeah, sure. 👍 Let's introduce those events and try to use them to detect WebLN capabilities and move on from there...

PR #869, opened 2022-05-08
feat: use settings for first language locale detection

### LINKED ISSUE #743: feat: load locale from settings stored under extension's storage
currently, we use ```browser_language_detector``` to detect language, store language under local storage and load the locale from the local storage.

https://github.com/getAlby/lightning-browser-extension/pull/711 saves locale in extension storage but we don't load this locale for use.

We can override the normal behaviour of browser_language_detector and load locale which is stored under extension storage. we can create a custom detector for that https://github.com/i18next/i18next-browser-languageDetector/issues/140

1. Create a custom detector, which loads locale from extension's storage (for existing user)
2. fallback to local storage if it's not an existing user or there is no locale present in extension storage (when new user use alby)

```
 detection: {
      // order and from where user language should be detected
      order: ['customDetector', 'querystring', 'cookie', 'localStorage', 'navigator', 'htmlTag']
    }
```

**Additional context**
we have to discuss the cases when custom locale present in extension storage will be used! 

In the current setup.
1. create a new wallet, go under settings, uncheck ```WebEnhancement``` button
2.  remove existing Alby, open the same wallet using lndhub URL, see ```WebEnhancement``` but still checked!

in this case, we are not loading saved settings. clarification is needed in which cases we can load locale stored in extension's storage and how currently saved settings are loaded back

secondl1ght:
#### Link this PR to an issue
Fixes #743 

As requested and described in the related issue, I have setup i18n to use the language selected by Alby users in the Settings screen first and then fallback to the other default methods if the user has not made a selection yet.

I added a translation to the web interface and tested switching languages in the Settings screen, the language persists even after a page refresh.

#### Checklist:

#### Notes:

I did a `git commit --no-verify` to get past the TypeScript errors I was receiving. I tried a few different things to resolve them but was having issues, maybe TypeScript wizard @dylancom could have a look, or @pavanjoshi914?


pavanjoshi914:
on page refresh, the locale will persist cause the I18next browser language detector stores locale itself in local storage. so that is not enough to test this PR
 
image

the best way can be to check cross browser locale persistence of the particular alby account! cause that is the major reason we are doing load and store locale from settings

pavanjoshi914:
but currently, that is not possible! I think we don't provide lndhub URL during account creation anymore

bumi:
cross-browser test would be to use the same google/firefox account on different browsers. Settings are synced to the browser storage. (this is unrelated to lndhub or the connectors)

A reason to store this settings in the "settings" is also to have a unified settings storage. and not have some settings in some local storage and some other settings somewhere else.

secondl1ght:
Hey @dylancom would you be able to have a look at the TypeScript errors here and point me in the right direction to solve them?

src/i18n/i18nConfig.ts
```diff
@@ -3,6 +3,30 @@ import { initReactI18next } from "react-i18next";
 import LanguageDetector from "i18next-browser-languagedetector";
 import en from "./locales/en.json";
 import hi from "./locales/hi.json";
+import api from "~/common/lib/api";
+
+export const settings = {
```
escapedcat, line 8: According to https://github.com/i18next/i18next-browser-languageDetector/blob/master/index.d.ts#L75 `addDetector` is [taking `dectector` (`settings`) as `CustomDetector`](https://github.com/i18next/i18next-browser-languageDetector/blob/master/index.d.ts#L64-L68).

So tried:

```ts
import type CustomDetector from "i18next-browser-languagedetector";

export const settings: CustomDetector = {
  name: "settings",

  cacheUserLanguage(lng, options) {
    // options -> are passed in options
    // lng -> current language, will be called after init and on changeLanguage
    // store it
    api.getSettings().then((setting) => {
      i18n.changeLanguage(setting.locale);
    });
  },

  lookup(options) {
    // options -> are passed in options
    api.getSettings().then((setting) => {
      i18n.changeLanguage(setting.locale);
    });
  },
};
```

Which then gives me:
image

Ugh. Looks like it's not using the right type or not importing the right one?
Ideas @dylancom ?

escapedcat: Looks like `import type { CustomDetector } from "i18next-browser-languagedetector";` works better (bit confused about this tbh). Now `lookup` needs to return `string | string[] | undefined`. 

image

Not sure what lookup really needs to return here but maybe that helps. This would be nice to pair ;)

secondl1ght: Ok sounds good, we can pair on this next week to finish it off. :)

escapedcat: Let's do it! It's important to understand how to tackle these things.

secondl1ght:
Me and @escapedcat paired on this to finish it off, @dylancom can you have one more look? And then I think it is ready to be merged!

bumi (approved):

bumi:
@pavanjoshi914 do you approve?

pavanjoshi914:
> @pavanjoshi914 do you approve?

checking into this

src/i18n/i18nConfig.ts
```diff
@@ -1,10 +1,34 @@
...
 import hi from "./locales/hi.json";
 
+export const settings: CustomDetector = {
+  name: "settings",
+
+  lookup(_options) {
+    api.getSettings().then((setting) => {
+      i18n.changeLanguage(setting.locale);
+    });
+    return undefined;
```
bumi, line 17: in the [example](https://github.com/i18next/i18next-browser-languageDetector#interface) the language is returned. 
what is this `undefined` doing here? 

I can imagine that when it returns `undefined` it actually says that it did not detect a local and another detector will be used. https://github.com/i18next/i18next-browser-languageDetector/blob/386bc3db9b9e941fa0377840924076530ae60c61/src/index.js#L65

and this PR here might work because it is doing async `i18n.changeLanguage` - but then we don't need this detector class, do we?

bumi (changes requested):

src/i18n/i18nConfig.ts
secondl1ght: Me and @escapedcat returned undefined here when we were fixing the TS error, you are right though it makes sense to return the language to match the example. I have updated this to return the language.

src/i18n/i18nConfig.ts
```diff
@@ -1,13 +1,39 @@
...
+export const settings: CustomDetector = {
+  name: "settings",
+
+  lookup(_options) {
+    let settingsLang;
+    api.getSettings().then((setting) => {
+      i18n.changeLanguage(setting.locale);
+      settingsLang = setting.locale;
+    });
+    return settingsLang;
```
bumi, line 22: I am pretty sure this does not work. did you test it?

api.getSettings() is called asynchronously. this means the return here happens before api.getSettings() finished. 
settingsLang can not be available here.

bumi:
are we sure this can work? to me it seems `lookup()` is synchronous. but loading from the settings is async. So I am not sure if this can work?

src/i18n/i18nConfig.ts
secondl1ght: Yes you are right again, I tested this and `settingsLang` is undefined when it gets returned. But the functionality of this PR still works, it was working when we were returning `undefined`... should we say that it is OK because it achieves the goal of this PR which is to allow users to select the language settings via the Alby Settings page? Or should we debug further to try and understand this CustomDetector class better?

bumi: from how I understand it accidentally works because of a side-effect and not because of having this Detector. 

it only works because the lookup calls 
```
api.getSettings().then((setting) => {
  i18n.changeLanguage(setting.locale);
});
```

if that's how we change the language the we do not need to do that in such a detector. 
So I don't think this PR can be merged like this.

Somebody else looking at this code in a few months will also not understand this and will be confused.

secondl1ght: Yes that is how we change the language when the user selects one in the Settings, but the problem is - will that persist on refresh? I thought the reason we needed this custom detector is that i18n will not detect the Alby user settings when they close the browser or on refresh? I can test it without this custom detector and see if it still works though.

bumi: as you said your custom detector does not work. lookup does return `undefined` so the next detector will be used etc. 

it only works because this custom detector executes some code as a side effect. it executes code to load the language form the settings and calls i18n.changeLanguage. 
but the detector is the wrong place to trigger that.

secondl1ght: Ok I think I have removed all the not needed code now, I tested it and it looks like it works - you can select a language in the Alby settings and then refresh and the language you selected will persist.

I guess we over-thought this from the beginning, we didn't need all this fancy custom detector class stuff in there...

Please have a look and let me know what you think now, thanks!

bumi: actually that looks good to me! good job! 

@pavanjoshi914 can you review this and confirm?

escapedcat (approved):

src/i18n/i18nConfig.ts
```diff
@@ -1,13 +1,18 @@
...
 import { initReactI18next } from "react-i18next";
+import api from "~/common/lib/api";
 
 import en_common from "./locales/en/common.json";
 import hi_common from "./locales/en/common.json";
 import en_components from "./locales/en/components.json";
 import en from "./locales/en/translation.json";
 import hi from "./locales/hi/translation.json";
 
+api.getSettings().then((setting) => {
```
escapedcat, line 12: We're worried that this could lead to a race condition. Depending on how long it takes to get the settings this works or might fail and the LanguageDetector wins.  
Mayeb there's a way to combine this to make sure it's save to use?
@pavanjoshi914 what do you think?

secondl1ght:
Hey @bumi 👋🏼, me and @escapedcat have another variation for you to look at :P.

We moved the code into an async function and await so that we can be sure of the order of execution, I tested it locally and it works the same as before. Let me know what you think about this now.

Thanks!

src/i18n/i18nConfig.ts
```diff
@@ -26,18 +22,25 @@ export const resources = {
...
+    // init i18next
+    // for all options read: https://www.i18next.com/overview/configuration-options
+    .init({
+      //debug: true,
+      fallbackLng: "en",
+      resources,
+    });
+
+  const setting = await api.getSettings();
+  i18n.changeLanguage(setting.locale);
```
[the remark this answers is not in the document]
secondl1ght: wouldn't we always want to overwrite what is there to use the Alby settings so that we trigger i18n to use that language?

src/i18n/i18nConfig.ts
```diff
@@ -26,18 +22,25 @@ export const resources = {
...
+};
+
+loadLanguage();
```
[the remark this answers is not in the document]
secondl1ght: I think this code is run every time the page is refreshed or every time a top level component is rendered, you can see all the files it is imported into, there are 10 total.

src/i18n/i18nConfig.ts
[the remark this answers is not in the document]
escapedcat: Talked to bumi. Can we do this in the new `SettingsProvider`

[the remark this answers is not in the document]
bumi: can we do this now in the new `SettingsProvider`?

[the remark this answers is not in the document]
secondl1ght: So you just want to move this code into that file?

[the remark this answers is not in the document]
bumi: not sure how it best works. but the SettingsProvider loads the settings already. So yes, maybe we just move this there and do the `changeLanguage` after the settings are loaded there?  
but maybe you can ask @lisabaut maybe she has an idea how to do this best.

escapedcat:
Closing this in favour of #1265

secondl1ght:
> Closing this in favour of #1265

Thanks for taking this one across the finish line @escapedcat 😄.

PR #876, opened 2022-05-10
fix: support for urls with trailing slashes

reneaaron:
#### Link this PR to an issue
The current urlMatcher requires GitHub URLs to end with a slash which is not the case (anymore?).

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "crypto-browserify": "^3.12.0",
     "crypto-js": "^4.1.1",
     "dayjs": "^1.11.1",
-    "dexie": "^3.2.1",
+    "dexie": "^3.2.2",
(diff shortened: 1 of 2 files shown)


bumi:
huch...hmmm, didn'we talk about this or chat about that... good that you found that quickly!

reneaaron:
@bumi Yeah, I also thought that we discussed that. 

I'm currently verifying with the reporter of the problem if there is another issue. 🤔 Will mark the MR as ready as soon as we found the root cause. 👍

reneaaron:
Nope, works for me. Did you notice any problems yourself? One user reported the problem still persists for him so I'll try to setup a debugging session...

bumi:
nope it worked for me also. no idea what could be the problem.

PR #881, opened 2022-05-10
chore: update webextension-polyfill

dylancom:

Upgrade from 0.8.0 to 0.9.0. This breaks our tests initially, so a small change to the setup was needed.

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -4,5 +4,5 @@
 // learn more: https://github.com/testing-library/jest-dom
 import "@testing-library/jest-dom";
 
-// fix "This script should only be loaded in a browser extension."
+// fix "This script should only be loaded in a browser extension." e.g. https://github.com/mozilla/webextension-polyfill/issues/218
 if (!chrome.runtime.id) chrome.runtime.id = "history-delete";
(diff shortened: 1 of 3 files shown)


jest.setup.ts
```diff
@@ -3,3 +3,6 @@
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 import "@testing-library/jest-dom";
+
+// fix "This script should only be loaded in a browser extension."
```
bumi, line 7:
```suggestion
// fix "This script should only be loaded in a browser extension." e.g. https://github.com/mozilla/webextension-polyfill/issues/218
```

bumi (approved):

escapedcat:
Funny, prepared that here as well :P  
https://github.com/getAlby/lightning-browser-extension/pull/768

bumi:
oh I also did not realize there was an update in the depfu PR. sorry.

dylancom:
I was looking for a way to avoid adding extra tsconfig rules (by adding "@types/chrome").
But should have committed here actually, sorry

PR #885, opened 2022-05-11
Support paths on the LND URL

### LINKED ISSUE #884: LND URL does not accept any path
**Describe the bug**
Currently it is not possible to run the LND REST API under an additional path. 
E.g. BTCPay Server runs LND behind NGINX under the `lnd-rest/btc/` location. 

We are resetting the `pathname` here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/lnd.ts#L239

but we should add the lnd path. 

Note: we have to be careful of double slashes `//` (e.g. user enters the URL with a trailing slash, or without)
 
see also:
https://twitter.com/Bumi/status/1524331302366941185

**To Reproduce**
1. Add a new LND account with location (see tweet linked above)

**Information About Alby**
- Alby Version: 1.10.0
- Wallet connected with Alby: LND

bumi:
#### Link this PR to an issue
Fixes #884

- New feature/bugfix (non-breaking change which adds functionality)

Currently we replace the path - assuming that the REST API runs on the root path.
This change now appends the LND API path

Tried to add a new account which was mentioned on twitter. 
Or: 

* add an LND account with a path
* check the background script console that the HTTP request is made to the correct URL with path: e.g. `/my/path/v1/getInfo`  (for this check you do not need to have an LND running on that path, the request will fail, but you see if the URL was as expected)

### DIFF
--- a/src/extension/background-script/connectors/lnd.ts
+++ b/src/extension/background-script/connectors/lnd.ts
@@ -236,7 +236,7 @@ class Lnd implements Connector {
     defaultValues?: Record<string, unknown>
   ): Promise<Type> {
     const url = new URL(this.config.url);
-    url.pathname = path;
+    url.pathname = `${url.pathname.replace(/\/$/, "")}${path}`;
     let body = null;
     const headers = new Headers();
     headers.append("Accept", "application/json");


escapedcat:
Can you describe a way to test this? Could add a test later on.

escapedcat:
Not sure if it's worth creating a pure unit-test for this. Would rather create an integration test. What do you think? Could be add a test-env with a path?

bumi:
I think both is not so easy to setup. the unit test would need some mocks and the integration test needs some non default lnd server setup.

I would leave it for now 🙈

escapedcat (approved):

PR #901, opened 2022-05-18
refactor: input label should be inline

dylancom:

Currently when hovering above an amount field, will already show stepper arrows because the label's take up all width.
- Removed making the label a block element (keeping it the default inline)


escapedcat (approved):
Hm, I expected Prettier should have taken care of this

PR #906, opened 2022-05-19
feat(i18n) Add i18nextProvider for popup, optimize namespaces, add Unlock screen translations

thebrandonlucas:
#### Link this PR to an issue
Work in Progress: #781

    - Removed namespaces for simplification, since we're only using single file
        per language. Namespace now defaults to "translation"
    - Added "Unlock" screen strings to en.json
    - Added i18nextProvider component in Popup.js so that the extension
        can use translations

Visual testing along with `npm run test-ext` and `npm run test-app`

#### Checklist:


thebrandonlucas:
A few notes/questions on this one:

- It appears the use case for namespaces is if you have [multiple files to switch between](https://react.i18next.com/guides/multiple-translation-files):  I removed the use of namespaces since the top level of each file for a particular translation is considered a namespace. I figured we'd just go with one file per language for now but am happy to change it back if it gets too verbose and we decide to split it apart

- For strings used throughout the app, would we rather duplicate them but have them nested in the location they are used for clarity, or put them all in one spot (i.e. "common"). Similarly, should "actions" such as pressing a button, be it's own separate category or just nested in the location its used (I defaulted to the latter).

- For strings that had to be split up (i.e. because there was an <a> tag or something in between), I used the format "part1", "part2", etc., [as is given in an example by i18n here](https://react.i18next.com/latest/using-with-hooks#translation-files)

- The popup didn't work with i18next until I added a [`i18nextProvider` component to it](https://react.i18next.com/latest/i18nextprovider). The docs say you only need to use this for supporting multiple instances, so @escapedcat  @pavanjoshi914 let me know if you know of better ways to do this.

Thanks!

escapedcat:
> * It appears the use case for namespaces is if you have [multiple files to switch between](https://react.i18next.com/guides/multiple-translation-files):  I removed the use of namespaces since the top level of each file for a particular translation is considered a namespace. I figured we'd just go with one file per language for now but am happy to change it back if it gets too verbose and we decide to split it apart

I'm a bit worried if we stuff everything in one file we sooner or later gonna split it up anyways. I like their `common` example from the link you provided. I would suggest to go with splitting now instead of doing it later (been there done that).  
Thank you to mention this and being so thoughtful!

> * For strings used throughout the app, would we rather duplicate them but have them nested in the location they are used for clarity, or put them all in one spot (i.e. "common"). Similarly, should "actions" such as pressing a button, be it's own separate category or just nested in the location its used (I defaulted to the latter).

See common mention above. Regarding `actions` I'm hesitating. While input-labels can be re-used all over the app actions seem to be for view/context-focused. We could split them for now and if we notice that more general actions exists, we can move them into a `common` space?
 
> * For strings that had to be split up (i.e. because there was an  tag or something in between), I used the format "part1", "part2", etc., [as is given in an example by i18n here](https://react.i18next.com/latest/using-with-hooks#translation-files)

👍 

> * Full pages worked, but the popup didn't work with i18next until I added a [`i18nextProvider` component to it](https://react.i18next.com/latest/i18nextprovider). The docs say you only need to use this for supporting multiple instances, so @escapedcat  @pavanjoshi914 let me know if you know of better ways to do this.

No sure, but it looks good to me. @dylancom might have some feedback to this.

thebrandonlucas:
@escapedcat Sounds good! Added capabilities for multiple different namespace files using `common`. Also added `keyPrefixes` in files where we reuse prefixes a lot to avoid repetition.

escapedcat (approved):

src/app/screens/connectors/NewWallet/index.tsx
```diff
@@ -118,19 +138,25 @@ export default function NewWallet() {
           <div className="mt-6 dark:text-white">
             <p>
               <strong>
-                Your Alby account is ready. <br />
+                <br />
```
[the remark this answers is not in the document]
thebrandonlucas: Oops! Thanks for the catch! Fixed.

escapedcat (approved):

secondl1ght (commented):
I am no i18n expert, but this looks good to me!

escapedcat:
@thebrandonlucas maybe don't further add things to this PR but create "draft"-PRs based on this to extend step by step. Otherwise thid will get bigger and bigger. Let's get this merged first.

Thanls for your efforts!

thebrandonlucas:
> @thebrandonlucas maybe don't further add things to this PR but create "draft"-PRs based on this to extend step by step. Otherwise thid will get bigger and bigger. Let's get this merged first.
> 
> Thanks for your efforts!

Good call, will do

src/i18n/locales/hi/common.json
```diff
@@ -0,0 +1,3 @@
+{
+  "password": ""
```
bumi, line 2: let's copy the english version to make sure no text is missing but in the worst case (when a translation is missing we show the english text)

bumi (commented):
we should also document conventions in the wiki: https://github.com/getAlby/lightning-browser-extension/wiki/Internationalization-i18n-(Developer-Guide)

src/app/router/Welcome/Welcome.tsx
```diff
@@ -2,6 +2,7 @@ import { useEffect, useState } from "react";
 import { HashRouter as Router, useRoutes, useLocation } from "react-router-dom";
 import { useTranslation } from "react-i18next";
 import i18n from "~/i18n/i18nConfig";
+import {} from "~/i18n/namespaces";
```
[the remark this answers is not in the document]
thebrandonlucas: Looking into this, I think it’s better to delete so that default will beescapedValue: true. The docs default to using it because [React escapes HTML by default](https://stackoverflow.com/questions/33644499/what-does-it-mean-when-they-say-react-is-xss-protected). However, [as noted here](https://stackoverflow.com/questions/33644499/what-does-it-mean-when-they-say-react-is-xss-protected#comment92875465_33644981), there are other ways to inject malicious code that React does not protect against. Since we’re using strings in a variety of ways and since it doesn’t change the behavior I don’t see a reason to not escape by default, unless there’s some performance difference or something I’m missing.

src/i18n/i18nConfig.ts
```diff
@@ -21,6 +30,9 @@ i18n
     //debug: true,
     fallbackLng: "en",
     resources,
+    interpolation: {
+      escapeValue: false,
```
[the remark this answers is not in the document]
thebrandonlucas: Looking into this, I think it’s better to delete so that default will be `escapedValue: true`. The docs default to using it because [React escapes HTML by default](https://stackoverflow.com/questions/33644499/what-does-it-mean-when-they-say-react-is-xss-protected). However, [as noted here](https://stackoverflow.com/questions/33644499/what-does-it-mean-when-they-say-react-is-xss-protected#comment92875465_33644981), there are other ways to inject malicious code that React does not protect against. Since we’re using strings in a variety of ways and since it doesn’t change the behavior I don’t see a reason to not escape by default, unless there’s some performance difference or something I’m missing. Thanks for calling this out!

src/i18n/locales/hi/common.json
thebrandonlucas: Good idea, done!

thebrandonlucas:
> we should also document conventions in the wiki: https://github.com/getAlby/lightning-browser-extension/wiki/Internationalization-i18n-(Developer-Guide)

Done!

src/i18n/i18nConfig.ts
```diff
@@ -30,9 +30,6 @@ i18n
     //debug: true,
     fallbackLng: "en",
     resources,
-    interpolation: {
```
escapedcat: We don't want this anymore?  
Here it looks like we might want this? https://react.i18next.com/legacy-v9/step-by-step-guide#set-the-global-wait-option

escapedcat: Or is this just legacy?

escapedcat: Hm, I guess we don't need to set this anymore.

thebrandonlucas: The docs are a bit confusing for this library admittedly, I'm guessing they suggest turning it off for some sort of optimization purpose and because it's "not needed" for React, but I see no reason not to if there's any chance it can make things safer/makes the code cleaner and if there's no downside. That said I can add it back if there's something I'm missing!

escapedcat: I saw your comment and the research after I wrote this. All good. Thanks a lot for your efforts!

thebrandonlucas: Happy to 😄

escapedcat:
Checked the branch out and had a look. lgtm!

escapedcat:
@thebrandonlucas thanks for your patience! One more review and we're good to go.  
Maybe wait with a rebase till that is done.

Would you have time to look into this?: #964  
If not, no worries. You already did a lot!

thebrandonlucas:
> @thebrandonlucas thanks for your patience! One more review and we're good to go. Maybe wait with a rebase till that is done.
> 
> Would you have time to look into this?: #964 If not, no worries. You already did a lot!

@escapedcat Absolutely! I can look at that today :) Should I also look into the merge conflict?

escapedcat:
> Should I also look into the merge conflict?

Whatever you prefer. I'm just worried that there might be conflicts again. Might be better to first get the last approval, then get rid of conflicts one (last) time and merge this. Up to you :)

thebrandonlucas:
> Whatever you prefer. I'm just worried that there might be conflicts again. Might be better to first get the last approval, then get rid of conflicts one (last) time and merge this. Up to you :)

Sounds good! I'll do that

dylancom:
The provider is not needed, it just worked in this case because our i18n initialization code was never called otherwise. (as i18n.t isn't used directly in the popup)
Adding `import "~/i18n/i18nConfig";` to Popup/index.tsx makes sure it does get initialized. (which should also be added to the other router's root file)

dylancom:
I think we need general translations for very common words such as "done", "next", "succes" etc.

src/i18n/locales/en/translation.json
```diff
@@ -0,0 +1,144 @@
+{
+  "welcome": {
```
escapedcat, line 2: Eventually we will need to split this file up otherwise it will grow to big I'm worried. But no need to do it now.

escapedcat (approved):

src/i18n/locales/en/translation.json
thebrandonlucas: any thoughts on how to determine where to make the split? Should it be done by top-level section? (i.e. "welcome", "choose_connector", etc.)

escapedcat: That could work. We don't need to decide now. Let's see how this grows and split after.

PR #907, opened 2022-05-19
chore: update browser targets

escapedcat:
Update browser targets to be more flexible


bumi (approved):

escapedcat:
Looks like this is an issue. Will revert.

PR #915, opened 2022-05-20
fix: publish correct payment notification for keysend payments

bumi:
Before we did not store the keysend payments properly

Currently keysend payments did not have all the required information and have not been properly persisted and shown in the transaction list.

### DIFF
--- a/src/extension/background-script/events/persistPayments.js
+++ b/src/extension/background-script/events/persistPayments.js
@@ -4,7 +4,6 @@ const persistSuccessfullPayment = async (message, data) => {
   const recipientName = data.origin.name;
   const host = data.origin.host;
   const paymentResponse = data.response;
-  const paymentRequestDetails = data.paymentRequestDetails;
   const route = paymentResponse.data.route;
   const { total_amt, total_fees } = route;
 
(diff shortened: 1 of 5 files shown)


src/extension/background-script/events/persistPayments.js
```diff
@@ -4,18 +4,17 @@ const persistSuccessfullPayment = async (message, data) => {
   const recipientName = data.origin.name;
```
escapedcat, line 4: Can this be switched to TS without bigger issues?

bumi: good point. 
but I just tried it and it is a bit of a bigger thing because the values are currently all `any` and have to be defined. 
wanna pair on this next week?

escapedcat: Yay pairing! Yes, let's!

bumi: I merge this fix now, and we pair on TS for a new PR next week? ok?

escapedcat (approved):

src/extension/background-script/events/persistPayments.js
escapedcat: Yup

PR #924, opened 2022-05-21
chore: add check to not run tests against the production user API

bumi:
This tries to add a simple check to not accidentally run the end to end tests against the production app.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -7,6 +7,10 @@ import { devices } from "@playwright/test";
  */
 // require('dotenv').config();
 
+if (process.env.WALLET_CREATE_URL === "https://getalby.com/api/users") {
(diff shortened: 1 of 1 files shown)


escapedcat (approved):

playwright.config.ts
```diff
@@ -7,6 +7,10 @@ import { devices } from "@playwright/test";
  */
 // require('dotenv').config();
 
+if (process.env.WALLET_CREATE_URL === "https://getalby.com/api/users") {
+  console.log("----\nDo not run tests against production");
```
escapedcat, line 11: nit: `console.error`  
But we can change this next time we tocuh this.

bumi:
nice one! emojis FTW! :) thanks

PR #927, opened 2022-05-22
fix: remaining budget display

krsjoseph:
#### Link this PR to an issue
Fixes #ISSUE-NUMBER

@bumi from our pairing. I can't seem to find the PR number/

Fix displaying and calculating remaining budget for Allowances

#### Checklist:

### DIFF
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -1,7 +1,9 @@
 import { GearIcon } from "@bitcoin-design/bitcoin-icons-react/filled";
 import { CrossIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
 import { useState, useEffect } from "react";
+import type { FormEvent } from "react";
 import Modal from "react-modal";
+import { toast } from "react-toastify";
 import { useSettings } from "~/app/context/SettingsContext";
 import utils from "~/common/lib/utils";
 import { getFiatValue } from "~/common/utils/currencyConvert";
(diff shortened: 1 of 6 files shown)


src/extension/background-script/actions/allowances/update.js
```diff
@@ -2,11 +2,16 @@ import db from "../../db";
```
escapedcat, line 2: You wanna try to switch this file to TS? If there's an issue we could pair on that.

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -40,6 +41,13 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
   }
 
   async function updateAllowance() {
+    if (parseInt(budget) < allowance.remainingBudget) {
```
escapedcat, line 44: Would be nice if we could add a Jest test for this.

escapedcat (commented):
🎉 nice!

src/extension/background-script/actions/allowances/update.js
escapedcat: Will be done here:
https://github.com/getAlby/lightning-browser-extension/pull/1046

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -39,10 +42,18 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
   }
 
   async function updateAllowance() {
+    if (parseInt(budget) > allowance.remainingBudget) {
```
escapedcat, line 45: This is an important part. It was `<` before I changed it. I believe this is correct.

bumi: what is this check here? if the user tries to set a budget that is smaller than the remaining budget then we show an error? 

then it should be `<` ?

escapedcat:
Should wait for https://github.com/getAlby/lightning-browser-extension/pull/1046

escapedcat:
This is now:
- improving tests
- improving UX by using a form for the modal (send "on enter" works)
- improving type

bumi (approved):
utACK

PR #928, opened 2022-05-23
Chore/deps 20220523

escapedcat:
deps related to
- unit tests
- e2e tests
- storybook


escapedcat:
Depfu would soon create a PR with these but I wanted to get these out of the way.

bumi (approved):

PR #929, opened 2022-05-23
e2e Tests for LNBits and BlueWallet

escapedcat:
re-adding commented out tests

### DIFF
--- a/tests/e2e/createWallets.spec.ts
+++ b/tests/e2e/createWallets.spec.ts
@@ -5,37 +5,57 @@ const { getByText, getByLabelText } = queries;
 
 import { loadExtension } from "./helpers/loadExtension";
 
+const commonCreateWalletUserCreate = async () => {
+  const user = USER.SINGLE();
+  const { page: welcomePage, browser } = await loadExtension();
+
+  // go through welcome page
+  const $document = await getDocument(welcomePage);
+
+  // go through welcome page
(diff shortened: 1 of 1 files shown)


tests/e2e/createWallets.spec.ts
```diff
@@ -66,103 +66,107 @@ test.describe("Create or connect wallets", () => {
...
+      getByText($document, "Do you have a lightning wallet?")
+    );
 
-  //   // click at "LNDHub (BlueWallet)"
-  //   const createNewWalletButton = await getByText(
-  //     $document,
-  //     "LNDHub (Bluewallet)"
-  //   );
-  //   createNewWalletButton.click();
+    // click at "LNDHub (BlueWallet)"
```
kiwiidb, line 150: What if Bluewallet is down again and this causes the tests to fail?

escapedcat: Then we could create an issue and need to skip this test till it's fixed.

bumi:
Does it make sense to put the first `Protect your wallet` steps into a helper function? this looks repetitive.

escapedcat:
> Does it make sense to put the first `Protect your wallet` steps into a helper function? this looks repetitive.

Good point. Added them into the same file for now:
- `commonCreateWalletUserCreate`
- `commonCreateWalletSuccessCheck`

bumi (approved):

PR #931, opened 2022-05-23
feat: add function to abort polling

dylancom:

The receive screen makes API calls (maxAttempts: 20) to check if an invoice was paid. Currently there is no mechanism to abort those API calls when the receive screen gets closed (unmounted). This PR adds a shouldStopPolling option to the current poll implementation.


bumi (approved):
some magic :) 

utACK

PR #942, opened 2022-05-25
refactor(account): message-types #652

### LINKED ISSUE #652: TS: Clean up and generalize types for messages (focus: account)

Currently lot's of `background-script` code isn't TS yet.  
When some files were switched to TS `Message` (and `Sender`) types weren't nicely defined.  
Message-types should be defined and re-used. The app should be able to know what to expect.

- https://github.com/getAlby/lightning-browser-extension/pull/636/files#diff-0029b68659877433cec643410ec71d719a7cf163141ca11e7b0ced59716530caR6-R12
   i.e. a MessageType like this can be defined here and imported in places where these being send
- https://github.com/getAlby/lightning-browser-extension/pull/636/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aL51
  `Message`-arg can be a defined AccountType in the future with i.e. the type from above

**Describe alternatives you've considered**

Happy for alternative approaches.

**Additional context**

- Start with [existing account TS-files](https://github.com/getAlby/lightning-browser-extension/tree/master/src/extension/background-script/actions/accounts) first to close this issue
- Further Message-types can follow when switching more files to TS

**Are you working on this?**

Yes, will be finished with several smaller PRs

escapedcat:
#### Link this PR to an issue
Fixes #652

#### Type of change

- refactor

#### Describe the changes you have made in this PR

Aligned `Message` types for `Account`-messages

Tested if all actions still work manually

#### Checklist:


src/extension/background-script/actions/accounts/remove.js
```diff
@@ -1,15 +0,0 @@
-import state from "../../state";
```
dylancom, line 1: This is used in https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Onboard/TestConnection/index.tsx#L23

dylancom (changes requested):

src/extension/background-script/actions/accounts/remove.js
escapedcat: Arg, thanks. Will re-add and check.

escapedcat: Right, re-added it. But this does the same thing, no?:  
https://github.com/getAlby/lightning-browser-extension/blob/refactor/652_account-types/src/extension/background-script/actions/accounts/delete.ts  

I assume we could just use `delete` instead of `remove` or am I missing something here?

escapedcat: Alright, will adjust in another PR later on.

src/extension/background-script/actions/accounts/add.ts
```diff
@@ -31,7 +17,10 @@ const add = async (
 
   const accountId = uuidv4();
   newAccount.config = encryptData(newAccount.config, password);
-  accounts[accountId] = newAccount;
+  accounts[accountId] = {
+    id: accountId,
+    ...newAccount,
```
dylancom, line 22: potentially ...newAccount could overwrite id. Maybe switch the order?

src/extension/background-script/actions/accounts/add.ts
```diff
@@ -31,7 +17,10 @@ const add = async (
 
   const accountId = uuidv4();
   newAccount.config = encryptData(newAccount.config, password);
-  accounts[accountId] = newAccount;
+  accounts[accountId] = {
```
dylancom, line 20: Are we allowed to mutate state directly, or should we use a new accounts object?

escapedcat: Just used it like it were. Updated now.

src/extension/background-script/actions/accounts/add.ts
escapedcat: Good point. According to TS that can not happen, but it's good to start with the spread anyways, thanks!

PR #948, opened 2022-05-25
feat: update alerts to use new toast component

### LINKED ISSUE #934: replace `alert()` with toastify alerts
We recently introduced react-toastify for nicer alerts: https://github.com/getAlby/lightning-browser-extension/pull/852

we should use this in all frontends instead of the browser native `alert`

secondl1ght:
#### Link this PR to an issue
Fixes #934 

I have updated all instances of `alert()` and changed it to the corresponding `toast()`.

I did not test every single toast call but I did test to make sure toast works.

#### Checklist:

### DIFF
--- a/src/app/components/DevMenu/index.tsx
+++ b/src/app/components/DevMenu/index.tsx
@@ -1,7 +1,7 @@
 import { Component } from "react";
+import { toast } from "react-toastify";
 
 import utils from "~/common/lib/utils";
-import { toast } from "react-toastify";
 
 class DevMenu extends Component {
   reset() {
--- a/src/app/components/DevMenu/index.tsx
+++ b/src/app/components/DevMenu/index.tsx
@@ -1,17 +1,18 @@
 import { Component } from "react";
 
 import utils from "~/common/lib/utils";
+import { toast } from "react-toastify";
 
 class DevMenu extends Component {
   reset() {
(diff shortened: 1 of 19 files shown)


escapedcat (approved):

bumi (changes requested):
Does the toast messages work in the storybook stories? 
I think there is no toastcontainer, so maybe we need to leave it there to `alert`. the alerts there are only example implementations anyway

secondl1ght:
Sounds good, maybe we could add a toastcontainer to the stories but because it is just examples like you said then let's just revert back to `alert`. I am not sure how many people use the Storybook stories anyway.

bumi (commented):
I think it is a good practice to have the npm imports above the local imports. 
This is how it currently is in the code. 

Can you move the `import { toast } from "react-toastify";` above the local imports?

bumi (approved):

escapedcat:
@secondl1ght @bumi @dylancom what do you think about using a plugin to align all our import sortings? i.e. https://github.com/trivago/prettier-plugin-sort-imports

bumi:
I don't have a strong opinion there. 
less dependencies is better, but if it's helpful :)

secondl1ght:
@escapedcat I like prettier, so I'd be up for it.

PR #911, opened 2022-05-20
chore: reduce browser targets

escapedcat:
https://browserslist.dev/?q=bGFzdCA1IENocm9tZSB2ZXJzaW9ucywgbGFzdCA1IEVkZ2UgdmVyc2lvbnMsIGxhc3QgNSBPcGVyYSB2ZXJzaW9ucywgbGFzdCA1IEZpcmVmb3ggdmVyc2lvbnM%3D

Friend noticed this should be further reduced

PR #953, opened 2022-05-26
fix: revert browser targets

bumi:
for some reason this causes an issue and the extension does not work.
error: Could not establish connection. Receiving end does not exist.

this was introduced in: https://github.com/getAlby/lightning-browser-extension/pull/911/

image
image

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -7,7 +7,12 @@
         "useBuiltIns": false,
         // Do not transform modules to CJS
         "modules": false,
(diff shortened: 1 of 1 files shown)


bumi:
confirmed by @kiwiidb merging to unblock

escapedcat:
Bad Babel, bad bad Babel! Thanks, will have a look. Weird that everything worked when I build and tried. What browser/version did you try?

PR #963, opened 2022-05-28
fix: add descriptive message when method is not supported

### LINKED ISSUE #961: Correct error message when keysend is not supported
**Describe the bug**
Currently BlueWallet's LndHub does not support keysend payments. 
Our lndhub connector still tries to send a keysend API call which fails with a 404 but we do not give the user a good error message.

We should tell the user that this method is not supported by the connected account. 

**To Reproduce**
1. Connect a BlueWallet lndhub
2. Try to do a keysend payment

**Expected behavior**
Indicate that the connected account does not support the requested method.

dylancom:
#### Link this PR to an issue
Fixes #961

Shows: "keysend not supported by the connected account." when not available. Instead of 404.

### DIFF
--- a/src/extension/background-script/connectors/lndhub.ts
+++ b/src/extension/background-script/connectors/lndhub.ts
@@ -352,6 +352,10 @@ export default class LndHub implements Connector {
 
       if (axios.isAxiosError(e)) {
         const errResponse = e.response as AxiosResponse;
+        if (errResponse?.status === 404) {
+          const method = path.replace("/", "");
+          throw new Error(`${method} not supported by the connected account.`);
+        }
         const errorMessage = `${errResponse?.data.message}\n(${e.message})`;
         throw new Error(errorMessage);
       }


src/extension/background-script/connectors/lndhub.ts
```diff
@@ -352,6 +352,10 @@ export default class LndHub implements Connector {
 
       if (axios.isAxiosError(e)) {
         const errResponse = e.response as AxiosResponse;
+        if (errResponse?.status === 404) {
+          const method = path.replace("/", "");
+          throw new Error(`${method} not supported by the connected account.`);
```
escapedcat, line 357: Will the user be informed by this as well?

dylancom: Yes, the user will see this in the new Toast

escapedcat (approved):

bumi (approved):
utACK. looks good!

Issue #956, opened 2022-05-27
Review lnurl-auth compatibility with Zeus, Blixxt,...


bumi:
**Describe the bug**
LNURL-auth with LND does not generate the same key as Zeus.

**To Reproduce**
0a. Connect a LND node
1a. Visit stacker.news
2a. login with lightning

0b. Use Zeus and connect the same LND
1b. Visit stacker.news
2b. login with lightning

Those should log you into the same account.

**Expected behavior**
The lnurl-auth should create the same keys.

**Information About Alby**
- Alby Version: 1.10.0
- Alby installed through: Chrome store
- Wallet connected with Alby: LND

bumi:
There seems to be in issue with the hmac generation:

we use crypto.js hmacSHA256
zeus uses HMAC from [fast-sha256](https://github.com/dchest/fast-sha256-js)


```js
const hashingKey = '...'
//Alby
const AlinkingKeyPriv = hmacSHA256("getalby.com", hashingKey).toString(Hex);
console.log({ AlinkingKeyPriv });

//Zeus
const ZlinkingKeyPriv = new sha256HMAC(Buffer.from(hashingKey, "hex"))
  .update(stringToUint8Array("getalby.com"))
  .digest();
console.log({ ZlinkingKeyPriv });
```

PR #968, opened 2022-05-29
fix: correct lnurl-auth key generation

bumi:
We did not use the correct LNURL-auth linking key which caused Alby to generate incompatible lnurl-auth logins.
This change fixes this but also allows users to enable the old method to still be able to login to their accounts.

### Link this PR to an issue

issue: #956 

### How has this been tested?

Use the same account in Zeus and Alby.
Using LNURL-auth from both should login to the same account.

### Checklist


src/app/screens/Settings.tsx
```diff
@@ -77,6 +78,22 @@ function Settings() {
...
+        <Setting
+          title="Legacy LNURL-Auth"
+          subtitle="The key generation for LNURL-auth has changed. Due to an error Alby was not compatible with other implementations. This was fixed, but now different keys are used. If you used LNURL-auth to login before you can still enable the old method. This option will be removed later, make sure to switch to the new login."
+        >
+          {!loading && (
+            <Toggle
+              checked={settings.legacyLnurlAuthKey}
+              onChange={() => {
+                saveSetting({
+                  legacyLnurlAuthKey: !settings.legacyLnurlAuthKey,
```
escapedcat, line 90: This is just a boolean , right? `legacyLnurlAuthKey` could sound like it's a key. Maybe `isUsingLegacyLnurlAuthKey` could make this more clear?

escapedcat:
Do users know when they need to use this? How will we communicate this?

kiwiidb:
This is getting quite strange. I am again getting the "receiving end does not exist" error on this PR 😱

src/app/screens/Settings.tsx
bumi: the other one `legacyLnurlAuth` is also wrong, because that actually has nothing to do with lnurl auth :) 

yep, isUsingLegacyLnurlAuthKey sounds good. 

though I hope that we soon remove this

bumi:
> Do users know when they need to use this? How will we communicate this?

we have no good way to do that. maybe we can use an `onUpdateAvailable` listener to show a message for the release that includes this? 
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onUpdateAvailable

I'd try to do this soon to prevent new users from having this problem.

kiwiidb:
Managed to get it to work and verify that it works as advertised on stacker news.
It's a bit sad that we have to do this for the 2nd time but it should be soon, yes.

kiwiidb (approved):
Merge soon, give users the chance to migrate, and hopefully remove soon as well.

dylancom (approved):

escapedcat (approved):

kiwiidb (approved):
(Sorry, wrong PR.)

src/extension/background-script/migrations/index.ts
```diff
@@ -0,0 +1,62 @@
...
+      state.setState({
+        settings: newSettings,
+      });
+      // state is saved with the setMigrated call
+    }
+  },
+};
+
+const migrate = async () => {
+  // Object.keys(migrations).forEach((name: string) => {
```
escapedcat, line 52: What's this?

escapedcat (commented):
Pretty cool! The index-db migration doesn't help with this?

src/extension/background-script/state.ts
```diff
@@ -90,7 +96,7 @@ const state = createState<State>((set, get) => ({
   lock: async () => {
     const connector = get().connector;
     if (connector) {
-      connector.unload();
+      await connector.unload();
```
bumi, line 99: unload is async

this should not be in this PR 🙈

bumi:
> Pretty cool! The index-db migration doesn't help with this?

As this does not touch the DB and we do not change the DB version I don't think so.
This does not touch dexie

src/extension/background-script/migrations/index.ts
escapedcat: Can it be removed then?

src/extension/background-script/state.ts
```diff
@@ -13,6 +13,7 @@ import type Connector from "./connectors/connector.interface";
 interface State {
   account: Account | null;
   accounts: Accounts;
+  migrations: Array<string> | null;
```
escapedcat, line 16: So far we use `string[]` for these, no?

src/extension/background-script/state.ts
```diff
@@ -59,6 +64,7 @@ const state = createState<State>((set, get) => ({
   connector: null,
   account: null,
   settings: DEFAULT_SETTINGS,
+  migrations: [],
```
escapedcat, line 67: Could be an array of possible "migrations"? `typeof`... migration-constants?

src/extension/background-script/state.ts
escapedcat: I think these minor improvements are ok here

bumi: not necessarily minor... one word with potential big implications.

src/extension/background-script/state.ts
bumi: how do I do that?

escapedcat (approved):

src/extension/background-script/migrations/__tests__/legacylnurlauth.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      totalBudget: 500,
+      createdAt: "123456",
+      payments: [],
+      paymentsAmount: 0,
+      paymentsCount: 0,
+      percentage: "0",
+      usedBudget: 0,
+      tag: "",
+    },
+  ];
```
lisabaut, line 33: Hard to understand which data was altered for the test.

Why not define one basic object at the top of file like:

```TS
const allowance: Allowance = {
  enabled: true,
  host: "numbergoup.ngu",
  id: 1,
  imageURL: "https://numbergoup.ngu/favicon.ico",
  lastPaymentAt: "0",
  lnurlAuth: true,
  name: "ngu",
  remainingBudget: 500,
  totalBudget: 500,
  createdAt: "123456",
  payments: [],
  paymentsAmount: 0,
  paymentsCount: 0,
  percentage: "0",
  usedBudget: 0,
  tag: "",
};
```

And then for each test just spread in the data you need for this specific test run:

```TS
  const mockAllowances = [
    {
      ...allowances,
      lnurlAuth: false,
    },
  ];
```

... etc

src/extension/background-script/migrations/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const migrate = async () => {
+  // going forward we can iterate through the the migrations object above and DRY this up:
+  // Object.keys(migrations).forEach((name: string) => {
+  if (shouldMigrate("migrateisUsingLegacyLnurlAuthKeySetting")) {
+    console.info(
+      "Running migration for: migrateisUsingLegacyLnurlAuthKeySetting"
+    );
+    await migrations["migrateisUsingLegacyLnurlAuthKeySetting"]();
+    await setMigrated("migrateisUsingLegacyLnurlAuthKeySetting");
```
lisabaut, line 61: Many times the same-very-long-string. 

You could save it at top of the file as `const IS_USING_LEGACY = "migrateisUsingLegacyLnurlAuthKeySetting"`

and then use this const everyhwere:

```TS

const migrations = {
  [IS_USING_LEGACY]: async () => {
    const { settings } = state.getState();

 // .... 

  if (shouldMigrate(IS_USING_LEGACY)) {
    console.info(`Running migration for: ${IS_USING_LEGACY}`);
    await migrations[IS_USING_LEGACY]();
    await setMigrated(IS_USING_LEGACY);
  }
```

lisabaut (approved):
Thanks, see some minor recommendations below

src/extension/background-script/migrations/__tests__/legacylnurlauth.test.ts
bumi: +1 thought about that, too. but thought we need general fixtures for that? (same is used in other places?)
how do we best load fixtures?

src/extension/background-script/migrations/index.ts
bumi: I guess then I can also just use a shorter name? :)

src/extension/background-script/migrations/__tests__/legacylnurlauth.test.ts
```diff
@@ -0,0 +1,114 @@
...
+import type { Allowance } from "~/types";
+
+import migrate from "../index";
+
+describe("With lnurlAuth enabled allowance", () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+    state.getState().reset();
+    db.allowances.toCollection().delete();
+  });
```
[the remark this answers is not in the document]
bumi: so outside of the describe block? is this then still only executed for that file or how does that work?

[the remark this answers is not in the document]
lisabaut: Yes, scoped for the file. If every test needs the same forEach just put it at the top of the file
see also => https://jestjs.io/docs/setup-teardown

src/extension/background-script/migrations/__tests__/legacylnurlauth.test.ts
lisabaut: There are several ways.
Why not add a `fixtures` folder next to the `types` folder?
And then `import {allowanceFixture } from 'scr/fixtures/allowances'` for example

escapedcat: Good point. In the beginning we didn't have a lot of tests but now the amount of "fixtures" is growing and we could define them once and reuse them. I can help with that.

escapedcat:
Rebased and added i18n

src/extension/background-script/migrations/__tests__/legacylnurlauth.test.ts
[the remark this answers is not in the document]
bumi: thanks!

src/extension/background-script/migrations/index.ts
escapedcat: Currently it's ok to handle because it only accepts migration types. But we could have a look again once more migrations are needed.

Issue #964, opened 2022-05-28
Strings doesn't change to new language which are translated using I18n.t()


pavanjoshi914:
**Describe the bug**

Currently, strings that are translated using the I18n.t() method don't get translated to a new language set using a language switcher.

Reason::
https://github.com/getAlby/lightning-browser-extension/pull/883/files/503db64662e9aae93eb40515e6c87ebf576f8e27#
This will create a different instance of i18n as compared to the instance used by the locale switcher from react-i18next.
now if you change the language using the language switcher translated version of strings will not get reflected which are solved using i18n.t().

**To Reproduce**
Steps to reproduce the behaviour:
1.  use an alternate language translation for the string which is translated using I18n.t'
2. change language using the language switcher
3. See no change in the translation of string

**Expected behaviour**
The string must get translated after the language is changed via the language switcher

escapedcat:
Thanks! Pinged thebrandonlucas. Maybe he finds time to have a look otherwise we'll have a look.

thebrandonlucas:
Thanks for pointing this out @pavanjoshi914! Looking into this

thebrandonlucas:
This should be fixed now! #969 @pavanjoshi914 @escapedcat let me know if it works for you :)

escapedcat:
Closing this :)  
Thanks @thebrandonlucas !

PR #969, opened 2022-05-29
fix: i18n strings not updating

thebrandonlucas:

Fixed bug in #964. Added an "onLanguageChange" listener for strings that were declared outside React component. I had to abstract these into functions so they could be called again when the listener detected a change (otherwise the strings passed to the component would never change on an update).

[Here is where I found the solution](https://github.com/i18next/react-i18next/issues/835#issuecomment-486210116), which is from the creator of Reacti18next.

### Link this PR to an issue

#964 

### How has this been tested?

Visual testing to ensure strings are translating properly.

Also ran E2E and unit tests. 

2 E2E failed due to timeout issues, although I don't think these are due to my changes.


src/app/router/Welcome/Welcome.tsx
```diff
@@ -71,6 +85,25 @@ function App() {
...
+  i18n.on("languageChanged", () => {
+    connectorRoutes = getConnectorRoutes();
+    routes = getRoutes(connectorRoutes);
+
+    // Update name to new language only, don't update status
+    const tempSteps: Step[] = [];
+    for (let i = 0; i < routes.length; i++) {
+      tempSteps.push({
+        id: routes[i].name,
+        status: steps[i].status,
```
thebrandonlucas, line 98: I'm open if anyone has ideas on how to simplify this, but I think either way we'd have to do something like this to update the strings on language change

src/app/router/Welcome/Welcome.tsx
```diff
@@ -71,6 +85,25 @@ function App() {
...
+      });
+    }
+    setSteps(tempSteps);
+
```
thebrandonlucas, line 102: I'm open if anyone has ideas on how to simplify this, but I think either way we'd have to do something like this to update the strings on language change

escapedcat:
Checked it out and tried. Looks good.

> 2 E2E failed due to timeout issues, although I don't think these are due to my changes.

They can be flakey. If they are passing in CI it's all good.

src/app/router/Welcome/Welcome.tsx
```diff
@@ -71,6 +85,25 @@ function App() {
...
   const routesElement = useRoutes(routes);
 
+  const [languageChanged, setLanguageChanged] = useState(false);
+  i18n.on("languageChanged", () => {
+    connectorRoutes = getConnectorRoutes();
+    routes = getRoutes(connectorRoutes);
+
+    // Update name to new language only, don't update status
+    const tempSteps: Step[] = [];
+    for (let i = 0; i < routes.length; i++) {
```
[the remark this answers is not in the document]
thebrandonlucas: Good point, Done!

bumi:
can you fix the conflicts? 
and a check from @dylancom would be good.

thebrandonlucas:
@bumi Merge conflicts fixed!

@dylancom this should be ready to check now

escapedcat (approved):

bumi:
utACK
I'll do a full detailed test in master. 
great work on all this!

PR #971, opened 2022-05-30
Revert "fix: revert browser targets"

escapedcat:


escapedcat:
🎶 One More Time 🎶 🎶   
Let's revert the revert of the revert 😁

kiwiidb:
Can be merged, confirmed that the problem was on my side.

escapedcat:
> Can be merged, confirmed that the problem was on my side.

More like on 🔥 🦊 side...

kiwiidb (approved):

escapedcat:
> why would it now work?

Because the issue was Firefox-reload, not the code. Let's have a look together to double check?

bumi:
hmm...

escapedcat:
😱

PR #972, opened 2022-05-30
refactor(account): info message-type #652

escapedcat:
Refactoring account-info:
- TS
- tests


src/extension/background-script/actions/accounts/info.ts
```diff
@@ -0,0 +1,38 @@
+import type { MessageAccountInfo } from "~/types";
+import type { AccountInfoRes } from "~/common/lib/api";
+import state from "~/extension/background-script/state";
+
+const info = async (message: MessageAccountInfo) => {
+  // WHAT TO DO WITH THIS?
```
escapedcat, line 6: This `TODO!` comment was already there. What to do with this?

src/extension/background-script/actions/accounts/__tests__/info.test.ts
```diff
@@ -0,0 +1,50 @@
...
+import state from "~/extension/background-script/state";
+import type { MessageAccountInfo } from "~/types";
+
+import infoAccount from "../info";
+
+jest.mock("~/extension/background-script/state");
+
+const mockState = {
+  getConnector: () => ({
+    getInfo: () => Promise.resolve({ data: { alias: "getalby.com" } }),
```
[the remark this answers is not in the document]
escapedcat: You mean for tests? Uhm, let's do it once we need it? Would lie to avoid pre-optimization here.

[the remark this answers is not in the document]
bumi: sorry, should have marked the comment correct. I did not mean in this PR... that was just a general thought.

bumi (approved):

src/common/lib/api.ts
```diff
@@ -17,11 +17,11 @@ import {
 } from "./cache";
 import utils from "./utils";
 
-interface AccountInfoRes {
+export interface AccountInfoRes {
+  balance: { balance: string | number };
```
dylancom, line 21: I think we have to make a decision here. Always string or always number. (used to be string)

escapedcat: Currently we get `0` when balance it's zero but a `string` when there's a balance. Will check if we can adjust the `0` balance handling.

bumi: +1 we need to check amount handling. I am not sure if number is good enough? what's with big.js or something like that?

escapedcat: Hm, maybe pre TS nobody noticed:

- LND  
      `0` as strings:    
      image

- LNDHUB
      `0` as number:
       image

Should we keep `number | string` for now or should we check the connectors and parse anything `number` to be ` string`?

escapedcat: On the other hand this is how `AccountInfo` is typed currently:  
image

escapedcat: > +1 we need to check amount handling. I am not sure if number is good enough? what's with big.js or something like that?

Not too experienced with this. As long as we don't have an issue here I think we're good for now. In this case `balance` is always sats, right? So we won't end up with an issue because of i.e. `0.1324422` sats?

PR #973, opened 2022-05-30
fix: purgecss ignores max-screen classes

dylancom:

Currently max-screen-width-${maxWidth} classes are ignored, because PurgeCSS doesn't handle dynamically created classes. (unless you explicitly define them in your tailwind setup).
This change makes sure that the classes are fully written and can be found by PurgeCSS

### DIFF
--- a/src/app/components/Container.tsx
+++ b/src/app/components/Container.tsx
@@ -1,11 +1,29 @@
+type MaxWidth = "sm" | "md" | "lg" | "xl" | "2xl";
+
 type Props = {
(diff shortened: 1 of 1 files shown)


escapedcat (approved):
Nice, how did you notice that?

dylancom:
After a pull from master and a fresh dev build I noticed that all screens in the options where wider than they should be.
During making this Container component there were still some fully written max-w-screen-lg for e.g. within screens so PurgeCSS didn't ignore it (and this bug slipped in).

PR #982, opened 2022-05-31
feat: show it when there is no budget set

dylancom:
- Currently the allowance menu sits empty on the right and doesn't indicate what it can do.
This change shows there is no budget "0 / 0" just like on the Publisher screen and gives you a hint what you could do with the menu.

Before:
Schermafbeelding 2022-05-31 om 12 06 51

After:
Schermafbeelding 2022-05-31 om 12 04 12

### DIFF
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -136,26 +136,21 @@ function Home() {
       <>
         <div className="px-4 pb-5">
           <div className="flex justify-between items-center py-3">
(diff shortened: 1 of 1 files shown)


escapedcat:
That's an improvement for sure. Would it be good to show something actionable there? Like "Yoyo, Stacker news again, srsly? Better create an allowance now!" :P

dylancom:
Ye definitely, we need something actionable there. But we need to rethink how to make allowances more clear overall (so for now I aligned it with our Publisher screen).

PR #984, opened 2022-06-01
Refactor/652 account remove delete

escapedcat:
- Switch `remove` account to `delete` account because it was the same code
- `delete` has more functionality but can handle this as well. Extended the tests
- Switched some code from Promise/then to async/await

### DIFF
--- a/src/app/screens/Onboard/TestConnection/index.tsx
+++ b/src/app/screens/Onboard/TestConnection/index.tsx
@@ -13,36 +13,44 @@ export default function TestConnection() {
     name: string;
     balance: number;
   }>();
-  const [errorMessage, setErrorMessage] = useState();
+  const [errorMessage, setErrorMessage] = useState("");
   const [loading, setLoading] = useState(false);
   const { t } = useTranslation(["welcome"]);
 
   const navigate = useNavigate();
 
-  function handleEdit(event: React.MouseEvent<HTMLButtonElement>) {
-    utils.call("removeAccount").then(() => {
-      navigate(-1);
-    });
+  async function handleEdit(event: React.MouseEvent<HTMLButtonElement>) {
+    await utils.call("deleteAccount");
+    navigate(-1);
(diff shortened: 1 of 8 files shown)


src/app/screens/Onboard/TestConnection/index.tsx
```diff
@@ -13,36 +13,44 @@ export default function TestConnection() {
...
-        const { balance: resBalance } = response.balance;
-        const balance =
-          typeof resBalance === "number" ? resBalance : parseInt(resBalance);
+    try {
+      const response = await api.getAccountInfo();
 
-        setAccountInfo({ alias, balance, name });
-      })
-      .catch((e) => {
+      if ("error" in response) {
```
bumi, line 32: when is this the case? doesn't getAccountInfo throw an error in the error case?

escapedcat: Yeah, thanks. Had a look again and changed it.

src/app/screens/Onboard/TestConnection/index.tsx
```diff
@@ -13,36 +13,39 @@ export default function TestConnection() {
...
+      const { balance: resBalance } = response.balance;
+      const balance =
+        typeof resBalance === "number" ? resBalance : parseInt(resBalance);
 
-        setAccountInfo({ alias, balance, name });
-      })
-      .catch((e) => {
+      setAccountInfo({ alias, balance, name });
+    } catch (e) {
+      if (typeof e === "string") {
```
dylancom, line 39: In which case will a string be thrown?

escapedcat: Good point. Checked how we do it in other places and adjusted.

dylancom (approved):

im-adithya:
@escapedcat ig you forgot to delete the `remove.js` file :P

PR #990, opened 2022-06-03
fix: button text overflows in menu

im-adithya:

Added `overflow: hidden` to `menuItemButton` styling

**Before**
Screenshot from 2022-06-03 22-06-54

**After**
Screenshot from 2022-06-03 22-10-38

### Checklist


escapedcat:
Instead of just cutting it off hard it could be nice to try using [overflow ellipsis](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow). Could be combined with a default [`title`-attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) which would show the full name on hover.

im-adithya:
@escapedcat can you please check how the recent commit is working on your system. I'm facing a weird issue because of which the wallet icon is not being displayed before the account name (as you can see in the above screenshot) and is now (after the recent commit) of variable widths despite having the `w-6` style added.

Screenshot from 2022-06-04 13-13-01

(Mostly it should be an issue on my side with the CSS but it still was a bit weird before too, so I'm asking just to make sure everything's fine!)

> Could be combined with a default [title-attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) which would show the full name on hover.
I was thinking of the same, would help in increasing accessibility. Will add it after resolving the above issue!

escapedcat:
> I'm facing a weird issue because of which the wallet icon is not being displayed before the account name (as you can see in the above screenshot)

My understanding is that it was removed on purpose because it didn't provide any value to the user. See: #916

> > Could be combined with a default [title-attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) which would show the full name on hover.
> 
> I was thinking of the same, it would help in increasing accessibility! Will add it after resolving the above issue!

👍

im-adithya:
> See: https://github.com/getAlby/lightning-browser-extension/pull/916

Not the badge, I was talking about the wallet icon^

escapedcat:
> Not the badge, I was talking about the wallet icon^

Oh, I see. Maybe @secondl1ght knows

bumi:
I think the icon should be there. seems to be because of the length of the text?

im-adithya:
Ah, must be it. Applying shrink-0 should fix it then!

im-adithya:
Made all the necessary changes, ready for review!

Screenshot from 2022-06-04 19-59-29

src/app/components/AccountMenu/index.tsx
```diff
@@ -83,9 +83,12 @@ function AccountMenu({ title, subtitle, showOptions = true }: Props) {
                   selectAccount(accountId);
                 }}
                 disabled={loading}
+                title={account.name}
               >
-                <WalletIcon className="w-6 h-6 -ml-0.5 mr-2 opacity-75 text-gray-700 dark:text-neutral-300" />
-                {account.name}&nbsp;
+                <WalletIcon className="w-6 h-6 -ml-0.5 mr-2 shrink-0 opacity-75 text-gray-700 dark:text-neutral-300" />
+                <div className="overflow-hidden text-ellipsis whitespace-nowrap">
```
escapedcat, line 89: Could this be a `span`? Block element within this element seems not semantic.

escapedcat (changes requested):

src/app/components/AccountMenu/index.tsx
im-adithya: Done!

escapedcat (approved):

bumi (approved):

PR #991, opened 2022-06-03
feat: highlight selected account in menu

im-adithya:

I have added a `selected` prop (as active is already in use because of Headless UI) to highlight the selected button in menu.

Screenshot from 2022-06-03 23-07-14

### Note

Do we have to update the stories to include the new selected prop? 
(`/src/app/components/Menu/Menu.stories.tsx`)

### Checklist

### DIFF
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -83,6 +83,7 @@ function AccountMenu({ title, subtitle, showOptions = true }: Props) {
                   selectAccount(accountId);
                 }}
                 disabled={loading}
+                selected={accountId === auth.account?.id}
               >
                 <WalletIcon className="w-6 h-6 -ml-0.5 mr-2 opacity-75 text-gray-700 dark:text-neutral-300" />
                 {account.name}&nbsp;
--- a/src/app/components/AccountMenu/index.test.tsx
+++ b/src/app/components/AccountMenu/index.test.tsx
@@ -3,6 +3,7 @@ import userEvent from "@testing-library/user-event";
 import { act } from "react-dom/test-utils";
 import { BrowserRouter } from "react-router-dom";
 import * as AccountsContext from "~/app/context/AccountsContext";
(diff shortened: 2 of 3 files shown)


src/app/components/AccountMenu/index.tsx
```diff
@@ -83,6 +83,7 @@ function AccountMenu({ title, subtitle, showOptions = true }: Props) {
                   selectAccount(accountId);
                 }}
                 disabled={loading}
+                selected={accountId === auth.account?.id}
```
escapedcat, line 86: Would you mind extending this test with the new `selected` prop?:  
https://github.com/getAlby/lightning-browser-extension/blob/ed6bb1aa3ae39ed674abc8f686fb3f64ef867541/src/app/components/AccountMenu/index.test.tsx  

i.e. add a test "Current account is being highlighted"

escapedcat (changes requested):

escapedcat:
In general I'm always a bit conflicted using `bold` for highlights because the text will be wider. If a text is already long this might break things. Not sure if this can be the case here though. Just a thing to keep in mind and maybe test it.

nit: Not sure if this a bugfix. It's more like a new feature I think.

Thanks for your work!

im-adithya:
> Would you mind extending this test with the new selected prop?:

Done! :D

> nit: Not sure if this a bugfix. It's more like a new feature I think.

Changed the commit message and title!

im-adithya:
Also we should wait for #992 to be merged so that I can change it here as well!
(i.e. `fireEvent` to `userEvent`)

im-adithya:
> so that I can change it here as well! (i.e. `fireEvent` to `userEvent`)

Done!

bumi (approved):
thanks! looks good to me. 
I think the bold can easily be changed when we maybe chat about the design with @dvoroneca

escapedcat (approved):

escapedcat:
@im-adithya fyi, @dylancom changed it a bit here to avoid changing "all" menus:  #1001

PR #993, opened 2022-06-04
chore: rename type to action

### LINKED ISSUE #759: Rename message `type` to `action`
We had started this before. I think `action` is a better name and matches that we have an `actions` folder in the background script

The `type` or `action` in the content-script message to the background script defines which "action" function should be executed. The core of this is [here](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/index.js#L86-L89) 

I think `action` is more descriptive than `type`. But we have to be careful to find all places where the message is used or where a message is published.

im-adithya:

Renamed message `type` to `action`

### Link this PR to an issue

Closes #759

- `chore`: Renamed variable (updating grunt tasks etc; no production code change)

### How has this been tested?

Ran all unit and e2e tests.

### Checklist

### DIFF
--- a/src/common/lib/msg.ts
+++ b/src/common/lib/msg.ts
@@ -10,7 +10,7 @@ const msg = {
       .sendMessage({
         application: "LBE",
         prompt: true,
-        action: action,
+        action,
         args: args,
         origin: { internal: true },
         ...overwrites,
--- a/src/common/lib/msg.ts
+++ b/src/common/lib/msg.ts
@@ -2,7 +2,7 @@ import browser from "webextension-polyfill";
 
 const msg = {
   request: (
-    type: string,
+    action: string,
     args?: { [key: string]: string | number },
     overwrites?: { [key: string]: string }
   ) => {
@@ -10,7 +10,7 @@ const msg = {
       .sendMessage({
         application: "LBE",
         prompt: true,
-        type: type,
+        action: action,
         args: args,
         origin: { internal: true },
         ...overwrites,
--- a/src/extension/background-script/actions/ln/index.js
+++ b/src/extension/background-script/actions/ln/index.js
@@ -11,7 +11,7 @@ import verifyMessage from "./verifyMessage";
 
 const connectorCall = (method) => {
   return async (message, sender) => {
-    console.log(`Lightning call: ${message.type}`);
+    console.info(`Lightning call: ${message.action}`);
     const connector = await state.getState().getConnector();
 
     if (!connector) {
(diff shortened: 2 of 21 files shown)


src/extension/background-script/actions/ln/index.js
```diff
@@ -11,7 +11,7 @@ import verifyMessage from "./verifyMessage";
 
 const connectorCall = (method) => {
   return async (message, sender) => {
-    console.log(`Lightning call: ${message.type}`);
+    console.log(`Lightning call: ${message.action}`);
```
escapedcat, line 14: nit: change this to `console.info`

escapedcat (approved):

src/extension/background-script/actions/ln/index.js
im-adithya: Done!

src/extension/background-script/index.js
```diff
@@ -83,7 +83,7 @@ const routeCalls = (message, sender) => {
   }
   const debug = state.getState().settings.debug;
 
-  const action = message.type; //|| message.action; // TODO: what is a good message format to route to an action?
+  const action = message.action; //|| message.action; // TODO: what is a good message format to route to an action?
```
bumi, line 86: Because this is a tricky change and if we missed a place where we used `type` the app will break. 

I suggest to do something like this here: 

```js
if (message.type) { 
  console.debug("Invalid message, using type", message);
}
const action = message.action || message.type;
```

this would give us a log entry and would not break the app. 
what do you think?

im-adithya: That should help as we have a limited number of unit tests!

im-adithya: Done!

escapedcat (approved):

dylancom:
@bumi have to get a bit used to this, coming from "Redux" where actions are objects that hold a "type" and a payload.

escapedcat:
> @bumi have to get a bit used to this, coming from "Redux" where actions are objects that hold a "type" and a payload.

Is there a better naming which could make things less confusing?

dylancom:
We are for sure not bound by Redux terminology hehe, I understand to align it with the background actions though.

src/common/lib/msg.ts
```diff
@@ -2,15 +2,15 @@ import browser from "webextension-polyfill";
...
+    action: string,
     args?: { [key: string]: string | number },
     overwrites?: { [key: string]: string }
   ) => {
     return browser.runtime
       .sendMessage({
         application: "LBE",
         prompt: true,
-        type: type,
+        action: action,
```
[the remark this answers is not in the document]
im-adithya: Done!

PR #1000, opened 2022-06-06
Feat/597 currency support

### LINKED ISSUE #597: Make fiat currency configurable and display amounts additionally in fiat
Additionally to the satoshi amounts we should show fiat amounts for the not-yet hyper-bitcoinized users.

The favorit currency should be configurable and additional to the satoshi amounts we should convert the amount to fiat.

escapedcat:
Fixes #597 

### Tasks
- [x] rebase
- [x] prettier
- [x] rename NumberField
- [x] create tests for conversion utils

### Design issues
image  
image
image
 ✅  Will be solved by follow up issues.


src/app/components/AccountMenu/index.tsx
```diff
@@ -15,7 +15,10 @@ import Menu from "../Menu";
 
 export type Props = {
   title: string;
-  subtitle: string;
+  subtitle: {
```
escapedcat, line 18: nit: `balances`

src/app/components/AccountMenu/index.tsx
```diff
@@ -62,8 +65,14 @@ function AccountMenu({ title, subtitle, showOptions = true }: Props) {
         <div className="text-xs text-gray-700 dark:text-neutral-400">
           {title || <Skeleton />}
         </div>
-        <div className="text-xs dark:text-white">
-          {subtitle || <Skeleton />}
+        <div className="flex justify-between">
+          {/* THIS DESIGN BREAKS IF TOOOOO MUCH SATS */}
+          <div className="text-xs dark:text-white">
```
escapedcat, line 70: Not all `div`s

src/app/components/AccountMenu/index.tsx
```diff
@@ -62,8 +65,14 @@ function AccountMenu({ title, subtitle, showOptions = true }: Props) {
         <div className="text-xs text-gray-700 dark:text-neutral-400">
           {title || <Skeleton />}
         </div>
-        <div className="text-xs dark:text-white">
-          {subtitle || <Skeleton />}
+        <div className="flex justify-between">
+          {/* THIS DESIGN BREAKS IF TOOOOO MUCH SATS */}
```
escapedcat, line 69: Could be tackled in the future!

src/app/screens/Receive.tsx
```diff
@@ -201,29 +219,32 @@ function Receive() {
...
-                2 hours
-              </option>
-              <option key="180" value="180">
-                3 hours
-              </option>
-            </Select>
-          </div>
-        </div> */}
+                {/*
+                // DO WE NEED THIS?
```
escapedcat, line 223: delete!

src/common/constants.ts
```diff
@@ -1,2 +1,171 @@
 export const ABORT_PROMPT_ERROR = "Prompt was closed";
 export const USER_REJECTED_ERROR = "User rejected";
+export const CURRENCIES = [
```
escapedcat, line 3: Add comment regarding splitting it up in the future

src/app/styles/index.css
```diff
@@ -44,3 +44,18 @@ main {
 .ReactModal__Overlay--before-close {
   opacity: 0;
 }
+
+
+
+/* Hide number spinner */
+/* Webkit */
+#numberInput input::-webkit-outer-spin-button,
```
escapedcat, line 52: No `id` should be used!

src/extension/background-script/state.ts
```diff
@@ -29,13 +35,15 @@ interface BrowserStorage {
...
 
-export const DEFAULT_SETTINGS = {
+export const DEFAULT_SETTINGS: SettingsStorage = {
   websiteEnhancements: true,
   legacyLnurlAuth: false,
   userName: "",
   userEmail: "",
   locale: i18n.resolvedLanguage,
   theme: "system",
+  currency: "USD" as SupportedCurrencies,
```
escapedcat, line 45: remove the `as`

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  min,
+  max,
+  suffix,
+  endAdornment,
+}: React.InputHTMLAttributes<HTMLInputElement> & Props) {
+  const inputEl = useRef<HTMLInputElement>(null);
+  const outerStyles =
+    "shadow-sm rounded-md border border-gray-300 dark:border-gray-800 bg-white dark:bg-black transition duration-300";
+
+  const inputNode = (
```
escapedcat, line 36: Wondering if our existing default `Input` can/should be used here

src/app/components/form/DualCurrencyField/index.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter } from "react-router-dom";
+
+import type { Props } from "./index";
+import DualCurrencyField from "./index";
+
+const props: Props = {
+  endAdornment: <div />,
```
escapedcat, line 8: afaik can be removed

src/app/context/AuthContext.tsx
```diff
@@ -27,9 +31,16 @@ interface AuthContextType {
 
 const AuthContext = createContext({} as AuthContextType);
 
+// @TODO: open new issue
+// rename to "accountProvider"?
+// proposae in a new PR
 export function AuthProvider({ children }: { children: React.ReactNode }) {
   const [account, setAccount] = useState<AuthContextType["account"]>(null);
   const [loading, setLoading] = useState(true);
+  const [balancesDecorated, setBalancesDecorated] = useState<{
```
dylancom, line 40: TS understands the type already by the initial empty string values you provide.

src/common/utils/currencyConvert.ts
```diff
@@ -0,0 +1,101 @@
...
+    style: "currency",
+    currency: currency,
+  });
+  return {
+    satsBalance: `${balance} sats`,
+    fiatBalance: localeFiatValue,
+  };
+};
+
+const getFiatBtcRate = async (
```
bumi, line 36: can we cache this? right now a request is made every time an input number is changed. 

wondering also if this should be done in the background script and the rate should be stored in the settings storage?

src/app/components/AccountMenu/index.tsx
```diff
@@ -16,11 +16,14 @@ import Menu from "../Menu";
 
 export type Props = {
   title: string;
-  subtitle: string;
+  balances: {
+    satsBalance: string | null;
```
dylancom, line 20: isn't it always a string? (empty string by default)

src/app/screens/Settings.tsx
```diff
@@ -124,42 +132,95 @@ function Settings() {
...
+        <Setting
+          title="Currency"
+          subtitle="Change the currency display within Alby"
+        >
+          {!loading && (
+            <div className="w-64">
+              <Select
+                name="currency"
+                value={settings.currency}
+                onChange={async (eventent) => {
```
dylancom, line 145: eventent?

dvoroneca:
## I have a design challenge to display large amounts of SATS and the symbol vs acronym.

Firstly,
- If the amount exceeds 6 digits, wouldn't it be nice to display amounts in BTC?
- Separate every 3 digits with a "," or with space?
- How many digits after the "." would be nice to show if we show SATS in BTC?

Here are some designs to talk about:
Screenshot 2022-06-08 at 14 43 58

----------------------

Secondly, 
What do you think works best as a FIAT value currency indicator- acronyms (USD, EUR, INR) or symbols ($, €, ₹ )

Screenshot 2022-06-08 at 14 57 29

src/app/components/form/DualCurrencyField/index.tsx
escapedcat: @dylancom what do you think?

src/common/utils/currencyConvert.ts
escapedcat: Tried to address this here: https://github.com/getAlby/lightning-browser-extension/pull/1000/files#diff-8532ac3fc9044b3365001d3ea67200d23b63d5edfd398f7d772954117e92e11aR56

But I guess you're right. Instead of `€36,139.34` we could write `~€36,139.34` and only get the _exact_ rate every i.e. x minutes.  
Easiest way I could imagine is setting a long debounce (i.e. 1 hour) on this function.

The background idea might work as well. Not sure how we would throttle the calls. Maybe just the same way?  
Let's talk?

escapedcat:
@dvoroneca 

1. Large amounts of SATS
   - I was wondering about the width change of that dropdown. Is this wanted? Should we align this somehow?  
   - Regarding huge numbers or switching to BTC I think this should be planned in a separate issue. Could work nicely though!

1. The symbol vs acronym
   I don't have a real preference there. I assume people are used to their local currency symbol and the place it should displayed (before or after the amount).

src/app/context/AuthContext.tsx
```diff
@@ -27,9 +31,16 @@ interface AuthContextType {
 
 const AuthContext = createContext({} as AuthContextType);
 
+// @TODO: open new issue
```
escapedcat, line 34: @dylancom I just saw this...  
https://github.com/getAlby/lightning-browser-extension/blob/chore/yarn-action-3-only/src/app/context/AccountsContext.tsx  
Should we move everything to that?

src/types.ts
```diff
@@ -237,4 +241,13 @@ export interface SettingsStorage {
...
+  currency: SupportedCurrencies;
+  exchange: SupportedExchanges;
 }
+
+export type SupportedExchanges = "coindesk" | "yadio";
+
+// Supported currencies by Coindesk and yadio
+// FYI: yadio is i.e. not supporting "ISK", maybe more?
+// https://github.com/AryanJ-NYC/bitcoin-conversion/blob/master/src/index.ts#L143
+export type SupportedCurrencies = typeof CURRENCIES[number];
```
escapedcat, line 253: @dylancom should `currencies` and `exchanges` somehow be `enums`?

src/app/context/AuthContext.tsx
dylancom: Yes, we could combine those.

src/types.ts
dylancom: https://www.vendure.io/docs/typescript-api/common/currency-code/

dylancom: Seems like a practice.

src/app/components/form/DualCurrencyField/index.tsx
dylancom: If it's easy to add a second value in (new prop), we should re-use it.

bumi (commented):
I think it is a bit confusing that we have number formatting (for en) for the fiat numbers but not the satoshi number. Can we use `toLocaleString` there, too? 

that `toLocaleString` seems pretty good actually. 
Good to see that it also seems to get the precision correct. (currencies have 0 - 3 decimal points). 


---

when I change the currency in the settings, the account menu shows the correct currency but the rate is not updated (it uses still the number of the previously selected currency). A bit of an edge case maybe? - how is the debounce function actually behaving in that case?)

src/app/context/AuthContext.tsx
escapedcat: Added a todo

escapedcat:
> I think it is a bit confusing that we have number formatting (for en) for the fiat numbers but not the satoshi number. Can we use `toLocaleString` there, too?

Yeah, we will do this in #1022

> when I change the currency in the settings, the account menu shows the correct currency but the rate is not updated (it uses still the number of the previously selected currency). A bit of an edge case maybe? - how is the debounce function actually behaving in that case?)

No, you're right. I tackled this. It works, but it's not very beautiful. We could improve it with #1021

src/common/utils/currencyConvert.ts
```diff
@@ -0,0 +1,122 @@
...
+  let response;
+
+  if (exchange === "yadio") {
+    response = await axios.get(
+      `https://api.yadio.io/exrates/${currency.toLowerCase()}`
+    );
+
+    const data = await response?.data;
+
+    return currencyJs(data.BTC, {
```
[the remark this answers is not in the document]
escapedcat: You're right! Removed the dep and used the returned float-rates directly.

src/app/components/form/DualCurrencyField/index.tsx
escapedcat: https://github.com/getAlby/lightning-browser-extension/issues/1042

Issue #946, opened 2022-05-25
Allow users to go "back" to the general home from the website contextual screen


bumi:
Once there is an allowance for a website we show a publisher card and only the transactions made on the page on the home screen.

For some users it is confusing because this looks different to the "general" home screen.
We could add a way to go "back" to the general home from the contextual home screen.

PR #1005, opened 2022-06-07
feat: go back to general home from contextual home

dylancom:

Allow users to switch from the AllowanceView, back to the DefaultView.

### Link this PR to an issue

#946

### DIFF
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -291,7 +291,7 @@ function Home() {
     <div>
       {allowance && (
         <Header
-          title="Back to Home"
+          title={allowance.host}
           headerLeft={
             <IconButton
               onClick={() => setAllowance(null)}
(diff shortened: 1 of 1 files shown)


escapedcat (approved):

src/app/screens/Home/index.tsx
```diff
@@ -286,6 +289,17 @@ function Home() {
 
   return (
     <div>
+      {allowance && (
+        <Header
+          title="Back to Home"
```
bumi, line 294: how about we use `allowance.host` here? 

In the other screens we use the title showing the "name" of the current page. e.g. on the send screen it is "send".

dylancom: Ye I agree, the back button speaks for itself aswel.

escapedcat (approved):

secondl1ght (approved):

bumi (approved):

Issue #669, opened 2022-03-08
Export wallets after creation


kiwiidb:
**Is your feature request related to a problem? Please describe.**
There is no way to export a wallet after it’s created. So if you forget to backup for some reason, you can’t do it later. Or if you first install in Chrome, but then want it in Firefox as well, you can’t easily get the LNDhub connection string.

**Describe the solution you'd like**
At least for LNDhub wallets, we could have the option to export the connection string at any time (password protected). This way, your computer and mobile version of the wallet act like backups of eachother.

I don't think we need this for other connectors, but we could consider it.

**Describe alternatives you've considered**
- Inspect `/auth` network requests to extract the connection string.
- Look in the storage to extract the connection string.

**Are you working on this? (No)**

MoritzKa:
Feedback from another user: 

- A botton to export this URL / download QR Code image
- A warning on not sharing this

bumi:
We now got an accounts list. We can have an account detail page that shows the credentials and related information

bumi:
I think the easiest to do this is to add an "export" menu item and then show the lndhub connection details in a modal. 
As we show the details after the account is added: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/connectors/NewWallet/index.tsx#L127-L141

<img width="1008" alt="image" src="https://user-images.githubusercontent.com/318/172261321-8b10017e-90a1-4622-ac69-a67e9a86b95b.png">

im-adithya:
Any changes? @bumi 
![Screenshot from 2022-06-07 19-31-14](https://user-images.githubusercontent.com/64399555/172399763-c5967753-0482-436b-90f6-750329111cfa.png)
![Screenshot from 2022-06-07 19-30-15](https://user-images.githubusercontent.com/64399555/172399641-69a8382b-22cd-4001-b7a3-4e1a801c9636.png)

bumi:
I don't think we have the data for the lightning address there. so that would be just the export QR code AND we should add it as a string in an input field. - that allows users to copy it.

escapedcat:
At a later stage we could think about adding a "copy to clipboard" button to an input. Not sure if we already have this somewhere. I.e. https://bionicjulia.com/blog/creating-copy-to-clipboard-button-in-react  
Should not be done in this issue.

PR #1012, opened 2022-06-08
feat: add export account

im-adithya:

Introduces Export option in the accounts list.

### Link this PR to an issue

#669

If we have lightning address:

Screenshot from 2022-06-08 17-22-47

If we don't:

Screenshot from 2022-06-08 17-22-53

### How has this been tested?

Ran all tests

### Checklist


src/app/screens/Accounts/index.tsx
```diff
@@ -195,6 +236,70 @@ function AccountsScreen() {
...
+                      Your lightning address:
+                      <br />
+                    </strong>
+                  </p>
+                  {lndHubData.lnAddress && <p>{lndHubData.lnAddress}</p>}
+                </div>
+              )}
+              <div className="flex justify-center space-x-3 items-center dark:text-white">
+                <div className="flex-1">
+                  <strong>To use your wallet on your mobile:</strong>
```
escapedcat, line 278: If both sentences are `p` we could get rid of the `br`?

src/app/screens/Accounts/index.tsx
```diff
@@ -195,6 +236,70 @@ function AccountsScreen() {
             </div>
           </form>
         </Modal>
+
```
escapedcat, line 239: @dylancom you think by now we should move out these two modals into separate component-files?

src/app/screens/Accounts/index.tsx
escapedcat: Same for the part above this

src/extension/background-script/actions/accounts/export.ts
```diff
@@ -0,0 +1,26 @@
+import { decryptData } from "~/common/lib/crypto";
+import state from "~/extension/background-script/state";
+import type { MessageAccountEdit } from "~/types";
+
+const exportAccount = async (message: MessageAccountEdit) => {
```
escapedcat, line 5: The type is wrong. Would be good to add a new type and create a test like we do for other account-actions that use TS already.

src/extension/background-script/actions/accounts/export.ts
```diff
@@ -0,0 +1,26 @@
...
+  const accounts = state.getState().accounts;
+  const password = state.getState().password as string;
```
escapedcat, line 7: We should try to avoid `ad`. It's not good practice.

src/extension/background-script/actions/accounts/export.ts
escapedcat: Ah you created a type but don't use it here :)

escapedcat (changes requested):

src/extension/background-script/actions/accounts/export.ts
im-adithya: You mean `as`? It is causing a typescript error below where we use it in the `decryptData` function...

escapedcat: Arrr, yeah, `as`... pity, ok

src/extension/background-script/actions/accounts/export.ts
im-adithya: Oops forgot to change :P

src/app/screens/Accounts/index.tsx
im-adithya: Makes sense, done!

src/extension/background-script/actions/accounts/export.ts
escapedcat: Could you add a test for this as well, like we have for the other ones?:  
https://github.com/getAlby/lightning-browser-extension/tree/master/src/extension/background-script/actions/accounts/__tests__

im-adithya: Yup! Currently working on the same locally, will push it soon!

src/extension/background-script/actions/accounts/export.ts
```diff
@@ -0,0 +1,32 @@
+import { decryptData } from "~/common/lib/crypto";
+import state from "~/extension/background-script/state";
+import type { MessageAccountExport } from "~/types";
+
+const exportAccount = async (message: MessageAccountExport) => {
```
bumi, line 5: maybe we should call this action differently? this is not "exporting" the account data it is returning the config. or rather the full account details?

show? details?

src/extension/background-script/actions/accounts/export.ts
```diff
@@ -0,0 +1,32 @@
...
+        accounts[accountId].config as string,
+        password
+      );
+
+      return {
+        data: lndHubData,
+      };
+    } else {
+      return {
+        error: `Account: ${accountId} not an LndHub Account; cannot be exported`,
```
bumi, line 22: theoretically we could do this for all accounts. it is just returning the decrypted config. 

we can only do the check for the lndhub on the frontend. later we then could also display the details of other connector types. 

what do you think?

src/app/screens/Accounts/index.tsx
```diff
@@ -195,6 +236,70 @@ function AccountsScreen() {
...
+                  <p>
+                    <strong>Your lightning address:</strong>
+                  </p>
+                  {lndHubData.lnAddress && <p>{lndHubData.lnAddress}</p>}
+                </div>
+              )}
+              <div className="flex justify-center space-x-3 items-center dark:text-white">
+                <div className="flex-1">
+                  <p>
+                    <strong>To use your wallet on your mobile:</strong>
```
bumi, line 276: how about this text?
```
Tip: Use this wallet with your mobile device

Import this wallet into Zeus or BlueWallet by scanning the QRCode. 
```

src/extension/background-script/actions/accounts/export.ts
im-adithya: There's already an Info type, so I guess we can go with 'Details'...

src/extension/background-script/actions/accounts/export.ts
im-adithya: Yeah, I added this while adding tests, this is checked on the frontend only. Will change!

src/extension/background-script/actions/accounts/export.ts
escapedcat: `info`, `details`, `export`... seems all a bit confusing. `display`, also confusing, because it indicates what we do with the received data and not what this data is. Comparing `info` and this it seems like the main difference is, that we show something `decrypted`? `decryptedDetails`?
What do you think?

im-adithya: Makes sense, shall we go with this then @bumi?

bumi: yep!

im-adithya: Done! Ready for review!

src/app/screens/Accounts/index.tsx
```diff
@@ -46,7 +60,21 @@ function AccountsScreen() {
...
     getAccounts();
-    closeModal();
+    closeEditModal();
+  }
+
+  async function exportAccount({ id, name }: AccountAction) {
+    setLoading(true);
+    setTimeout(() => {
+      setExportModalIsOpen(true);
+    }, 50);
```
escapedcat, line 70: Sorry, one last thing. Could you add the comment:

```
                            /**
                             * @HACK
                             * @headless-ui/menu restores focus after closing a menu, to the button that opened it.
                             * By slightly delaying opening the modal, react-modal's focus management won't be overruled.
                             * {@link https://github.com/tailwindlabs/headlessui/issues/259}
                             */
                            setTimeout(() => {
                              setModalIsOpen(true);
                            }, 50);
```

Eventually we want to get rid of this hack and need to find all occurrences 🤞

im-adithya: Done! Ready to merge now!

im-adithya: Btw there are three (including above) occurences if this hack: 
`src/app/components/AllowanceMenu/index.tsx` (1)
`src/app/screens/Accounts/index.tsx` (2)
And issue (https://github.com/tailwindlabs/headlessui/issues/259) doesn't seem to be resolved.
Gotta do something

escapedcat: Problem is, that the issue is closed. Maybe we should open a new one with a codesandbox example or something?

src/app/screens/Accounts/index.tsx
```diff
@@ -195,6 +242,72 @@ function AccountsScreen() {
...
+              <Loading />
+              <span>waiting for LndHub data...</span>
+            </div>
+          )}
+          {!loading && (
+            <div className="p-5 border-t border-b border-gray-200 dark:bg-surface-02dp dark:border-neutral-500">
+              {lndHubData.lnAddress && (
+                <div className="dark:text-white mb-6">
+                  <p>
+                    <strong>Your lightning address:</strong>
```
bumi, line 274:
```suggestion
                    <strong>Your Lightning Address:</strong>
```

I think we write it like this everywhere else

bumi (approved):
got one last thing. and if you fix the merge conflict we are good to go!

escapedcat:
Please fix conflicts. You think we should move the modals in to a separate file(s)?

im-adithya:
The `edit` action also uses modals... will make a separate PR for both!

bumi (approved):
we should add i18n (@im-adithya maybe we do this in a new PR?)

im-adithya:
> we should add i18n (@im-adithya maybe we do this in a new PR?)

Sure!

PR #1013, opened 2022-06-08
fix: request.type being used instead of request.action

riccardobl:
This typo prevents extractLightningData() from being called.


bumi:
thanks! this was introduced in the last merged PR. 🙈

PR #1014, opened 2022-06-08
Use github api to fetch lightningData

riccardobl:
This fetches the gh user bio using github rest apis instead of parsing the HTML .
Using the HTML parser seems to be unreliable somehow, i couldn't even manage to get it to work in my case, it might be due to the theme i use, my region or something else, I am not really sure, however the selectors seem to be wrong.

This PR uses the official github apis to do the same thing in a more reliable manner.

I can do the same thing for the repos, but i am not sure which places need to be searched for the ln address. Maybe the repo description? and also the readme? if someone can confirm this i will extend the PR to cover that case as well

### DIFF
--- a/src/extension/content-script/batteries/GitHub.ts
+++ b/src/extension/content-script/batteries/GitHub.ts
@@ -17,7 +17,7 @@ const battery = async (): Promise<void> => {
     lightningData = await getLightningDataFromRepository(username, repo);
   }
 
-  // Search lightning data in the user profile if the current page is a profile page
+  // Search lightning data in the user profile or organization profile if the current page is a profile page
   // or if the current page is a repository page but the lightning data is not set at repository level
   if (!lightningData && username) {
     lightningData = await getLightningDataFromProfile(username);
--- a/src/extension/content-script/batteries/GitHub.ts
+++ b/src/extension/content-script/batteries/GitHub.ts
@@ -4,11 +4,15 @@ import getOriginData from "../originData";
 import setLightningData from "../setLightningData";
 
 const urlMatcher = /^https:\/\/github.com\/([^/]+)(\/([^/]+))?$/;
+const validationRegex = /^[a-z0-9_.-]+$/i;
 
 const battery = async (): Promise<void> => {
   const urlParts = document.location.pathname.split("/");
   const username = urlParts[1];
   const repo = urlParts[2];
+
+  if (username && !validationRegex.test(username)) return; // invalid username
+  if (username && repo && !validationRegex.test(repo)) return; // invalid repo
 
   let lightningData = null;
 
--- a/src/extension/content-script/batteries/GitHub.ts
+++ b/src/extension/content-script/batteries/GitHub.ts
@@ -3,75 +3,115 @@ import setLightningData from "../setLightningData";
 
 const urlMatcher = /^https:\/\/github.com\/([^/]+)(\/([^/]+))?$/;
 
-const battery = (): void => {
+const battery = async (): Promise<void> => {
   const urlParts = document.location.pathname.split("/");
   const username = urlParts[1];
   const repo = urlParts[2];
 
+  let found = false;
+
   if (username && repo) {
-    handleRepositoryPage(username);
-  } else if (username) {
-    handleProfilePage();
+    found = await handleRepositoryPage(username, repo);
+  }
+
+  if (!found && username) {
+    await handleProfilePage(username);
   }
 };
 
-function parseElement(elementSelector: string) {
-  const text = document.querySelector<HTMLElement>(elementSelector)?.innerText;
-  if (text) {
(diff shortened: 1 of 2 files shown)


riccardobl:
> I can do the same thing for the repos, but i am not sure which places need to be searched for the ln address. Maybe the repo description? and also the readme? if someone can confirm this i will extend the PR to cover that case as well

Ok i think i figured this out.

The updated code searches for ln addresses in the repo description, repo readme, user bio and user readme (ie the readme that can appear in the user profile).

Also in case the repo doesn't have a ln address but the owner has, it will fallback and propose to donate to the owner of the repo.

This should cover all cases nicely. Let me know what you think.

src/extension/content-script/batteries/GitHub.ts
```diff
@@ -3,75 +3,115 @@ import setLightningData from "../setLightningData";
 
 const urlMatcher = /^https:\/\/github.com\/([^/]+)(\/([^/]+))?$/;
 
-const battery = (): void => {
+const battery = async (): Promise<void> => {
   const urlParts = document.location.pathname.split("/");
   const username = urlParts[1];
   const repo = urlParts[2];
 
+  let found = false;
```
escapedcat, line 11: `found` what? Could this be named more expressive?

src/extension/content-script/batteries/GitHub.ts
```diff
@@ -3,75 +3,115 @@ import setLightningData from "../setLightningData";
...
-  const text = document.querySelector<HTMLElement>(elementSelector)?.innerText;
-  if (text) {
-    const match = text.match(/(⚡:?|lightning:|lnurl:)\s?(\S+@\S+)/i);
-    if (match) return match[2];
+  if (!found && username) {
+    await handleProfilePage(username);
   }
+};
+
+function parseText(text: string) {
```
escapedcat, line 22: is this i.e. `findLightningAddressInText`?

riccardobl: Yes it was called parseElement before. 
I've renamed it to findLightningAddressInText as you suggested

src/extension/content-script/batteries/GitHub.ts
riccardobl: Sure.
I've refactored this entire code to be cleaner, let me know if you like this more.

riccardobl:
Out of an abundance of caution i've added a regex to check for the validity of repo and username before using them in the api calls

src/extension/content-script/batteries/GitHub.ts
escapedcat: Cool thanks!

src/extension/content-script/batteries/GitHub.ts
escapedcat: Ah ok, missed that, sorry. Thanks for changing it!

src/extension/content-script/batteries/GitHub.ts
```diff
@@ -1,77 +1,125 @@
...
+  }
+
+  if (!address) {
+    // search in profile's readme
+    const userRepoReadmeData = await fetch(
+      `https://api.github.com/repos/${username}/${username}/readme`
+    ).then((res) => res.json());
+
+    if (userRepoReadmeData.download_url) {
+      const userRepoReadmeContent = await fetch(
```
bumi, line 62: I think instead of downloading the file we already get the content of that file in `userRepoReadmeData.content`. 
it is a base64 encoded string.

but we need to think that anything can be in there, the user controls the repo an could upload any file.
maybe we just check the bio for now?

src/extension/content-script/batteries/GitHub.ts
```diff
@@ -1,77 +1,125 @@
...
+  }
+
+  if (!address) return null; // address was not found
+
+  return {
+    method: "lnurl",
+    recipient: address,
+    ...getOriginData(),
+    description: userProfile.bio ?? "",
+    name: userProfile.name ?? userProfile.login,
```
bumi, line 76: should this be `userProfile.name || userProfile.login` ?

src/extension/content-script/batteries/GitHub.ts
```diff
@@ -1,77 +1,125 @@
...
+  }
+
+  if (!address) {
+    // searcg in readme
+    const repoReadmeData = await fetch(
+      `https://api.github.com/repos/${owner}/${repo}/readme`
+    ).then((res) => res.json());
+
+    if (repoReadmeData.download_url) {
+      const repoReadmeDataContent = await fetch(
```
bumi, line 106: also here we get the content of the file in the `repoReadmeDataContent.content`

src/extension/content-script/batteries/index.ts
```diff
@@ -35,7 +35,7 @@ async function extractLightningData() {
   );
 
   if (match) {
-    match.battery();
```
bumi, line 38: why do we need to wait here? was there a problem?

PR #1017, opened 2022-06-09
Add twitch support

riccardobl:
Search for ln address in channel bio.
Test channel: https://www.twitch.tv/rblb0/

### DIFF
--- a/src/extension/content-script/batteries/Twitch.ts
+++ b/src/extension/content-script/batteries/Twitch.ts
@@ -1,11 +1,7 @@
-import browser from "webextension-polyfill";
-
 import getOriginData from "../originData";
-import setLightningData from "../setLightningData";
+import { findLightningAddressInText, setLightningData } from "./helpers";
 
 const urlMatcher = /^https:\/\/(?:www\.)?twitch.tv\/([^/]+).+$/;
-const validationRegex = /^[a-z0-9_.-]+$/i;
-const clientIdExtractor = /clientId="([A-Z0-9]+)"/i;
 
 /* eslint-disable  @typescript-eslint/no-explicit-any */
 async function twitchApiCall(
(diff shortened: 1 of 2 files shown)


fczuardi:
@riccardobl On your test channel there is a :zap: emoji on the contents of the page because you don't have recent streams, I used a different test page here https://www.twitch.tv/sedentarismo where the user only see the lightning address if navigating to the "About" page. Do you know if there is some setting that the content creator can use to always display the lightning address on the main channel page?

riccardobl:
> Looking at the source code of the page with devtools, there is a `<meta name="twitter:description" content="⚡rblb@getalby.com">` under the `<head>` tag. I wonder if we could use that instead of the proposed API fetch

It seems twitch is doing some sort of weird partial page update, if you try to open a video directly: https://www.twitch.tv/videos/1534401704 you'll see the twitter:description tag is not set to the channel bio, even if you navigate back to the about section or channel home.

I've found another api ViewerFeedback_Creator that should work everywhere, however it needs the channel ID and i haven't found a way to obtain it from the streams pages yet.

riccardobl:
I've found a possible solution that should work everywhere

src/extension/content-script/batteries/Twitch.ts
```diff
@@ -0,0 +1,174 @@
...
+          clientId,
+          1,
+          "ChannelVideoCore",
+          "cf1ccf6f5b94c94d662efec5223dfb260c9f8bf053239a76125a58118769e8e2",
+          {
+            videoID: video,
+          }
+        )
+      )[0];
+      if (
```
lujakob, line 89: you could simplify this condition with Typescript optional chaining. See:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

src/extension/content-script/batteries/Twitch.ts
```diff
@@ -0,0 +1,174 @@
...
+          clientId,
+          1,
+          "ChannelShell",
+          "580ab410bcd0c1ad194224957ae2241e5d252b2c5173d8e0cce9d32d5bb14efe",
+          {
+            login: channel,
+          }
+        )
+      )[0];
+      if (channelData && channelData.data && channelData.data.userOrError) {
```
lujakob, line 111: Also could apply optional chaining here.

bumi:
yay, thanks for the PR! 🚀 
@im-adithya you know the batteries well, can you do a review?

im-adithya:
Nice PR! Thanks @riccardobl 🚀 

Just a few suggestions:
- Please rebase (to remove conflicts) and use the common `findLightningAddressInText` function from the helpers file.
- Optional chaining wherever possible (as @lujakob suggested above)
- And please separate the handling of the videos and channels pages to increase readability (you can take the GitHub battery as an example)!

Also, is it possible to make an `extractClientId` function during the start of the battery and proceed only if we get one (and pass it on everywhere we make an API call)? I was thinking of avoiding the usage of browser storage. (it's not worth caching if we can find it by looping through the script tags anyways(?))

im-adithya:
@riccardobl are you still working on this?

im-adithya (approved):

PR #1019, opened 2022-06-09
Instagram support

riccardobl:
A lot of artists use instagram to show their work, so i figured this might be an important addition. 

Parsing instagram wasn't very easy, since they generate most of their dom with random classes and no id, the selector used are pretty weird, but i tried to make them as solid as possible. 

Also note: when there are two different selectors for the same thing is because the layout changes when the window width goes below a certain value, so i had to fetch both.


reneaaron:
We have removed support for 3rd party integrations in the meantime 😞

PR #1028, opened 2022-06-10
E2ebattery

riccardobl:
This PR extends e2e tests to detect when batteries break.
I provided my own test links for now, they should be eventually replaced by something that is under the control of the core team.

Current coverage:
- [x] Twitter
- [x] Twitch
- [x] Instagram *
- [x] Github repo description
- [x] Github repo readme
- [x] Github user bio
- [ ] Github user readme 
- [x] Github repo to user bio fallback
- [ ] Github repo to user readme fallback
- [ ] Medium
- [ ] Monetization
- [ ] Peertube
- [ ] Podcastindex
- [ ] Vimeo
- [ ] Youtube Channel
- [ ] Youtube Video
 
<sub>**it seems IG hates any form of automated interaction, so i had to add some random delays to make it somewhat reliable. Also, since it limits the interactions for anonymous users, the script tries to login with the provided* **ALBY_E2E_INSTAGRAM_USERNAME** and **ALBY_E2E_INSTAGRAM_PASSWORD**
 *env variables, if they are set. It is better to use a throw away account here, since it might trigger some anti-bot filter , and it might be against their ToS (i haven't actually checked...)....*</sub>

The missing tests are mostly due to the fact that i need to setup an account for each of them, so if you happen to know the public link to an account that is already configured (or you want to make a new one...), feel free to share it here and i will look into adding it to the missing tests.

Note: i merged my other PRs https://github.com/getAlby/lightning-browser-extension/pull/1019 ,  https://github.com/getAlby/lightning-browser-extension/pull/1017 , https://github.com/getAlby/lightning-browser-extension/pull/1014 in this PR to write the tests for them. I will rebase the code to whatever the current master is before the merge.

### DIFF
--- a/src/extension/content-script/batteries/GitHub.ts
+++ b/src/extension/content-script/batteries/GitHub.ts
@@ -1,77 +1,125 @@
+import { Battery } from "~/types";
+
 import getOriginData from "../originData";
 import setLightningData from "../setLightningData";
 
-const urlMatcher = /^https:\/\/github.com\/([^/]+)(\/([^/]+))?$/;
+const urlMatcher = /^https:\/\/github.com\/([^/]+)(\/([^/]+))?.*$/;
+const validationRegex = /^[a-z0-9_.-]+$/i;
 
-const battery = (): void => {
+const battery = async (): Promise<void> => {
   const urlParts = document.location.pathname.split("/");
   const username = urlParts[1];
   const repo = urlParts[2];
 
+  if (username && !validationRegex.test(username)) return; // invalid username
+  if (username && repo && !validationRegex.test(repo)) return; // invalid repo
+
+  let lightningData = null;
+
+  // Search lightning data in the repository if the current page is a repository page
   if (username && repo) {
-    handleRepositoryPage(username);
-  } else if (username) {
-    handleProfilePage();
+    lightningData = await getLightningDataFromRepository(username, repo);
   }
+
+  // Search lightning data in the user profile or organization profile if the current page is a profile page
+  // or if the current page is a repository page but the lightning data is not set at repository level
+  if (!lightningData && username) {
+    lightningData = await getLightningDataFromProfile(username);
+  }
+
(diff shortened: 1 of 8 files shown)


riccardobl:
the debug=true hardcoded variables here
https://github.com/getAlby/lightning-browser-extension/blob/80d279fb71bad895874cdee8d0cf18570a438d67/src/extension/content-script/index.js#L36-L37
and  here
https://github.com/getAlby/lightning-browser-extension/blob/80d279fb71bad895874cdee8d0cf18570a438d67/src/extension/content-script/setLightningData.ts#L6-L7
are going to be replaced with some code that check if the extension is in production or development, once i figure out how to do it

riccardobl:
I've changed the code to disable the instagram test  if ALBY_E2E_INSTAGRAM_USERNAME or ALBY_E2E_INSTAGRAM_PASSWORD are missing. 
This is to avoid breaking the CI in PRs where gh secrets are not shareable anyway.

escapedcat:
This is really cool! I'm a bit worried this is getting a big too big for just one PR.
Would you mind creating one PR with the general and just one battery test. Including some documentation how everything is supposed to work? This would be super helpful for others and we could focus on the general setup first and then move in more tests with follow-up PRs?

riccardobl:
I can, but it isn't so big actually, 90% of the commits here are merges of my other PRs (as they add batteries..) maybe we should wait for see if they get rejected or merged first.

And i still haven't figured out why this code timeouts on gh action while it works fine locally

riccardobl:
Thank you again for helping me solve the timeout issue @escapedcat . I moved the general implementation to this PR https://github.com/getAlby/lightning-browser-extension/pull/1041 and I will refactor this one to extend the test coverage on top of #1041

im-adithya:
Oops! Didn't notice this! But I think #1077 will help in the GitHub Repo Readme and Profile Readme fallback as we can't fetch it from API.

reneaaron:

PR #1037, opened 2022-06-13
fix: url should include port number

dylancom:

Use host instead of hostname to include the port number.
This prevents `GET https://umbrel.local/v1/getinfo net::ERR_CONNECTION_REFUSED` errors


dylancom:
@bumi Yes, also updated other connectors.

bumi:
aaaand.. now that we got the e2e test we should add a test for that. 
that is pretty much the same as for LND but with a lndconnect string. the rest is exactly the same.

bumi:
great, let's do the same for start9 and mynode and then we do a patch release.

bumi (approved):

PR #1041, opened 2022-06-14
e2e tests for batteries

riccardobl:
This PR extends e2e tests to detect when batteries break.
As suggested, this PR covers just the basic implementation with a test only for the *monetization* battery. The tests for the other batteries will be added by this other PR https://github.com/getAlby/lightning-browser-extension/pull/1028 at a later time.

EDIT: i've added some code to test if the extension is built and also to skip the e2e batteries test (with a warning) if the extension is built only in production mode. This is to avoid breaking current scripts in case they are expecting to test the extension by building only the production mode.


tests/e2e/helpers/loadExtension.ts
```diff
@@ -1,16 +1,35 @@
...
+      return false;
+  }
+}
+
+/**
+ * Load the extension
+ * @param production specify if you want to load the production or development mode
+ */
+export const loadExtension = async (production?) => {
+  if(typeof production=="undefined")production=process.env.CI;
```
escapedcat, line 24: Uhm, I get what you are doing here but it's confusing to me.  
Is there any way, we could make this simpler?  

I believe tests should run against a prod version of the code. In this case we need to "inject" special test relevant mocking code.
Is it possible to i.e. do a production build but inject that code as well? Maybe instead of using `process.env.NODE_ENV` something like `process.env.MOCK_E2E_CODE`. Just an idea.

tests/e2e/testBatteries.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+ */
+test.describe("Test Batteries", () => {
+  for (const [unitName, unitData] of Object.entries(Units)) {
+    test(unitName, async () => {
+      let page;
+
+      try{
+        ({ page,  } = await loadExtension(false));        
+      }catch(e){ // continue without failing the test if the extension is not found
+        console.warn("Batteries test cannot continue:",e);
```
escapedcat, line 136: In the error case `.error` can be used.

src/extension/content-script/setLightningData.ts
```diff
@@ -3,6 +3,15 @@ import utils from "~/common/lib/utils";
 import { Battery } from "~/types";
 
 const setLightningData = (data: [Battery]): void => {
+  // Pass lightningData to e2e tests
+  if (process.env.NODE_ENV == "development") {
```
escapedcat, line 7: This will also be triggered when just developing locally? If yes, can we try to avoid it?

tests/e2e/testBatteries.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+        await button.click();
+        if (selectionData.waitNavigator) {
+          await page.waitForNavigation({ waitUntil: 'networkidle2' })
+        } else {
+          await delay(100);
+        }
+        rr();
+        return;
+      }
+      console.log("Looking for button ...", selectionData);
```
escapedcat, line 65: Better use `.info` instead og `.log`. Log always looks like forgotten debug code.  
Same for other usages.

escapedcat:
❤️  First up, thanks for your awesome work! 🚀   
Takes some time to wrap my head around it.  

So far I did not try it locally just yet.  
I made some general questions-kinda comments first. Have a look and let me know if something isn't clear please.

Further general points:

### Linting 

Looks like we should add the `tests` directory to these npm-scripts:
```
"lint:js": "eslint src --ext .js,.jsx,.ts,.tsx --max-warnings 0",
"lint:js:fix": "eslint src --ext .js,.jsx,.ts,.tsx --fix",
"format": "prettier --check 'src/**/*.(js|ts|jsx|tsx)'",
"format:fix": "prettier --loglevel silent --write 'src/**/*.(js|ts|jsx|tsx)'",
```
To make sure we have an aligned format/linting throughout the important code.
Could you give that a try?  
If it's too complicated to fix upcoming issues let me know. I can support with that.

### JS doc

We don't really need to jsdoc comments. We have TypeScript for this.

.github/workflows/playwright.yml
```diff
@@ -30,6 +32,11 @@ jobs:
         with:
           cmd: build:chrome # will run `yarn build` command
 
+      - name: Build Dev
```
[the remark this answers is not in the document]
riccardobl: yes, since the hooks to get the lightning data is added only in development mode to avoid exposing internal apis to random pages in production

src/extension/content-script/setLightningData.ts
riccardobl: Yes it will. I don't see why it should be avoided since this is part of the testing that is useful also to test local changes

tests/e2e/helpers/loadExtension.ts
riccardobl: That would make a third build that is an unmarked production build with exposed apis, i think it would be more confusing.

tests/e2e/testBatteries.spec.ts
riccardobl: should be resolved

tests/e2e/testBatteries.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+  if (unitData.cookieButtonSelector) await clickButton(page, unitData.cookieButtonSelector);
+
+  const lightningData = (await getLightningData(page) ?? [])[0];
+  console.info("Received lightning data", lightningData)
+  if (!lightningData) throw `Lightning data not found for unit ${unit} :(`;
+  for (let k in unitData.output) {
+    const v = unitData.output[k];
+    if (typeof v === "object") {
+      if (v.startsWith) {
+        if (!lightningData[k].startsWith(v.startsWith)) throw "Invalid " + k + ". Got " + lightningData[k] + " but " + v.startsWith + "... was expected";
```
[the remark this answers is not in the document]
riccardobl: Right, i tried to use template literals everywhere but i've missed some places. I am normally a java developer and we like to have a lot of + in our strings. Should be solved

tests/e2e/testBatteries.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+ *        $: "selector", // optional, "button" will be used if not specified 
+ *        innerText: "text content of the button", // optional
+ *        waitNavigator: true, // wait for page to be idle after clicking
+ *        scrollDown: true // scroll the page down before clicking (this is used eg. for the cookies notice)
+ *      }
+ */
+async function clickButton(page, selectionData) {
+  const selector = typeof selectionData == "object" && selectionData.$ ? selectionData.$ : "button";
+  const start = Date.now();
+  return new Promise<void>((rr) => {
```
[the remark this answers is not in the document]
riccardobl: replaced with resolve()

riccardobl:
Thanks for the review. 

# Linting
I am not sure how to fix the linting and formatting since i've never used these tools. 
But since this is not related to this PR it think it should be fixed with a separated commit by someone who knows how, i can rebase the pr on top of that.

# JS doc
I've been asked to comment my code, how am i supposed to do it then?

src/extension/content-script/setLightningData.ts
escapedcat: > also to test local changes

Can you document how this would be done? I.e. here: https://github.com/getAlby/lightning-browser-extension#e2e-tests-via-playwright-using-testing-library

- add new headline
- explain that we have special code in `development` to test batteries
- how to use this locally

For people looking a this the first time it might be hard to understand how and when to use this.

escapedcat (changes requested):

tests/e2e/helpers/loadExtension.ts
escapedcat: What do you mean with "unmarked"?  
The reason I prefer prod builds for E2E testing is that some issues only appear once the code is built.

escapedcat:
> I am not sure how to fix the linting and formatting since i've never used these tools. But since this is not related to this PR it think it should be fixed with a separated commit by someone who knows how, i can rebase the pr on top of that.

Added eslint/prettier. Please have a look and see if everything still works as expected.

> I've been asked to comment my code, how am i supposed to do it then?

I think it's less about the function attributes but on how to actually use these special injected mock-functions during development. [As mentioned here](https://github.com/getAlby/lightning-browser-extension/pull/1041#discussion_r897527361). You can remove the jsdocs comments.  

You did an awesome job with this PR coming up with a special idea on how to tackle battery testing. I'm worried if the concept of you idea is not well documented it will be hard to maintain or hard to be used by others.

I can also give it a try, extend the README and you check my draft and correct it. What do you think?

src/extension/content-script/setLightningData.ts
riccardobl: Are you sure i should document it in the readme? The only meaningful way to use this is with the [getLightningData](https://github.com/getAlby/lightning-browser-extension/pull/1041/files#diff-800d7b264f1edb5e8353d2c7aac0efb98b2169ed6709e7b0cc2fdb89f09acea4R122-R137) in testBatteries.spec.ts, and it is just the standard js cross-origin communication that is used by the extension everywhere to communicate with the background scripts.

To add a test here you only need to change [the Unit object on the start of the script](https://github.com/getAlby/lightning-browser-extension/pull/1041/files#diff-800d7b264f1edb5e8353d2c7aac0efb98b2169ed6709e7b0cc2fdb89f09acea4R4-R32) there is no need to change the code normally.

I think you are misunderstanding this as a way for the developer to debug the lightning data in realtime during the development but this is limited to this specific test. If not, do you have an idea on how to document this in a way that is not confusing for someone that is not interested in the internal functions working?

tests/e2e/helpers/loadExtension.ts
riccardobl: Only testBatteries.spec.ts will run in development mode, i think it is a decent compromise, the extension is tested in production while the batteries in development mode. 

If we want to make things proper we should have a third build that is for *testing*  with production code+testing hooks, but since it requires some extensive change to the build script, i think it should probably go in its own PR, and then "testing" can be used instead of "development" . 
With unmarked i mean there would be two types of production builds, the one that is really production and the one with injected code and they will be both marked as a production, i fear this might lead to testing code being leaked into code that is expected to be production only.

tests/e2e/testBatteries.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+}
+
+/**
+ * Retrieve the lightning data of the page by querying the extension
+ * @param page page object to query
+ * @returns the lightning data
+ */
+const getLightningData = async (page): Promise<void> => {
```
escapedcat, line 122: My `void` return type is wrong. Maybe @dylancom has an idea what this should be.

riccardobl:
> I can also give it a try, extend the README and you check my draft and correct it. What do you think?

Sure, fine by me.  Imo this is not needed since it is an internal api [as i commented here](https://github.com/getAlby/lightning-browser-extension/pull/1041#discussion_r897580792), but extra documentation doesn't hurt, it is just that i am not sure how to write it (i am bad with documentation).

tests/e2e/testBatteries.spec.ts
riccardobl: Oh, right... it should probably be 'any'

escapedcat: Yeah, it was `any` before. Would like to avoid `any`. But if we don't have a choice we can use it, comment it and turn of linting for this line.

tests/e2e/helpers/loadExtension.ts
escapedcat: Got it. I'm ok with rethinking this in a later PR if the others are ok with it.  
If we create a new npm-script to create a prod build with injected mock-code I'm ok with that. Only `yarn build` will ever be used for the acutal published extension.

tests/e2e/testBatteries.spec.ts
riccardobl: Mh, the actual type should be Battery from import {Battery} from '~/types', but this file is not in the test source path. We could redefine a type that is a subset of battery but that will add extra maintenance cost, leaving it as any is not too bad imo

reneaaron:

Issue #989, opened 2022-06-03
TS: Clean up and generalize types for messages (focus: allowance)


escapedcat:
### Is your feature request related to a problem? Please describe

Similar to #652 but focusing on the allowance section

### Describe the solution you'd like

All files should be switched to TS

### Additional context

Add tests if possible/usefull

- [x] add.js
- [x] delete.js
- [x] enable.js
- [x] get.js
- [x] getById.js
- [x] list.ts
- [x] update.js

### Are you working on this?

Yes

escapedcat:
Thanks @im-adithya !

PR #1044, opened 2022-06-14
refactor(allowance): add action

im-adithya:

Switching allowance/add to TS

### Link this PR to an issue

Relates to #989

### How has this been tested?

Should add tests, would be done after initial review

### Checklist


src/extension/background-script/actions/allowances/add.ts
```diff
@@ -12,7 +14,7 @@ const add = async (message, sender) => {
     .first();
 
   if (allowance) {
-    await db.allowances.update(allowance.id, {
+    await db.allowances.update(allowance.id as number, {
```
escapedcat, line 17: It could be `undefined`, yes? Could we just return if there's no `id`? Would be great if we can avoid `as` because it's not good practice.

escapedcat (commented):
🎉

src/extension/background-script/actions/allowances/add.ts
im-adithya: Done!

escapedcat (approved):

src/extension/background-script/actions/allowances/delete.ts
```diff
@@ -3,6 +3,7 @@ import type { MessageAllowanceDelete } from "~/types";
 
 const deleteAllowance = async (message: MessageAllowanceDelete) => {
   const id = message.args.id;
+  if (!id) return;
```
bumi, line 6: this probably should return an error like `{ error: "id is missing" }`

to make sure that all actions always return the same response format

src/types.ts
```diff
@@ -321,13 +332,13 @@ export interface PaymentResponse
 }
 
 export interface Allowance {
-  createdAt: string;
```
escapedcat, line 324: bumi and me just talked about this..
```
createdAt: string;
lastPaymentAt: string;
lnurlAuth: string;
```
This is how it's saved in the index-db of most users we assume.
So changing this to "correct" types looks great and works in this scenario but in reality this might break because i.e. `lnurlAuth` will be `"true"` and not `true`.  

We should first make sure we cover the real world types which are being used currently before we change them. If we change them we might also need to migrate the index-db for the users.

Is this understandable? You think this is an issue? Happy to talk via Slack.

im-adithya: Got it! Reverting the changes,
> saved in the index-db of most users we assume

didn't know this can be a breaking change!

escapedcat (approved):

escapedcat (dismissed):

src/extension/background-script/actions/allowances/add.ts
```diff
@@ -28,6 +32,14 @@ const add = async (message, sender) => {
       totalBudget: totalBudget,
       remainingBudget: totalBudget,
       lastPaymentAt: 0,
+      createdAt: Date.now().toString(),
```
bumi, line 35:
```suggestion
      createdAt: Date.now(),
```
for the payments we use numbers. 
and I think we should keep the timestamp here as a number.

src/extension/background-script/actions/allowances/add.ts
```diff
@@ -28,6 +32,14 @@ const add = async (message, sender) => {
...
+      lnurlAuth: false,
+      payments: [],
```
bumi, line 37: I don't think this can go in the DB. there is no payments array in the DB.
same is for paymentsAmount, paymentsCount, percentage, usedBudget

bumi (changes requested):

src/extension/background-script/actions/allowances/add.ts
im-adithya: But wouldn't this break for existing users? cc @escapedcat

src/extension/background-script/actions/allowances/add.ts
im-adithya: I didn't get this... this would be done for new allowances, so this shouldn't break right?

src/extension/background-script/actions/allowances/add.ts
bumi: seems we do not save this so far. At least I did not see it and this change is added in this PR

src/extension/background-script/actions/allowances/add.ts
bumi: those are the fields of the allowances table:https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/db.ts#L15

`allowances.add` is a dexie function that stores data in that table. 

seems we use an "Allowance" type that does not represent the data of the DB but the "allowance"... 
but now when calling `allowances.add` we pass in all the Allowance attributes - and not just the one that are relevant for the DB.

I don't know what dexie does with that, but to me that's wrong.

im-adithya: Oh! Got it now! I was thinking we don't save the above things just to save space while working in get action but didn't know we don't even allow them all together in the DB!

im-adithya: I guess we can make the types optional `?:` that would make sense as we are using those only when needed.

im-adithya: > but I'd say we should use different types a special type for the data stored in the DB

That would make things easier actually...

(Waiting for feedback from them before I proceed)

escapedcat: I agree with bumi here. I wasn't aware that the allowances we save in dexie ware not the same that we use withing the app. Unlike blocklist and payments. I split up the type and adjusted the code. Please have a look at this commit:
https://github.com/getAlby/lightning-browser-extension/pull/1044/commits/e4a1fe5d4047eb22c383f9873d618eae82668713

src/extension/background-script/actions/allowances/add.ts
escapedcat: Looking at the current code it should be number and won't break things because it was always a number I believe

src/types.ts
bumi: I did not find code that writes the createdAt so far. https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/allowances/add.js#L23-L31

ideally the timestamp is a number (like lastPaymentAt)

bumi (approved):

escapedcat:
YOLO 🤞

PR #1045, opened 2022-06-14
refactor(allowance): delete action

im-adithya:

Switching allowance/delete to TS

### Link this PR to an issue

Relates to #989

### How has this been tested?

Should add tests, would be done after initial review

### Checklist


escapedcat:
Would be nice to add a unit-test.

src/extension/background-script/actions/allowances/delete.ts
```diff
@@ -0,0 +1,12 @@
+import type { MessageAllowanceDelete } from "~/types";
+
+import db from "../../db";
```
escapedcat, line 3: Can we use `~` as well here?

src/extension/background-script/actions/allowances/delete.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const deleteAllowance = async (message: MessageAllowanceDelete) => {
+  const id = message.args.id;
+  await db.allowances.delete(parseInt(id));
```
escapedcat, line 7: Allowances `id` per type definitions are `string` but here we need to switch it to a `number`? Not sure if I understand.

escapedcat: Maybe have a look at this: https://github.com/getAlby/lightning-browser-extension/pull/988/files#diff-1b329da99ce24a934532ca1ee401d135a5d5917619b62db9711da859ce47f987R64-R67  
not sure what's correct really.

bumi: why do we need those parseInt() 

can we get rid of those parseInt and number/string checks? those are primary keys and we should know what type it is and be able to use the correct type?

escapedcat: I think this came when switching to TS. Once account/allowance is switched let's have a look if we can assume that `id` is always a string and create a PR.

escapedcat (approved):
❤️

src/extension/background-script/actions/allowances/__tests__/delete.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        internal: true,
+      },
+      args: {
+        id: 2,
+      },
+    };
+
+    await db.allowances.bulkAdd(mockAllowances);
+
+    expect(await deleteAllowance(message)).toStrictEqual({
```
bumi, line 65: also here, let's test that the allowance is actually deleted from the db.

bumi (changes requested):

bumi (approved):

PR #1049, opened 2022-06-15
Test/allowance list

escapedcat:
✅  ~~Issue:~~

```sh
  ● account all › get all accounts

    TypeError: _db.default.allowances.toCollection is not a function

       6 |
       7 |   const dbAllowances = await db.allowances
    >  8 |     .toCollection()
         |      ^
       9 |     .reverse()
      10 |     .sortBy("lastPaymentAt"
```

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -6,13 +6,13 @@ module.exports = {
   verbose: true,
   transform: {
     "^.+\\.[t|j]sx?$": "babel-jest",
-    "^.+\\.mjs?$": "babel-jest", // dexie needs this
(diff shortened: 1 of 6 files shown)


bumi:
no idea how related that is. but there are tools like https://www.npmjs.com/package/fake-indexeddb that might help with testing those dexie DB things

escapedcat:
> no idea how related that is. but there are tools like https://www.npmjs.com/package/fake-indexeddb that might help with testing those dexie DB things

Thanks for the reminder! Works like a charm!  
In this case `allowances` is mocked as `[]` which is ok for now. Couldn't find docs on how to mock specific entries.

bumi (approved):

PR #1053, opened 2022-06-16
chore: rename delete to remove

### LINKED ISSUE #955: Replace "Delete" account with "Unlink"  account
User can remove accounts in Alby when they click on "Delete". However, they might think that they really delete an account and not just unlink it what they actually do. 

Replace "Delete" account with "Unlink"  account

**Describe alternatives you've considered**
"Remove" Account

**Additional context**
image

No

im-adithya:

Renames delete to remove and updates the alert message.
Follows this https://github.com/getAlby/lightning-browser-extension/issues/955#issuecomment-1153553635

### Link this PR to an issue

Closes #955

- `chore`: Renaming a variable

Screenshot from 2022-06-16 15-42-15

### How has this been tested?

Ran all tests

### Checklist

### DIFF
--- a/src/app/screens/Accounts/index.tsx
+++ b/src/app/screens/Accounts/index.tsx
@@ -50,14 +50,18 @@ function AccountsScreen() {
   }
 
   async function removeAccount({ id, name }: AccountAction) {
-    if (window.confirm(`Are you sure you want to delete account: ${name}?`)) {
+    if (
+      window.confirm(
+        `Are you sure you want to remove account: ${name}? \nThis can not be undone. If you used this account to login to websites you might loose access to those.`
+      )
+    ) {
       let nextAccountId;
       let accountIds = Object.keys(accounts);
       if (auth.account?.id === id && accountIds.length > 1) {
         nextAccountId = accountIds.filter((accountId) => accountId !== id)[0];
       }
 
-      await api.deleteAccount(id);
+      await api.removeAccount(id);
       accountIds = accountIds.filter((accountId) => accountId !== id);
 
       if (accountIds.length > 0) {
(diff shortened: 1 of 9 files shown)


src/app/screens/Accounts/index.tsx
```diff
@@ -50,14 +50,18 @@ function AccountsScreen() {
   }
 
   async function removeAccount({ id, name }: AccountAction) {
-    if (window.confirm(`Are you sure you want to delete account: ${name}?`)) {
+    if (
+      window.confirm(
+        `Are you sure you want to remove account: ${name}? \nThis can not be undone. If you used this account to login to websites and you might loose access to those.`
```
bitcoinuser, line 55: Needs the (And)?

im-adithya: Oops! Fixed it!

im-adithya:
Ready!

bitcoinuser:
I changed my mind. I think it's better use the conceit of Archive account. So you will always be able to undo the action. Same thing of what Blockstream Green wallet does.

What do you think?

im-adithya:
Or else we can combine both of these and could give an option to "archive instead?" below in a modal component instead of a native alert...

bitcoinuser:
Are you sure you want to archive the account: main?
If you used this account to login on sites using Alby, you might not be able to access them.
This action always can be undone.

bitcoinuser:
> Or else we can combine both of these and could give an option to "archive instead?" below in a modal component instead of a native alert... As remove option should always be there for the user at any point of time, wdyt?

I think two options is confusing. Just my opinion.

escapedcat:
Bumi approved this so I think we're good here for now.  
You can always re-add an account. "Archieve" sounds like I can see a list of archived accounts somewhere. Or it would be "greyed out".
Let's go with remove for now. Let's move further discussions to the [design slack channel](https://bitcoindesign.slack.com/archives/C02591ADXM2).

Issue #1051, opened 2022-06-16
Update extension manifest from v2 to v3


escapedcat:
### Is your feature request related to a problem?

Update extension [to new manifest version](https://github.com/getAlby/lightning-browser-extension/blob/master/src/manifest.json#L2) to make sure it will not be an issue to be listed in Chrome the extension store in the future.

### Describe the solution you'd like

Extension ships with v3 manifest

### Describe alternatives you've considered

none

### Additional context

> They've also put out a new spec that extensions need to conform to (Manifest v3)

[_Add any other additional context or screenshots about the feature request here_](https://github.com/joule-labs/joule-extension/issues/311#issuecomment-1156823457)

### Are you working on this?

No

PR #1061, opened 2022-06-19
Update manifest.json

secondl1ght:

Update manifest version.

### Link this PR to an issue

#1051 

### Checklist

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -1,5 +1,5 @@
 {
-  "manifest_version": 2,
+  "manifest_version": 3,
   "name": "Alby - Bitcoin Lightning Wallet",
   "version": "0.0.0",
 


secondl1ght:
`npx playwright test` is failing here on GitHub but I ran it locally and all the tests passed - looks like it is timing out here.

bumi:
careful: afaik v3 defines a completely different architecture of the extension. 
there is no background-script anymore, but it introduces service-workers. 

https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/

bumi:
actually let's close this one. because it will be open for long and changing the number will be easily done when we know what exactly needs to be done. 

did you actually run and use Alby with this change?

secondl1ght:
Sounds good - thanks for the heads up, yes I ran it locally and it appeared to be working (no errors in browser console or backpage console). I didn't run through all the functionality though.

Issue #960, opened 2022-05-28
LNDHub does not re-authenticate after an access_token is expired


bumi:
**Describe the bug**
lndhub does API calls with an access_token. This access_token is initially created from login/password. 
When an API call fails with an bad auth error we refresh the token and retry the call. 
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/lndhub.ts#L359-L362

This currently does NOT work for Alby wallets. Alby wallets return a HTTP status 401 (the correct HTTP status for such errors) But in our connector code we only do the retry if the response was a HTTP 200 (which is what the bluewallet lndhub returns).

we need to update the code that it retries in both cases.

**To Reproduce**
You need an expired access_token. This happens after 1 week.

**Information About Alby**
- Alby Version: 1.10.0
- Alby installed through: chrome
- Wallet connected with Alby: BlueWallet's LndHub vs. Alby's LndHub

PR #1072, opened 2022-06-23
fix(lndhub): expired alby token #960

escapedcat:

Takes care of expired Alby tokens

### Link this PR to an issue

#960

### How has this been tested?

Using https://www.charlesproxy.com  
Old school, but worked :D

### Checklist

### DIFF
--- a/src/extension/background-script/connectors/lndhub.ts
+++ b/src/extension/background-script/connectors/lndhub.ts
@@ -29,6 +29,12 @@ interface Config {
   url: string;
 }
 
+const defaultHeaders = {
+  Accept: "application/json",
+  "Access-Control-Allow-Origin": "*",
+  "Content-Type": "application/json",
+};
+
 export default class LndHub implements Connector {
   config: Config;
   access_token?: string;
@@ -50,7 +56,7 @@ export default class LndHub implements Connector {
   }
 
   async getInfo(): Promise<GetInfoResponse> {
-    const data = await this.request<{ alias: string }>(
+    const { alias } = await this.request<{ alias: string }>(
       "GET",
       "/getinfo",
       undefined,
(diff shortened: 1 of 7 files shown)


src/extension/background-script/connectors/lndhub.ts
```diff
@@ -353,15 +352,31 @@ export default class LndHub implements Connector {
...
+
         if (errResponse?.status === 404) {
           const method = path.replace("/", "");
           throw new Error(`${method} not supported by the connected account.`);
         }
+
+        if (errResponse?.status === 401) {
+          console.warn(
+            "TOKEN MIGHT BE EXPIERED, LET'S TRY TO GET YOU LOGGED IN"
+          );
```
bumi, line 364:
```suggestion
```

as this is a normal case I would not log anything here. (we also did not use all-caps messages so far).
if we want to log something we could do a console.info() in the authorize() function?

src/extension/background-script/connectors/lndhub.ts
```diff
@@ -375,9 +390,11 @@ export default class LndHub implements Connector {
         throw new Error(data.message);
       }
     }
+
     if (defaultValues) {
-      data = Object.assign(Object.assign({}, defaultValues), data);
```
bumi, line 379: those default values are left-overs from joule, there default values had been used. we don't really do that I think.

bumi (approved):
utACK.

src/extension/background-script/connectors/lndhub.ts
escapedcat: Removed

src/extension/background-script/connectors/lndhub.ts
escapedcat: Removed it all connectors where it's not used. [Only connector were it's used is lnd](https://github.com/getAlby/lightning-browser-extension/blob/fix/960_expired-lndhub-alby-token/src/extension/background-script/connectors/lnd.ts/#L199-L205).

PR #1073, opened 2022-06-23
feat: extend webln sendPayment method

pavanjoshi914:

1. add metadata as an optional parameter to the  ```SendPayment``` function

2. code works when metadata is passed or not passed

### How has this been tested?
Tested by making payments with and without passing metadata from the client-side using prototype

image

### Checklist

### DIFF
--- a/src/extension/ln/webln/index.ts
+++ b/src/extension/ln/webln/index.ts
@@ -59,11 +59,11 @@ export default class WebLNProvider {
     return this.execute("lnurl", { lnurlEncoded });
   }
 
-  sendPayment(paymentRequest: string) {
+  sendPayment(paymentRequest: string, metadata?: string) {
     if (!this.enabled) {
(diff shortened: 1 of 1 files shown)


pavanjoshi914:
@bumi ?

bumi:
I think we need to propose this change to the webln spec, ideally other webln wallets will also implement that.
So I think a first step is to write a proposal to the spec for this. This means need to explain the idea and reason for this and also specify what can be used as metadata. 
Did you already write something on this?

pavanjoshi914:
> s means need to explain th

https://gist.github.com/pavanjoshi914/53f28f9d40ed3bc9f3028a2117a417a4 
If this looks good to you! I can finalize and upload this as a new spec-proposal for webln

pavanjoshi914:
@bumi  can you review that gist?

pavanjoshi914:
@bumi

Issue #389, opened 2021-11-17
Show "receive transactions" in the transaction list of the popup and the "website" menu


MoritzKa:
Currently Alby does not show WebLN receive transactions in the transaction lists (of the pop up and the "⚡️ Websites")

We only show sending transactions made from Alby because of the metadata that we add.

To be considered here:
- ~show an entry when a website requests an invoice (same as when the website sends a payment) (webln.makeInvoice vs. webln.sendPayment)~
- ~sync the data with the transactions from the node (did we actually receive the payment?)~
- get (latest n) incoming transactions from the node via API call

bumi:
I think somehow this should be the full audit system that logs every action a user made on Alby.

MoritzKa:
confirmed.

dvoroneca:
Hi Guys, here is my take on how the transactions list, including the incoming transactions. 
Feel free to leave comments in [figma](https://www.figma.com/file/kdkEo1fpGvaiMifkfZQQjN/Recent-Transactions?node-id=0%3A1). 

There are versions that need feedback. 

1. Updated open transaction view V1 and V2.
2. Incoming, Outgoing, and Invoice expended views.

MoritzKa:
According to #https://github.com/getAlby/meta/issues/15

tasks:
- API call to get incoming transactions
- design tab for incoming transaction

dvoroneca:
As an outcome of the refinement meeting, here are the mockups for the tabs in the recent transactions lists on the "Transactions list" and transaction history on the "Contextual Screen"

![transactions](https://user-images.githubusercontent.com/35034985/174791888-e552b97c-fd95-4133-a2aa-0fb6a21a079e.png)

-----------

## Transaction History:

![List-1](https://user-images.githubusercontent.com/35034985/174792207-94466786-cace-46c5-87f0-bd113a9cf1e3.png)
![List](https://user-images.githubusercontent.com/35034985/174792215-6a65d213-5ca7-43b8-b0d5-2d574195ffde.png)

-------------------

## Opened view:
![open V3](https://user-images.githubusercontent.com/35034985/174792307-34c34b1a-a02f-4ab9-92bf-2febaef68956.png)

-------------------

## Contextual Screen:
![Extension Recent Transactions-1](https://user-images.githubusercontent.com/35034985/174792408-fbf85110-6474-4f20-a733-86fbbc970f46.png)
[...]

PR #1080, opened 2022-06-24
feat: add getInvoices #389

escapedcat:

- invoices call for LND
- invoices call for LNDHub
- invoice-tx list to home screen
- tabs to switch between incoming/outgoing txs
- types

#### Todos

- [x] translations
- [x] tests (partly done)
- [x] tab design

### Link this PR to an issue

#389

image

### How has this been tested?

- so far only manually

### Checklist

### DIFF
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -86,7 +86,7 @@ export default function TransactionsTable({ transactions }: Props) {
                   <Disclosure.Panel>
                     <div className="mt-1 ml-9 text-xs text-gray-600 dark:text-neutral-400">
                       {tx.description}
-                      <p>Fee: {tx.totalFees} sats</p>
+                      {tx.totalFees && <p>Fee: {tx.totalFees} sats</p>}
                       {tx.preimage && (
                         <p className="truncate">Preimage: {tx.preimage}</p>
                       )}
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -103,6 +103,7 @@ export default function TransactionsTable({ transactions }: Props) {
                           <li>app_name: {tx.boostagram.app_name}</li>
                           <li>name: {tx.boostagram.name}</li>
                           <li>podcast: {tx.boostagram.podcast}</li>
+                          <li>sender_name: {tx.boostagram.sender_name}</li>
                           <li>
                             message:{" "}
                             {tx.boostagram.message &&
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -103,7 +103,11 @@ export default function TransactionsTable({ transactions }: Props) {
                           <li>app_name: {tx.boostagram.app_name}</li>
                           <li>name: {tx.boostagram.name}</li>
                           <li>podcast: {tx.boostagram.podcast}</li>
-                          <li>message: {tx.boostagram.message}</li>
+                          <li>
+                            message:{" "}
+                            {tx.boostagram.message &&
+                              decodeURI(tx.boostagram.message)}
+                          </li>
                         </ul>
                       )}
                     </div>
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -49,12 +49,10 @@ export default function TransactionsTable({ transactions }: Props) {
                     </div>
                     <div className="overflow-hidden mr-3">
                       <div
-                        className={`
-                      text-sm font-medium text-gray-900 truncate dark:text-white ${
-                        !tx.title && "invisible"
-                      }`}
+                        className="
+                      text-sm font-medium text-gray-900 truncate dark:text-white"
                       >
-                        {tx.title}
+                        {tx.title || "\u00A0"}
                       </div>
                       <p className="text-xs text-gray-600 capitalize dark:text-neutral-400">
                         {tComponents(`transactionsTable.${tx.type}`)} -{" "}
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -54,7 +54,7 @@ export default function TransactionsTable({ transactions }: Props) {
                     </div>
                     {tx.badges && (
                       <div className="ml-6 space-x-3">
-                        {tx.badges.map((badge, i) => (
+                        {tx.badges.map((badge) => (
                           <Badge
                             key={badge.label}
                             label={badge.label}
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -8,7 +8,7 @@ import { Transaction } from "~/types";
 
 import Badge from "../Badge";
 
-type Props = {
+export type Props = {
   transactions: Transaction[];
 };
 
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -48,7 +48,12 @@ export default function TransactionsTable({ transactions }: Props) {
                       {tx.type && renderIcon(tx.type)}
                     </div>
                     <div className="overflow-hidden mr-3">
-                      <div className="text-sm font-medium text-gray-900 truncate dark:text-white">
+                      <div
+                        className={`
+                      text-sm font-medium text-gray-900 truncate dark:text-white ${
+                        !tx.title && "invisible"
+                      }`}
+                      >
                         {tx.title}
                       </div>
                       <p className="text-xs text-gray-600 capitalize dark:text-neutral-400">
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -76,11 +76,13 @@ export default function TransactionsTable({ transactions }: Props) {
                           ~{tx.totalAmountFiat}
                         </p>
                       </div>
-                      <Disclosure.Button className="block h-0 mt-2 text-gray-500 hover:text-black dark:hover:text-white transition-color duration-200">
-                        <CaretDownIcon
-                          className={`${open ? "rotate-180" : ""} w-5 h-5`}
-                        />
-                      </Disclosure.Button>
+                      {tx.type !== "received" && (
+                        <Disclosure.Button className="block h-0 mt-2 text-gray-500 hover:text-black dark:hover:text-white transition-color duration-200">
+                          <CaretDownIcon
+                            className={`${open ? "rotate-180" : ""} w-5 h-5`}
+                          />
+                        </Disclosure.Button>
+                      )}
                     </div>
                   </div>
                   <Disclosure.Panel>
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -52,7 +52,8 @@ export default function TransactionsTable({ transactions }: Props) {
                         {tx.title}
                       </div>
                       <p className="text-xs text-gray-600 capitalize dark:text-neutral-400">
-                        {tx.type} - {tx.date}
+                        {tComponents(`transactionsTable.${tx.type}`)} -{" "}
+                        {tx.date}
                       </p>
                     </div>
                     {tx.badges && (
@@ -112,21 +113,30 @@ export default function TransactionsTable({ transactions }: Props) {
                       {tx.boostagram && (
                         <ul>
                           <li>
-                            label={tComponents("transactionsTable.sender_name")}
+                            {tComponents(
+                              "transactionsTable.boostagram.sender_name"
+                            )}
                             : {tx.boostagram.sender_name}
                           </li>
                           <li>
-                            label={tComponents("transactionsTable.message")}:{" "}
+                            {tComponents(
+                              "transactionsTable.boostagram.message"
+                            )}
+                            :{" "}
                             {tx.boostagram.message &&
                               decodeURI(tx.boostagram.message)}
                           </li>
                           <li>
-                            label={tComponents("transactionsTable.app_name")}:{" "}
-                            {tx.boostagram.app_name}
+                            {tComponents(
+                              "transactionsTable.boostagram.app_name"
+                            )}
+                            : {tx.boostagram.app_name}
                           </li>
                           <li>
-                            label={tComponents("transactionsTable.podcast")}:{" "}
-                            {tx.boostagram.podcast}
+                            {tComponents(
+                              "transactionsTable.boostagram.podcast"
+                            )}
+                            : {tx.boostagram.podcast}
                           </li>
                         </ul>
                       )}
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -76,14 +76,14 @@ export default function TransactionsTable({ transactions }: Props) {
                           ~{tx.totalAmountFiat}
                         </p>
                       </div>
-                      {[tx.type && "sent", "sending"].includes(tx.type) ||
-                        (tx.type === "received" && tx.boostagram && (
-                          <Disclosure.Button className="block h-0 mt-2 text-gray-500 hover:text-black dark:hover:text-white transition-color duration-200">
-                            <CaretDownIcon
-                              className={`${open ? "rotate-180" : ""} w-5 h-5`}
-                            />
-                          </Disclosure.Button>
-                        ))}
+                      {([tx.type && "sent", "sending"].includes(tx.type) ||
+                        (tx.type === "received" && tx.boostagram)) && (
+                        <Disclosure.Button className="block h-0 mt-2 text-gray-500 hover:text-black dark:hover:text-white transition-color duration-200">
+                          <CaretDownIcon
+                            className={`${open ? "rotate-180" : ""} w-5 h-5`}
+                          />
+                        </Disclosure.Button>
+                      )}
                     </div>
                   </div>
                   <Disclosure.Panel>
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -4,6 +4,7 @@ import {
   CaretDownIcon,
 } from "@bitcoin-design/bitcoin-icons-react/filled";
 import { Disclosure } from "@headlessui/react";
+import { useTranslation } from "react-i18next";
 import { Transaction } from "~/types";
 
 import Badge from "../Badge";
@@ -13,6 +14,8 @@ export type Props = {
 };
 
 export default function TransactionsTable({ transactions }: Props) {
+  const { t: tComponents } = useTranslation("components");
+
   function renderIcon(type: string) {
     function getIcon() {
       const iconClasses = "h-3 w-3";
@@ -88,10 +91,18 @@ export default function TransactionsTable({ transactions }: Props) {
                   </div>
                   <Disclosure.Panel>
                     <div className="mt-1 ml-9 text-xs text-gray-600 dark:text-neutral-400">
-                      {tx.description}
-                      {tx.totalFees && <p>Fee: {tx.totalFees} sats</p>}
+                      <p>{tx.description}</p>
+                      {tx.totalFees && (
+                        <p>
+                          {tComponents("transactionsTable.fee")}: {tx.totalFees}{" "}
+                          sats
+                        </p>
+                      )}
                       {tx.preimage && (
-                        <p className="truncate">Preimage: {tx.preimage}</p>
+                        <p className="truncate">
+                          {tComponents("transactionsTable.preimage")}:{" "}
+                          {tx.preimage}
+                        </p>
                       )}
                       {tx.location && (
                         <a href={tx.location} target="_blank" rel="noreferrer">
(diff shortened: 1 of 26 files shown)


src/app/screens/Home/index.tsx
```diff
@@ -94,6 +100,33 @@ function Home() {
...
+    if (index === 1) {
+      await loadInvoices();
+    }
+  }
+
+  async function loadInvoices() {
+    setLoadingInvoices(true);
+    const result = await api.getInvoices();
+    const invoices: Transaction[] = result.invoices
+      .filter((invoice) => invoice.settled)
```
escapedcat, line 113: Do we want to display unpaid invoices?

src/extension/background-script/connectors/connector.example
```diff
@@ -34,7 +34,7 @@ class ConnectorExample implements Connector {
     // Add your own implementation.
   }
 
-  getBalance(): Promise<GetBalanceResponse> {
+  getBalance(): Promise<GetBalanceResponse> { // we could remove this
```
escapedcat, line 37: We never use this. `balance` is being taken via `getInfo`

src/app/components/TransactionsTable/index.tsx
```diff
@@ -76,26 +76,39 @@ export default function TransactionsTable({ transactions }: Props) {
...
+                      {tx.location && (
                         <a href={tx.location} target="_blank" rel="noreferrer">
                           {tx.location}
                         </a>
-                      ) : (
-                        ""
+                      )}
+                      {tx.boostagram && (
+                        <ul>
+                          <li>app_name: {tx.boostagram.app_name}</li>
```
escapedcat, line 103: This is very basic and not yer translated. What should we show here from the boostagram-data?

src/app/screens/connectors/NewWallet/index.tsx
```diff
@@ -13,6 +13,7 @@ import utils from "~/common/lib/utils";
 
 const walletCreateUrl =
   process.env.WALLET_CREATE_URL || "https://app.regtest.getalby.com/api/users";
+// process.env.WALLET_CREATE_URL || "https://getalby.com/api/users";
```
escapedcat, line 16: Remove

src/common/lib/utils.ts
```diff
@@ -170,6 +172,21 @@ const utils = {
         });
     });
   },
+  getBoostagramFromInvoice: (
+    custom_records:
+      | {
+          "696969": string;
```
escapedcat, line 178: What else can be in here? Do we need to care now?

MoritzKa:
I tried to add my getalby.com account but it shows zero sats
image

src/app/components/TransactionsTable/index.tsx
bumi: I think sender_name and message is the most important one. 

Is there a way we can add the rest in a popup or something? 

raw data in getalby.com
image

src/app/screens/Home/index.tsx
```diff
@@ -94,6 +100,33 @@ function Home() {
     }
   }
 
+  async function onTabChangeHandler(index: number) {
+    if (index === 1) {
+      await loadInvoices();
```
bumi, line 105: should we add a check there that this is only done if the invoices are NOT yet loaded?

Right now it always reloads when I change the tab

src/app/screens/Home/index.tsx
```diff
@@ -94,6 +100,33 @@ function Home() {
...
+
+    const invoices: Transaction[] = result.invoices
+      .filter((invoice) => invoice.settled)
+      .map((invoice) => ({
+        ...invoice,
+        title: invoice.memo,
+        date: dayjs(invoice.settleDate).fromNow(),
+      }));
+
+    for await (const invoice of invoices) {
```
bumi, line 121: what is this `await` here?

src/app/screens/Home/index.tsx
```diff
@@ -94,6 +100,33 @@ function Home() {
...
+  async function loadInvoices() {
+    setLoadingInvoices(true);
+    const result = await api.getInvoices();
+
+    const invoices: Transaction[] = result.invoices
+      .filter((invoice) => invoice.settled)
+      .map((invoice) => ({
+        ...invoice,
+        title: invoice.memo,
+        date: dayjs(invoice.settleDate).fromNow(),
```
bumi, line 118: somehow I would like that this data would be coming in a usable format from the action. that less filtering/transformation is done in the "frontend".
This would make it more re-usable.

maybe something to refactor in the future?

src/app/screens/Home/index.tsx
```diff
@@ -286,45 +321,88 @@ function Home() {
...
-                        rel="noreferrer"
-                      >
-                        {payment.name}
-                      </a>
+
+            <Tab.Group onChange={onTabChangeHandler}>
+              <Tab.List className="mb-2">
+                {Object.keys({
+                  [t("transaction_list.tabs.outgoing")]: true,
+                  [t("transaction_list.tabs.incoming")]: true,
```
bumi, line 331: why is this not? 
```
[ t("transaction_list.tabs.outgoing"), t("transaction_list.tabs.incoming")].map(...
```

src/app/screens/connectors/NewWallet/index.tsx
bumi: TODO

src/common/lib/utils.ts
bumi: I think the following are interesting for us to show: 

* 32768:  https://github.com/lightning/blips/blob/master/blip-0002.md#messages
* 7629169: https://github.com/lightning/blips/blob/master/blip-0010.md

I think others can be ignored/not shown.

@kiwiidb what do you say?

src/common/lib/utils.ts
```diff
@@ -170,6 +172,21 @@ const utils = {
         });
     });
   },
+  getBoostagramFromInvoice: (
```
bumi, line 175: the parameter that is passed in here is a custom_record object? 
If so I'd rename that function.

src/extension/background-script/connectors/lnd.ts
```diff
@@ -223,11 +227,82 @@ class Lnd implements Connector {
...
+      last_index_offset: string;
+      first_index_offset: string;
+    }>("GET", "/v1/invoices", undefined, {});
+
+    const invoices: Invoice[] = data.invoices
+      .map((invoice, index): Invoice => {
+        const custom_records =
+          invoice.htlcs[0] && invoice.htlcs[0].custom_records;
+
+        const boostagram = utils.getBoostagramFromInvoice(custom_records);
```
bumi, line 284: wondering if this is the right place to do this, or if the connector should return the raw data and then on a "higher level" we make sense of the data / extract the boostagram. 

I'd move that to the action level actually.

src/app/components/TransactionsTable/index.tsx
escapedcat: Added the `sender_name`. `message` is already there.  
If we really want to show all the detailed info we could add this later on. I'm unsure for what users all the details are interesting. I feel like we could tackle this with the future transaction designs.
What do you think @dvoroneca ?

src/app/screens/Home/index.tsx
escapedcat: Yeah, I was following the payment-example:  
https://github.com/getAlby/lightning-browser-extension/blob/feat/389_tx-lists-incoming-invoices/src/app/screens/Home/index.tsx#L238
Can be something for the future I believe. I.e. as soon as we show payments/invoices somewhere else and also need to use the i.e. `dayjs`-function

src/extension/background-script/connectors/lnd.ts
escapedcat: Moved this to the action

src/common/lib/utils.ts
escapedcat: Let's zoom

src/extension/background-script/connectors/connector.example
bumi: I think we use the `getBalance()` function. 

also this is unrelated to this PR. let's check this separately, this PR is already big and adding a bunch of things.

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -9,6 +9,37 @@ interface Route {
...
 }
 
+export interface Invoice {
+  id: string;
+  memo: string;
+  type: "received";
+  settled: boolean;
+  settleDate: number;
+  totalAmount: string;
+  totalAmountFiat?: string;
```
bumi, line 19: the connector probably does not know about fiat amounts.

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -9,6 +9,37 @@ interface Route {
...
+  preimage: string;
+  custom_records?: {
+    "696969": string;
+    "7629169": string;
+    "5482373484": string;
+  };
+  boostagram?: {
```
bumi, line 26: imo the connector should not know about boostagrams. 

The connector should return low-level lightning data and then in an action or some "higher level" we parse boostagrams, check fiat amounts, and other things.

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -31,6 +62,12 @@ export type GetBalanceResponse = {
   };
 };
 
+export type GetInvoicesResponse = {
+  data: {
+    invoices: Invoice[];
```
bumi, line 67: the response here should be a minimal(?) set of data that can be returned by all connectors. 

I actually would use a different type here for this. I think the connector interfaces ideally do not know anything about the rest of the alby application.

src/extension/background-script/connectors/lnbits.ts
```diff
@@ -56,6 +56,11 @@ class LnBits implements Connector {
     });
   }
 
+  // not yet implemenetd
+  getInvoices() {
+    return new Error("Has not been implemneted on this connector: getInvoices");
```
bumi, line 61:
```suggestion
    console.error(`${this.constructor.name} does not implement the getInvoices call`);
    return new Error("Not yet supported with the currently used account.");
```

Was wondering if the error message should be more descriptive for the user?

src/extension/background-script/connectors/lnd.ts
```diff
@@ -223,11 +227,76 @@ class Lnd implements Connector {
...
+          invoice.htlcs[0] && invoice.htlcs[0].custom_records;
+
+        return {
+          custom_records,
+          id: `${invoice.payment_request}-${index}`,
+          memo: invoice.memo,
+          preimage: invoice.r_preimage,
+          settled: invoice.settled,
+          settleDate: parseInt(invoice.settle_date) * 1000,
+          totalAmount: invoice.value,
```
bumi, line 287: is LND returning strings here?

escapedcat: Yes :(

https://api.lightning.community/?javascript#v1-invoices
image

src/extension/background-script/connectors/connector.interface.ts
escapedcat: Good point. I split this up into two types.

src/extension/background-script/connectors/lnbits.ts
escapedcat: Good point. Added.

escapedcat:
> what do you mean? I guess you have zero sats in your wallet? this is not releated to the list if incoming payments in this PR, is it?

We had a look. We assume the issue was that the downloaded build version from the comment above will connect to regtest if you create/add your Alby wallet. If you have sats on your prod wallet you won't have those on regtest. You know what I mean?

src/app/screens/Home/index.tsx
escapedcat: Should be refactored in a TODO. Once this and further TS-switches for allowances are merged.

escapedcat:
Need to finalize and add i18n

src/app/screens/Home/index.tsx
escapedcat: https://github.com/getAlby/lightning-browser-extension/issues/1158

src/app/screens/Home/index.tsx
```diff
@@ -286,45 +326,88 @@ function Home() {
...
+                        "w-1/2 rounded-lg py-2.5 font-bold transition duration-150",
+                        "focus:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:ring-orange-bitcoin",
+                        "hover:bg-gray-50 dark:hover:bg-surface-16dp",
+                        selected
+                          ? "text-orange-bitcoin"
+                          : "text-gray-700  dark:text-neutral-200"
+                      )
+                    }
+                  >
+                    {category}
```
bumi, line 351: I have here two different styles:

default when I open the popup
image
when I clock in incoming: 
image

src/app/screens/Home/index.tsx
```diff
@@ -94,6 +100,36 @@ function Home() {
...
 
+  async function onTabChangeHandler(index: number) {
+    if (index === 1) {
+      await loadInvoices();
+    }
+  }
+
+  async function loadInvoices() {
+    // incoives have already been loaded
+    if (invoiceTransactions) return invoiceTransactions;
```
bumi, line 111: switching accounts does not reload the invoices.

we also don't do this with the outgoing ones because there we only show the ones in our DB (no matter from which account)

src/app/screens/Home/index.tsx
```diff
@@ -94,6 +100,36 @@ function Home() {
...
+
+    setLoadingInvoices(true);
+    const result = await api.getInvoices();
+
+    const invoices: Transaction[] = result.invoices
+      .filter((invoice) => invoice.settled)
```
bumi, line 117: we could at some point add this to the action. thinking of pagination we might need arguments there anyway.

```
api.getInvoices({settled: true});

File: src/app/components/TransactionsTable/index.tsx [hunk 20]
```diff
@@ -49,12 +52,13 @@ export default function TransactionsTable({ transactions }: Props) {
                         {tx.title}
```
bumi, line 52: I think when title is blank then the type and date below move up. 
maybe we should always use a fixed height here? (always add a &nbsp; at the end 😱 )

src/app/components/TransactionsTable/index.tsx
```diff
@@ -76,26 +80,65 @@ export default function TransactionsTable({ transactions }: Props) {
                           ~{tx.totalAmountFiat}
                         </p>
                       </div>
-                      <Disclosure.Button className="block h-0 mt-2 text-gray-500 hover:text-black dark:hover:text-white transition-color duration-200">
-                        <CaretDownIcon
-                          className={`${open ? "rotate-180" : ""} w-5 h-5`}
-                        />
-                      </Disclosure.Button>
+                      {([tx.type && "sent", "sending"].includes(tx.type) ||
```
bumi, line 83: when having an a class object... I would write here something like `tx.hasMetadata?` (in ruby) 

but we probably need this check only here, so it's fine.

src/extension/background-script/connectors/lnd.ts
```diff
@@ -223,11 +227,75 @@ class Lnd implements Connector {
     });
   };
 
-  getTransactions = () => {
-    return this.request("GET", "/v1/payments", undefined, {
-      transactions: [],
-    });
-  };
+  async getInvoices(): Promise<GetInvoicesResponse> {
+    const data = await this.request<{
```
bumi, line 231: we miss pagination here.
`index_offset` `num_max_invoices`

there is also a `reversed` parameter which we probably want to use instead of the reverse() in JS.

if we set `reversed` to true, then we should get the newest ones and probably also can ignore pagination for now.

src/extension/background-script/router.ts
```diff
@@ -19,6 +19,7 @@ const routes = {
     lnurl: webln.lnurl,
     makeInvoice: webln.makeInvoiceOrPrompt,
     verifyMessage: ln.verifyMessage,
+    getInvoices: ln.invoices,
```
bumi, line 22: this should not be here. getInvoices is currently not a webln call and can not be accessed through webln

src/types.ts
```diff
@@ -265,24 +271,25 @@ export interface IBadge {
...
   preimage: string;
-  badges: IBadge[];
+  subTitle?: string | React.ReactNode;
+  title: string | React.ReactNode;
   totalAmount: string;
-  totalFees: string;
-  description: string;
-  location: string;
-  totalAmountFiat: string;
+  totalAmountFiat?: string;
```
bumi, line 289: those strings are wrong here I think. 
those should be numbers, but I think we currently store the value we get from the connector: in lnd this is a string in lndhub it's a number...

src/app/screens/Home/index.tsx
```diff
@@ -30,9 +32,13 @@ function Home() {
   const [isBlocked, setIsBlocked] = useState<boolean>(false);
   const [currentUrl, setCurrentUrl] = useState<URL | null>(null);
   const [payments, setPayments] = useState<Transaction[]>([]);
+  const [invoiceTransactions, setInvoiceTransactions] = useState<
```
bumi, line 35: this sounds a bit doppeltgemoppelt.

and we do `invoiceTransactions(invoices)`

src/app/screens/Home/index.tsx
```diff
@@ -94,6 +100,36 @@ function Home() {
...
+      .filter((invoice) => invoice.settled)
+      .map((invoice) => ({
+        ...invoice,
+        title: invoice.memo,
+        date: dayjs(invoice.settleDate).fromNow(),
+      }));
+
+    for (const invoice of invoices) {
+      const totalAmountFiat = await getFiatValue(invoice.totalAmount);
+      invoice.totalAmountFiat = totalAmountFiat;
```
bumi, line 126: also here maybe a class object with computed attributes would be nice. 

`invoice.totalAmountFiat` could take care of this on its own whenever it is accessed. 
Now we everywhere need to iterate and do those calls manually

bumi (commented):
(one thing that I keep thinking about is that we now show account specific incoming invoices but global outgoing invoices...)

src/app/components/TransactionsTable/index.tsx
escapedcat: Fixed

src/app/screens/Home/index.tsx
escapedcat: Done

src/app/screens/Home/index.tsx
escapedcat: Done

src/extension/background-script/router.ts
escapedcat: Removed

src/types.ts
escapedcat: Switched it to number

src/app/components/TransactionsTable/index.tsx
bumi: lol, did not know that GH interprets `&nbsp;` here in comments.

src/app/screens/Home/index.tsx
bumi: not sure how this is done in JS in 2022. 

```
class Invoice {

   get totalAmountFiat() {
    return getFiatValue(this.totalAmount);
  }
}
```

but actually I am wondering if we should remove the fiat amounts here? 
If we make fiat enabled/disabled configurable then we would need a check here everywhere and if we actually want the fiat values of the time of the transaction we need to store it in the DB and for the payments that are not in our DB we do not have historic data.

escapedcat: Yeah, but we won't tackle this in this PR, right?  
If we want to show historic data we should do a brainstorming.
Lisa also had some comments to this.

src/app/components/TransactionsTable/index.tsx
bumi: image

src/extension/background-script/connectors/lnd.ts
bumi: it seems it is somewhat working for me. 
I get new transactions BUT still need the reverse() ? 

image

src/types.ts
bumi: string | number would be currently correct.

lnd returns strings (and then we save strings)
lndhub returns numbers (and then we save numbers)

src/app/screens/Home/index.tsx
bumi: no. but the more I use it the more I am in favor of removing fiat in the transaction list. wondering what others say.

MoritzKa: My priorities for the transaction list would be
1. btc with historical fiat prices
2. only btc
3. btc with current fiat prices

src/app/screens/Home/index.tsx
```diff
@@ -286,45 +324,89 @@ function Home() {
...
+                {[
+                  t("transaction_list.tabs.outgoing"),
+                  t("transaction_list.tabs.incoming"),
+                ].map((category) => (
+                  <Tab
+                    key={category}
+                    className={({ selected }) =>
+                      classNames(
+                        "w-1/2 rounded-lg py-2.5 font-bold transition duration-150",
+                        "focus:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:ring-orange-bitcoin",
```
bumi, line 341: I actually like it better without the ring.

src/app/screens/Home/index.tsx
bumi: any objections to remove it?

src/i18n/locales/en/translation.json
```diff
@@ -140,13 +140,22 @@
...
       "allowance_view": {
+        "recent_transactions": "Recent Transactions",
         "allowance": "Allowance",
         "sats_used": "sats used",
         "no_transactions": "No transactions on <0>{{name}}</0> yet."
       },
       "default_view": {
-        "no_transactions": "No transactions yet."
+        "recent_transactions": "Recent Transactions",
+        "no_transactions": "No transactions yet.",
```
bumi, line 157:
```suggestion
        "no_transactions": "No transactions made with Alby, yet.",
```

maybe we should indicate that no transactions have been made with Alby?
because in the incoming tab we show all, but in the outgoing tab we only show the ones made with Alby. 
This is somewhat confusing....

src/app/screens/Home/index.tsx
bumi: 1. is not possible (we do not have the incoming payments in our DB and we do not have historical rates) - seems breez and zeus do not show amounts. Phoenix and bluewallet have rates (not sure if historical, can't imaging bluewallet has historical rates)

src/extension/background-script/connectors/lnd.ts
escapedcat: As discussed. Reversing the reversed result...

src/types.ts
escapedcat: Had a look.  
So `totalAmountFiat` on `Transaction` is always a `string.  
`totalAmount` on `Payment` is `number | string`, but we handle this already.
I think we're good here.

src/app/screens/Home/index.tsx
escapedcat: Will remove after this is merged. This came in with the original Fiat PR.

src/app/screens/Home/index.tsx
escapedcat: Done

PR #1084, opened 2022-06-25
chore: fix react-modal warning

im-adithya:

We don't support screen readers so might as well turn off the`ariaHideApp` prop functionality of the react-modal which hides the background content from being accessed by screen readers when the modal is open.

### Link this PR to an issue

--

- `chore`

--

### How has this been tested?

Ran all tests

### Checklist

### DIFF
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -77,6 +77,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
         </Menu.List>
       </Menu>
       <Modal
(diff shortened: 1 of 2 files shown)


escapedcat:
> We don't support screen readers

I don't know about screen readers and extensions or if anyone is using Alby with a screen reader. In general I think we shouldn't turn off aria labels on purpose. On the other hand if we don't really care at the moment than it's understandable to turn it off. And maybe take care of this later on.
@bumi what do you think?

escapedcat:
We don't have knowledge regarding screen readers and extensions. If someone has issues using Alby we're happy to improve this.

Issue #793, opened 2022-04-14
Make it more recognizable that sites are offering tipping support


reneaaron:
I just noticed that there is a different Alby icon showing up in the browser whenever a website supports tipping. (metatag, but also on GitHub, Youtube, etc)

Inactive
![image](https://user-images.githubusercontent.com/100827540/163378598-4c832a38-32c9-48c3-83ec-b45ad8f5a53b.png)

Active
![image](https://user-images.githubusercontent.com/100827540/163378659-b7ea5116-5564-485e-adff-f932cfab0fdd.png)

I think it would be great if the active icon would indicate tipping support a bit more obvious. (by having a different color maybe?)

**Are you working on this?**
If you agree I'll prepare some suggestions and make a PR.

bitcoinuser:
Agree

bumi:
maybe @johnsBeharry or @GBKS have some ideas? 
There is also the possibility to add a colored badge to the icon (like those unread messages badges)

GBKS:
I think the problem really is the dark background. On a light background, the difference between the two icons is more obvious. Can we show different icons for dark backgrounds?

For reference, here is the [source](https://www.figma.com/file/YVMGfhSHNkK7IWWkCeGUvp/Alby?node-id=47%3A227) for the different Alby head states.

bumi:
@secondl1ght can you check if we can customize the browser action icon based on the theme?
I think there is something like `theme_icons`

bumi:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action

reneaaron:
Here is the difference on light themes:

✅ Active
![image](https://user-images.githubusercontent.com/100827540/163562368-87ffed03-0c21-4363-bb5f-a1ff19829afa.png)

❌ Inactive
![image](https://user-images.githubusercontent.com/100827540/163562394-a9ee3537-748b-4673-9ee8-220d35173d66.png)

To me the the 2 icons are too similar for users to recognize their function, also considering the size they are displayed. I used Alby for multiple weeks when I found out about different icons while browsing the source code 😅

blastshielddown:
Another way to approach this might be to use a standard icon/indicator on the service/site itself to show that Alby (or another WebLN wallet) is supported. Even something as simple as a chip that says "WebLN" with a checkmark or green dot. Would be curious to hear what better UX experts think of this.

Rough example:
[...]

PR #1086, opened 2022-06-25
Feature: Update extension icons

reneaaron:

Implemented #793 

#### Light
image
image
image

#### Dark
image
image
image

### How has this been tested?

 - [x] Checked different states in Chrome & Firefox

### Checklist


reneaaron:
@GBKS I implemented a first version of the new icons and attached some screenshots of how they currently look in different themes. I really like how they turned out, what do you think?

escapedcat:
If it's not too much effort it would be great if we could switch the `js` files to `ts`. Not much left till we switched the whole app.

reneaaron:
Good point. Do we just update the files we touch during a PR or is it better to do it in a separate PR? I guess it would probably be easier to review if we keep them separate since you are not distracted with "unrelated" changes in other files? 🤔 

Either way: I'd love to get some help with that. 🙌

escapedcat:
> Good point. Do we just update the files we touch during a PR or is it better to do it in a separate PR? I guess it would probably be easier to review if we keep them separate since you are not distracted with "unrelated" changes in other files? 🤔

It depends :D  
In my last projects we i.e. had the rule "if it's not yet TS it has to be when the file is touched". In that way the app gets switched faster.
We also had "if a certain view doesn't have a test at all it needs to have a test after it has been touched".
In our case we could think about "if you touch the file and it's not yet using translation-strings you need to switch out the English strings for translation-strings".

> Either way: I'd love to get some help with that. 🙌

Let's pair on that :)

GBKS:
Thanks for implementing these. I think they look good. Although the blue and green states on dark mode might be nicer if the eye balls were dark blue/green and the pupil white. Will take another look at that.

GBKS:
I think these two look a bit better contrast-wise for the dark mode. At a small size, it's easier to distinguish the eye ball from the pupil, and it's more consistent with the yellow one.

image

reneaaron:
> I think these two look a bit better contrast-wise for the dark mode. At a small size, it's easier to distinguish the eye ball from the pupil, and it's more consistent with the yellow one.

Thanks for your review and the new icons for the dark mode, they look great! 👍 

image
image

GBKS:
Awesome, I'm excited to try them out IRL.

src/extension/background-script/actions/setup/index.js
```diff
@@ -1,7 +1,14 @@
 import reset from "./reset";
```
escapedcat, line 1: Switch to `ts`?

src/extension/background-script/actions/setup/setIcon.ts
```diff
@@ -0,0 +1,59 @@
...
+  Tipping = "alby_icon_blue",
+  Active = "alby_icon_green",
+}
+
+const setIconMessageHandler = async (
+  message: SetIconMessage,
+  sender: Runtime.MessageSender
+): Promise<{ data: boolean }> => {
+  // Under some circumstances a Tab may not be assigned an ID
+  if (!sender.tab || !sender.tab.id) {
```
escapedcat, line 19: Does this work?:

```js
if (!sender.tab?.id) {
```

src/extension/background-script/actions/setup/setIcon.ts
```diff
@@ -0,0 +1,59 @@
+import browser, { Runtime } from "webextension-polyfill";
+import { SetIconMessage } from "~/types";
```
escapedcat, line 2: Could we rename this to `MessageSetIcon`? So far all message-types start with `Message`

src/extension/background-script/actions/setup/setIcon.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+};
+
+const setIcon = async (icon: string, tabId: number): Promise<void> => {
+  const currentIcon = tabIcons.get(tabId);
+
+  // The active icon has priority over tipping
+  if (
+    currentIcon &&
+    currentIcon == ExtensionIcon.Active &&
```
escapedcat, line 38: `===` should be prefered because it's type-safe. Should check eslint rules I guess.  
Context https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

src/extension/background-script/index.ts
```diff
@@ -76,7 +76,15 @@ const handleInstalled = (details) => {
 
 // listen to calls from the content script and calls the actions through the router
 // returns a promise to be handled in the content script
-const routeCalls = (message, sender) => {
+const routeCalls = (
```
escapedcat, line 79: `type` was replaced by `action`. Should be possible to remove `type`.  

To keep things aligned maybe try:

```ts
import type { MessageDefault } from "~/types";

interface Message extends MessageDefault {
  args?: unkown;
}

...

const routeCalls = (
  message: Message
  ...
```

Usually `args` is an object I think

src/types.ts
```diff
@@ -70,13 +70,15 @@ export interface Battery extends OriginData {
   icon: string;
 }
 
-// deprecated message type,please stop using this
+/**
+ * @deprecated Use MessageDefault instead
```
escapedcat, line 74: ❤️

src/extension/background-script/actions/setup/setIcon.ts
```diff
@@ -0,0 +1,59 @@
+import browser, { Runtime } from "webextension-polyfill";
+import { MessageSetIcon } from "~/types";
+
+import state from "../../state";
+
+const tabIcons = new Map<number, string>();
+
+enum ExtensionIcon {
```
escapedcat, line 8: 👏

escapedcat (approved):

src/extension/background-script/index.ts
```diff
@@ -76,7 +76,15 @@ const handleInstalled = (details) => {
...
+  message: {
```
bumi, line 80: aren't there more attributes in this message? e.g. args?

src/extension/background-script/index.ts
```diff
@@ -9,14 +10,24 @@ import state from "./state";
 
 let isFirstInstalled = false;
 
+declare global {
```
bumi, line 13:
```suggestion
// when debugging is enabled in development mode a window.debugAlby object is defined that can be used within the console. This is the type interface for that
declare global {
```

do we need a comment there?

src/extension/background-script/actions/setup/setIcon.ts
```diff
@@ -0,0 +1,59 @@
+import browser, { Runtime } from "webextension-polyfill";
+import { MessageSetIcon } from "~/types";
+
+import state from "../../state";
+
+const tabIcons = new Map<number, string>();
```
bumi, line 6: can we add a comment what we store here? 

the number is always unique per tab?
And the data stays there even if the tab is closed, correct? (but not a problem because it is not much?)

bumi (approved):
it works for me! \o/

src/extension/background-script/index.ts
reneaaron: There are, but they are not used in this function but passed along into the router:

image

There is the generic message type `Messsage` (it has been deprecated) and separate types for each message type. There is currently no common base type for messages anymore that could be used in here 🤔

src/extension/background-script/actions/setup/setIcon.ts
reneaaron: > can we add a comment what we store here?

Sure, will do! 👍

According to the docs the tabId is unique, yes:

➡️ https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab

> The tab's ID. Tab IDs are unique within a browser session. The tab ID may also be set to [tabs.TAB_ID_NONE](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/TAB_ID_NONE) for browser windows that don't host content tabs (for example, devtools windows).

>  And the data stays there even if the tab is closed, correct? (but not a problem because it is not much?)

That's a good point. We could add a `onRemoved` listener to clean up this map after tabs have been closed. I don't think it's a problem (as you said it's not much data being stored in there) but it would feel a bit cleaner. Let me know if I should add it...

src/extension/background-script/index.ts
escapedcat: This shouldn't block this PR I think. If the deprecated type works just use it. At least files were switched to TS. That's the important step. Aligning message types can be done later. I'm happy to work on this later on.
The main goal of switching to individual message types was to clean up and identify what we currently use and also to use these in Jest tests. Maybe a Union-type works better later on.

src/extension/background-script/actions/setup/setIcon.ts
bumi: no code is better than no code... so I think it is fine for now. less event handler are also better I think

PR #1088, opened 2022-06-26
build: switch babel to swc

escapedcat:
Replace babel with [swc](https://swc.rs/docs/usage/swc-loader).

> SWC is 20x faster than Babel on a single thread and 70x faster on four cores.


bumi:
we want no babel! yay! 🎉 drinks on me when this is merged.

.swcrc
```diff
@@ -0,0 +1,21 @@
...
+      "tsx": true
+    },
+    "transform": {
+      "react": {
+        "runtime": "automatic"
+      }
+    }
+  },
+  "env": {
+    "targets": [
```
escapedcat, line 14: Currently there's an issue with [browserslist](https://github.com/browserslist/browserslist) support so we set it directly in here

jest.config.ts
```diff
@@ -15,6 +16,9 @@ module.exports = {
             },
           },
         },
+        env: {
+          targets: "defaults",
```
escapedcat, line 20: Use default-targets for jest tests.

bumi (approved):
works on my machine! :shipit:

escapedcat:
yolo!

Issue #1016, opened 2022-06-09
Allowances: Edit budget should use DualCurrencyField instead of TextField with `type=number`


escapedcat:
### Is your feature request related to a problem?

#1000 will add [a new Input-type](https://github.com/getAlby/lightning-browser-extension/pull/1000/files#diff-e341c06a28185f14a891bdb91afc6aa3ecaffb6f2f78dea6afb5e4b7d891d2c7) for sats which also displays the fiat value while typing.  

When editing an allowance (budget) this isn't being used yet:  

![image](https://user-images.githubusercontent.com/1016218/172776936-d7e74739-3b1d-4afb-ab23-8994c299d873.png)
 
### Describe the solution you'd like

The new field should be used there as well.  
An example how to use it can be found on the "Receive" view  :
![image](https://user-images.githubusercontent.com/1016218/172777343-8ae9e630-b1a0-47ec-82f6-9598d7d469d5.png)

### Additional context

Maybe have a look through the code if other fields need to be updated as well.  
If yes you could create issues for these and create PRs if you want to.

### Are you working on this?

No

secondl1ght:
This is completed with #1089

PR #1089, opened 2022-06-27
refactor: use new input component to support dual currency on edit budget screen

secondl1ght:
On the Edit Allowance screen I removed the TextField component in favor of the new DualCurrencyField component to support fiat values on this input element.

### Link this PR to an issue

#1016 

BEFORE:

image

AFTER:

image

### How has this been tested?

Ran locally and tested setting a new budget, checked console for any errors and verified new budget was set.

### Checklist


secondl1ght:
NOTE: I also added a new prop to the DualCurrencyField component to support adding a hint text which will appear below the input element (it is optional). The TextField component had this functionality so I thought we should also have it in this new component.

escapedcat:
Would be nice if the hint has the same style as the hint from the textfield-input

secondl1ght:
I pushed a WIP fix for the unit test FAIL that is happening, it looks like the currency setting is undefined in this unit test environment. Which is weird because this setting has a default of USD. So I am trying to set this setting before the test runs but it is still not working. Any ideas here?

image
image

secondl1ght:
> Would be nice if the hint has the same style as the hint from the textfield-input

Good call, I updated this to be the case.

secondl1ght:
Merge conflicts resolved, unit test fixed and this is ready for final review and merging now.

escapedcat (approved):

PR #1098, opened 2022-06-28
fix: show allowances without budget

bumi:
Allowances without budget should be listed but do not have an "active" badge

### DIFF
--- a/src/app/screens/Publishers/index.test.tsx
+++ b/src/app/screens/Publishers/index.test.tsx
@@ -1,0 +1,44 @@
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter } from "react-router-dom";
+
+import { AccountsProvider } from "../../context/AccountsContext";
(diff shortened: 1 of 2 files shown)


escapedcat (approved):

src/app/screens/Publishers/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { AccountsProvider } from "../../context/AccountsContext";
+import Publishers from "./index";
+
+jest.mock("~/common/lib/utils", () => {
+  return {
+    call: jest.fn(() => ({
+      allowances: [
+        {
+          host: "publisher.regtest.getalby.com",
+          name: "publisher regtest getalby",
```
bumi, line 13: we need some more fun mock data :)

bumi (commented):
thanks for adding the test! \o/ looks pretty simple actually.

PR #1104, opened 2022-06-30
perf(jest): remove ts-jest

escapedcat:

- Remove ts-jest to improve speed
  - We still had this in to handle the path-alias' we created
  - Switched to manual mapping
- Remove ts-node dep
  - Switched jest config files back to js
  - Added types and usign those via comment option

- perf

### How has this been tested?

- unit tests ✅ 
- build ✅


jest.config.ts
```diff
@@ -15,6 +12,7 @@ module.exports = {
             },
           },
         },
+        target: "es2016",
```
escapedcat, line 15: Added this to avoid the weird `class` error

escapedcat:
e2e test failing for other reasons

bumi (approved):
looks good. 
(from my limited experience)

escapedcat:
> looks good. (from my limited experience)

We should just use this for all upcoming PRs :P

PR #1107, opened 2022-07-02
feat: change type of metadata for further rendering on screens

pavanjoshi914:
change type of metadata for further rendering


pavanjoshi914:
@bumi

PR #1114, opened 2022-07-05
test: publishers screen after wallet creation

escapedcat:

Add a test to open the publishers screen.

### DIFF
--- a/tests/e2e/createWallets.spec.ts
+++ b/tests/e2e/createWallets.spec.ts
@@ -47,33 +47,43 @@ const commonCreateWalletSuccessCheck = async ({ page, $document }) => {
   await waitFor(() => getByText($document, "Success!"));
 };
 
+const createAlbyWallet = async ({ page, $document, user }) => {
+  // click at "Create Alby Wallet"
+  const createNewWalletButton = await getByText($document, "Alby Wallet");
+  createNewWalletButton.click();
+
+  await waitFor(() => getByText($document, "Your Alby Lightning Wallet"));
+
+  // type user email
+  const emailField = await getByLabelText($document, "Email Address");
+  await emailField.type(user.email);
+
+  // type user password and confirm password
+  const walletPasswordField = await getByLabelText($document, "Password");
+  await walletPasswordField.type(user.password);
+
+  // click create a wallet button
+  const createWalletButton = await getByText($document, "Continue");
+  createWalletButton.click();
+
+  await page.waitForResponse(() => true);
+
+  await waitFor(() => getByText($document, "Your Alby account is ready."));
+};
+
 test.describe("Create or connect wallets", () => {
+  test("successfully creates an Alby wallet", async () => {
+    const { user, browser, page, $document } =
+      await commonCreateWalletUserCreate();
+    await createAlbyWallet({ page, $document, user });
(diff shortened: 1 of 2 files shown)


tests/e2e/createWallets.spec.ts
```diff
@@ -8,10 +8,10 @@ const { getByText, getByLabelText } = queries;
 
 const commonCreateWalletUserCreate = async () => {
   const user = USER.SINGLE();
-  const { page: welcomePage, browser } = await loadExtension();
+  const { page: page, browser } = await loadExtension();
```
escapedcat, line 11: Renaming `welcomePage` to `page` because it's the general puppeteer page object which can change during a test.

tests/e2e/createWallets.spec.ts
```diff
@@ -76,16 +70,28 @@ test.describe("Create or connect wallets", () => {
...
-    await welcomePage.waitForResponse(() => true);
+    await page.waitForResponse(() => true);
 
     await waitFor(() => getByText($document, "Your Alby account is ready."));
 
-    await commonCreateWalletSuccessCheck({ browser, welcomePage, $document });
+    await commonCreateWalletSuccessCheck({ page, $document });
+
+    // reload and expect to be on publishers screen
+    await Promise.all([
```
bumi, line 80: Why do we add this in this test? 
can this be a new test

escapedcat: Yeah, you're right. But in reality this won't work without adding a wallet first.  
We could fake it, split this into two tests and make it dependent on each other:
https://playwright.dev/docs/api/class-test#test-describe-serial

bumi: hmm.. but won't we need for anything an account first? (so anything would go into this? ;))

bumi: should be, no? in the end it's just an entry in the browser storage (setting some state)

escapedcat: Doesn't seem to be too simple to use i.e. `browser.storage...` within puppeteer because in our the code it's added via the webextension-polyfill. I believe that's what we would need, right?  
[This is easier with i.e. localstoarge](https://stackoverflow.com/a/56141229/1667461).

bumi: yeah, I see. this is the same problem as in the other e2e test for the batteries. 
hmmmmmmmmmmmm....

tests/e2e/createWallets.spec.ts
```diff
@@ -35,27 +35,21 @@ const commonCreateWalletUserCreate = async () => {
...
+  await page.waitForResponse(() => true);
   await waitFor(() => getByText($document, "Success!"));
-
-  await browser.close();
 };
 
 test.describe("Create or connect wallets", () => {
-  test("successfully creates an Alby wallet", async () => {
-    const { user, browser, welcomePage, $document } =
+  test("successfully creates an Alby wallet and opens publishers screen", async () => {
```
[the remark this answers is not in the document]
escapedcat: I'll take a look.  
The jest-puppeteer lib maintainer also replied. Maybe there's a way to tackle this better in the future.

bumi (approved):
let's get it in.

PR #1118, opened 2022-07-06
chore: revamp enable screen

im-adithya:

Redesign of "Enable" screen

- `chore`: Revamp (non-breaking change improving design)

## 1) Enable Screen

<p>

</p>

## 2) lnurlAuth Screen

<p>

</p>

## 3) lnurlPay Screen

<p>

</p>

## 4) lnurlWithdraw Screen

<p>

</p>

## 5) MakeInvoice Screen

Screenshot from 2022-07-28 17-42-56

## 6) Publisher Screen

Screenshot from 2022-07-28 14-54-00

Screenshot from 2022-07-28 14-48-40

## 7) ConfirmKeySend Screen

Screenshot from 2022-07-28 17-54-35

## 8) ConfirmPayment Screen

Screenshot from 2022-07-28 16-56-35

## 9) ConfirmSignMessage Screen

Screenshot from 2022-07-28 13-19-13

## 10) Send Satoshis (Home) Screen

<p>

</p>

### How has this been tested?

Tested manually

### Note
Have to check the success screens before merging!

### Checklist


escapedcat:
My understanding was that we want to get rid of the background color because it can look very weird.

im-adithya:
Made changes! Here's the new one:
Screenshot from 2022-07-07 16-48-36

src/app/components/NewPublisherCard/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  children?: React.ReactNode;
+};
+
+export default function PublisherCard({ title, image, url, children }: Props) {
+  return (
+    <div className="h-full flex flex-col justify-center items-center bg-gray-100 dark:bg-surface-02dp">
+      <img
+        className="w-20 h-20 m-2 shrink-0 bg-white border-solid border-2 border-white object-cover rounded-full shadow-2xl"
+        src={image}
+        alt=""
```
escapedcat, line 17: Would be good to set an `alt` what's this supposed to display for semantic reasons.  
Instead of add the image via JS is it possible to default the image-prop with the default-image and overwrite it if a image-prop is being handed in?
Using JS to set an image seems unsemantic.
Maybe I don't understand the full context here.

src/app/components/NewPublisherCard/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+        onError={(e) => {
+          const target = e.target as HTMLImageElement;
+          target.onerror = null;
+          target.src = DEFAULT_IMAGE;
+        }}
+      />
+      <div className="overflow-hidden text-center">
+        <h2
+          className="text-xl my-2 leading-5 font-bold dark:text-white truncate"
+          title={title}
```
escapedcat, line 27: Why do we need a title-attr on the h2 which is displaying the tile anyways?

escapedcat (changes requested):

bitcoinuser:
> ### Describe the changes you have made in this PR
> Redesign of "Enable" screen
> 
> ### Type of change (Remove other not matching type)
> * `chore`: Revamp (non-breaking change improving design)
> 
> ### Screenshots of the changes (If any)
> Screenshot from 2022-07-06 17-19-17
> 
> ### How has this been tested?
> Tested manually
> 
> ### Checklist
> * [x]  My code follows the style guidelines of this project and performed a self-review of my own code
> * [x]  New and existing tests pass locally with my changes
> * [x]  I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

Hi,

I was chatting with @bumi about my experience and doubts about this part of Alby and I would like to give some feedbacks.
Problems: 
1- I thought that the permission in case I have two accounts on Alby would be for the two accounts. Was not clear to me that is just for the account selected on top left.
2- It's easy for users forget to select the correct account on top left.

Feedback:

1- I think for this kind of window, it's not necessary show the account switch button that exist on top left. People usually ignore it when making actions on this kind of window.
2-I think should be able to grant permission for more than one account if more than one account exist on Alby.
3- Below the text (Request approval for transactions), should have a text like: (Select the accounts to grant permissions). The connect button should stay disabled until the user select the accounts that will receive grant permission.
4- Should be able to edit the grant permission if the user wants, like disable permission on a site for one or the two accounts. 

Thanks

bitcoinuser:
> 3- Below the text (Request approval for transactions), should have a text like: (Select the accounts to grant permissions). The connect button should stay disabled until the user select the accounts that will receive grant permission.

Or if the user has two or more accounts, after clicking on connect button, Alby show a window warning the user has two accounts and give him the option to choose which ones he would like to grant permission.

bitcoinuser:
Another idea is remove the Cancel button and rename the Connect button to Select accounts.

escapedcat:
@bitcoinuser thanks for your feedback! We won't tackle this in the PR I believe. I'm worried your feedback might in this way.
Are you on the [bitcoin design slack](https://github.com/getAlby/lightning-browser-extension#join-the-conversation)? You could join our channel there and talk to @dvoroneca and bumi directly. They can open up an issue to tackle these things in the future if needed.

src/app/components/NewPublisherCard/index.tsx
im-adithya: I have tried this in a project previously, but I think this kind of fallback using ternary would only work when the image URL is undefined; in cases where the link is valid but the image is broken (which can happen more often here), the onError works to replace the image with the fallback

src/app/components/NewPublisherCard/index.tsx
im-adithya: Thought it would help on truncate, but ig the design allows more than 1 line of text! Changing it!

escapedcat:
Tested with [starblocks.acinq.co](https://starblocks.acinq.co/):
- Should the URL have `test-overflow: ellipsis` and a title tag to show the full link there?
    image
- "Pay now" screen doesn't fit in the prompt. It's too high and cuts off some next below.
    The "old" prompt will show a scrollbar. This "new" one doesn't seem to do this.
    image image
    See old:
      image
    (Tested "old" on Chrome, "new" on Chromium. Not sure if that's an issue)
- Looks like all occurrences of ConformOrCancel/PublisherCard have been replaced, apart from this newly merged file:
    https://github.com/im-adithya/lightning-browser-extension/blob/task-revamp/src/app/screens/LNURLChannel.tsx  
    This can also use the new ones I assume.

src/app/screens/LNURLAuth.tsx
```diff
@@ -27,16 +27,26 @@ function LNURLAuth({ details, origin }: Props) {
...
+        <PublisherCard title={origin.name} image={origin.icon} />
+      </div>
+      <div className="flex flex-col justify-between h-3/5">
+        <dl className="m-6 shadow bg-white dark:bg-surface-02dp p-4 rounded-lg">
           <dt className="font-semibold text-gray-500">
             {origin.name} asks you to login to
           </dt>
-          <dd className="mb-6 dark:text-white">{details.domain}</dd>
+          <dd className="mb-6 dark:text-white">
+            {"https://google.com" + details.domain}
```
escapedcat, line 40: What's google doing here? :)

escapedcat (changes requested):

src/app/screens/LNURLAuth.tsx
im-adithya: Oops! xD

im-adithya:
Addressed all issues, ready for review!

bumi:
good job @im-adithya 

@dvoroneca can you look at this, and use this? we need to see how we step by step implement the new components. 
to me the publisher screen takes too much screen space right now.

escapedcat:
Actions to finish this PR:

1. @dvoroneca will have a look how to maybe adjust the banner
2. We need to come up with a final copy-text for:
     image

dvoroneca:
### 1) Enable Screen

Reduce the card size, place it on a card
Connect with www podverce fm
----------
### 2) lnurlAuth Screen

Reduce the card size and add a layout variant
Send Payment

-----------
### 3) lnurlPay Screen
Reduce the card size and add a layout variant

Send Payment-3
------------

### 4) lnurlWithdraw Screen
Reduce the card size and add a layout variant

Send Payment-1

----------

### 5) MakeInvoice Screen

Reduce the card size and add a layout variant

Send Payment

-----------

### 6) Publisher Screen
# 👍

-----------

### 7) ConfirmKeySend Screen

Reduce the card size and add a layout variant

Send Payment

-------------

### 8) ConfirmPayment Screen

Reduce the card size and add a layout variant

Send Payment-1

---------------

### 9) ConfirmSignMessage Screen

Reduce the card size and add a layout variant

Send Payment-6

--------------

### 10) Send Satoshis (Home) Screen
# 👍

dvoroneca:
> Actions to finish this PR:
> 
> 1. @dvoroneca will have a look how to maybe adjust the banner
> 2. We need to come up with a final copy-text for:
>    image
Connect with www podverce fm

bumi:
@dvoroneca thanks, placing it on a card is also progress towards your next screens that show the sent amount and the recipient?

im-adithya:
> ### 9) ConfirmSignMessage Screen
> Reduce the card size and add a layout variant
> 
> Send Payment-6

The Withdraw heading is an error, I assume?

im-adithya:
Screenshot from 2022-08-02 16-33-11
Screenshot from 2022-08-02 16-37-50
<p>




</p>
<p>


</p>

Screenshot from 2022-08-02 16-34-31

Screenshot from 2022-08-02 16-34-40

dvoroneca:
Nice, thanks. You've also edited the headers 👍

MoritzKa:
While testing I experienced this behavior:

Chrome: Uploading chrome_OHbX04ad8U.mp4…

Firefox: jp8mbT7242


I used these [nightlies](https://github.com/getAlby/lightning-browser-extension/actions/runs/2781795078)

src/app/router/Prompt/Prompt.tsx
```diff
@@ -46,6 +46,18 @@ const routeParams: {
...
+  routeParams.args = {
+    amountEditable: true,
+    memoEditable: true,
+    invoiceAttributes: {
+      destination: "test string",
+      customRecords: {},
+      valueSat: "string",
+    },
+    origin: {},
+  };
```
bumi, line 60: be very very very careful with such test data hacks. and actually never use those. if you need something like that then we need to look at the flow to make it easier that in development this actually can be tested. 

using the app in a real world context is most important also during development.

bumi (changes requested):

src/app/components/ConfirmOrCancel/index.tsx
```diff
@@ -1,48 +1,53 @@
...
-      </p>
-
-      <a
-        className="underline text-sm text-gray-600 dark:text-neutral-400"
-        href="#"
-        onClick={onCancel}
-      >
-        Cancel
-      </a>
+      {onBlock ? (
```
bumi, line 40: I think the block section we only need on the enable screen.
thus we probably don't need that in here, do we?

im-adithya: Yes, but the catch is that the link should be placed inside the component...
Can passing it as a child work?

im-adithya:
> While testing I experienced this behavior:
> 
> Chrome:
> 
>  chrome_OHbX04ad8U.mp4 
> Firefox: jp8mbT7242
> 
> I used these [nightlies](https://github.com/getAlby/lightning-browser-extension/actions/runs/2781795078)

Sorry for that, it should work now!

src/app/components/ConfirmOrCancel/index.tsx
bumi: what do you mean with inside the component? 

What I mean is that we do not have it inside the component, but just add it to the enable screen - or wherever it is needed

im-adithya: But currently, we are using flex and justify-between to push the ConfirmOrCancel screen to the bottom, so we would have to have the block link inside it, otherwise, the workaround would be to use a prop to disable showing the usual "only_trusted" string and then add it separately and wrap the component and block link in a div. Should I do this, then?

Screenshot from 2022-08-02 16-33-11 (copy)

bumi: no I would not add new props. Rather less props / less options and less code is always better (less complexity that needs to be managed) 

@escapedcat do you have an idea? 

I would actually like to have those text out of the ConfirmOrCancel component and also have the "block" option only on the enable screen. So I don't think the ConfirmOrCancel should know about the blocking

reneaaron:
First of all: I think this is a huge improvement, especially the enable flows and publisher cards look a lot cleaner to me. 💯

I noticed some minor things: 

1) Default screen for publishers with tipping 
image

While the button seems to be a bit odd in this screen. Maybe we could also use the header from here? 
image

2) Rounding of logo container
I've seen different cases where the logo of the publisher is cropped quite badly: 
image
image

Do you think we could go with lesser rounding here so those logos do not get cropped anymore? (here with `rounded-lg`) /cc @dvoroneca 

image

dvoroneca:
Thanks @reneaaron for pointing this issues out. Let's redefine the logo cropping to 8 px corner radius and see if this looks better. 

About the button, also agree that instances that have a button should have a wide state.

👍

im-adithya:
> Default screen for publishers with tipping

Fixed! The actual idea was to have the second screen, but somehow it didn't fetch the description properly, resulting in the small single-row component.

> Let's redefine the logo cropping to 8 px corner radius and see if this looks better.

Done!

im-adithya:
> Default screen for publishers with tipping

Fixed! The actual idea was to have the second screen, but somehow it didn't fetch the description properly, resulting in the small single-row component.

> Let's redefine the logo cropping to 8 px corner radius and see if this looks better.

Done!

MoritzKa:
If the website does not offer an og:image the payment button looks off
image

MoritzKa:
if I log in to a website. Can we switch the button from "Connect" to "Login"
chrome_EA731kITrT

MoritzKa:
I suspect the "do not ask for this site again" button can have bad UX effects
image

I clicked this button on https://getalby.com/demo and also confirmed the connection. Then I deleted the connection in the extension again. Since then I there is no Alby prompt window again if I press the "Pay 25 Sats" on that website. But since I deleted the connection I cannot use that button anymore.

bumi:
> I clicked this button on https://getalby.com/demo and also confirmed the connection. Then I deleted the connection in the extension again. Since then I there is no Alby prompt window again if I press the "Pay 25 Sats" on that website. But since I deleted the connection I cannot use that button anymore.

not sure if I understand this, because it is the point of the block feature that you are not prompted anymore. 
but this is also unrelated to this PR and related to the blocklist feature - did you use this in the current version?

escapedcat:
> > I clicked this button on https://getalby.com/demo and also confirmed the connection. Then I deleted the connection in the extension again. Since then I there is no Alby prompt window again if I press the "Pay 25 Sats" on that website. But since I deleted the connection I cannot use that button anymore.

@MoritzKa will be tackled here: https://github.com/getAlby/lightning-browser-extension/issues/1225
I can do that right after we merge this.

MoritzKa:
> > I clicked this button on https://getalby.com/demo and also confirmed the connection. Then I deleted the connection in the extension again. Since then I there is no Alby prompt window again if I press the "Pay 25 Sats" on that website. But since I deleted the connection I cannot use that button anymore.
> 
> not sure if I understand this, because it is the point of the block feature that you are not prompted anymore. but this is also unrelated to this PR and related to the blocklist feature - did you use this in the current version?

I thought I could not pay on the website anymore if I block further requests. But I did not see the "Enable now" button in the extension. All good!

src/app/screens/MakeInvoice.tsx
```diff
@@ -83,64 +86,68 @@ function MakeInvoice({
...
-                    <Dt>Amount</Dt>
-                    <Dd>{invoiceAttributes.amount}</Dd>
-                  </dl>
-                )}
-                {error && <p className="mt-1 text-red-500">{error}</p>}
-              </div>
-              <div className="mt-4">
-                {memoEditable && (
+    <div className="h-full flex flex-col overflow-y-auto no-scrollbar">
+      <div className="text-center text-xl font-semibold dark:text-white py-2 border-b border-gray-200 dark:border-neutral-500">
```
bumi, line 90: should this be a header component? 
we have some duplication there in div and css classes.

src/app/components/Button/index.tsx
```diff
@@ -29,7 +31,10 @@ export default function Button({
       className={classNames(
         direction === "row" ? "flex-row" : "flex-col",
         fullWidth && "w-full",
-        fullWidth ? "px-0 py-2" : "px-7 py-2",
+        fullWidth ? "px-0 py-2" : "",
+        halfWidth && "w-1/2",
+        halfWidth ? "px-0 py-2 mx-2" : "",
+        !(halfWidth || fullWidth) && "px-7 py-2",
```
bumi, line 37: hmmm, I think this is a bit hard to read
what means fullWidth vs. halfWidth
and why can both be true?

MoritzKa:
Not super bad; just one observation I had. 
I enable Alby on a website (e.g. makers.bolt.fun); then I delete the connection in the extension window; but I can still use the website and make a payment. Only when I reload the page the "enable" prompt appears again. 
I would expect that the "enable" prompt appears with having to reload the page.

https://user-images.githubusercontent.com/13707997/182706173-0ebee45a-9e5c-43d0-ab96-7a218c87b11c.mp4

bumi:
that has always been like that. and you're right that's not ideal. can you create a new issue (this is not part of this PR)

bumi:
some suggestions @im-adithya @escapedcat: https://github.com/im-adithya/lightning-browser-extension/pull/6

src/app/components/PublisherCard/index.tsx
```diff
@@ -1,73 +1,91 @@
...
 }: Props) {
-  const { data } = usePalette(image);
   return (
     <div
-      className="p-4 bg-gray-300"
-      style={{
-        backgroundColor: data.vibrant,
-        backgroundImage: `linear-gradient(${data.vibrant}, rgba(0, 0, 0, 0.15))`,
-      }}
+      className={
```
escapedcat, line 27: See if you can use the existing `classNames` helper like i.e. here:  
https://github.com/getAlby/lightning-browser-extension/blob/feat/1216_refactor-prompt-popup-lnurl-handling/src/app/components/Button/index.tsx#L29

src/app/components/PublisherCard/index.tsx
```diff
@@ -1,73 +1,91 @@
...
-            target.src = DEFAULT_IMAGE;
-          }}
-        />
-        <div className="overflow-hidden">
-          <h2
-            className="text-lg leading-5 font-bold text-white truncate"
-            title={title}
+      <img
+        className={
+          "m-2 shrink-0 bg-white border-solid border-2 border-white object-cover rounded-lg shadow-2xl " +
```
escapedcat, line 35: For these smaller optional classes you can use template literals like i.e. here:
https://github.com/getAlby/lightning-browser-extension/blob/feat/1216_refactor-prompt-popup-lnurl-handling/src/app/components/Badge/index.tsx#L11

src/app/screens/ConfirmKeysend/index.tsx
```diff
@@ -107,17 +106,16 @@ function Keysend(props: Props) {
...
-        description={origin.description}
-        image={origin.icon}
-      />
-      <div className="py-4">
-        <Container maxWidth="sm">
-          {!successMessage ? (
-            <>
-              <div className="mb-8">
+    <div className="h-full flex flex-col overflow-y-auto no-scrollbar">
+      <div className="text-center text-xl font-semibold dark:text-white py-2 border-b border-gray-200 dark:border-neutral-500">
```
escapedcat, line 110: As bumi mentioned below. Looks like the header-component HTML is the same for all screens. Just the text is changing, right? If that's the case we should move it into it's own component.

src/app/screens/ConfirmPayment/index.tsx
```diff
@@ -100,19 +103,19 @@ function ConfirmPayment(props: Props) {
...
-        <Container maxWidth="sm">
-          {!successMessage ? (
-            <>
-              <h1 className="dark:text-white font-bold mb-4">
-                Approve payment
-              </h1>
-              <div className="mb-6">
+    <div className="h-full flex flex-col overflow-y-auto no-scrollbar">
+      <div className="text-center text-xl font-semibold dark:text-white py-2 border-b border-gray-200 dark:border-neutral-500">
+        Approve Payment
```
escapedcat, line 108: See above

src/app/screens/ConfirmSignMessage/index.tsx
```diff
@@ -55,58 +54,65 @@ function ConfirmSignMessage(props: Props) {
...
-              <dl className="shadow bg-white dark:bg-surface-02dp p-4 rounded-lg mb-8">
-                <dt className="font-semibold text-gray-500">
-                  {originRef.current.host} asks you to sign:
-                </dt>
-                <dd className="mb-6 dark:text-white">{messageRef.current}</dd>
-              </dl>
-
+    <div className="h-full flex flex-col overflow-y-auto no-scrollbar">
+      <div className="text-center text-xl font-semibold dark:text-white py-2 border-b border-gray-200 dark:border-neutral-500">
+        Sign
```
escapedcat, line 59: See above

src/app/screens/ConfirmSignMessage/index.tsx
```diff
@@ -55,58 +54,65 @@ function ConfirmSignMessage(props: Props) {
...
+      </div>
+      {!successMessage ? (
+        <div className="h-full flex flex-col justify-between">
+          <div>
+            <PublisherCard
+              title={originRef.current.name}
+              image={originRef.current.icon}
+            />
+            <dl className="m-4 shadow bg-white dark:bg-surface-02dp p-4 rounded-lg">
```
escapedcat, line 68: Kinda looks like we can also move this into it's own component to keep all views consistent.

escapedcat (commented):
Moving out components doesn't need to block this PR for the release I think. I'm ok with merging this if everything else works and we tackle the component split in a follow-up PR.

src/app/components/Button/index.tsx
im-adithya: I didn't understand the question, here we check if neither of the two is set and apply the original thing just so as not to break things.

Earlier (before this PR):
`fullWidth ? "px-0 py-2" : "px-7 py-2"`

Full Width Button
181014351-cab1358f-9f30-47e1-97af-39ec7547cc38

Half Width Button
Screenshot from 2022-08-04 12-42-47

dvoroneca:
### Hi guys, happy this is merged now, however, there are a few tweaks regarding this update.

1. Too many symbols in a user profile name. Should be only one line of text.
Screenshot 2022-08-04 at 16 38 1

Group 27

-----------

2. The Header is still 20px text but should be 18px.

Screenshot 2022-08-04 at 17 28 33

PR #1124, opened 2022-07-08
refactor!: manifest v3 support #1051

### LINKED ISSUE #1051: Update extension manifest from v2 to v3

escapedcat:
# ~CURRENT FINDINGS~ ALL GOOD SO FAR

- ~Axios doesn't seem to work in background-scripts, might need to switch (back) to fetch~:
  https://stackoverflow.com/questions/66305856/typeerror-adapter-is-not-a-function-error-when-using-axios-and-webpack-in-chrom/70206333#70206333
- ~No (extrenal) images~
- ~No external API calls~
  image
- ~Just switching current code to use web workers doesn't work~  
  image

Trying to prepare the switch to v3

### Link this PR to an issue

Fixes #1051

- `refactor`: Breaking refactor

### How has this been tested?

...not yet

### Checklist


src/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
-  "manifest_version": 2,
+  "manifest_version": 3,
   "name": "Alby - Bitcoin Lightning Wallet",
-  "version": "0.0.0",
+  "version": "1.0.0",
```
bumi, line 4: fyi: the version is set by the package.json

escapedcat: Ah, thanks

src/manifest.json
```diff
@@ -35,10 +47,10 @@
     }
   },
 
-  "__chrome__minimum_chrome_version": "49",
+  "__chrome__minimum_chrome_version": "88",
   "__opera__minimum_opera_version": "36",
```
escapedcat, line 51: This version should be updated as well

escapedcat (commented):
Needs testing in Firefox

src/manifest.json
escapedcat: https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser#Opera_2020
v74 is based on chrome v88.
Updating to 74.

AnastasiaVolk:
The message when an incorrect password is entered contains extra information Screenshot 2022-10-11 at 12 59 50

bumi:
that's if you try to signup with a @getalby.com email. which is prevented now.

AnastasiaVolk:
@bumi ok, will do! 
signup was with a LN address instead of email
with a regular email it's ok

bumi:
there is no sigunp with a lightning address (you can login with a lightning address, but not signup) and there is no user support@

AnastasiaVolk:
Also, the Extension autolocks in a couple of minutes. And when it does (or if you lock it yourself) - it gives this when trying to create or login to my Alby accountScreenshot 2022-10-11 at 13 36 18

AnastasiaVolk:
and these messages about actions during the lock
Screenshot 2022-10-11 at 13 40 26

bumi:
that sounds like the background script/worker or such is no longer running. can you always post the logs? https://guides.getalby.com/overall-guide/alby-browser-extension/debugging/alby-on-google-chrome

AnastasiaVolk:
oh sure

AnastasiaVolk:
Screenshot 2022-10-11 at 15 42 09
is that ok if I do it like that?

AnastasiaVolk:
Screenshot 2022-10-11 at 15 55 35
Screenshot 2022-10-11 at 15 57 22

AnastasiaVolk:
Screenshot 2022-10-11 at 15 57 22
Screenshot 2022-10-11 at 15 55 35

escapedcat:
Thanks for testing @AnastasiaVolk !  
Apart from the auto-logout, did the basic functionality work as expected?  
As bumi said,this might be related to "servceWorkers" are being shut down, which we should be able to avoid as @bumi already pointed out via the metamask code.  
I'll focus on that next time I work on this.

bumi:
* https://developer.chrome.com/docs/extensions/mv3/mv2-sunset/
* https://twitter.com/search?q=manifest%20v3&src=typed_query&f=top
* https://github.com/search?q=%22chrome.storage.session%22&type=Code
* https://github.com/belaviyo/v3tov2

escapedcat:
- https://developer.chrome.com/docs/extensions/reference/storage/#property-session
-

escapedcat:
Steps to reproduce the "receiving end does not exist" error:
1. Install extension & password-pin and account
1. Do not open any devtools
1. Wait ~5 minutes
1. Open: chrome://extensions/
     and see "service worker (inactive)"
1. Click extension icon
1. See extension is locked and see error

Force the error:
1. Install extension & password-pin and account
1. Open: chrome://extensions/
     and see "service worker"
1. Open: chrome://inspect/#service-workers
     and see "Service Worker chrome-extension://bbi...mjic/js/background.bundle.js"
1. terminate that service worker
1. Open: chrome://extensions/
     and see "service worker (inactive)"
1. Click extension icon
1. See extension is locked and see error

escapedcat:
- `chrome.session....` needs to be replaced by browser

src/extension/background-script/actions/accounts/unlock.ts
```diff
@@ -31,7 +31,7 @@ const unlock = (message: MessageAccountUnlock) => {
   }
 
   // if everything is fine we keep the password in memory
-  state.setState({ password });
+  await chrome.storage.session.set({ password });
```
bumi, line 34: I would not like have code that we know will not work in here I think. 
This will be double work. 
If you go that route then also check that we are able to go that way and how. 

(especially when it is done in all the places https://github.com/getAlby/lightning-browser-extension/pull/1124/files#diff-1af05f60841943485ce3642d0c9cb25cfd58f148f6caec4162513a08bf571eb6R10 - all those need to change again)

escapedcat: Updated it. Tried to centralize the usage. Works. Not sure if this is how you use `state` but it's better than. Before. wdyt?  
Need to adapt the code for MV2 only in one place now.

jest.setup.js
```diff
@@ -13,6 +13,20 @@ if (!chrome.runtime.id) chrome.runtime.id = "history-delete";
 beforeAll(() => {
   // Enable the mocking in tests.
   server.listen();
+  const get = jest.fn();
+  const set = jest.fn();
+  global.chrome = {
```
bumi, line 18: Did you try the polify? actually I think no `chrome` should be in our code. we can use `browser`.

escapedcat: That's just for the jest tests. They run in a chrome env. We were already doing something similar in the tests.  
Problem with the webextension-polyfill is that it doesn't not yet support i.e. `...storage.session` (I guess MV3 in general: https://github.com/mozilla/webextension-polyfill/issues/329). [According to this comment](https://github.com/mozilla/webextension-polyfill/issues/329#issuecomment-1188822881) we could just use `chrome.` in all browsers already? I can try that.

bumi: ok, interesting. so more changes.

escapedcat: Using the polyfill for now but ignoring the missing type. eventually we can think about ditching the polyfill if `chrome` just works everywhere. But not now.

escapedcat:
Everything's green but adding commando connector.  
Fails with:
image

Narrowing it down to lnmessage. Could be this:
- https://github.com/aaronbarnardsound/lnmessage/blob/master/src/index.ts#L437
- https://github.com/aaronbarnardsound/lnmessage/blob/master/src/crypto.ts/#L20
- https://github.com/aaronbarnardsound/lnmessage/blob/master/src/crypto.ts/#L103

escapedcat:
Manifest adaption done 😱   
This needs testing now.

src/app/context/AccountContext.tsx
```diff
@@ -121,11 +121,14 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
         }
       })
       .catch((e) => {
-        toast.error(`An unexpected error occurred (${e.message})`);
+        toast.error(
+          `AccountContext: An unexpected error occurred (${e.message})`
```
bumi, line 125: why do we write "AccountContext" there? this is not really helpful for the user I guess.

src/extension/background-script/nostr/index.ts
```diff
@@ -16,23 +17,24 @@ class Nostr {
     return null;
   }
 
-  getPublicKey() {
+  async getPublicKey() {
     const publicKey = secp256k1.schnorr.getPublicKey(
-      secp256k1.utils.hexToBytes(this.getPrivateKey())
+      secp256k1.utils.hexToBytes(await this.getPrivateKey())
```
bumi, line 22: for me to learn: is this a good practice to use `await something` as function parameter? 
(back in the days this was problematic due to polyfills)

src/manifest.json
```diff
@@ -11,22 +12,46 @@
...
+      "resources": [
+        "js/inpageScript.bundle.js",
+        "js/inpageScriptWebLN.bundle.js",
+        "js/inpageScriptNostr.bundle.js"
+      ],
+      "matches": ["https://*/*"]
+    }
+  ],
+
   "permissions": [
```
bumi, line 33: if we have `__chrome__permissions` shouldn't those then be firefox__permission? or do those get overwritten?

src/app/context/AccountContext.tsx
escapedcat: So far this was helpful for us. Doubt the user cares or has a worse experience because of these tbh

escapedcat:
Testing:

- Works after service-worker is being shut down:
    https://github.com/getAlby/lightning-browser-extension/pull/1124#issuecomment-1305803446
- `webLn.enable()`
- Invoice creation
- Add commando connector

src/app/context/AccountContext.tsx
bumi: why is it helpful for us? why not use `console.error(e)`?

bumi:
I see this error on Chrome: 
image
is three the manifest file not perfectly correct?

bumi:
I assume: it should not be `content_security_policy.extension_pages` but rather something like: 
```
"content_security_policy": {
    "extension_pages": "...",
```

and google chrome tells this as error. (in the "errors" of the extension.) 
`Unrecognized manifest key 'content_security_policy.extension_pages'.`

escapedcat:
> and google chrome tells this as error. (in the "errors" of the extension.) `Unrecognized manifest key 'content_security_policy.extension_pages'.`

Thanks! Fixed. Didn't know about the "errors" button...

src/extension/background-script/nostr/index.ts
escapedcat: fixed a think, tested, seems to work

bumi:
Did we test this with all possible connectors already?

escapedcat:
> Did we test this with all possible connectors already?

I did not, not sure if @AnastasiaVolk tested with all connectors

bumi:
OK, likely we then need to test it in: 

- [ ] LND
- [ ] CLN/Commando
- [ ] Eclair
- [ ] LNBits
- [ ] BlueWallet

and because this is a big change that also can not be undone I think it would be good if we all check it out and try everything.

escapedcat:
Tested with LND (albydev/testnet) and CLN:

- nostr
- send
- invoice creation

CLN throws an error when sending to our ln-address but that's also the case on current master.

reneaaron:
Let's do a final round of testing and get this ready for the release. I just prepared a little testing matrix: 

|                | Chrome (Windows) | Chrome (Mac) | Chrome (Linux) | Firefox (Windows) | Firefox (Mac) | Firefox (Linux) |
|----------------|------------------|--------------|----------------|-------------------|---------------|-----------------|
| LND            |                  |              |                |                   |               |                 |
| LNDHub         |                  |              |                |                   |               |                 |
| Core Lightning |                  |              |                |                   |               |                 |
| LNbits         |                  |              |                |                   |               |                 |
| LNDHub         |                  |              |                |                   |               |                 |
| Kollider       |                  |              |                |                   |               |                 |
| Eclair         |                  |              |                |                   |               |                 |
| Umbrel         |                  |              |                |                   |               |                 |
| BTCPay Server  |                  |              |                |                   |               |                 |

I don't think we have to tick all boxes here, but I'd like to have some overview about what exactly we tested. Please just update this table accordingly ☝️ 

Things that should be tested: 
 - Invoice generation
 - Sending & receiving payments
 - WebLN integration
 - Nostr integration
 - Permissions 
 - Extension permissions (scan QR code)
 - TOR
 - Batteries (tipping on youtube, etc)

escapedcat:
noticed, will double check:
- popup doesn't close on "lock" (in chrome/osx)

im-adithya:
> popup doesn't close on "lock" (in chrome/osx)

I suspect this: https://github.com/getAlby/lightning-browser-extension/pull/2025/files#diff-3eda5a900b82096e98a0204c9e75d6b1d1b5e212de86fca7e6d107b123c0b412R125

escapedcat:
> > popup doesn't close on "lock" (in chrome/osx)
> 
> I suspect this: https://github.com/getAlby/lightning-browser-extension/pull/2025/files#diff-3eda5a900b82096e98a0204c9e75d6b1d1b5e212de86fca7e6d107b123c0b412R125

```
...).extension.getViews is not a function
```

Yup, let me have a look

escapedcat:
no e2e tests fail.. ugh... will check

escapedcat:
> no e2e tests fail.. ugh... will check

Works locally. Will retrty here.

src/extension/background-script/state.ts
```diff
@@ -132,10 +132,18 @@ const state = createState<State>((set, get) => ({
     } else {
       set({ mv2Password: null });
     }
-    const allTabs = browser.extension.getViews({ type: "tab" });
-    for (const tab of allTabs) {
-      tab.close();
-    }
+
+    const allTabs = await browser.tabs.query({ title: "Alby" });
```
bumi, line 136: is this correct, we query all open tabs and check for the title?

why is the filter() on the tab ids needed?

reneaaron: Follow the stackoverflow link. I think it's for TS to be happy 🤷

src/extension/background-script/actions/setup/reset.ts
```diff
@@ -3,11 +3,11 @@ import type { MessageReset } from "~/types";
 import state from "../../state";
 
 const reset = async (message: MessageReset) => {
+  await state.getState().password(null);
```
escapedcat, line 6: @im-adithya please check if this is correct. Just adjusted it according to the mv3 way of handling the password.

escapedcat:
## COMMANDO / LNMESSAGE UPDATE

### Status
Commando was already tested and checked ✅ above.

Now this didn't work anymore:
image

image

### Findings

- Commando connector worked here: https://github.com/getAlby/lightning-browser-extension/pull/1124/commits/43a4091b6622616493561c4b45e4f20169552599
- Commando connector did not work here anymore: https://github.com/getAlby/lightning-browser-extension/pull/1124/commits/dd85f9aa414176ee55ef75e598bdb931b2bc4e70

What happened in between?: [lnmessage v.01.0](https://github.com/aaronbarnardsound/lnmessage/releases/tag/0.1.0)

### Solution

Reverting back to lnmessage v0.0.19 is fixing this for now

PR #1133, opened 2022-07-11
refactor(allowance): get action

im-adithya:

Switching allowance/delete to TS

### Link this PR to an issue

Relates to #989

### How has this been tested?

Should add tests, would be done after review

### Checklist

### DIFF
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -194,75 +194,65 @@ function Home() {
           </h2>
           {allowance?.payments.length > 0 ? (
             <TransactionsTable
-              transactions={allowance.payments.reduce<Transaction[]>(
-                (acc, payment) => {
-                  if (!payment?.id) return acc;
+              transactions={allowance.payments.map((payment) => {
+                const {
+                  createdAt,
+                  description,
+                  host,
+                  id,
+                  location,
+                  name,
+                  preimage,
+                  totalAmount,
+                  totalFees,
+                } = payment;
 
-                  const {
-                    createdAt,
-                    description,
-                    host,
-                    id,
-                    location,
-                    name,
-                    preimage,
-                    totalAmount,
-                    totalFees,
-                  } = payment;
-
-                  acc.push({
-                    createdAt,
-                    description,
-                    id: `${id}`,
-                    location,
-                    name,
-                    preimage,
-                    host,
-                    totalAmount:
-                      typeof totalAmount === "string"
-                        ? parseInt(totalAmount)
-                        : totalAmount,
-                    totalFees:
-                      typeof totalFees === "number"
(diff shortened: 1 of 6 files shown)


escapedcat (dismissed):
🚀

src/extension/background-script/db.ts
```diff
@@ -1,11 +1,11 @@
 import Dexie from "dexie";
 import "fake-indexeddb/auto";
 import browser from "webextension-polyfill";
-import type { Allowance, Payment, Blocklist } from "~/types";
+import type { Allowance, Transaction, Blocklist } from "~/types";
 
 class DB extends Dexie {
   allowances: Dexie.Table<Allowance, number>;
-  payments: Dexie.Table<Payment, number>;
+  payments: Dexie.Table<Transaction, number>;
```
bumi, line 8: Why that? I don't think this is correct The data in the table does not match all the attributes of `Transaction`

@escapedcat ?

escapedcat: Good point. I thought this has been done for a reason but should have asked. @im-adithya can you elaborate?

im-adithya: I was getting a TS error as we have the `payments` type set to `Transaction[]` in types:

https://github.com/getAlby/lightning-browser-extension/blob/84a2d8fe8b77cf4a7170d65b09bec163eb2d6246/src/extension/background-script/actions/allowances/get.js#L22


https://github.com/getAlby/lightning-browser-extension/blob/84a2d8fe8b77cf4a7170d65b09bec163eb2d6246/src/types.ts#L325

im-adithya: And we have this component saying transactions specifically and also uses Transaction type, so I thought changing the payments type to Payment[] would be breaking things:
https://github.com/getAlby/lightning-browser-extension/blob/84a2d8fe8b77cf4a7170d65b09bec163eb2d6246/src/app/screens/Home/index.tsx#L197

escapedcat: Thanks!  
I think the payments in the db and the payments attached to an allowance should have the type `payment`.  
Currently only the transaction-table is using `transaction` I think.  
Following does that path I [created a new PR](https://github.com/im-adithya/lightning-browser-extension/pull/2) which to extend this PR. Currently there are a lot of changes, but once this one is rebased with our master it should only have one commit left.
I rebased this PR but I can't force push onto this one due to missing permissions. Not sure why. I believe maintainers usually can push into contributor-forks 🤷

escapedcat:
- Typecheck is failing
- Is this needed:
  ```ts
  totalAmount: number;
  ```
  hav ewe stored `totalAmount` in the index-db before? If a user has strings in there, will this break?

escapedcat (approved):

PR #1135, opened 2022-07-11
refactor(allowance): enable action

im-adithya:

Switching allowance/enable to TS

### Link this PR to an issue

Relates to #989

### How has this been tested?

Added tests

### Checklist

### DIFF
--- a/src/extension/background-script/actions/allowances/__tests__/enable.test.ts
+++ b/src/extension/background-script/actions/allowances/__tests__/enable.test.ts
@@ -2,7 +2,7 @@ import { Runtime } from "webextension-polyfill";
 import utils from "~/common/lib/utils";
 import db from "~/extension/background-script/db";
 import state from "~/extension/background-script/state";
-import type { MessageAllowanceEnable } from "~/types";
+import type { DbAllowance, MessageAllowanceEnable } from "~/types";
 
 import enableAllowance from "../enable";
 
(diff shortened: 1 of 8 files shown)


escapedcat:
No need to do it here but do you think we should/can type `const mockAllowances = [...`

im-adithya:
The `bulkAdd` accepts only Allowance[] type, so it gives off errors anyway. But yeah, we should type it as per practice!

escapedcat (approved):
Tested

src/extension/background-script/actions/allowances/__tests__/enable.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      args: {
+        host: "pro.kollider.xyz",
+      },
+    };
+
+    const sender: Runtime.MessageSender = {};
+
+    expect(await enableAllowance(message, sender)).toStrictEqual({
+      data: {
+        enabled: true,
```
lisabaut, line 119: Why not check here for `remember` too? 🤔

lisabaut (approved):
Thanks! 
Just one question of understanding below.

src/extension/background-script/actions/allowances/__tests__/enable.test.ts
im-adithya: It's the test case for an already enabled allowance, in that case we don't open the prompt and hence don't get the `remember` info.

lisabaut: 👍

PR #1138, opened 2022-07-12
Add support to load the exchange rate from the Alby server

bumi:

This add support to load the fiat exchange rates from the Alby server.
This means we no longer by default request data from third-party services (which has privacy implications for the user)

### How has this been tested?

This is a manually tested additional settings option.

### Checklist

### DIFF
--- a/src/app/screens/Settings.tsx
+++ b/src/app/screens/Settings.tsx
@@ -176,6 +176,7 @@ function Settings() {
                   });
                 }}
               >
+                <option value="alby">Alby</option>
                 <option value="coindesk">Coindesk</option>
                 <option value="yadio">yadio</option>
               </Select>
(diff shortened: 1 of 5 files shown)


escapedcat (approved):
Didn't k now we have this. Should always be teh first option. Where do we get the rate from?

bumi:
I added that because of that feedback. It is basically just a proxy for some of the API

src/app/screens/Settings.tsx
```diff
@@ -176,6 +176,7 @@ function Settings() {
                   });
                 }}
               >
+                <option value="alby">Alby</option>
```
escapedcat, line 179: Should we let people know what service we use in the background? i.e. 

> Alby (via Coindesk)

Or... Should all options be implemented via our proxy and the setting will someone let proxy know which service should be used?

Just wondering what users care about here.

PR #1140, opened 2022-07-12
fix: handle stored payment amounts in strings

bumi:
we seem to still store payment amounts in strings.
This causes that numbers are concatenated instead of summed up.

image

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -270,19 +270,19 @@ export type Transaction = {
 };
 
 export interface Payment {
-  id?: number;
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -270,19 +270,19 @@ export type Transaction = {
 };
 
 export interface Payment {
-  id?: number;
   allowanceId: string;
+  createdAt: string;
```
escapedcat, line 274: Just sorted this

PR #1157, opened 2022-07-16
fix: properly propagate errors from LND

bumi:

Errors in the LND#sendPayment call have not properly be propagated and the user did not get a proper error message.

### How has this been tested?

For example trying to pay an invoice from another network (e.g. regtest) causes now an error to be shown. 
not sure how to best write a test for this, but it looks good.

### Checklist


escapedcat (approved):
We don't have any tests for the ln-actions. Could add these step by step later on.

PR #1167, opened 2022-07-18
style: change sats icon to alby logo on onboard success screen

secondl1ght:

Updated the sats icon to the Alby logo on the onboard success screen.

### Link this PR to an issue

https://github.com/getAlby/lightning-browser-extension/pull/1153#issuecomment-1186837656

image

### How has this been tested?

Tested locally on light and dark modes.

### Checklist

### DIFF
--- a/src/app/components/ConnectorForm/index.tsx
+++ b/src/app/components/ConnectorForm/index.tsx
@@ -54,11 +54,18 @@ function ConnectorForm({
                 </video>
               </div>
             ) : (
-              <img
-                src="assets/icons/satsymbol.svg"
-                alt="sats"
-                className="w-64"
-              />
+              <>
+                <img
+                  src="assets/icons/alby_logo.svg"
+                  alt="Alby"
+                  className="block dark:hidden w-64"
+                />
+                <img
+                  src="assets/icons/alby_logo_dark.svg"
+                  alt="Alby"
+                  className="hidden dark:block w-64"
+                />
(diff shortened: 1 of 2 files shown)


escapedcat (approved):

MoritzKa:
ok for me based on the screenshot. 

Here we still have the sats logo right?
image

escapedcat:
Good point! @secondl1ght you wanna have a look where else we might need to switch this from "sats" to "bee"?

secondl1ght:
I noticed that but wasn't sure if we wanted to keep the sats symbol in some places? If not then I can update here as well.

escapedcat:
@MoritzKa do we need to keep it somewhere? For all other connectors we show specific connector images, right? We only use this for the Alby account.

secondl1ght:
> @MoritzKa do we need to keep it somewhere? For all other connectors we show specific connector images, right? We only use this for the Alby account.

Some other connectors also show the sats symbol, maybe we want to update all of them to use the connector specific icon?

MoritzKa:
In my opinion, showing the sats symbol in other connectors is okay for now until we´ll have a bigger redesign.

bumi:
+1 at some point these screens will be redesigned anyway. So I would not spend too much time (unless you think changing the image is such a big improvement)

(I cared a bit about the first page, because that's the fist thing the users sees after installing)

Issue #738, opened 2022-03-24
LNURL-channel support


bumi:
[LNURL-channel](https://github.com/fiatjaf/lnurl-rfc/blob/luds/02.md) describes a way of how wallets and services communicate to request a new channel.

A service like [blocktank](https://blocktank.synonym.to/) uses LNURL. Can we somehow handle those channel requests in Alby and make it easier for those services to open channels to the user's node.

bumi:
This wouldn't be supported by all connectors (which is fine). We can start with LND.

We need a new prompt: https://github.com/getAlby/lightning-browser-extension/blob/39ac28cfe3f2691ee487dfed8da151d6e681e076/src/extension/background-script/actions/lnurl/index.ts#L21-L32
and then add an ln action for channels: https://github.com/getAlby/lightning-browser-extension/tree/39ac28cfe3f2691ee487dfed8da151d6e681e076/src/extension/background-script/actions/ln

escapedcat:
- connectPeer method in connector
- https://api.lightning.community/?javascript#v1-peers
  - `perm: true`
- lnuri response need to be split (`node_key@ip_address:port_number`)
- wait for response and call the `callback`
- who else implemented this already? zeus?
- how can we tests this, which service implemented lnchannels already?
- https://requestbin.net/

MoritzKa:
[these apps](https://github.com/fiatjaf/lnurl-rfc#:~:text=ones%20listed%20below-,02,clich%C3%A9%2C%20SimpleBitcoinWallet%2C%20Zap%20Android%2C%20Zap%20Desktop%2C%20Zeus,-03) implemented it.

PR #1173, opened 2022-07-20
feat(lnurl): support channelRequest #738

escapedcat:

Support LNURLChannel via connectPeer.  
Currently only supports LND.

### Link this PR to an issue

#738

### How has this been tested?

Via devtools

### Checklist


src/extension/background-script/actions/lnurl/index.ts
```diff
@@ -20,10 +21,11 @@ async function lnurl(message: Message) {
 
   switch (lnurlDetails.tag) {
     case "channelRequest":
-      console.error("lnurl-channel is not implemented");
-      // TODO: add support for LNURL channel
-      // connectPeer method in connector
-      return { error: "not implemented" };
+      if ("public" in message) {
```
bumi, line 24: what is this check about? and why only a `return` in the else case?

src/extension/background-script/actions/ln/getBalance.js
```diff
@@ -0,0 +1,16 @@
+import state from "../../state";
+
+const getInfo = async (message, sender) => {
```
bumi, line 3: what is this getBalance.js for? where is it used.
and I think you have a c&p error here => `getInfo`?

bumi (commented):
I still think we should call it lnurlchannel and not lnurlopenchannel because the spec is called lnurl channel and it actually also does not open a channel (it connects a peer and does a channel request to a service)

src/app/screens/LNURLOpenChannel.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  async function confirm() {
+    try {
+      setErrorMessage("");
+      setLoadingConfirm(true);
+      await api.connectPeer({
+        host,
+        pubkey,
+      });
+      // todo: check success
```
escapedcat, line 39: This is outdated, yes?

bumi: +1 looks like 

though we do not check if the `connectPeer` call was successful?

src/extension/background-script/actions/lnurl/index.ts
bumi: hmm. now we got another else there. and all the other lnurl options are missing. 
should that then be 

```js 
lnurl(MessageLNURLChannel | MessageLNURLPay | MessageLNURLAuth | ...)
```

and where is that `public` coming from?

src/app/screens/LNURLOpenChannel.tsx
escapedcat: No, but doesn't this take care of it?
https://github.com/getAlby/lightning-browser-extension/pull/1173/files#diff-75c67e9b6d920ae59f2bb6ec2011595392fa96986711b39bb1e57f36ded8f5ceR21-R28

bumi: ah ok, we have that global `try {` where we catch that error. 🙈

PR #1175, opened 2022-07-20
chore: removing state subscription to persist to browser storage

bumi:
we now manually call the saveToStorage() method whenever needed.
That means we should not need the complexity of these state subscriptions


escapedcat (approved):

escapedcat:
@lisabaut also mentioned this. Nice!

Issue #1141, opened 2022-07-12
Add new setting: disable fiat amounts


bumi:
### Is your feature request related to a problem? Please describe

We recently introduced support to show fiat amounts. But maybe I do not want to use fiat amounts and thus it should be possible to disable the fiat amounts

### Describe the solution you'd like

We should have a setting to choose if fiat amounts should be shown and then (if fiat amounts should be shown) the fiat currency can be choosen.

bumi:
#1125

lisabaut:
@dvoroneca 

As discussed see here the current status of the new toggle which shows/hides the settings "currency" and "exchange":


| Setting enabled  | Setting disabled  |
|---|---|
|![Bildschirmfoto 2022-07-25 um 13 27 26](https://user-images.githubusercontent.com/11243967/180769767-3e4df8d2-e8d3-4bea-83e2-4ff9a2a6b6f0.png)| ![Bildschirmfoto 2022-07-25 um 13 41 10](https://user-images.githubusercontent.com/11243967/180769985-247518d8-5e4b-4cf9-8851-2fec601d99a4.png) |

Happy about some design feedback for better text and maybe new Headers etc 🙂

lisabaut:
I created a follow-up ticket fot the design https://github.com/getAlby/lightning-browser-extension/issues/1224

PR #1177, opened 2022-07-20
feat(app): Show and convert fiats if setting is enabled

lisabaut:

This is the main integration of the new settings feature.

This PR does:
- add a new boolean to the DEFAULT_SETTINGs => `showFiat`
- `showFiat` is `true` by default
- ~Settings: If in development mode show a "Show Fiat" toggle~ => Toggle is shown in production
- Settings: If `showFiat` is `false` => hide settings for "currency"  and "exchange" 
- Add new SettingsContext and Provider which gets the current settings from `api.getSettings` 
- Wrap all screens with this Provider
- Settings: get settings now via SettingsContext and not directly from `api.getSettings` no more
- Settings: ensure to update the SettingsContext state too when saving new settings
- Acounts: Divide `getSats` and `getFiatValues`
- Acounts: get `showFiat`  from SettingsContext
- Accounts: do not fetch fiats if `showFiat` is `false` => fiatValue remains or becomes an empty string
- Do the same for all other places where `getFiatValue` is used (Receive, Pay, Payment, KeySend, Home, Publisher)
- Ensure to not render any data if `fiatValues` is an empty string (AccountMenu, DualCurrency, Transactions)
- addes settings fixtures for tests
- and some minor refactorings

Please checkout this branch and toggle the Setting by yourself to check if something is missing 🙏 

### Next steps:
- [x] talk with Design about the Toggle in Settings view => waiting for update in ticket
- [x] show the Toggle in production mode => **FEEDBACK REQUIRED** if this PR should already render the toggle for production users
- [x] more usages of the hook `useSettings` on component level, see [comment](https://github.com/getAlby/lightning-browser-extension/pull/1177#discussion_r929754374) => new ticket https://github.com/getAlby/lightning-browser-extension/issues/1210

### Link this PR to an issue

Part of https://github.com/getAlby/lightning-browser-extension/issues/1141

### How has this been tested?

I added addiitonal tests where a Test File existed already.

| Setting enabled  | Setting disabled  |
|---|---|
|   |    |
|    |       |
|    |       |
|incoming_with  |  incoming_without  |

### Checklist

### DIFF
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -68,14 +68,16 @@ function AccountMenu({ title, balances, showOptions = true }: Props) {
           {title || <Skeleton />}
         </p>
 
-        {balances.satsBalance && balances.fiatBalance ? (
+        {balances.satsBalance ? (
           <p className="flex justify-between">
             <span className="text-xs dark:text-white">
               {balances.satsBalance}
             </span>
-            <span className="text-xs text-gray-600 dark:text-neutral-400">
-              ~{balances.fiatBalance}
-            </span>
+            {!!balances.fiatBalance && (
+              <span className="text-xs text-gray-600 dark:text-neutral-400">
+                ~{balances.fiatBalance}
+              </span>
+            )}
           </p>
         ) : (
           <Skeleton />
--- a/src/app/components/AllowanceMenu/index.test.tsx
+++ b/src/app/components/AllowanceMenu/index.test.tsx
@@ -1,9 +1,21 @@
 import { act, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { MemoryRouter } from "react-router-dom";
+import * as SettingsContext from "~/app/context/SettingsContext";
+import type { SettingsStorage } from "~/types";
 
 import type { Props } from "./index";
 import AllowanceMenu from "./index";
+
+const mockSettings = {
+  showFiat: true,
+} as SettingsStorage;
+
+jest.spyOn(SettingsContext, "useSettings").mockReturnValue({
+  settings: mockSettings,
+  isLoading: false,
+  updateSetting: jest.fn(),
+});
 
 jest.mock("~/common/lib/utils");
 
--- a/src/app/components/AllowanceMenu/index.test.tsx
+++ b/src/app/components/AllowanceMenu/index.test.tsx
@@ -1,15 +1,11 @@
 import { act, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { MemoryRouter } from "react-router-dom";
+import { settingsFixture as mockSettings } from "~/../tests/fixtures/settings";
 import * as SettingsContext from "~/app/context/SettingsContext";
-import type { SettingsStorage } from "~/types";
 
 import type { Props } from "./index";
 import AllowanceMenu from "./index";
-
-const mockSettings = {
-  showFiat: true,
-} as SettingsStorage;
 
 jest.spyOn(SettingsContext, "useSettings").mockReturnValue({
   settings: mockSettings,
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -2,6 +2,7 @@ import { GearIcon } from "@bitcoin-design/bitcoin-icons-react/filled";
 import { CrossIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
 import { useState, useEffect } from "react";
 import Modal from "react-modal";
+import { useSettings } from "~/app/context/SettingsContext";
 import utils from "~/common/lib/utils";
 import { getFiatValue } from "~/common/utils/currencyConvert";
 import type { Allowance } from "~/types";
@@ -17,18 +18,21 @@ export type Props = {
 };
 
 function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
+  const { isLoading: isLoadingSettings, settings } = useSettings();
+  const showFiat = !isLoadingSettings && settings.showFiat;
+
   const [modalIsOpen, setIsOpen] = useState(false);
   const [budget, setBudget] = useState("0");
   const [fiatAmount, setFiatAmount] = useState("");
 
   useEffect(() => {
-    if (budget !== "") {
+    if (budget !== "" && showFiat) {
       (async () => {
         const res = await getFiatValue(budget);
         setFiatAmount(res);
       })();
     }
-  }, [budget]);
+  }, [budget, showFiat]);
 
   function openModal() {
     setBudget(allowance.totalBudget.toString());
--- a/src/app/components/form/DualCurrencyField/index.tsx
+++ b/src/app/components/form/DualCurrencyField/index.tsx
@@ -4,7 +4,7 @@ import { classNames } from "~/app/utils";
 export type Props = {
   suffix?: string;
   endAdornment?: React.ReactNode;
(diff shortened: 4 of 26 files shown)


src/app/components/AccountMenu/index.tsx
```diff
@@ -68,14 +68,16 @@ function AccountMenu({ title, balances, showOptions = true }: Props) {
...
-        {balances.satsBalance && balances.fiatBalance ? (
+        {balances.satsBalance ? (
           <p className="flex justify-between">
             <span className="text-xs dark:text-white">
               {balances.satsBalance}
             </span>
-            <span className="text-xs text-gray-600 dark:text-neutral-400">
-              ~{balances.fiatBalance}
-            </span>
+            {!!balances.fiatBalance && (
```
escapedcat, line 76: Wondering if when this is active and the API takes a bit longer, will the loading-skeleton is already gone but the fiat value might still be i.e. `~0$`?

escapedcat (commented):
Really awesome PR! Thanks!

Checked it out. Not sure if everything works as supposed or if I'm doing something wrong:
image
After switching the toggle the AccountMenu still shows Fiat. Also after a hard reload. Does this work for you? 

Same for the Popup:
image

Weird

src/app/components/AllowanceMenu/index.test.tsx
```diff
@@ -1,10 +1,22 @@
...
 import { MemoryRouter } from "react-router-dom";
+import * as SettingsContext from "~/app/context/SettingsContext";
+import type { SettingsStorage } from "~/types";
 
 import type { Props } from "./index";
 import AllowanceMenu from "./index";
 
+const mockSettings = {
+  showFiat: true,
+} as SettingsStorage;
```
escapedcat, line 12: Would be nice if we could avoid `as`. Would it work to define all needed props and do `mockSettings:SettingsStorage = {`? Same for the other usages.

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -2,6 +2,7 @@ import { GearIcon } from "@bitcoin-design/bitcoin-icons-react/filled";
 import { CrossIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
 import { useState, useEffect } from "react";
 import Modal from "react-modal";
+import { useSettings } from "~/app/context/SettingsContext";
```
escapedcat, line 5: Maybe we could also replace this then? Not in this but in a next PR?:
https://github.com/getAlby/lightning-browser-extension/blob/feat/738_openchannel/src/app/screens/LNURLPay.tsx/#L100-L109

src/app/components/AllowanceMenu/index.test.tsx
lisabaut: Changes everywhere with adding a new fixtures folder for tests (=> see also the discussion [here](https://github.com/getAlby/lightning-browser-extension/pull/968#discussion_r928803305))

src/app/components/AllowanceMenu/index.tsx
lisabaut: Yes! Is noted and added to the TODOs above in the description

escapedcat (approved):

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -4,7 +4,7 @@ import { classNames } from "~/app/utils";
 export type Props = {
   suffix?: string;
   endAdornment?: React.ReactNode;
-  fiatValue: string | number;
+  fiatValue: string;
```
bumi, line 7: that means we will always have a formatted string in there? 

handling amounts is something I think we need to be careful to not be confusing: 
where do we use strings, numbers, bignumbers, ...? 
is it formatted? how to we calculate with it?

src/app/components/AllowanceMenu/index.tsx
bumi: + the other occurrences where we use `api.getSettings` (currencyConvert.ts, LNURLPay.tsx, utils/index.ts, batteries/index.ts, ...(?)

src/app/screens/Settings.tsx
```diff
@@ -220,12 +153,83 @@ function Settings() {
...
-                loading={loading}
-                disabled={loading}
+                loading={isLoading}
+                disabled={isLoading}
               />
             </div>
           )}
         </Setting>
+
+        {process.env.NODE_ENV === "development" && (
```
bumi, line 163: +1 show it in production mode :shipit:

bumi (approved):
enable it in production mode and then we merge this :shipit: 

what are the downsides of using the context? 
normally when the context changes everything will be re-rendered, correct? 
in the settings context the settings data also does not often change so this is not a problem? - in fact it actually only changes from the settings page currently when `updateSetting` is called? 

can you take care of changing the other usages of the settings that we there also use this new context?

bumi:
can the langauge also be changed from that settings context? 
https://github.com/getAlby/lightning-browser-extension/pull/869

so whenever the settings are loaded we call that `changeLanguage` function

src/app/components/form/DualCurrencyField/index.tsx
escapedcat: This is a valid correction I believe. They value is initaly being created by `getFiatValue` which always returns a string anyway.

escapedcat (approved):

escapedcat:
> can the language also be changed from that settings context? #869
> 
> so whenever the settings are loaded we call that `changeLanguage` function

Yeah, we can take a look once this is merged

src/app/components/form/DualCurrencyField/index.tsx
lisabaut: Yes exactly. Every fiatValue is [formatted here](https://github.com/getAlby/lightning-browser-extension/blob/665ccd0cf611e31f82b07a9a2953707d22bd9db9/src/common/utils/currencyConvert.ts#L96-L111) which always returns a string.

src/app/components/AllowanceMenu/index.tsx
lisabaut: The new hook `useSettings` works only on component /rendering level

I created a new ticket here as a follow up
https://github.com/getAlby/lightning-browser-extension/issues/1210

src/app/components/AccountMenu/index.tsx
lisabaut: Can you explain again what this Skeleton is good for? So far I did not experience any loading issues.

src/app/screens/Settings.tsx
lisabaut: Done ✅

lisabaut:
> enable it in production mode and then we merge this :shipit:
> 

=> done

> what are the downsides of using the context? normally when the context changes everything will be re-rendered, correct? in the settings context the settings data also does not often change so this is not a problem? - in fact it actually only changes from the settings page currently when `updateSetting` is called?
> 

=> yes, all components which read from the settings context with the hook `useSettings` will re-rerender. Which is is intended in this case

> can you take care of changing the other usages of the settings that we there also use this new context?

=> sure, to stay in scope I would do this in follow-up PRs . I assigned myself to this [ticket](https://github.com/getAlby/lightning-browser-extension/issues/1210), feel free to create more tickets and assign me if you have more places in mind where this context should be used.

src/app/components/AccountMenu/index.tsx
escapedcat: Not blocking. We can have a look once this is merged.

lisabaut:
I rebased again and fixed the new view for Incoming Transactions too. 
If the setting "sho fiat" is disabled => no fiats in incoming transactions.
See two new screenshots above.

PR #1178, opened 2022-07-20
feat: Store metadata in db

pavanjoshi914:

feat: store metadata in DB

store metadata in payments table 

covered cases -
1. non-empty invoice valid  metadata
2. non-empty invoice invalid metadata
3. non-empty invoice no metadata
4. empty invoice valid metadata
5. empty invoice invalid metadata
6. empty invoice no metadata
7. validate metadata before making payment
8. allow payments to happen only when metadata is valid
9. no metadata passed is equivalent to valid metadata

updates:
change metadata type to json 

updates:
create new table to store metadata
persist metadata in new table
add additional helpful fields such as AllowanceId, host etc to associate metadata table to other tables
store metadata JSON as well as common attributes in separate column to easily query the metadata
(querying metadata fields + rendering metadata through iterating over objects -  solutions for both done)

image

### How has this been tested?

using prototype and lightning regtest network

1. Existing functionality working fine
image

2. if metadata passed and if it's valid and stored in DB
image
image

3. if  no valid metadata is provided don't allow to proceed with payment

image

4. when no metadata is passed, the code works as before 
image

5. when no metadata is passed but the invoice is valid, the code work as before and allows the user to make payment
image
image

### Checklist

### DIFF
--- a/src/common/lib/utils.ts
+++ b/src/common/lib/utils.ts
@@ -97,11 +97,8 @@ const utils = {
   openUrl: (url: string) => {
     browser.tabs.create({ url });
   },
-  /**
-   *
-   * @param message
-   * @param message.metadata is optional, so that it doesnt conflict with lnurl openPrompt function call
-   */
+  // @param message
+  // @param message.metadata is optional, so that it doesnt conflict with lnurl openPrompt function call
   openPrompt: <Type>(message: {
     args: Record<string, unknown>;
     origin: OriginData | OriginDataInternal;
--- a/src/common/lib/utils.ts
+++ b/src/common/lib/utils.ts
@@ -85,6 +85,7 @@ const utils = {
       details: data.details,
       paymentRequestDetails: data.paymentRequestDetails,
       origin: message.origin,
+      metadata: message.metadata,
     });
   },
   openPage: (page: string) => {
@@ -96,9 +97,15 @@ const utils = {
   openUrl: (url: string) => {
     browser.tabs.create({ url });
   },
+  /**
+   *
+   * @param message
+   * @metadata is optional, so that it doesnt conflict with lnurl openPrompt function call
+   */
   openPrompt: <Type>(message: {
     args: Record<string, unknown>;
     origin: OriginData | OriginDataInternal;
+    metadata?: string;
     action: string;
   }): Promise<{ data: Type }> => {
     const urlParams = new URLSearchParams();
@@ -109,6 +116,9 @@ const utils = {
     // passing on the message origin to the prompt if present
     if (message.origin) {
       urlParams.set("origin", JSON.stringify(message.origin));
+    }
+    if (message.metadata) {
+      urlParams.set("metadata", JSON.stringify(message.metadata));
     }
     // action must always be present, this is used to route the request
     urlParams.set("action", message.action);
--- a/src/common/lib/utils.ts
+++ b/src/common/lib/utils.ts
@@ -100,7 +100,7 @@ const utils = {
   /**
    *
    * @param message
-   * @metadata is optional, so that it doesnt conflict with lnurl openPrompt function call
+   * @param message.metadata is optional, so that it doesnt conflict with lnurl openPrompt function call
    */
   openPrompt: <Type>(message: {
     args: Record<string, unknown>;
--- a/src/common/lib/utils.ts
+++ b/src/common/lib/utils.ts
@@ -118,7 +118,7 @@ const utils = {
       urlParams.set("origin", JSON.stringify(message.origin));
     }
     if (message.metadata) {
-      urlParams.set("metadata", JSON.stringify(message.metadata));
+      urlParams.set("metadata", message.metadata);
     }
     // action must always be present, this is used to route the request
     urlParams.set("action", message.action);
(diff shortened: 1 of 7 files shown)


pavanjoshi914:
@bumi

src/common/lib/utils.ts
```diff
@@ -96,9 +97,15 @@ const utils = {
   openUrl: (url: string) => {
     browser.tabs.create({ url });
   },
+  /**
+   *
+   * @param message
+   * @metadata is optional, so that it doesnt conflict with lnurl openPrompt function call
```
escapedcat, line 103: What's the benefit of the JDDoc comment if we use TS? I'm also confused because it looks like `openPrompt` is getting `message` and `metadata` but actually it's getting a message with optional metadata? Maybe I'm reading it wrong.

escapedcat (commented):
Look nice!

src/common/lib/utils.ts
pavanjoshi914: Not Sure, We were already using JSDoc format so made a quick comment using the same format

pavanjoshi914: Though I documented it in the wrong way corrected this! thanks for pointing out. It shall be ```@param message.metadata .....```

pavanjoshi914: shall I remove the whole comment then?  which format I shall follow currently?

escapedcat: I'd rather get rid of JSDoc comments as logn as we don't really do anything with it. @bumi ?

src/extension/background-script/db.ts
```diff
@@ -14,7 +14,7 @@ class DB extends Dexie {
       allowances:
         "++id,&host,name,imageURL,tag,enabled,totalBudget,remainingBudget,lastPaymentAt,lnurlAuth,createdAt",
       payments:
-        "++id,allowanceId,host,location,name,description,totalAmount,totalFees,preimage,paymentRequest,paymentHash,destination,createdAt",
+        "++id,allowanceId,host,location,name,description,totalAmount,totalFees,preimage,paymentRequest,metadata,paymentHash,destination,createdAt",
```
bumi, line 17: now we store metadata as a string. I was thinking if we store it as JSON then it should be easier to search for certain things. 

maybe it is also useful to use a new table for this and then split up the metadata into different columns? do you have thoughts on this?

src/extension/background-script/db.ts
```diff
@@ -14,7 +14,7 @@ class DB extends Dexie {
...
     });
     this.version(2).stores({
```
bumi, line 19: the DB above must never be changed. 
in Dexie you always create new versions to update. 
check the docs: https://dexie.org/docs/Dexie/Dexie.version()#sample

src/extension/background-script/db.ts
pavanjoshi914: Sure! I can experiment and make changes when we decide how and what content we render in the transaction list (already there is PR representing rendering of metadata), in whichever cases we need to query metadata.


Now I am thinking of puttin metadata in a format of JSON as object

IndexedDB stores javascript objects. So if your documents are JSON docs, they can be perfectly stored as objects (such as the result from for example JSON.parse(json)). If you choose to store the JSON as text instead, that would work as well, but in case you like to index certain properties, you must store them as objects.


JSON as a blob has few drawback according to the stackoverflow post

```
Putting a JSON blob in a column makes it virtually impossible to query (and painfully slow when you actually find a query that works)
```

Reference::

https://github.com/dexie/Dexie.js/issues/823
https://stackoverflow.com/questions/15367696/storing-json-in-database-vs-having-a-new-column-for-each-key


regarding making a relational table and storing metadata as one attribute per column might have its own complexities but sure until I try! but as dexie can store javascript objects, this might not needed if query is simple to extract fields from the metadata object

src/extension/background-script/db.ts
pavanjoshi914: making this change

src/common/lib/utils.ts
pavanjoshi914: done

pavanjoshi914:
@bumi

pavanjoshi914:
@bumi can you merge this! so that I can build next improvement on top of this

src/extension/background-script/db.ts
```diff
@@ -19,6 +19,10 @@ class DB extends Dexie {
...
       blocklist: "++id,host,name,imageURL,isBlocked,createdAt",
     });
+    this.version(3).stores({
+      payments:
+        "++id,allowanceId,host,location,name,description,totalAmount,totalFees,preimage,paymentRequest,metadata,paymentHash,destination,createdAt",
```
bumi, line 24: still thinking about leaving `payments` as it is and adding a new table metadata with the metadata entries associated to a payment.

this would leave the current data in payment untouched and we can easily always extend the metadata table with new columns independent of the rest of the app

pavanjoshi914: image

src/extension/background-script/db.ts
```diff
@@ -19,20 +20,27 @@ class DB extends Dexie {
     this.version(2).stores({
       blocklist: "++id,host,name,imageURL,isBlocked,createdAt",
     });
+    this.version(3).stores({
+      metadata:
+        "++id,allowanceId,host,metadata,paymentHash,createdAt,type",
```
bumi, line 25: I think this is great, we can expand that further if needed.

PR #1187, opened 2022-07-22
Fix url for raspiblitz v1.7.2

BtcContributor:
This PR fixes the wrong suggestion for Raspiblitz.

This has been tested with the latest Raspiblitz version v1.7.2.


MoritzKa:
Tested. It works.

MoritzKa (approved):

bumi:
so the current one is wrong?

MoritzKa:
yes, the current one is wrong. The sudo command shows another onion address.
I wrote the guide when I had an older Raspiblitz version installed.

PR #1204, opened 2022-07-28
chore: remove account export in onboarding

bumi:
People are not aware that this is sensitive data and share screenshot and videos
Export is now possible from the export function

image

### DIFF
--- a/src/app/screens/connectors/NewWallet/index.tsx
+++ b/src/app/screens/connectors/NewWallet/index.tsx
@@ -13,13 +13,14 @@ import utils from "~/common/lib/utils";
 const walletCreateUrl =
   process.env.WALLET_CREATE_URL || "https://app.regtest.getalby.com/api/users";
 
+interface LNDHubCreateResponse {
+  login: string;
+  password: string;
+  url: string;
+  lnAddress: string;
+}
+
 export default function NewWallet() {
-  const [lndHubData, setLndHubData] = useState({
-    login: "",
-    password: "",
-    url: "",
-    lnAddress: "",
-  });
   const [email, setEmail] = useState("");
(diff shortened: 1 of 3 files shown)


kiwiidb:
You mean: export is "now" possible from the export function?

kiwiidb (approved):
LGTM

bumi:
uuups. yes, it's possible

escapedcat (approved):

bumi:
I've removed the empty success page and directly redirect to the final success page. 

@escapedcat how do I deal with removing translation strigs?

src/app/screens/connectors/NewWallet/index.tsx
```diff
@@ -63,24 +65,22 @@ export default function NewWallet() {
...
           `${t("pre_connect.errors.create_wallet_error")} ${e.message}`
         );
-      })
-      .finally(() => {
         setLoading(false);
       });
   }
 
-  async function next(event: React.FormEvent<HTMLFormElement>) {
+  async function next(lndhub: LNDHubCreateResponse) {
```
escapedcat, line 80: Would it be easier to understand if we call this i.e. `postSignUp` instead of next?

escapedcat (approved):

src/app/screens/connectors/NewWallet/index.tsx
bumi: it was next before, so I did not change it. do you care, should I change it?

escapedcat: No, not blocking. I'll keep this in mind.

escapedcat:
Green!

PR #1218, opened 2022-07-30
Chore/deps swc 220 plus workaround

escapedcat:
Latest swc updates break our Jest setup. There's is a workaround but we still need to adjust our setup.  
Staying on `v205` is currently working for us.

## SWR and Jest

See this discussion for SWR with Jest
=> https://github.com/swc-project/swc/discussions/5151#discussioncomment-3149154


bumi:
what is this one about?

escapedcat:
> what is this one about?

Adjusted the description

bumi:
is this still active or can we close it?

escapedcat:
> is this still active or can we close it?

We should do this

bumi:
so then let's solve this. July is already very old.

escapedcat:
It's about priorities. Happy to touch this after the manifest.

lisabaut (approved):
Thanks, I added the link to the swr discussion to the description for later documentation.

I think, your current fix with removing `spyOn` is ok for now.

escapedcat:
> Thanks, I added the link to the swr discussion to the description for later documentation.
> 
> I think, your current fix with removing `spyOn` is ok for now.

Thanks, already watching that discussion.

Issue #1216, opened 2022-07-30
Support all LNURLs from the send popup-screen


escapedcat:
### Is your feature request related to a problem?

<img width="200" alt="image" src="https://user-images.githubusercontent.com/1016218/181875990-b60580fc-5f6e-4cd3-b2d2-2d386925f497.png">

- The input says: "Paste invoice, lnurl or lightning address"
- Not all LNURLs are actually supported through this

#### Follow up issues

- [ ] #1268
- [ ] #1269

### Describe the solution you'd like

LNURLs work via the prompt and one via the popup. But at least all prompt-supported ones should also work via the popup.

### Additional context

- current popup code:
    - https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Popup/Popup.tsx#L36
    - https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Send.tsx#L37-L39
- prompt code:
    - https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Prompt/Prompt.tsx/#L74-L121

The prompt code get's the details via the router. This should also be done in the popup router. If the functionality is the same it could be moved out.

### Are you working on this?

No

bumi:
when I look at this then this manual loading of data from the URL and passing it in the props looks wrong to me: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Prompt/Prompt.tsx/#L28-L49

we have those two situations: 
1. the LNURL* components are rendered from the prompt. The prompt is opened from the `utils.openPrompt` function. There we pass the details in a query string. 
2. the LNURL* components are rendered in the popup and options page. There we navigate to those components.

Can we use some `state` that either gets populated when the prompt is loaded or passed in with the `navigate()` call? - instead of the props?

bumi:
or can we load some page in the prompt which then also just navigates to the correct page? - to make sure the LNURL* components (and others) are always loaded the same way? 

`navigate("/lnurlChannel", {state: lnurlDetails })`

bumi:
(actually we have this problem with keysend and other components, too

reneaaron:
This should also cover lnurl-auth links, right? That would allow cross-device sign-ins (logging in on your mobile phone while Alby is installed your desktop PC), too! 💯

escapedcat:
The base approach was merged. Further support will be done in split out tickets. See description above.

Issue #792, opened 2022-04-13
Support keysend payments in all the tipping features.


bumi:
**Is your feature request related to a problem? Please describe.**
Currently we support sending tips through lnurl/lightning addresses. People have to have lightning address and add that one to their social media profiles (Youtube, GitHub, Twitter, etc.) 
This leaves out users who do not support LNURL. 

**Describe the solution you'd like**
Additionally to LNURL payments we can also support keysend payments. Users can either provide a lightning address OR a node pubkey in their profiles. 
Alby checks if it is a pubkey and uses keysend otherwise it uses the current lnurl pay flow. 

We would need to adjust the regex in all the batteries (e.g. https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/content-script/batteries/GitHub.ts#L21) 

In the lightning data we already have the `method` key which is currently hard-coded to lnurl: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/content-script/batteries/GitHub.ts#L42

For keysend data we would probably need to navigate to the keysend screen: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/index.tsx#L100-L115

(alternatively we could think of a new special tipping screen?)

joelklabo:
PR #3487 addresses this — adds pubkey detection to the Monetization battery so keysend tipping works via `<meta name="lightning">` tags.

Supports compressed (02/03 prefix) and uncompressed (04 prefix) pubkeys, plus `key=value` format with custom TLV records. 13 tests covering all detection paths.

Builds and passes all existing tests.

patoo0x:
PR #3497 adds keysend tipping support for this issue: https://github.com/getAlby/lightning-browser-extension/pull/3497

**What was built:**
- `isPubKey()` helper that recognises compressed (02/03) and uncompressed (04) Lightning node pubkeys
- Monetization battery updated to auto-detect bare pubkeys in `<meta name="lightning">` tags and set `method: "keysend"`, routing to the existing Keysend screen via the already-implemented `PublisherLnData` logic
- 24 tests (all passing) covering both the helper and the full detection pipeline

Note: PR #3487 also addresses this — our implementation differs in that `isPubKey` is a separately exported utility (making it testable in isolation) and the detection logic handles the edge case where key=value format lacks an explicit method field but has a pubkey address.

PR #1223, opened 2022-08-01
feat: support keysend in lightning meta tag

bumi:

This adds support for keysend payments with the lightning meta tag. 
I also think that the meta tag should follow a similar naming and usage as the `podcast:value` tag. 

the meta tag can now define: 
* `method`: lnurl or keysend
* `address`: the recipient (lnurl/lightning address or node key)
* `customKey`/`customValue`: custom record for keysend

### ToDo:

- [ ] update meta tag spec
- [ ] remove passing data by search params (see comments in #1216)

### Link this PR to an issue

#792 

### How has this been tested?

example meta tag:

```
<meta name="lightning" content="method=keysend; address=030a58b8653d32b99200a2334cfe913e51dc7d155aa0116c176657a4f1722677a3; customKey=696969; customValue=3wQCCrfOAMYNzOh1sL05" />
```

### Checklist


bumi:
How do I do this in proper TS: https://github.com/getAlby/lightning-browser-extension/pull/1223/files#diff-cdee507dd9e4243d8b288907c7c1d3924972f7c47c0a08b5c3d16fe079efad30R16

im-adithya:
> How do I do this in proper TS: https://github.com/getAlby/lightning-browser-extension/pull/1223/files#diff-cdee507dd9e4243d8b288907c7c1d3924972f7c47c0a08b5c3d16fe079efad30R16


Not quite sure, but this can be one way to avoid errors:
```
  const obj: Record<string, string>= {};
           ------------------------
  tokens.forEach((t) => {
    const kv = t.split("=").map((e) => e.trim());
    obj[kv[0]] = kv[1];
  });
  return obj as unknown as Battery;
             -----------
```

escapedcat:
>   return obj as unknown as Battery;

I guess this would be a ok workaround for now till we took care of how to handle transforming URL-params into matching types later.

bumi:
> I guess this would be a ok workaround for now till we took care of how to handle transforming URL-params into matching types later.

those are actually unrelated of the URL-params.
this data is just loaded and used in the Home screen to show the "Send sats" button.

PR #1228, opened 2022-08-01
refactor: better context integrations

lisabaut:

This is a small refactoring PR:

- wrap all views in `<Providers>` component as all use the same providers
- remove all providers from `Welcome` screen as none of the child components need/read any data from any context
- component `AccountMenu` is already reading the current accountInfo from the `AccountContext` with the hook `useAccount()` 
- => no need to pass the `title` and `balances` via multiple Parent Components as the AccountMenu can read them by itself already
- Fix updating the account name after Editing the name on the Accounts Screen with calling `fetchAccountInfo` from AccountContext

Immediately update account name in all occasions

Bildschirmfoto 2022-08-01 um 16 32 27

### How has this been tested?

Only tested manually. Let me know if I can write better tests. i am not familiar with storybook.

### Checklist

### DIFF
--- a/src/app/components/AccountMenu/AccountMenu.stories.tsx
+++ b/src/app/components/AccountMenu/AccountMenu.stories.tsx
@@ -4,12 +4,7 @@ import { AccountsProvider } from "~/app/context/AccountsContext";
 
 import AcountMenu from ".";
 
-export const Default = () => (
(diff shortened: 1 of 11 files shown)


escapedcat (approved):
lgtm

Issue #1226, opened 2022-08-01
Publisher screen: TransactionList partly broken


escapedcat:
### Describe the bug

- FIAT missing
- tx info is broken

### Screenshots

![image](https://user-images.githubusercontent.com/1016218/182123003-8f56f107-809c-4dc9-8034-bbe9ddc72833.png)

### To Reproduce

In current master branch:

1. Visit to i.e. a page you made payments before
1. Click on Alby extension
1. See transaction list

### Expected behavior

Looks like transaction list in home-screen

### Alby information

master branch

### Are you working on this issue?

Yes

bumi:
I think this is a bug that must not go live on this and must be fixed. 
what do you need to answer "Are you working on this issue? - No, but I could" ? 

I am worried that with this issue it lays around for too long.

bumi:
this is not good: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/index.tsx#L275-L278

bumi:
this seems to me to be an issue because on all the different places we handle the payments differently and always map() over payments to do thing. 

With this setup I think we will always miss places when we make changes.
I think the payments must be returned by an action and they must always return the same expected value.

bumi:
I think this one either needs to load the payments for an allowance or all payments: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/index.tsx#L87-L101

so when an allowance is `set` the payments must be loaded. 

we must fix this before the release.

I don't know how to trigger that load after a `setAllowance` call? 
hackhack:
```diff
lightning/lightning-browser-extension@master ↑  » cat d.diff
diff --git a/src/app/screens/Home/index.tsx b/src/app/screens/Home/index.tsx
index f04c6ecb..b048f205 100644
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -59,6 +59,7 @@ function Home() {
       const result = await api.getAllowance(url.host);
       if (result.enabled) {
         setAllowance(result);
+        loadPayments();
       }
       const blocklistResult = await api.getBlocklist(url.host);
       if (blocklistResult.blocked) {
@@ -86,7 +87,16 @@ function Home() {

   const loadPayments = useCallback(async () => {
     try {
-      const result = await api.getPayments({ limit: 10 });
+      let result;
+      console.log({ allowance });
+      if (allowance) {
+        result = await api.getPaymentsByHost({
[...]

PR #1230, opened 2022-08-02
fix(home): allowance payments in tx-list #1226

escapedcat:

- Transform allowance payments to transactions before being used with the tx-list
- We already did something similar with the general payments

### Link this PR to an issue

#1226

image

### How has this been tested?

Manually

### Checklist


src/app/screens/Home/index.tsx
```diff
@@ -86,19 +86,35 @@ function Home() {
...
   const loadPayments = useCallback(async () => {
     try {
-      const result = await api.getPayments({ limit: 10 });
-      for await (const payment of result.payments) {
-        const totalAmountFiat = settings.showFiat
-          ? await getFiatValue(payment.totalAmount)
-          : "";
-        payment.totalAmountFiat = totalAmountFiat;
+      if (allowance) {
+        const payments: Transaction[] = allowance.payments.map((payment) => ({
```
bumi, line 90: I would change this and do `api.getPayments` call with a host (we don't set an allowance Id).

then these two cases do not need to be handled differently and in this screen they are exactly the same. 

see my comment with that diff

bumi (changes requested):

Issue #1192, opened 2022-07-25
Changing the language does not work


AnastasiaVolk:
**Bug description**

In the settings there is an option to change language, but it doesn't work

**Screenshots**

![Image](https://user-images.githubusercontent.com/108762093/180831201-bc43a9fa-d94c-4bb7-ac82-7a97e309ce1f.png)

**Steps to reproduce the behaviour:**

1. Click on Alby Extension Icon
2. Pop up window appears
3. Click on the "menu" button in the right
4. The menu appears
5. Click on "settings"
6. "Settings" page appears
7. In the line "Language" try to change it

**Expected behavior**

The language changes to the chosen one

**Suggestion**

Either remove the option for changing or make it work

**Alby information**

Alby Version: [v1.12.0]
Alby installed through Google Chrome

**Device information:**

OS: [macOS Monterey 12.0.1 ] 
Browser [chrome]  Browser Version [Chrome Version 103.0.5060.114]

**Are you working on this issue?**

No

escapedcat:
Not all languages have been translated completely yet. This will happen over time [via Weblate](https://hosted.weblate.org/projects/getalby-lightning-browser-extension/getalby-lightning-browser-extension/). I think this can be closed.

AnastasiaVolk:
Maybe it makes sense to remove this option before it's done? It gives the impression of a very raw product

escapedcat:
True, I understand that. Doubt translations will be 100% finished soon. We kinda wanted this in so people are aware of it.

Maybe we could improve the info here:
![image](https://user-images.githubusercontent.com/1016218/180947866-4cf02e47-7ded-428a-9df8-a5b1603f7c51.png)
1. Change to "Translations are not 100% finished yet. Help us translate Alby into your language!
2. "translate Alby into your language" link this to the Weblate service we use for translations

AnastasiaVolk:
Yes, then it turns out to be even useful! I love it!

P.S. I think translations are very important, actually. For example, most Russians don't speak English at all. Neither do Spaniards. It might even stop them from using the product.

secondl1ght:
I agree that the current setup is not correct, we should only show a language in the selector when the entire app has been translated. It doesn't make sense to show it otherwise, so we could only show English for now and then when we have full translations add the other languages.

escapedcat:
For the entire app to be translatable we need to switch all strings to i18n placeholders first. You up for it?

secondl1ght:
[...]

PR #1244, opened 2022-08-05
refactor: update copy, add link and remove unfinished language selector

secondl1ght:

Updated copy for language selector in Settings based on user feedback, also added a link to the Weblate for translations, and removed unfinished languages to avoid further confusion. 

### Link this PR to an issue

#1192 

image

### How has this been tested?
Ran locally

### Checklist

### DIFF
--- a/src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
+++ b/src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
@@ -32,7 +32,6 @@ export default function LocaleSwitcher() {
   return (
     <Select name="locale" value={dropdownLang} onChange={languageHandler}>
       <option value="en">English</option>
-      <option value="hi">हिन्दी</option>
     </Select>
   );
 }
--- a/src/app/components/Setting/index.tsx
+++ b/src/app/components/Setting/index.tsx
@@ -13,8 +13,7 @@ function Setting({ title, subtitle, children, link }: Props) {
           {title}
         </span>
         <p className="text-gray-500 dark:text-neutral-500 text-sm">
-          {subtitle}
-          {link && <a href={link[0]}>{link[1]}</a>}
+          {subtitle} {link && <a href={link[0]}>{link[1]}</a>}
         </p>
       </div>
       {children}
--- a/src/app/components/Setting/index.tsx
+++ b/src/app/components/Setting/index.tsx
@@ -1,11 +1,10 @@
 type Props = {
   title: string;
-  subtitle: string;
+  subtitle: string | React.ReactNode;
   children: React.ReactNode;
-  link?: string[];
 };
 
-function Setting({ title, subtitle, children, link }: Props) {
+function Setting({ title, subtitle, children }: Props) {
   return (
     <div className="py-4 flex justify-between items-center">
       <div>
@@ -13,7 +12,7 @@ function Setting({ title, subtitle, children, link }: Props) {
           {title}
         </span>
         <p className="text-gray-500 dark:text-neutral-500 text-sm">
-          {subtitle} {link && <a href={link[0]}>{link[1]}</a>}
+          {subtitle}
         </p>
       </div>
       {children}
--- a/src/app/components/Setting/index.tsx
+++ b/src/app/components/Setting/index.tsx
@@ -2,9 +2,10 @@ type Props = {
   title: string;
   subtitle: string;
   children: React.ReactNode;
+  link?: string[];
 };
 
-function Setting({ title, subtitle, children }: Props) {
+function Setting({ title, subtitle, children, link }: Props) {
   return (
     <div className="py-4 flex justify-between items-center">
       <div>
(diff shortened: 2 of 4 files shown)


escapedcat:
Would this work to avoid an extra `link`-prop?:
https://stackoverflow.com/a/70769183/1667461

secondl1ght:
> Would this work to avoid an extra `link`-prop?: https://stackoverflow.com/a/70769183/1667461

Can you pass a React component in as a component prop though, I thought you could not? I can try and see.

src/app/components/Setting/index.tsx
```diff
@@ -2,17 +2,18 @@ type Props = {
   title: string;
   subtitle: string;
   children: React.ReactNode;
+  link?: string[];
 };
 
-function Setting({ title, subtitle, children }: Props) {
+function Setting({ title, subtitle, children, link }: Props) {
```
bumi, line 8: I am unhappy about this. that' is the problem with such abstractions... you loose flexibility. 
If we now add custom code or attributes here and there it increases a lot of the complexity. 
No other Setting needs the link, so I would actually avoid adding this. is that link really that valuable?
and if? can we pass in html as the subtitle? (without removing escaping)

bumi (changes requested):

src/app/components/Setting/index.tsx
bumi: `Trans` sounds interesting. 

I am also OK with removing the link

secondl1ght: I was able to remove the prop and use the `Trans` comp.

secondl1ght:
Refactored to remove the new prop and use the `Trans` component.

src/app/screens/Settings.tsx
```diff
@@ -99,7 +99,17 @@ function Settings() {
             </p>
           )}
         </Setting>
-        <Setting title={t("language.title")} subtitle={t("language.subtitle")}>
+        <Setting
+          title={t("language.title")}
+          subtitle={
```
escapedcat, line 104: Would be great if we could avoid the extra link prop.  
Have a look at this: https://www.createit.com/blog/i18next-react-links-inside-translations/  
This seems more flexible. We might have these kinda situations more often in the future with multiple links.

secondl1ght: What do you mean the remove extra link prop? I had a look at that article and im not sure what problem this is solving for what we are doing here?

src/i18n/locales/en/translation.json
```diff
@@ -205,7 +205,8 @@
       },
       "language": {
         "title": "Language",
-        "subtitle": "Alby goes international! help us translate Alby in your language"
+        "subtitle": "Translations are not 100% finished yet.",
```
escapedcat, line 208:
```js
{
  "gdpr_policy_links": "I agree to <link1>Privacy Policy</link1> and <link2>Terms and Conditions</link2>",
  "gdpr_required_msg": "Please accept Privacy Policy"
}
```

This looks like we wouldn't need `link` and `subtitle` but only `subtitle` if we use this right. 
I hope we can do this:
```js
{
  "subtitle": "Translations are not 100% finished yet. <link>Help us translate Alby into your language!</link>",
}
```

bumi:
I am kinda lost here. what is this `Trans` doing exactly? 
and do we really need that link?

escapedcat:
> I am kinda lost here. what is this `Trans` doing exactly?

[Trans helps us to handle html-tags](https://www.createit.com/blog/i18next-react-links-inside-translations/) within translations strings. We already use it in other places.  
Getting familiar with its usage will help with all further situations where we have to deal with html-tags.

> and do we really need that link?

I believe it's a good starting point for people who might be interested in adding languages.

bumi:
ok, then can this be merged?

escapedcat:
> ok, then can this be merged?

No, we should get rid of the extra `link`-prop

escapedcat (approved):

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
```diff
@@ -32,7 +32,6 @@ export default function LocaleSwitcher() {
   return (
     <Select name="locale" value={dropdownLang} onChange={languageHandler}>
       <option value="en">English</option>
-      <option value="hi">हिन्दी</option>
```
bumi, line 35: we actually got many languages now and some are pretty much completely translated (from what can be translated right now)
Ideally we load those even dynamically here

bumi (commented):
If we show the current languages in the dropdown, then I am ok

src/app/components/LocaleSwitcher/LocaleSwitcher.tsx
escapedcat: Yeah, possibly. On the other hand a new language would need to wait for a new release or a patch release. Not that hard to manage, right? Let's talk about it.

bumi: not hard, but right now it's not there. let's add it.

PR #1246, opened 2022-08-06
Geyser battery

secondl1ght:

I have added another battery detector for Geyser crowdfunding projects, you can now fund projects directly with Alby.

### Link this PR to an issue

No related issue but this was discussed on Twitter with @bumi 

image
image

### How has this been tested?

Ran locally and was able to get Alby to detect the lightning address and other attributes on the Geyser website.

### Checklist

### Note

I am putting in a PR with Geyser as well the finalize this - it will work currently but I am adding more selectors so that we can use a better avatar and project title as seen in the sceenshots.

### DIFF
--- a/src/extension/content-script/batteries/GeyserProject.ts
+++ b/src/extension/content-script/batteries/GeyserProject.ts
@@ -28,8 +28,7 @@ function handleProjectPage(name: string) {
         method: "lnurl",
         address: address,
         ...getOriginData(),
-        description:
-          "Geyser - global bitcoin crowdfunding that helps builders and creators fund their ideas.",
+        description: "geyser.fund",
         name: title ?? name,
         icon: icon ?? "https://geyser.fund/logo-brand.svg",
       },
--- a/src/extension/content-script/batteries/GeyserProject.ts
+++ b/src/extension/content-script/batteries/GeyserProject.ts
@@ -1,0 +1,41 @@
+import getOriginData from "../originData";
+import { setLightningData } from "./helpers";
+
+const urlMatcher = /^https:\/\/geyser.fund\/([^/]+)(\/([^/]+))?$/;
+
+const battery = (): void => {
+  const urlParts = document.location.pathname.split("/");
+  const project = urlParts[1];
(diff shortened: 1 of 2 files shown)


secondl1ght:
The failing test seems unrelated to this PR.

escapedcat:
> The failing test seems unrelated to this PR.

Yes, the publisher one is still flaky. Will check.

escapedcat (approved):
lgtm

bumi (approved):
this is great! 
there is one problem: geyser.fund loads the data asynchronously. This means the lightning address is not yet in the DOM when we search for the lightning address. 

that's why the icon does not change, but it works when you click on the Alby icon (then we do another check)

src/extension/content-script/batteries/GeyserProject.ts
```diff
@@ -0,0 +1,45 @@
...
+  )?.src;
+
+  if (address) {
+    setLightningData([
+      {
+        method: "lnurl",
+        address: address,
+        ...getOriginData(),
+        description:
+          "Geyser - global bitcoin crowdfunding that helps builders and creators fund their ideas.",
```
bumi, line 32: do we need this? it's a bit long imp. 
```suggestion
        description: "geyser.fund",
```

bumi (approved):

secondl1ght:
> this is great! there is one problem: geyser.fund loads the data asynchronously. This means the lightning address is not yet in the DOM when we search for the lightning address.
> 
> that's why the icon does not change, but it works when you click on the Alby icon (then we do another check)

Ah ok, I know Geyser is looking at pre-rendering the pages eventually because this causes other issues as well like poor SEO and open graph social sharing (so this will be solved at some point).

Also Geyser will add dynamic Alby `meta` tags to the projects too, so at that point we can depreciate this battery because Alby will be able to use the `meta` tags instead which will be better for maintainability in the future. But this solution works for now!

secondl1ght:
@bumi the Geyser PR to accommodate this has been merged - so the selectors for this PR are now live and ready to go!

PR #1250, opened 2022-08-06
fix: remove decodeURI from boostagram message

bumi:
the message is a random string provided by the user


escapedcat (approved):

escapedcat:
Thanks! I wasn't aware of this: https://reactjs.org/docs/introducing-jsx.html#jsx-prevents-injection-attacks  
decodeURI was a stupid attempt.
Adjusted the tests.

bumi:
ah, thanks for the test!!

PR #1256, opened 2022-08-08
fix: use container for responsiveness

im-adithya:

Modifies the Container component and uses it for responsiveness.

### Link this PR to an issue

Extends #1118 

Screenshot from 2022-08-08 21-28-50

### How has this been tested?

### Checklist

### DIFF
--- a/src/app/components/Container.tsx
+++ b/src/app/components/Container.tsx
@@ -2,10 +2,11 @@ type MaxWidth = "sm" | "md" | "lg" | "xl" | "2xl";
 
 type Props = {
   children: React.ReactNode;
+  isScreenView?: boolean;
   maxWidth?: MaxWidth;
 };
 
-function Container({ children, maxWidth = "lg" }: Props) {
+function Container({ children, isScreenView = false, maxWidth = "lg" }: Props) {
   // Avoid dynamically created class strings as PurgeCSS doesn't understand this.
   const getMaxWidthClass = (maxWidth: MaxWidth) => {
     switch (maxWidth) {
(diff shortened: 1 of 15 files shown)


escapedcat:
Tested with "Send". lgtm

src/app/components/Container.tsx
```diff
@@ -2,10 +2,11 @@ type MaxWidth = "sm" | "md" | "lg" | "xl" | "2xl";
 
 type Props = {
   children: React.ReactNode;
+  screen?: boolean;
```
escapedcat, line 5: Can we call it `isScreenView` to indicate it's boolean and it's related to views? Otherwise it's too generic I'm worried.

escapedcat (changes requested):

src/app/components/Container.tsx
im-adithya: Done!

src/app/components/Container.tsx
```diff
@@ -23,7 +24,13 @@ function Container({ children, maxWidth = "lg" }: Props) {
   };
 
   return (
-    <div className={`container ${getMaxWidthClass(maxWidth)} mx-auto px-4`}>
+    <div
+      className={`container mx-auto px-4 ${getMaxWidthClass(maxWidth)} ${
+        isScreenView
+          ? "h-full flex flex-col justify-between overflow-y-auto no-scrollbar"
```
bumi, line 30: no way we can avoid this two cases here? 
how did we do it before, or on the receive page?

bumi:
@im-adithya could you check the conflict?

src/app/components/Container.tsx
im-adithya: The problem is, right now, we want to push the ConfirmOrCancel to the bottom, and for that, we're using flex and justify-between vertically apart from that, there's nothing new (except for h-full to make sure the height used for distribution is the entire space available)

im-adithya:
Done!

escapedcat (approved):

src/app/components/Container.tsx
bumi: cc @escapedcat so maybe if that "pushing the ConfirmOrCancel to the bottom" is too complicated, we should rethink/drop that at some point.

escapedcat: Maybe we eventually will end up with a sticky bottom-bar for that? I'll mention this in our next ui/ux call

PR #1235, opened 2022-08-03
refactor(prompt): webln route handling #1216

escapedcat:
### Describe the changes you have made in this PR

- LNURL-components expect the same data-handling via prompt/popup
- data-handling is split from UI-rendering
- location state is being used (no props or url-query-params)
- finally all LNURLs should be supprted via prompt and popup

## TODOS
- [ ] types for `args` from searchparams
- [ ] check if `isPrompt` conditions can be handled in Parent Component (e.g. all `reject` handlings are the same)
- [ ] tests for all helpers in `Prompt.tsx`
- [ ] add Loading State for Parent Component before the Form can be rendered?
- [ ] add Tests for Parent Components

### Link this PR to an issue

#1216

### Type of change (Remove other not matching type)

- `feat`: New feature (non-breaking change which adds functionality)

### Screenshots of the changes

_Add screenshots to help explain your problem_

### How has this been tested?

_Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration_

### Checklist

- [ ] My code follows the style guidelines of this project and performed a self-review of my own code
- [ ] New and existing tests pass locally with my changes
- [ ] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

PR #1257, opened 2022-08-08
refactor(lnurlpay): replace searchParams with location props

lisabaut:

This PR simplifies the navigation to `/lnurlpay` with passing a location state to `navigate` from `react-router-dom`.

This PR does:
- convert `content-script` to TS and add a lot of types accordingly
- refactor how `Prompt` is reading the searchparams from `window`
- pass data from searchParams to all `Routes` via location state in `Prompt`
- add a new hook `useNavigationState` to read the location state 
- wrap `LNURLpay` in a Parent component which checks if `origin` and `args` in location state exist and fetchs `lnurlDetails` on missing data accordingly

This PR cherry-picked some commits of the already ongoing PR https://github.com/getAlby/lightning-browser-extension/pull/1235 which is already growing in complexity. 
To make a review more easy this PR focuses only on `LNURLPay` plus all the TypeScript changes and other refactorings. 
**This PR should also serve as a proof of concept.**

### Link this PR to an issue

#1216

### How has this been tested?

On a Website Console:
- `await webln.enable()`
- `await window.webln.lnurl("A-LNURL");`

In Extension:
- Popup
  - Open `Send`
  - paste in a `LNURL`
- Via i.e. embedded twitter LN-address

### Checklist


src/app/screens/Home/index.tsx
```diff
@@ -242,19 +244,21 @@ function Home() {
...
                     description: lnData[0].description,
                     icon: lnData[0].icon,
                   };
                   if (lnData[0].method === "lnurl") {
-                    navigate(
-                      `/lnurlPay?lnurl=${
-                        lnData[0].address
-                      }&origin=${encodeURIComponent(JSON.stringify(origin))}`
-                    );
+                    // HOW CAN THIS BE TESTED ON HOME !?!
```
bumi, line 255: go to some website, or twiter or some youtube video that has the lightning address in it and click on the Alby icon. 

https://twitter.com/escapedcat
https://www.youtube.com/watch?v=RFSBWrAllzw

src/common/lib/utils.ts
```diff
@@ -102,7 +102,7 @@ const utils = {
   openPrompt: <Type>(message: {
     args: Record<string, unknown>;
     origin: OriginData | OriginDataInternal;
-    action: string;
+    action: string; // TODO: here action should always match a route PATH in app/router/routes
```
escapedcat, line 105: I.e.:
```ts
  makeInvoice: "makeInvoice",
  LNURLChannel: "LNURLChannel",
  lnurlWithdraw: "lnurlWithdraw",
  confirmKeysend: "confirmKeysend",
  confirmPayment: "confirmPayment",
  confirmSignMessage: "confirmSignMessage",
```
But should be added step by step with further PRs, right?

src/extension/content-script/index.ts
```diff
@@ -0,0 +1,110 @@
...
+  // (the inpage script can not do that directly, but only the inpage script can make webln available to the page)
+  window.addEventListener(
+    "message",
+    (ev: { source: MessageEvent["source"]; data: WebLNEventData }) => {
+      // Only accept messages from the current window
+      if (ev.source !== window) {
+        return;
+      }
+
+      // QUESTION: when does happen @bumi?: ev.data.response
```
escapedcat, line 53: @bumi

src/app/screens/Home/index.tsx
escapedcat: This still works 👍

src/extension/content-script/index.ts
escapedcat: I think i.e. here:
https://github.com/getAlby/lightning-browser-extension/blob/feat/1216_refactor-prompt-popup-lnurl-handling--lnurlauth/src/extension/background-script/actions/lnurl/withdraw.ts#L14

src/app/router/routes.ts
```diff
@@ -0,0 +1,9 @@
+// PATHs names react-router is matched to in path={}
+export const PATHS = {
+  lnurlPay: "lnurlPay",
+} as const;
+
+// URLs which should match the router PATHs to call `navigate(URL)`
+export const URLS = {
+  lnurlPay: `/${PATHS.lnurlPay}`,
```
bumi, line 8: why do we have these URLS a and PATHS objects?

with that we write `path={PATHS.lnurlPay}` instead of `path="lnurlPay"` in the popup and prompt?

src/app/screens/LNURLPay/LNURLPayForm.tsx
```diff
@@ -43,21 +39,12 @@ const Dd = ({ children }: { children: React.ReactNode }) => (
   <dd className="mb-4 text-gray-600 dark:text-neutral-500">{children}</dd>
 );
 
-function LNURLPay(props: Props) {
+const LNURLPayForm: FC<Props> = ({ origin, details, isPrompt }) => {
```
bumi, line 42: what is FC?

src/app/screens/LNURLPay/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { useNavigationState } from "~/app/hooks/useNavigationState";
+import lnurlLib from "~/common/lib/lnurl";
+import getOriginData from "~/extension/content-script/originData";
+import { LNURLPayServiceResponse, OriginData } from "~/types";
+
+import LNURLPayForm from "./LNURLPayForm";
+
+const errorMessage = "Not a payRequest LNURL";
+
+const LNURLPay = () => {
```
bumi, line 12: Is this file LNURLPay specific? 
I think this would be the same for all LNURL-* screens: we mainly load the lnurl details and then render a component.

If I read this correct this component is translating the navState to props?
why can't we directly use the navState in the form component? 
then I need to make sure that I always pass the state to `navigate()`

src/app/screens/Send.tsx
```diff
@@ -35,8 +37,11 @@ function Send() {
       }
 
       if (lnurl) {
-        await lnurlLib.getDetails(lnurl); // throws if invalid.
-        navigate(`/lnurlPay?lnurl=${lnurl}`);
+        navigate(URLS.lnurlPay, {
+          state: {
```
bumi, line 41: shouldn't we say that we always pass the lnurlpay details to the state?

as you also removed the `getDetails` here the error case is no longer handled.

src/common/lib/utils.ts
bumi: not sure what this means. why that mapping? 

(btw. we should unify the names `LNURLChannel` vs. `lnurlWithdraw`)

bumi (commented):
I'd rather not touch things like the content-script here. This makes it really hard to review. The content script is here an unrelated change and only a move to TS (but the whole file looks new and thus it is not visible if/what changed there) 

The content-script and how these things work is also extremely critical. If you break something here all the lightning web apps will no longer work. 
Not something we should randomly change... also when I look back we had a few issues introduced when moving to TS - which is good, but we need to find those in reviews.

src/extension/content-script/index.ts
bumi: I don't think this should be touched here. We initially wanted to change the way lnurl-* routing works that people can also copy&paste the lnurl-pay string. 
Now it is touching all the routes and here now even the content-script and how messaging between inpage<->content<->background is done. (which is quite complex)

src/common/lib/utils.ts
escapedcat: As discussed. Not that list but only further LNURL-calls with nw PRs:

- lnurlAuth
- lnurlWithdraw
- LNURLChannel (rename this one to match casing)

To only "support" these first we can trigger the prompt-action from the Send-screen first. We can then refactor and make it feel more integrated (with the usual popup UX) later. We will do this with `lnurlAuth`

src/extension/content-script/index.ts
escapedcat: You mean this file in general or something specific? Lisa only added some types and switched it to `.ts`.
There's still some `FixMe/any` but step by step this will help us to make this easier to understand and keep i.e. actions aligned throughout the app.

src/types.ts
```diff
@@ -103,6 +105,47 @@ export interface MessageDefault {
   prompt?: boolean;
 }
 
+export type WebLNCall = typeof weblnCalls[number];
+export type WebLNEventData = {
+  action: WebLNCall;
+  application: "LBE";
+  args: Record<string, unknown> | Record<string, never>; // TODO: check if args can be undefined or an empty object?
+  prompt: true;
+  response?: FixMe; // QUESTION: when does this happen
```
escapedcat, line 114: Lisa, let's pair

src/common/lib/utils.ts
lisabaut: This is just for TypeScript.

The `action` in util `openPrompt`  is always matching a specific route name. 
Otherwise it would not work.

This is just the beginning of typing the route names [here for lnurpay](https://github.com/getAlby/lightning-browser-extension/pull/1257/files#diff-6d91cb5155c78128e7bcef234215937c2c128d06e51d23fff892313f968fd9f7R3) and the types [here](https://github.com/getAlby/lightning-browser-extension/pull/1257/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR127) and [here](https://github.com/getAlby/lightning-browser-extension/pull/1257/files#diff-6d91cb5155c78128e7bcef234215937c2c128d06e51d23fff892313f968fd9f7R3)

src/extension/content-script/index.ts
lisabaut: The file `content-script` is only converted to TS.
No functionality / code was changed. 
The only difference is that some types are added to let TypeScript understand the file.

src/app/router/routes.ts
lisabaut: Correct, the string "lnurlpay" is always the same.
This is only the beginning of a CONST collection to avoid repeating the same strings everywhere when creating `<Routes>` or when navigating with react-router 
and to import all PATHS later to the TS types to let TS know which routes name exist.

src/app/screens/LNURLPay/LNURLPayForm.tsx
lisabaut: Correct, just a differrent way of writing a Functional Component.

`function MyComponent(props: Props) {....`
`const MyComponent: FC<Props> = (props) => {....`

src/app/screens/LNURLPay/index.tsx
lisabaut: LNURLPay is rendered from three different point:

- PROMPT => `origin` and `args.lnurlDetails` exist from searchParams and are passed via navigation.state [here](https://github.com/getAlby/lightning-browser-extension/pull/1257/files#diff-6aa3c99dd75ad2803ee2a158c3686f6f611f1bbaa59183f9648b97b8cf7abc10R68)
- HOME => `origin` exists but no `lnurlDetails`, therefore `origin` and `lnurl` are passed via navigation.state [here](https://github.com/getAlby/lightning-browser-extension/pull/1257/files#diff-041371f0bc0d8e5acd8d60f55e0b217ee4320404c44f5ebe9a3d61db95753a45R257-R258)
- SEND => no `origin` and no `args.lnurlDetails` exist, therefore only the `lnurl` is passed via navigation.state [here](https://github.com/getAlby/lightning-browser-extension/pull/1257/files#diff-b78594d65e4874c4016ae236fd162b02d7ea484236ce7e9b6d0c2fcd7a053f8dR42)

The component `LNURLPay` is smart enough to check what data is in the navigation.state and gets missing data accordingly to render the view.

src/app/screens/Send.tsx
lisabaut: The component `LNURLPay` does exactly that check [here](https://github.com/getAlby/lightning-browser-extension/pull/1257/files#diff-0ae7727e94b1752f52e903368c32a80af0ae7d22e7d4d4e17bb6b7219fe7a076R37)

Therefore one can navigate to the view with passing a "lnurl"  only in the navigation.state.
The component does the rest.

src/types.ts
lisabaut: Sure !

src/extension/content-script/index.ts
bumi: I do not see here in the diff what was changed. and most bugs start with "only added" :) 
but I'd just want you to be aware that this is a critical piece and this initial rather small refactoring got a bit big and bigger when we throw in some ts fixes here and there. (I know we said we switch to TS when we touch things, but this should not be touched here, and ev.data.response is out of scope here)

src/app/router/routes.ts
bumi: but what's the difference to writing `PATHS.lnurlPay` instead of `lnurlPay` ? either I repeat a getter or a string. 
I don't care URLS + PATHS sounds complex to me (at least this right now puts `??` in my head)

src/app/screens/LNURLPay/index.tsx
bumi: imo the LNURLPay component should not be rendered with different details.
in cases where no lnurlDetails is present we already loaded the lnurl details before navigating to the component - so we can just pass it there. (that's send and home)

@escapedcat then you now have another component just to not use useNavigationState() in the other one? When use useNavigationState? and when not? 

Did you think about the other LNURL* components here?

src/app/screens/Send.tsx
bumi: @lisabaut yes, but I don't get the point of that we already load the lnurldetails here (and we must do that because we need to check if it is a valid LNURL and especially what kind `lnurlDetails.tag`) of lnurl this is to navigate to the correct route (lnurlpay, lnurlauth, etc.)

so why keep that complexity in the LNURLPay component?

lisabaut: Sure, the component would not take care of the other views.
This PR is a first version of showing how to use the navigation state with the the view "LNURPay" as a start for further views.

To sum up:

- the check if the lnurl is valid should be done in components HOME and SEND
- `lnurlLib.getDetails` should get missing details in components HOME and SEND 
- missing origin data should be created in components HOME and SEND
- `origin` & `args.lnurDetails` should be attached to the navigation.state when navigating to LNURLPay from HOME and SEND
- the wrapper component for the actual LNURLPay view should be removed
- the LNURLPay view should read the navigationState directly

src/common/lib/utils.ts
lisabaut: To sum up:

- remove the comment

src/app/screens/Send.tsx
bumi: now I am confused, why do we have it then if it should be removed again? (we already touch the lnurlpay component, don't we?)

lisabaut: Now I am confused too 😄 

The sum-up above is the way you want to change the code to remove "the complexity" in LNURLPay, right?

src/extension/content-script/index.ts
escapedcat: No diff because file-ending was changed.
Attached a diff file which shows changes without changing the file-ending.
[diff.txt](https://github.com/getAlby/lightning-browser-extension/files/9296075/diff.txt)

lisabaut: The commit history is very detailed to make a review more easy.
If you go through the commits, then you can see step by step what changed in this file too 
https://github.com/getAlby/lightning-browser-extension/pull/1257/commits

(useful Keyboard commands "n" and "p" to browse through the commits)

lisabaut:
Closing in favor of https://github.com/getAlby/lightning-browser-extension/pull/1264
which contains the feedback from the discussions above

PR #1264, opened 2022-08-10
refactor(lnurlpay): replace searchParams with location props

escapedcat:

Based on #1257, but with these changes:
- content-script stays js
- lnurDetails are being handed into the navState 

This is the removal of SearchParams in `LNURLPay` component.
Now the data `origin` and `details` are passed via the react-router navigation state from Prompt and from Components SEND and HOME in Pop-Up.

## Open tasks

- [x] adjust navState type
- [ ] move out similar code from Send and Home -> postponing this till further LNURL support has been added
- [x] Tests?

### Link this PR to an issue

https://github.com/getAlby/lightning-browser-extension/issues/1216

- `refactor`

### How has this been tested?

Manually LNURL via prompt, popup and home

On a Website Console:
- `await webln.enable()`
- `await window.webln.lnurl("A-LNURL");`

In Extension:
- Popup
  - Open `Send`
  - paste in a `"A-LNURL"`
- Via i.e. embedded twitter LN-address

### Checklist


lisabaut:
@escapedcat @bumi 

Please have a look again on this PR which should cover the review feedback from the former PR. 
=> See summary comment [here for reference](https://github.com/getAlby/lightning-browser-extension/pull/1257#discussion_r941297614). 

Also the conversion of the content script is removed. We can do this in another PR later at some point.
Same goes for the string collection of all possible navigation Paths.

This PR is still WIP, see the "Open Todos" above.

This PR serves as a proof of concept for using the navigation state.
If everyone agrees then we can continue to refactor all other LNURL components and make them reusable for Prompt and Pop-Up.

src/app/screens/Home/index.tsx
```diff
@@ -242,31 +245,54 @@ function Home() {
...
-                      `/lnurlPay?lnurl=${
-                        lnData[0].address
-                      }&origin=${encodeURIComponent(JSON.stringify(origin))}`
-                    );
+                    const lnurl = lnData[0].address;
+                    const lnurlDetails = await lnurlLib.getDetails(lnurl); // throws if invalid.
+
+                    if (lnurlDetails.tag === "payRequest") {
+                      // rather pick the key/value the component needs
+                      // the data coming from `getDetails` contains deep nested non-serializable values which results in a `null` navigation state
```
bumi, line 263: the non-serializable value is the URL here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/common/lib/lnurl.ts#L82
I think we don't need that and could simple pass keep the URL as string and init a URL object where needed. 

it could even be the case that this `url` value is currently not even accessed.

bumi (approved):
utACK. 

I'd remove the url from the lnurl details then we don't need the `pick`

src/app/screens/Home/index.tsx
escapedcat: Thanks for the info. Adjusted this 👍

src/app/screens/LNURLPay/index.test.tsx
```diff
@@ -0,0 +1,92 @@
...
...JmFiwoOadk8Jv2MhojWBqMoVlXFzpvmYviZggM6cXa78D3pZ3Q/orXGKItVVbHzZl2YX3iPi/4/0ge0fPxN6ee0CNFaYRTGip037HJ3+Lj84Zw+0s/pJERrySiMVVWx86Y1XdiRPphVprn17U9LP6uTEK0FozRWVcXOm+Zm4br0wax0eJ3/ivSzWoZoDUZxrKqKnTetif670gezyuzNDp30szoN0QrqY1VV7LxpTfTfkT6YVaaJwUs/q1UUHS0Dsaoqdt40NwvPSh/MivMt6We1qiKjZSRWVcXOm9bEw30FB7PK7Eo/q3UUFS1Dsaoqdt603c5/WcHBnDpNd/BF6We1riKiZSxWVcXOm7Z/fPCo9MGsMhdfevaj0s9qE1lHy2CsqoqdN891/nXpw1n+Yvg/SD+jbWQZLaOx6rHzhrku/ET8gJZME8OPpJ/RtrKKlvFYVRU7b5qL/rz0AS2bvaPwBelnNIYsopVBrKqKnTfPdeGf0od0wgvyJ+lnMybT0cokVj123jC9L5J/WvrZjE3vsy4nVlWl+Rzy2/nRXTxuHxL4JZKnvSTZ/kmj92UpI1ZVxc6btzc7dOIHds/489LPZEomopVprHrsvHGuC7+QP7Qwb6L/qfSzSEF1tDKPVY+dN+yd34J7R/TgunB84dfPvF/6WaSiMlqFxKqq2HnzdrqDM00Mb8gcnr+zf/vqI9LPIDVV0SooVj123rjL3eHjTfT/Snx4f97pDs5If3cpKqJVYKx67LxxLvon0v0tir+ze/vax6W/szTRaBUcqx47b9z+8cGjTQy/m/Lgms7//KnfHHxQ+rtqIRItYnUXO2/cldeuvK+JwbsY3hr3JfGvN7NwsZpX75L+jtokjRax+j/sfAYuv9w+7Gb+my6GV7c7OH9zbxa+Jv19tEsSLWK1FDufiebIf66J4Ycu+t+vcmBNF150nf/+TnftU9Kf3ZJJo0Ws1sLOZ+IbL/oPNd3hV90sXHddaO/OzIfdLnyJny/ZziTRIlZbYeeBJUaNFrECMLVRokWsAKSyVbSIFYDUNooWsQIgZa1oESsA0laKFrECoMXSaBErANosjBaxAqDVPdEiVgC063/aWvpzAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQI//AplAdntdLBX1AAAAAElFTkSuQmCC"]]',
```
bumi, line 22: as you don't check for the image I would not use one with an image here. (except if this is a fun easter egg image :)

escapedcat: That's just fiatjefs image. Just needed deafult data to test the main thing. If we later in have issues with images we can adjust the test.

bumi: it is just a bloated file then. no need for that image there.

escapedcat: Will remove with one of the next PRs

PR #1272, opened 2022-08-12
fix(lnurlpay): correct close for popup/prompt

escapedcat:

Align close actions for all actions. Those will not be reached by the modal because it is closing itself on success.

### Link this PR to an issue

#1216 

- `refactor`

### How has this been tested?
Manually tested with fiatjef and starblocks

### Checklist

### DIFF
--- a/src/app/screens/LNURLPay/index.tsx
+++ b/src/app/screens/LNURLPay/index.tsx
@@ -56,7 +56,6 @@ function LNURLPay() {
   const [successAction, setSuccessAction] = useState<
     LNURLPaymentSuccessAction | undefined
   >();
-  const [payment, setPayment] = useState<PaymentResponse | undefined>();
 
   useEffect(() => {
     if (showFiat) {
@@ -157,8 +156,6 @@ function LNURLPay() {
         }
       );
 
-      setPayment(paymentResponse);
-
       // Once payment is fulfilled LN WALLET executes a non-null successAction
       // LN WALLET should also store successAction data on the transaction record
       if (paymentInfo.successAction) {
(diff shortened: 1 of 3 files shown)


escapedcat:
> the success view in which context?
> the prompt must reply that the response is sent back to the `webln.lnurl` call for example.

Ah ok, hm, need to test this. Will switch the PR to draft.

bumi:
is this mergeable

src/app/screens/LNURLPay/index.tsx
```diff
@@ -205,9 +205,10 @@ function LNURLPay() {
   function close(e: MouseEvent) {
     e.preventDefault();
     if (navState.isPrompt) {
-      msg.reply(payment);
-    } else {
+      msg.reply(payment); // will thsi ever be reached? according to the comment in line 181 maybe not?
       window.close();
```
bumi, line 209: this must not be both, either msg.reply or window.close

bumi:
what's up with this one?

src/app/screens/LNURLPay/index.tsx
```diff
@@ -205,12 +202,9 @@ function LNURLPay() {
   }
 
   function close(e: MouseEvent) {
+    // will never be reached via prompt
     e.preventDefault();
-    if (navState.isPrompt) {
-      msg.reply(payment);
```
escapedcat, line 210: We already do this above:
```ts
  if (navState.isPrompt) {
    msg.reply(paymentResponse);
  }
```

bumi (approved):
utACK

PR #1276, opened 2022-08-12
missing label in LnurlWithdraw screen

### LINKED ISSUE #1274: Bug: LnurlWithdraw "Amount (Satoshi)" input is missing label
### Describe the bug

Inputs should always have a label if possible.
In this case there's a text which looks like a label but it's not `<label>`

### Screenshots

image

### To Reproduce

1. Visit to https://lnurl.fiatjaf.com/
1. Click on "lnurl-withdraw QR-code"
1. Alvy prompt opens
1. See `Amount (Satoshi)` isn't clickable.

### Expected behavior

`Amount (Satoshi)` is the label for the `input` below. They should be connected.

### Additional context

`dt`/`dd` are being used here. That's fine for normal text but not for a label/input combination.

### Are you working on this issue?

No

im-adithya:

Changes LnurlWithdraw screen to use TextField component

### Link this PR to an issue

Fixes #1274 

### How has this been tested?

Tested Manually at https://lnurl.fiatjaf.com/

### Checklist

### DIFF
--- a/src/app/screens/LNURLWithdraw/index.tsx
+++ b/src/app/screens/LNURLWithdraw/index.tsx
@@ -3,7 +3,7 @@ import Container from "@components/Container";
 import ContentMessage from "@components/ContentMessage";
 import PublisherCard from "@components/PublisherCard";
 import SuccessMessage from "@components/SuccessMessage";
-import Input from "@components/form/Input";
+import TextField from "@components/form/TextField";
 import axios from "axios";
 import { useState, MouseEvent } from "react";
 import { useNavigate } from "react-router-dom";
@@ -57,11 +57,18 @@ function LNURLWithdraw() {
 
   function renderAmount() {
     if (minWithdrawable === maxWithdrawable) {
-      return <p>{`${minWithdrawable / 1000} sats`}</p>;
+      return (
+        <ContentMessage
+          heading={`Amount (Satoshi)`}
+          content={`${minWithdrawable / 1000} sats`}
+        />
+      );
     } else {
       return (
-        <div className="mt-1 flex flex-col">
-          <Input
+        <dl className="my-4 p-4 shadow bg-white dark:bg-surface-02dp rounded-lg overflow-hidden">
+          <TextField
+            id="amount"
+            label="Amount (Satoshi)"
             type="number"
             min={minWithdrawable / 1000}
             max={maxWithdrawable / 1000}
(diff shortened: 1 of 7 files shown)


src/app/screens/LNURLWithdraw.tsx
```diff
@@ -61,19 +61,26 @@ function LNURLWithdraw(props: Props) {
...
+        <ContentMessage
+          heading={`Amount (Satoshi)`}
+          content={`${minWithdrawable / 1000} sats`}
+        />
+      );
     } else {
       return (
-        <div className="mt-1 flex flex-col">
-          <Input
+        <dl className="my-4 p-4 shadow bg-white dark:bg-surface-02dp rounded-lg overflow-hidden">
```
escapedcat, line 72: `dl` is not valid html with these child-components:
image
(plus some other issues)

Using a `div` here might solve the issue.

escapedcat:
Would be nice to have tests that checks if the correct component is being rendered in each situation.

im-adithya:
Done!

src/app/screens/LNURLWithdraw.tsx
im-adithya: Changed to div :+1:

escapedcat (approved):
Awesome!

src/app/screens/LNURLWithdraw/index.tsx
```diff
@@ -61,11 +61,18 @@ function LNURLWithdraw(props: Props) {
...
+          heading={`Amount (Satoshi)`}
+          content={`${minWithdrawable / 1000} sats`}
+        />
+      );
     } else {
       return (
-        <div className="mt-1 flex flex-col">
-          <Input
+        <div className="my-4 p-4 shadow bg-white dark:bg-surface-02dp rounded-lg overflow-hidden">
+          <TextField
```
bumi, line 73: should this even be a dual currency field? (that shows fiat amounts?)

escapedcat: Good point. @im-adithya wanna give that a try?

lisabaut (approved):
Thanks

src/app/screens/LNURLWithdraw/index.tsx
im-adithya: Sure, initially thought of using that, will change now!

bumi:
🚀

PR #1277, opened 2022-08-12
fix: trim invoice

### LINKED ISSUE #1252: Trim form data
### Is your feature request related to a problem? Please describe

I just copy / pasted an invoice into the payment screen and got this error: 

image

Turns out, there was a whitespace at the very front of the invoice (which was not visible anymore):

image

We could this prevent this problem by trimming values before they are processed as whitespaces in front or at the end of a string have no real use. 

### Are you working on this?

No.

im-adithya:

Trim the invoice

### Link this PR to an issue

Fixes #1252 

### How has this been tested?

Tested Manually

### Checklist


escapedcat:
Is it possible to add a test for this?

im-adithya:
I think we'll have to add a check for this after adding e2e tests for app screens

escapedcat:
Is it possible to create a unit test which types in an address with spaces and the value will only be the address?
- Pro: We have a basic unit test for this component
- Contra: We're testing if `trim` works...

im-adithya:
Not sure if we can add a unit test as we'll have to enter a value and then check... we do all that in e2e, right?

escapedcat:
> Not sure if we can add a unit test as we'll have to enter a value and then check... we do all that in e2e, right?

Currently we do not have e2e tests for this.  
Regarding a unit-test I thought we could do i.e. `type("  abc   ")` and `expect(...toHaveValue("abc"))` after.  
Does this make sense to you?

src/app/screens/Send.tsx
```diff
@@ -124,7 +124,7 @@ function Send() {
               value={invoice}
               disabled={loading}
               onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
-                setInvoice(event.target.value)
+                setInvoice(event.target.value.trim())
```
bumi, line 127:
```suggestion
                setInvoice(event.target.value.trim().replace(/^lightning:/i,""))
```

we should additionally remove a potential `lightning:` in the beginning. This is there if you copy a link.

bumi (approved):
I would prefer a unit test for this. 
e2e should test full integrations, but not such small units that check if a string is correctly edited.

escapedcat:
With the added commit form bumi let's add a unit-test

src/app/screens/Send/index.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      </MemoryRouter>
+    );
+
+    expect(
+      await screen.getByLabelText("Lightning Invoice")
+    ).toBeInTheDocument();
+    await act(async () => {
+      await user.type(
+        screen.getByLabelText("Lightning Invoice"),
+        "    sampleinvoice  "
```
escapedcat, line 34: Can we add the `lightning:` here as well? Then it's done. Thanks for you work!

im-adithya: Done!

escapedcat:
Thanks!

Issue #1269, opened 2022-08-11
Feat: Support all LNURLs via Popup - lnurlAuth


escapedcat:
### Is your feature request related to a problem?

Split out from #1216 

### Describe the solution you'd like

Auth/login should work via prompt and via "Send" within the popup.  
Main usecase is to i.e. login for your phone-account via Alby on desktop.

### Describe alternatives you've considered

none

### Additional context

none

### Are you working on this?

No

PR #1282, opened 2022-08-15
Feat/1269 lnurlauth popup

escapedcat:

lnurlAuth support via the "Send" within the popup

### Link this PR to an issue

#1269 

none

### How has this been tested?

manually with fiatjefs playground and stacker.news

### Checklist


src/extension/background-script/actions/ln/auth.ts
```diff
@@ -0,0 +1,16 @@
+import { MessageLnurlAuth } from "~/types";
```
bumi, line 1: the `ln` folder is used for actions that call LN functions on the connector. 
so this does not fit in here. we have a lnurl folder, don't we?

escapedcat: Yes, will move. We talked about it but I wasn't sure.

escapedcat: Have a look again

src/extension/background-script/actions/lnurl/lnurlAuth.ts
```diff
@@ -0,0 +1,16 @@
+import { MessageLnurlAuth } from "~/types";
+
+import { auth as lnurlAuth } from "../lnurl/auth";
+
+const auth = async (message: MessageLnurlAuth) => {
```
bumi, line 5: some tricky naming here. `auth as lnurlAuth` vs. the new `auth` here

bumi (approved):
utACK. I will test more of the whole LNURL* options in master

Issue #1283, opened 2022-08-15
i18n - Switch hardcoded strings to translations: Screens batch 1


escapedcat:
### Is your feature request related to a problem? Please describe

We can split this ticket. Please write your name behind the "screen" in the list or write a comment if you start working on it. And assign yourself to this issue if you can.
Just to avoid people accidentally work on the same screen.

Switch these screens and related components to be translatable:

- [x] Accounts - #1291
- [x] Enable - #1297
- [x] Home - #1298
- [x] Onboard
- [x] Options - #1300
- [x] Publishers - #1295
- [x] Send - #1290
- [x] Receive - #1290

### Describe the solution you'd like

All hardcoded English strings should be replaced by

### Additional context

- Maybe unit-tests need to be adjusted to handle i18n. [Like we did here](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/TransactionsTable/index.test.tsx#L125-L127) maybe.
- Don't create one big PR. Keep them small so they are easy to review and can be merged fast and do not create big conflicts.

### Are you working on this?

No

im-adithya:
The options screen is done already!

im-adithya:
Same with Onboarding!
(except for a small word `sats` if we want it to be translated as well (if yes, then we can just wait till the Home screen PR gets merged as it includes `sats` in tCommon to borrow from there!))

escapedcat:
No, all good, just wanted to double check it. I'll mark them as done. Thanks!

PR #1291, opened 2022-08-16
feat: i18n for account screen

im-adithya:
i18n for Account screen

### Link this PR to an issue

#1283 

### Checklist

### DIFF
--- a/src/app/screens/Accounts/index.tsx
+++ b/src/app/screens/Accounts/index.tsx
@@ -89,11 +89,7 @@ function AccountsScreen() {
   }
 
   async function removeAccount({ id, name }: AccountAction) {
-    if (
-      window.confirm(
-        `Are you sure you want to remove account: ${name}? \nThis can not be undone. If you used this account to login to websites you might loose access to those.`
-      )
-    ) {
+    if (window.confirm(t("remove.confirm", { name }))) {
       let nextAccountId;
       let accountIds = Object.keys(accounts);
       if (auth.account?.id === id && accountIds.length > 1) {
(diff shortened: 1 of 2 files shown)


src/app/screens/Accounts/index.tsx
```diff
@@ -210,12 +206,14 @@ function AccountsScreen() {
           closeTimeoutMS={200}
           isOpen={editModalIsOpen}
           onRequestClose={closeEditModal}
-          contentLabel="Edit account name"
+          contentLabel={t("edit.name_label")}
```
escapedcat, line 209: According to the modal docs this is:

>   /* String indicating how the content container should be announced
     to screenreaders */}
     
Should we call it `screenreader.label`?
Should we prefix all `modal`-intrenal strings with `modal`?

im-adithya: > Should we call it screenreader.label?

Makes sense!

> Should we prefix all modal-intrenal strings with modal?

We can. Currently, I'm using `label` for the text which makes the modal popup (in this case, it's in a drop-down: edit, export, delete) and then `title` for the modal title...

escapedcat: Hm, yeah, not too important. Let's just do the screenreader-change and merge.

im-adithya: Just came to know after working on other PRs that we can actually use tCommon for these labels. `common.actions` should have these to avoid repetition... Wdyt?

And we can just have title for strings like "Edit Account"

escapedcat: For really common "words" that makes sense yeah. But not action-labels "in general" maybe?

escapedcat: Oh, yes, sure, sorry!

escapedcat (approved):

Issue #1286, opened 2022-08-15
After trying to enter the incorrect password, when entering with the correct one, messages about incorrect entry appear (one for each try)


AnastasiaVolk:
**Description:**

After trying to enter the password incorrectly, when entering with the correct password, messages about incorrect entry appear (one for each try)

**Screenshots:**
![Image](https://user-images.githubusercontent.com/108762093/183482148-51f6a064-3268-49d5-b6b5-3700fcd19c8c.png)

**Steps to reproduce:**

 1. Click on the Extension icon
2. Input incorrect password
3. Input correct password
4. The extension unlocks and notification about incorrect password appears (as many as were incorrect password attempts)

**Alby information**

Alby Version: [v1.13.0]
Alby installed through Google Chrome

**Device information:**

OS: [macOS Monterey 12.0.1 ] 
Browser [chrome]  Browser Version [Chrome Version 103.0.5060.114]

**Are you working on this issue?**

No

escapedcat:
Some background
- Toast is trggered by
    https://github.com/getAlby/lightning-browser-extension/blob/master/src/common/lib/utils.ts#L30
- Initial message comes from here
    https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/accounts/unlock.ts#L26
- The unlock screen needs a message to show it here
    https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Unlock/index.tsx#L87

PR #1294, opened 2022-08-16
feat(unlock): render Toasts on Errors

lisabaut:

As discussed below:
The problem of the [bug ticket](https://github.com/getAlby/lightning-browser-extension/issues/1286) is not that the Toast Messages of an Invalid Password should surpressed if the screen is not Unlock (because Unlock screen has its own error rendering).
Instead, the problem is that the Screen Unlock is not aware of Toasts at all. 
This has to be changed and toasts should appear on this screen too.

unlock

### Link this PR to an issue

Closes https://github.com/getAlby/lightning-browser-extension/issues/1286

- Feature

### Checklist

### DIFF
--- a/src/app/router/Options/Options.tsx
+++ b/src/app/router/Options/Options.tsx
@@ -77,7 +77,15 @@ function Options() {
               }
             />
           </Route>
-          <Route path="unlock" element={<Unlock />} />
+          <Route
+            path="unlock"
+            element={
+              <>
+                <Unlock />
+                <ToastContainer />
+              </>
+            }
+          />
         </Routes>
       </HashRouter>
     </Providers>
--- a/src/app/router/Popup/Popup.tsx
+++ b/src/app/router/Popup/Popup.tsx
@@ -36,7 +36,15 @@ function Popup() {
             <Route path="keysend" element={<Keysend />} />
             <Route path="confirmPayment" element={<ConfirmPayment />} />
           </Route>
-          <Route path="unlock" element={<Unlock />} />
+          <Route
+            path="unlock"
+            element={
+              <>
+                <Unlock />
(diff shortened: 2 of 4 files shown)


escapedcat:
> Another idea is to remove the toast.error call from lib/utils.ts completely
and ensure that all components which call any action from background script are showing toasts on errors.

Yeah, I feel like calling a toast form the utils is confusing. Might be better to do that explicitly from the components.

bumi:
hmm... the problem here is that the user does not get a toast error when an invalid password  is entered but that the toast message is only shown then on the next screen.

I don't think this justifies another `if` there. 

Why don't we add the ToastContainer to the unlock screen? 

automatically showing a toast for all the error responses feels a bit dirty, true. 
we probably can also trigger the toast here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Unlock/index.tsx#L47 
but somehow I would not refactor that now, it's not that big of a problem, is it?

escapedcat:
> Why don't we add the ToastContainer to the unlock screen?

If this works, this might be the best solution for now

lisabaut:
> hmm... the problem here is that the user does not get a toast error when an invalid password is entered but that the toast message is only shown then on the next screen.
> 
> I don't think this justifies another `if` there.
> 
> Why don't we add the ToastContainer to the unlock screen?
> 

😄 Because I assumed that this is on purpose. And followed the ticket description.

Sure, let's add Toasts to Unlock if this is what the actual problem is.

I change this PR to draft, change the PR title, and will take care of this next week then

lisabaut:
@escapedcat @bumi 
PR is updated as discussed.

escapedcat (approved):
🍞

Issue #1302, opened 2022-08-18
Error handling for LNURL-withdraw


reneaaron:
### Describe the bug

Withdraws can fail for various reasons. (not enough balance, no channel, etc) Currently it seems there is no error handling and all of the withdraws are confirmed with a success message:

![image](https://user-images.githubusercontent.com/100827540/185404916-0e91fd6b-3d06-46bf-b66c-a41d3a06b431.png)

### To Reproduce

1. Open the Alby extension popup, click on "Send"
2. Enter a LNURL-withdraw link that doesn't work (go to LNBits, enable LNURL-w, create a new link and do not fund the wallet)
3. See success message (but no incoming transaction as the withdraw failed)

### Expected behavior

I want to see a proper error message if the withdraw failed. (+ if possible some advanced information on what went wrong, if we have this info)

### Are you working on this issue?

No.

bumi:
I think this is released? can you check again?

PR #1315, opened 2022-08-21
fix(lnurlwithdraw): handle error case #1302

escapedcat:

Handle failign lnurlWithdraw response and display error

### Link this PR to an issue

#1302

### How has this been tested?

### Checklist

### DIFF
--- a/src/app/screens/LNURLWithdraw/index.test.tsx
+++ b/src/app/screens/LNURLWithdraw/index.test.tsx
@@ -6,6 +6,8 @@ import { act } from "react-dom/test-utils";
 import { MemoryRouter } from "react-router-dom";
 import { settingsFixture as mockSettings } from "~/../tests/fixtures/settings";
 import * as SettingsContext from "~/app/context/SettingsContext";
+import { useNavigationState } from "~/app/hooks/useNavigationState";
+import { makeInvoice } from "~/common/lib/api";
 import type { LNURLWithdrawServiceResponse, OriginData } from "~/types";
 
 import LNURLWithdraw from "./index";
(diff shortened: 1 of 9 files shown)


src/app/screens/LNURLWithdraw/index.tsx
```diff
@@ -47,19 +47,24 @@ function LNURLWithdraw() {
...
-      await axios.get(details.callback, {
+      const response = await axios.get(details.callback, {
         params: {
           k1: details.k1,
           pr: invoice.paymentRequest,
         },
       });
 
-      setSuccessMessage("Withdraw request sent successfully.");
+      if (response.data.status === "OK") {
```
bumi, line 63: I'd do a .toUpperCase() here. 
what if a service returns `status: "ok"`

```suggestion
      if (response.data.status && response.data.status.toUpperCase() === "OK") {
```

src/app/screens/LNURLWithdraw/index.test.tsx
```diff
@@ -26,6 +30,16 @@ const mockDetails: LNURLWithdrawServiceResponse = {
   domain: "lnurl.fiatjaf.com",
 };
 
+const mockDetailsLnBits: LNURLWithdrawServiceResponse = {
+  tag: "withdrawRequest",
+  callback:
+    "https://legend.lnbits.com/withdraw/api/v1/lnurl/cb/D7paixQdqcsm3VJrstczsQ?id_unique_hash=a5BgmGCDZfz2eMQCofxquP",
```
bumi, line 36: I'd prefer to use real mock values and data that we control
also here it is not obvious from the URL that this link does not work.

src/app/screens/LNURLWithdraw/index.test.tsx
```diff
@@ -26,6 +30,16 @@ const mockDetails: LNURLWithdrawServiceResponse = {
...
+  k1: "XJnHjDAPwSkfZzUeT3wiDj",
+  minWithdrawable: 10000,
+  maxWithdrawable: 10000,
+  defaultDescription: "vouchers",
```
lisabaut, line 40: Rebase or this will fail because of missing `url` after merge

src/app/screens/LNURLWithdraw/index.tsx
lisabaut: Hm, according to the docs of `axios` the `status` is always a number.

Either check for `statusText` (not recommended, see docs)
or check for `status > 200` for GET requests

https://axios-http.com/docs/res_schema

src/types.ts
```diff
@@ -329,7 +329,7 @@ export interface LNURLWithdrawServiceResponse {
   payLink?: string;
   minWithdrawable: number; // Min amount (in millisatoshis) the user can withdraw from LN SERVICE, or 0
   maxWithdrawable: number; // Max amount (in millisatoshis) the user can withdraw from LN SERVICE, or equal to minWithdrawable if the user has no choice over the amounts
-  domain: string;
+  domain?: string;
```
lisabaut, line 332: How come that `domain` is optional now? 
Did you spot a difference on different methods to do a withdraw?

tests/fixtures/msw/handlers.ts
```diff
@@ -0,0 +1,42 @@
...
+            updateduk: "Jun 10, 2022 at 10:06 BST",
+          },
+          disclaimer:
+            "This data was produced from the CoinDesk Bitcoin Price Index (USD). Non-USD currency data converted using hourly conversion rate from openexchangerates.org",
+          bpi: {
+            USD: {
+              code: "USD",
+              rate: "29,991.8360",
+              description: "United States Dollar",
+              rate_float: 29991.836,
```
lisabaut, line 22: Looking forward to fixtures 😄

lisabaut (commented):
I do have some comments below and will review again after rebase as here a re some files doubled from `master`

src/app/screens/LNURLWithdraw/index.tsx
escapedcat: Thanks, updated. Guess I'm more used to consistent repsonses :P

src/types.ts
escapedcat: Double checked, thanks!

src/app/screens/LNURLWithdraw/index.test.tsx
escapedcat: > I'd prefer to use real mock values and data that we control 

Please elaborate, not sure what you mean with "real mock data" and data we control. What should we use for example?

> Also here it is not obvious from the URL that this link does not work.

This is just a test for the UI. Not sure how this is relevant here.

lisabaut (approved):
Thanks

src/app/screens/LNURLWithdraw/index.test.tsx
bumi: why not use "https://example.com/invalid-link"

when I read that code it looks like a very real link and we do not know what happens behind it.

PR #1335, opened 2022-08-25
test(e2e): basic receive & send to ln-address

escapedcat:


bumi (approved):
utACK
one thought was if we could  move the unlock code into a function to reuse it?

escapedcat:
> utACK one thought was if we could move the unlock code into a function to reuse it?

Good point, done!

Issue #770, opened 2022-04-02
Distinguish between receiver and initiator of a payment


reneaaron:
**Describe the bug**
See https://github.com/getAlby/lightning-browser-extension/pull/762#discussion_r839345174

The only situation where we know something about the receiver is when we pay to a lightning address in directly. 

| Use-Case                 | Initiator | Receiver |
|--------------------------|-----------|----------|
| Pay to Lightning Adress  | Alby      | ✅        |
| Pay an invoice in Alby   | Alby      | ❌        |
| Pay via WebLN (Invoice)  | Website   | ❌        |

The UI of Alby should reflect what we know about the payment being made. 

***Affected areas***
 - Payment dialogs
 - Payment notifications

**To Reproduce**

***Pay an invoice in Alby***
1. Open Alby
2. Paste an invoice

The UI currently indicates that one is paying to Alby, while Alby is only the initiator of the payment. 

![image](https://user-images.githubusercontent.com/100827540/161384716-cba84293-f1fe-4b56-8e07-e94fa93c3156.png)
![image](https://user-images.githubusercontent.com/100827540/161384747-954b3aa0-d70c-46d6-b089-f2372739249c.png)

***Pay via WebLN***
1. Pay an invoice via WebLN

![image](https://user-images.githubusercontent.com/100827540/161384826-d7fa386b-6f79-4958-b81c-fb763513294b.png)

The UI currently indicates that one is paying to the website, while the website ist only the initiator of the payment.

**Are you working on this issue? No**

escapedcat:
Looking into this.

Filling this would be helpful:


> ### Expected behavior
> 
> _A clear and concise description of what you expected to happen_

bumi:
maybe let's pair on this?

And I think this should be tackled when we do @dvoroneca new designs. One major topic here is that we don't show the Alby publisher card as seen in the screenshot above. (we currently do that because we reused a component that expects `origin data` that typically comes from the website that initiates the payment. )

so we must check which components are used in these flows and what we can reuse and what we need to adjust.

PR #1339, opened 2022-08-26
Fix/770 show relevant/correct auth-data in auth-notfications

escapedcat:

- Handle empty origin for auth-notifications

### Link this PR to an issue

#770

- Login via `Send` without `Origin`
  image
- Login via `Prompt` with `Origin`
  image

### How has this been tested?

Manually and added a basic unit-test

### Checklist

### DIFF
--- a/src/app/screens/LNURLAuth/index.tsx
+++ b/src/app/screens/LNURLAuth/index.tsx
@@ -98,9 +98,9 @@ function LNURLAuth() {
               )}
 
               <ContentMessage
-                heading={`${
+                heading={`Do you want to login to ${
                   origin ? origin.name : details.domain
-                } asks you to login to`}
+                }?`}
                 content={details.domain}
               />
             </div>
--- a/src/app/screens/LNURLAuth/index.tsx
+++ b/src/app/screens/LNURLAuth/index.tsx
@@ -37,7 +37,7 @@ function LNURLAuth() {
         lnurlDetails: details,
       });
 
-      if (navState.isPrompt && origin.host) {
+      if (navState.isPrompt && origin?.host) {
         const allowance = await api.getAllowance(origin.host);
 
         if (allowance.lnurlAuth === false) {
@@ -120,11 +120,13 @@ function LNURLAuth() {
         </>
       ) : (
         <Container isScreenView maxWidth="sm">
-          <PublisherCard
-            title={origin.name}
-            image={origin.icon}
-            url={details.domain}
-          />
+          {origin && (
+            <PublisherCard
+              title={origin.name}
+              image={origin.icon}
+              url={details.domain}
+            />
+          )}
           <div className="my-4">
             <SuccessMessage message={successMessage} onClose={close} />
           </div>
--- a/src/app/screens/LNURLAuth/index.tsx
+++ b/src/app/screens/LNURLAuth/index.tsx
@@ -20,6 +20,7 @@ function LNURLAuth() {
   const navigate = useNavigate();
 
   const navState = useNavigationState();
+
   const details = navState.args?.lnurlDetails as LNURLAuthServiceResponse;
   const origin = navState.origin as OriginData; // this action will always have an `origin` set, just the type is optional to support usage via PopUp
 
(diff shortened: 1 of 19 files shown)


src/app/screens/LNURLAuth/index.tsx
```diff
@@ -20,12 +20,15 @@ function LNURLAuth() {
   const navigate = useNavigate();
 
   const navState = useNavigationState();
+
+  // console.log({ navState });
+
```
lisabaut, line 25: Remove logs

src/app/screens/LNURLAuth/index.tsx
```diff
@@ -20,12 +20,15 @@ function LNURLAuth() {
...
   const details = navState.args?.lnurlDetails as LNURLAuthServiceResponse;
   const origin = navState.origin;
 
-  if (!origin) {
-    throw new Error("Origin is missing");
-  }
+  // if (!origin) {
+  //   throw new Error("Origin is missing");
+  // }
```
lisabaut, line 31: Remove logs

src/app/screens/LNURLAuth/index.tsx
```diff
@@ -34,17 +37,17 @@ function LNURLAuth() {
   const [successMessage, setSuccessMessage] = useState("");
 
   async function confirm() {
-    if (!origin) {
-      throw new Error("Origin is missing");
-    }
+    // if (!origin) {
+    //   throw new Error("Origin is missing");
+    // }
```
lisabaut, line 42: Remove logs

src/app/screens/Send/index.tsx
```diff
@@ -56,7 +56,7 @@ function Send() {
         if (lnurlDetails.tag === "login") {
           navigate("/lnurlAuth", {
             state: {
-              origin: originData,
+              // origin: originData,
```
lisabaut, line 59: Remove this line

src/extension/background-script/events/notifications.ts
```diff
@@ -57,14 +57,30 @@ const paymentFailedNotification = (
   });
 };
 
-const lnurlAuthSuccessNotification = (message: FixMe, data: FixMe) => {
+const lnurlAuthSuccessNotification = (
+  message: "lnurl.auth.success",
+  data: AuthNotificationData
+) => {
+  // console.log("lnurlAuthSuccessNotification", message, data);
```
lisabaut, line 64: Remove

src/types.ts
```diff
@@ -67,6 +67,16 @@ export interface PaymentNotificationData {
   };
 }
 
+export interface AuthResponseObject {
+  reason?: string;
+  status: string;
+}
+export interface AuthNotificationData {
+  authResponse?: AuthResponseObject;
+  origin?: OriginData;
```
lisabaut, line 76: So many optional values, a comment would be good like

Is only defined if ...
Is always undefined if ...

src/types.ts
```diff
@@ -277,7 +287,7 @@ export interface MessageConnectPeer extends MessageDefault {
 
 export interface MessageLnurlAuth {
   args: {
-    origin: OriginData;
+    origin?: OriginData;
```
lisabaut, line 290: Same here, in which context can origin not be defined?

lisabaut (approved):
Thanks.

Just for understanding:

`origin` can be undefined when coming via PopUp, right?
If this is the main thing in this PR I would change the title to:

"lnurlauth: do not show default origin data in pop-up" or alike

and add comments accordingly

src/app/screens/LNURLAuth/index.tsx
```diff
@@ -88,13 +89,18 @@ function LNURLAuth() {
...
+                  title={origin.name}
+                  image={origin.icon}
+                  url={details.domain}
+                />
+              )}
+
               <ContentMessage
-                heading={`${origin.name} asks you to login to`}
+                heading={`${
+                  origin ? origin.name : details.domain
```
bumi, line 102: maybe we rephrase the text here because now we might show twice the domain. 
maybe something like: Do you want to login to: ...?

src/types.ts
```diff
@@ -67,6 +67,16 @@ export interface PaymentNotificationData {
   };
 }
 
+export interface AuthResponseObject {
+  reason?: string;
+  status: string;
```
bumi, line 72: btw. some people were saying we should simply return what the service returns. 
Now they miss some data in the webln.lnurl(auth-str) response.

src/app/screens/LNURLAuth/index.tsx
escapedcat: Yes, it would. Updated it.

src/types.ts
escapedcat: Extended it. Have a look. Not sure if that's what you meant.

Issue #1021, opened 2022-06-10
Currency conversion: Replace decounce by using app cache


escapedcat:
### Is your feature request related to a problem? 

- #1000 introduced rate (BTC/Fiat) fetching
- We currently use debounce to limit rate fetching
- Every app-view (Popup, Prompt, Options, etc.) is still doing a call though

### Describe the solution you'd like

- We could reduce fetching further by saving the rate [globally in the cache](https://github.com/getAlby/lightning-browser-
extension/blob/feat/597_currency-support/src/common/lib/cache.ts)
- Via timestamp we could check if the rate is older than 1 minute, if yes, we fetch
- result can be used globally in the app (Popup, Prompt, Options, etc.)

### Describe alternatives you've considered

-

### Additional context

Blocked till #1000 is merged

### Are you working on this?

No

PR #1354, opened 2022-08-29
refactor(currencyconvert): use localstorage for rate

escapedcat:

We decided it's ok to get the rate in the background-script if it's older than 10 minutes.
In all the frontend screens (i.e. PopUp, Prompt, tab-view) we will only get the rate once because they chance that that view stays open is low. And every time the view is being opened the rate will come from the background-script.

- instead of using debounce and still do a rate-API call on every "view" (prompt, popup, etc) we use an action in the background to cache the result in the browser-storage and only do an API-call if it's older than 10 minutes
- adjust all `isLatestRate` occurrences because those are not needed anymore
- Leaving SWR out of focus for now because it feels like pre-optimization and not a big issue

### Link this PR to an issue

#1021

- `refactor`: New feature (non-breaking change which adds functionality)

none

### How has this been tested?

Manually and via unit-tests

### Checklist


bumi (commented):
hmm. can we maybe implement the caching on a lower layer? 
Similar to in the swr function we have a custom cache implementation. 

and maybe more calls need to be cached? 


some braindump pseudo code:
```js
needsCache = () => {
 ///
}

cached = memoize(needsCache, {
  key: (args) => `foo-${args[0]`, 
  ttl: 13213,
});

cached = memoize(needsCache, {
  swr: true
});

function memoize(call, optons) { 
  loadAndCache = () => {
    call(...arguments).then(response => {
       save(respone, key
  }
  return async () => {
    getCachedValue();
    if (cached && !expired) {
      if (options.swr) { loadAndCache() }; 
       return Promise.resolve(cached);
    } 
    return await loadAndCache();
  } 
}

```

bumi:
Does this also work when we want to show fiat rates in the notifications, too?
this is then probably another local storage then (which does not feel ideal to me)?

src/common/lib/cache.ts
```diff
@@ -24,3 +30,42 @@ export const removeAccountFromCache = async (id: string) => {
...
 };
+
+export const getCurrencyRateFromCache = async (currency: CURRENCIES) => {
+  let currencyRateCache: { rate?: number; timestamp?: number } = {};
+  const result = await browser.storage.local.get(["currencyRate"]);
+
+  if (result?.currencyRate) {
+    currencyRateCache = JSON.parse(result.currencyRate);
+
+    const currentTime = dayjs();
```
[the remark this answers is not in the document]
bumi: ping

src/common/lib/cache.ts
```diff
@@ -24,3 +30,42 @@ export const removeAccountFromCache = async (id: string) => {
...
+    const rateTimestamp = dayjs(currencyRateCache?.timestamp);
+    const rateTimestampPlusOneMinute = dayjs(rateTimestamp).add(1, "minute");
```
bumi, line 43: and we can cache that for let's say 10 min

src/common/lib/cache.ts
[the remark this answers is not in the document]
escapedcat: We're already using dayjs in the project. It's easier to read.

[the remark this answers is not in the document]
bumi: hmm. 

```js
const currentTime = dayjs();
const rateTimestamp = dayjs(currencyRateCache?.timestamp);
const rateTimestampPlusOneMinute = dayjs(rateTimestamp).add(10, "minute");
const isRateCurrent = currentTime.isSameOrBefore(
 rateTimestampPlusOneMinute
);
```
isn't that the same? so a simple number comparison? 

```js
isRateCurrent = false;
if (currencyRateCache?.timestamp) { 
  isRateCurrent = (currencyRateCache.timestamp + 10 * 60 * 1000) > Date.now();
}
```
I just dislike dependencies. and 3 dayjs instances.

or maybe then 

```js
const isRateCurrent = dayjs().isSameOrBefore(dayjs(currencyRateCache?.timestamp).add(10, "minute"));
```

src/extension/background-script/actions/cache/getCurrencyRate.ts
```diff
@@ -0,0 +1,85 @@
...
+    timestamp: Date.now(),
+  };
+
+  await browser.storage.local.set({
+    currencyRate: JSON.stringify(currencyRate),
+  });
+};
+
+const getFiatBtcRate = async (currency: CURRENCIES): Promise<number> => {
+  const result = await getSettings();
```
bumi, line 23: as you're here in the background script you can use the state directly and don't need to go through the messaging

src/extension/background-script/actions/cache/getCurrencyRate.ts
```diff
@@ -0,0 +1,85 @@
...
+dayjs.extend(isSameOrBefore);
+
+const storeCurrencyRate = async (rate: number) => {
+  const currencyRate = {
+    rate,
+    timestamp: Date.now(),
+  };
+
+  await browser.storage.local.set({
+    currencyRate: JSON.stringify(currencyRate),
```
bumi, line 18: is this enough? I think we need to store for what currency this is. (use USD/EUR/etc. somehow in the key)

src/common/utils/currencyConvert.ts
```diff
@@ -17,61 +15,11 @@ const getCurrencySettings = async () => {
...
   amountInBtc: number | string,
-  convertTo: CURRENCIES,
-  isLatestRate?: boolean
+  convertTo: CURRENCIES
 ) => {
-  const rate = isLatestRate
-    ? await getFiatBtcRate(convertTo)
-    : await debouncedGetFiatBtcRate(() => getFiatBtcRate(convertTo));
-
+  const { rate } = await getCurrencyRate({ currency: convertTo });
```
bumi, line 22: is this call is done for every amount that we want to convert?

src/extension/background-script/actions/cache/getCurrencyRate.ts
escapedcat: Thanks! Done

src/extension/background-script/actions/cache/getCurrencyRate.ts
escapedcat: Didn't think of this, yes! Changed

src/common/utils/currencyConvert.ts
escapedcat: Yes, is there an issue with this? What are you worried about?

bumi: it's sending some message, doing some storage lookups, etc. then for every amount that we want to display on the page. 
Even though it might work I am not sure if this is a good idea? 

we load the settings on each page load, correct? can't we include the current rate for the configured currency in that settings call and use it from there? - just thinking out loud.

PR #1358, opened 2022-08-30
feat: add webln on start script

bumi:
this initializes an empty object on "document_start" to indicate that webln is loading and will be available.
This makes it easier for apps to check if the user has Alby / an webln enabled browser.

manifest docs:  https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts

codepen: https://codepen.io/bumibumi/pen/xxWvrqd?editors=1111

### DIFF
--- a/src/extension/content-script/onstart.js
+++ b/src/extension/content-script/onstart.js
@@ -15,11 +15,10 @@ async function onstart() {
       return;
     }
     const scriptEl = document.createElement("script");
-
-    var inlineScript = document.createTextNode(
-      "if (!window.webln) { window.webln = { loading: true }; }"
+    scriptEl.setAttribute(
+      "src",
+      browser.extension.getURL("js/inpageScriptWebLN.bundle.js")
     );
-    scriptEl.appendChild(inlineScript);
     scriptEl.setAttribute("type", "text/javascript");
     container.appendChild(scriptEl);
   } catch (err) {
--- a/src/extension/content-script/onstart.js
+++ b/src/extension/content-script/onstart.js
@@ -1,3 +1,4 @@
+import injectScript from "./injectScript";
 import shouldInject from "./shouldInject";
 
 async function onstart() {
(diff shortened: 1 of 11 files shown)


src/extension/content-script/onstart.js
```diff
@@ -0,0 +1,28 @@
+import shouldInject from "./shouldInject";
```
escapedcat, line 1: Could we try to add new files directly as TS? Even if all types are "FixMe". At least the base is existing.

src/extension/inpage-script/webln.js
```diff
@@ -0,0 +1,5 @@
+import WebLNProvider from "../ln/webln";
```
escapedcat, line 1: Same for this one

src/extension/inpage-script/index.js
```diff
@@ -3,7 +3,11 @@ import { ABORT_PROMPT_ERROR, USER_REJECTED_ERROR } from "~/common/constants";
 import WebLNProvider from "../ln/webln";
 
 if (document) {
-  window.webln = new WebLNProvider();
+  // window.webln is normally loaded onstart (see onstart.js)
```
reneaaron, line 6: Why do we need that? Is there any reason the `onstart.js` isn't being executed?

src/manifest.json
```diff
@@ -99,6 +99,12 @@
       "matches": ["*://*/*"],
       "run_at": "document_end",
       "js": ["js/contentScript.bundle.js"]
```
reneaaron, line 101: Should we maybe rename the `contentScript.bundle.js` to `contentScriptOnEnd.bundle.js` for consistency?

src/extension/inpage-script/index.js
bumi: I don't know. there are certain `if` cases where webln will not be enabled and because it is crucial I thought it does not hurt to do this check again.

src/manifest.json
bumi: +1

reneaaron:
@bumi I've been using this build for ~1 week and didn't notice any sites malfunctioning with the new injection strategy. Maybe it would make sense to merge it and ask the rest of the team to do some testing before we release it? 

WebLN is available instantly with this detection code 👍 (doesn't even need to register events / check intervals)

https://codepen.io/reneaaron/pen/zYWVXMy?editors=0011

Issue #1348, opened 2022-08-27
Alby integration for SoundCloud


AnastasiaVolk:
### Description

For SoundCloud it is possible to integrate https://getalby.com/yourname to the main page. The link is working and when you click on it, also the Extension activates . But if the LN address is being added in the track description (same way as it done for YouTube videos), it wouldn’t activate the Extension as it does for YouTube videos

### How would like to see it

LN address in the track description on SoundCloud activates the Extension same way it does on YouTube

### Screenshot
Here is the screenshot of SoundCloud profile, the track description page on SC and a link for it (https://soundcloud.com/monvol)

<img width="1440" alt="Screenshot 2022-08-26 at 12 56 31" src="https://user-images.githubusercontent.com/108762093/187022115-4255978c-d606-49c6-93b5-23a03588af37.png">

<img width="1307" alt="Screenshot 2022-08-27 at 11 35 23" src="https://user-images.githubusercontent.com/108762093/187022400-83ca9a4e-4a0a-4d93-bb91-14821a5254a5.png">

### Are you working on this?

No

escapedcat:
Good idea! Same for https://www.mixcloud.com/ maybe.

AnastasiaVolk:
Now in value4value section we need to add SoundCloud (and Mixcloud)
Aa well as in the Integration in the Account on getalby.com

AnastasiaVolk:
It is not working for Mixcloud

escapedcat:
We figured it out, all good

PR #1364, opened 2022-08-31
SoundCloud Battery

im-adithya:

Added integration for SoundCloud

### Link this PR to an issue

#1348

### Demo

https://user-images.githubusercontent.com/64399555/187685931-cea0251d-711d-4085-840d-9ff1cfdef69e.mp4

### Checklist

- [x] My code follows the style guidelines of this project, and performed a self-review of my own code

### DIFF
--- a/src/extension/content-script/batteries/SoundCloud.ts
+++ b/src/extension/content-script/batteries/SoundCloud.ts
@@ -1,0 +1,86 @@
+import getOriginData from "../originData";
+import { findLightningAddressInText, setLightningData } from "./helpers";
+
+const urlMatcher =
+  /^https:\/\/soundcloud.com\/([^/]+)(\/([^/]+))?(\/([^/]+))?(\/([^/]+))?$/;
+
+function battery(): void {
+  const urlParts = document.location
+    .toString()
+    .replace("https://soundcloud.com", "")
+    .split("/");
+  const username = urlParts[1];
+
+  if (username && !urlParts[2]) {
+    handleProfilePage();
+  } else if (urlParts[2] === "sets" || urlParts[3] === "sets") {
+    handleTrackPage();
+  }
+}
+
+function handleTrackPage() {
+  const description = document.querySelector<HTMLDivElement>(
+    ".truncatedAudioInfo__content .sc-text-body"
+  );
+  if (!description) return;
+
+  const address = findLightningAddressInText(description.innerText);
+  if (!address) return;
+
+  setLightningData([
+    {
+      method: "lnurl",
+      address: address,
+      ...getOriginData(),
+      description:
+        document.querySelector<HTMLDivElement>(
+          ".soundTitle__titleHeroContainer"
+        )?.innerText ?? "",
+      name:
+        document.querySelector<HTMLAnchorElement>(
+          ".userBadge__username .userBadge__usernameLink"
+        )?.innerText ?? "",
+      icon:
+        document
+          .querySelector<HTMLSpanElement>(`.userBadge__avatar span.sc-artwork`)
+          ?.style.backgroundImage.slice(4, -1)
+          .replace(/"/g, "") ?? "",
+    },
+  ]);
+}
+
+function handleProfilePage() {
+  const webProfiles = document.querySelector<HTMLDivElement>(".web-profiles");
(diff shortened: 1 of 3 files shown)


src/extension/content-script/batteries/helpers.ts
```diff
@@ -5,8 +5,8 @@ import type { Battery } from "~/types";
 import { ExtensionIcon } from "../../background-script/actions/setup/setIcon";
 
 export const findLightningAddressInText = (text: string): string | null => {
-  const match = text.match(/(⚡:?|lightning:|lnurl:)\s?(\S+@[\w-.]+)/i);
-  if (match) return match[2];
+  const match = text.match(/((⚡|⚡️):?|lightning:|lnurl:)\s?(\S+@[\w-.]+)/i);
```
im-adithya, line 8: Surprisingly, these two are different characters!

escapedcat: You have some context/docs to this?

bumi: I think we ran into this problem before. some link with info to the character would be good.

bumi:
@AnastasiaVolk does this look good?

AnastasiaVolk:
@bumi [  ] (https://www.loom.com/share/883c525429ff4cbfb38af3fabdcbd91c) all good!

AnastasiaVolk:
Now the important thing the donation link appears in the wallet dashboard on getalby.com [#207](https://github.com/getAlby/getalby.com/issues/207) and after SC and MC are needed to be added to value4value section (As well as in the Integration section in the Account on getalby.com)

src/extension/content-script/batteries/helpers.ts
escapedcat: Nice, can you add a comment with the link to the regex?

bumi:
@im-adithya we should try one thing here: 
extract the information from a `https://getalby.com/p/[LN-ADDRESS]` link. for example: 

https://getalby.com/p/monvol -> should be a ln address monvol@getalby.com 

if we do this in the `helpers.ts` then this would extract that link automatically from the other platforms, too. 

so we check for either the ⚡️that we currently have OR check for `https://getalby.com/p/(\w+)`

this allows users to actually put the link there which then works for users that do not have the Alby extension installed.

im-adithya:
This is great! Didn't know about this earlier; added support now!

bumi:
for that to work we need to check the href of the profile links. 

`document.querySelector(".web-profiles a[href*='getalby.com']").href` 
image
soundcloud uses some redirect there, so we have to get the URL from that href: `"https://gate.sc/?url=https%3A%2F%2Fgetalby.com%2Fmonvol&token=7f1dd6-1-1662050178940" `

image


@AnastasiaVolk can you update your link to getalby.com/p/monvol? - when publishing this we should use the `/p/` for technical reasons.

AnastasiaVolk:
@bumi done, working ) then it should be the same with YouTube..

bumi:
yes, we always use the /p/... URL - that makes it easier for us to identify those profile URLs. 
Will be shown on getalby.com - as per your other issue.

escapedcat (approved):

src/extension/content-script/batteries/SoundCloud.ts
```diff
@@ -0,0 +1,102 @@
...
+          ?.style.backgroundImage.slice(4, -1)
+          .replace(/"/g, "") ?? "",
+    },
+  ]);
+}
+
+function handleProfilePage() {
+  const webProfiles = document.querySelector<HTMLDivElement>(".web-profiles");
+  const linkElement = document.querySelector<HTMLAnchorElement>(
+    ".web-profiles a[href*='getalby.com']"
```
escapedcat, line 55: This is just a general fallback if i.e. no ⚡ has been found?

im-adithya: Yes, not just as a fallback but also because some people prefer to add it directly like this:
Screenshot from 2022-09-05 12-19-47
While others add a link:
Screenshot from 2022-09-05 12-19-21
Apart from this (web profiles), I've also added the description to be checked for an address if they put it there by any chance!

PR #1366, opened 2022-08-31
Mixcloud Battery

im-adithya:

Added integration for Mixcloud

### Link this PR to an issue

#1348

### How has this been tested?

Tested Manually

### Checklist

### DIFF
--- a/src/extension/content-script/batteries/Mixcloud.ts
+++ b/src/extension/content-script/batteries/Mixcloud.ts
@@ -1,0 +1,73 @@
+import axios from "axios";
+
+import getOriginData from "../originData";
+import { findLightningAddressInText, setLightningData } from "./helpers";
+
+const urlMatcher = /^https:\/\/www.mixcloud.com\/([^/]+)(\/)?([^/]+)?(\/)?$/;
+
+function battery(): void {
+  const urlParts = document.location.pathname.split("/");
+  const username = urlParts[1];
+  const showName = urlParts[2];
+
+  if (username && !showName) {
+    handleProfilePage(username);
+  } else if (showName) {
+    handleShowPage(username);
+  }
+}
+
+async function handleShowPage(username: string) {
+  const descriptionElement = document.querySelector<HTMLMetaElement>(
+    'head > meta[property="og:description"]'
+  );
+  const imageUrl = document.querySelector<HTMLMetaElement>(
+    'head > meta[property="og:image"]'
+  )?.content;
+  if (!descriptionElement) return;
+
+  const address = findLightningAddressInText(descriptionElement.content);
+
+  if (!address) {
+    handleProfilePage(username);
+    return;
+  }
+
+  setLightningData([
+    {
+      method: "lnurl",
+      address: address,
+      ...getOriginData(),
+      description: descriptionElement.content ?? "",
+      name: document.title.split(" | Mixcloud")[0] ?? "",
+      icon: imageUrl ?? "",
+    },
+  ]);
+}
+
+async function handleProfilePage(username: string) {
+  const userResponse = await axios.get(`https://api.mixcloud.com/${username}`);
+  if (!userResponse) return;
+  const userInfo = await userResponse.data;
+
+  const address = findLightningAddressInText(userInfo.biog);
+
+  if (!address) return;
+
+  setLightningData([
+    {
+      method: "lnurl",
+      address: address,
+      ...getOriginData(),
+      description: userInfo.biog,
+      name: userInfo.name ?? "",
+      icon: userInfo.pictures.medium ?? "",
+    },
+  ]);
+}
+
+const mixcloud = {
+  urlMatcher,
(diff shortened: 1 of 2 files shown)


escapedcat (approved):
utACK

src/extension/content-script/batteries/Mixcloud.ts
```diff
@@ -0,0 +1,75 @@
+import axios from "axios";
+
+import getOriginData from "../originData";
+import { findLightningAddressInText, setLightningData } from "./helpers";
+
+const urlMatcher = /^https:\/\/www.mixcloud.com\/([^/]+)(\/)?([^/]+)?(\/)?$/;
+
+function battery(): void {
+  const urlParts = document.location.pathname.split("/");
+  const username = urlParts[1];
```
bumi, line 10: are we able to get those directly from the regex?

src/extension/content-script/batteries/Mixcloud.ts
```diff
@@ -0,0 +1,75 @@
+import axios from "axios";
+
+import getOriginData from "../originData";
+import { findLightningAddressInText, setLightningData } from "./helpers";
+
+const urlMatcher = /^https:\/\/www.mixcloud.com\/([^/]+)(\/)?([^/]+)?(\/)?$/;
```
bumi, line 6: are those basically all mixcloud urls?

src/extension/content-script/batteries/Mixcloud.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (username && !showName) {
+    handleProfilePage(username);
+  } else if (showName) {
+    handleShowPage(username, showName);
+  }
+}
+
+async function handleShowPage(username: string, showName: string) {
+  const showResponse = await axios.get(
+    // The filter can be generated here: https://api.stackexchange.com/docs/users-by-ids
```
bumi, line 22: c&p error?

src/extension/content-script/batteries/Mixcloud.ts
im-adithya: Yes, the Artist page: https://www.mixcloud.com/djcity/
and the Track/Show page: https://www.mixcloud.com/djcity/spinelli/

src/extension/content-script/batteries/Mixcloud.ts
im-adithya: Oops! Fixed!

src/extension/content-script/batteries/Mixcloud.ts
im-adithya: Yes! The Mixcloud API links are the same as the original ones but with `api.mixcloud.com/username/showname` instead of `www.mixcloud.com/username/showname`

src/extension/content-script/batteries/Mixcloud.ts
bumi: those are three match groups? would checking for the two be enough?

im-adithya: Yes, those are the only places where we can find content by creators

AnastasiaVolk:
Tested it on Mixcloud - doesn't work

bumi:
@AnastasiaVolk can you work with @im-adithya on this? 
You have an artist account there and @im-adithya can update the code to make it work. :)

im-adithya:
> Tested it on Mixcloud - doesn't work

Oops! Sorry for that.

> You have an artist account there and @im-adithya can update the code to make it work. :)

Yup!

bumi:
why do you use the API here? it seems quite easy to extract the LN address from the DOM? 

e.g. https://www.mixcloud.com/NVolkova/ 

and if we use the API then we should probably only and always load it from the user. (to make sure the user only has to add it once)

im-adithya:
> why do you use the API here? it seems quite easy to extract the LN address from the DOM?

It is, but there is some issue with the page load. The content doesn't load by the time the scripts run, resulting in null. This must be a problem with their website as our extension starts running scripts only after page load...

I changed it for the Show Page, though, forgot that we can also extract from meta tags. But the profile page doesn't have information in the head.

> and if we use the API then we should probably only and always load it from the user.

Not sure if I understand this

im-adithya:
Also, it should be working now!
@AnastasiaVolk, can you please check again?

AnastasiaVolk:
@im-adithya still doesn't :(

im-adithya:
@AnastasiaVolk are you using a Mac? It seems to work fine on Linux and Windows...

AnastasiaVolk:
@im-adithya yes, I'm on Mac

escapedcat:
Works for me. Wrote @AnastasiaVolk . Let's figure out what the issue might be.

AnastasiaVolk:
@im-adithya yes, it was a wrong build. All is working and it is super cool! Thanks @im-adithya

PR #1369, opened 2022-09-01
Chore/comments and deletions

escapedcat:

1. comment on unused components
    - as long as they are not used we do not need to write tests/translate/etc
1. deleted unused helpers
    - looks like some helpers were created for the fiat-conversion which is done with other code now
    - removing unused sorting helpers

Let me know if some of this code should stay. If so we should comment it and add tests for it later on.


bumi:
can we delete the components when they are not used?

escapedcat:
> can we delete the components when they are not used?

Yeah, I guess. Thought about that as well.
We can always recover those from git if we eventually need those. Although they might be outdated anyways by then.
Will delete.

PR #1324, opened 2022-08-22
feat: show fiat in Confirm Payment

lisabaut:
### Describe the changes you have made in this PR

This PR renders the Fiat Amount of a confirmable payment (if the setting "showFiat" is true and fiatAmount exists accordingly).

~Design came from https://github.com/getAlby/lightning-browser-extension/issues/1306#issuecomment-1222261734~
=> Design changed, see Screenshot and comments below

### Link this PR to an issue

Closes https://github.com/getAlby/lightning-browser-extension/issues/1306

### Type of change (Remove other not matching type)

- `feat`: New feature (non-breaking change which adds functionality)

### Screenshots of the changes (If any)

Bildschirmfoto 2022-08-24 um 10 27 26

### How has this been tested?
usign Stackernews and invoke payments

### Checklist

- [x] My code follows the style guidelines of this project and performed a self-review of my own code
- [x] New and existing tests pass locally with my changes
- [x] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

PR #1377, opened 2022-09-01
fix: calculate fiat amount from correct value

bumi:


src/app/screens/ConfirmPayment/index.tsx
```diff
@@ -42,12 +42,26 @@ function ConfirmPayment(props: Props) {
...
       (async () => {
-        const res = await getFiatValue(budget);
-        setFiatAmount(res);
+        if (invoice.satoshis) {
+          const res = await getFiatValue(invoice.satoshis);
+          setFiatAmount(res);
+        }
+      })();
+    }
+  }, [invoice.satoshis, showFiat]);
```
lisabaut, line 56: Combine the dependency check with

```suggestion
     if (showFiat && invoice.satoshis) {
      (async () => {
          const res = await getFiatValue(invoice.satoshis);
          setFiatAmount(res);
      })();
    }
  }, [invoice.satoshis, showFiat]);
```

src/app/screens/ConfirmPayment/index.tsx
```diff
@@ -42,12 +42,26 @@ function ConfirmPayment(props: Props) {
...
+
+  useEffect(() => {
+    if (showFiat) {
+      (async () => {
+        if (budget) {
+          const res = await getFiatValue(budget);
+          setFiatBudgetAmount(res);
+        }
```
lisabaut, line 64: Same here:

```suggestion
    if (showFiat && budget) {
      (async () => {
          const res = await getFiatValue(budget);
          setFiatBudgetAmount(res);
```

lisabaut (approved):
Thanks, See some suggestions below.

**Question:**
I am not sure what exactly this is fixing
but does this have something to do with this PR: [feat: show fiat in Confirm Payment](https://github.com/getAlby/lightning-browser-extension/pull/1324/files)
?

If so, is this also relevant then for ConfirmKeySend?

src/app/screens/ConfirmPayment/index.tsx
bumi: lol, true

escapedcat (approved):

bumi:
@lisabaut it has something to do with that. 
The wrong fiatValue has been used there. `fiatValue` is the fiat value of the suggested budget and NOT the fiat value of the payment amount.  The suggested budget is 10 times the invoice amount.

src/app/screens/ConfirmPayment/index.test.tsx
```diff
@@ -79,12 +83,16 @@ describe("ConfirmPayment", () => {
       );
     });
 
+    expect(screen.getByText("25 sats")).toBeInTheDocument();
```
bumi, line 86: is this specific enough? because we have two fiat values in the document.

escapedcat: Made sure the second value is really not there in the beginning.

PR #1388, opened 2022-09-04
case insensitive monetization and more consistent alby metatags

bumi:
this make the check for the lightning metatag case in-sensitive. 

Checks for `name="alby:name"` and "alby:description" metatags to load the website information.
currently we only checked for `property="alby:name"` this feels a bit inconsistent because for the lightning meta tage we use the `name` attribute.


escapedcat (approved):
So insensitive! :P

PR #1397, opened 2022-09-06
Feature/better tor support

bumi:
### Describe the changes you have made in this PR

if a user has Tor installed or is using the Tor Browser there is no
need to install the companion app. - it actually also does not work.
So users of the Tor Browser should not install the companion app
and we need to use the normal connectors (and not the native connectors)

Sadly I don't know of a good way to detect the Tor Browser automatically.
(except of doing a HTTP request, which is slow and more complicated)

### Link this PR to an issue

#670
also closes #1162 #1182

### Type of change (Remove other not matching type)

- `fix`: Bug fix (non-breaking change which fixes an issue)
- `feat`: New feature (non-breaking change which adds functionality)

### How has this been tested?

Use Alby with the Tor Browser and add a node that is behind Tor

### Checklist

- [x] My code follows the style guidelines of this project and performed a self-review of my own code
- [x] New and existing tests pass locally with my changes
- [x] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

PR #1404, opened 2022-09-08
chore: linting and unit-test

escapedcat:

- linting for: #1397
- unit-test for: https://github.com/getAlby/lightning-browser-extension/commit/31018f73294c242f1a3534eb8a9c749fc07da800

### DIFF
--- a/src/app/screens/Onboard/TestConnection/index.tsx
+++ b/src/app/screens/Onboard/TestConnection/index.tsx
@@ -47,14 +47,13 @@ export default function TestConnection() {
       setErrorMessage(message);
     } finally {
       setLoading(false);
+      setErrorMessage("");
       clearTimeout(timer);
     }
   }, [t]);
 
   useEffect(() => {
-    (async () => {
-      await loadAccountInfo();
-    })();
+    loadAccountInfo();
   }, [loadAccountInfo]);
 
   return (
--- a/src/app/screens/Onboard/TestConnection/index.tsx
+++ b/src/app/screens/Onboard/TestConnection/index.tsx
@@ -1,7 +1,7 @@
 import Button from "@components/Button";
 import Card from "@components/Card";
 import Loading from "@components/Loading";
-import React, { useState, useEffect, useCallback } from "react";
+import React, { useState, useEffect } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import api from "~/common/lib/api";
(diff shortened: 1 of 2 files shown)


src/app/screens/Onboard/TestConnection/index.tsx
```diff
@@ -49,11 +49,13 @@ export default function TestConnection() {
       setLoading(false);
       clearTimeout(timer);
     }
-  }
+  }, [t]);
```
bumi, line 52: why was this not needed before? 
did you test this very well? (I had strange errors in some places when I had to do this callback,useEffect stuff - and I still don't understand it) (e.g. did you test the timeout?)
you are missing the test connection for the options page.

escapedcat: > why was this not needed before?

Define "before"

> did you test this very well? (I had strange errors in some places when I had to do this callback,useEffect stuff - and I still don't understand it) (e.g. did you test the timeout?) 

Let me test. Just opened this to see if the pipeline will be green.

> you are missing the test connection for the options page.

What do you mean? With this PR I just wanted to fix the warning.

bumi: please always make sure that things work when you change thing on such critical components. They simply have to work, if it's broken for some reason then people will have issues adding accounts. Checking the pipeline is (and will never be) enough.
This also helps you to understand what goes on and how things are used. 

So here for example you seem to miss that there are two similar flows (welcome + options) both have a test-connection screen - and sadly are not very DRY. So you have to make this change in the other screen, too.

escapedcat: Tested the welcome/onboard testconnection with this and it seems to work well.  
The same way is not working for the options screen so I'm leaving that for now and will have a look with Lisa when she's back.
Would you mind testing this as well?

src/app/screens/Onboard/TestConnection/index.tsx
```diff
@@ -47,13 +47,14 @@ export default function TestConnection() {
       setErrorMessage(message);
     } finally {
       setLoading(false);
+      setErrorMessage("");
```
bumi, line 50: why that? finally runs always (no matter if it is successful or an error happens). So does this not overwrite the `setErrorMessage` in the error case in line 47?

escapedcat: True, I just noticed that if the timeout error is showing and a success is happening during that, error and success show at the same time. Adjusted it.

bumi:
so imo this is broken. It does no longer show an error message and I also don't get why we make this change in one of the components (onboard) and not the other one (options).  
it seems like we don't know what we're doing here? so I'd rather not change this.

PR #1406, opened 2022-09-08
fix: publishercard in lnurlpay screen

im-adithya:

The recent PR has changed the prop values in the LNURLPay screen which caused this:

Screenshot from 2022-09-08 18-31-41

This PR corrects it to this:

Screenshot from 2022-09-08 18-33-08

### Checklist


escapedcat:
We talked, I assume this will be updated

src/app/components/PublisherCard/index.tsx
```diff
@@ -23,6 +23,10 @@ export default function PublisherCard({
   isSmall = true,
   children,
 }: Props) {
+  if (!title) {
+    title = description as string;
```
bumi, line 27: 👍

src/app/screens/LNURLPay/index.tsx
```diff
@@ -275,8 +276,8 @@ function LNURLPay() {
     return (
       <Container maxWidth="sm">
         <PublisherCard
-          title={getRecipient()}
-          description={navState.origin?.description}
+          title={navState.origin?.name as string}
```
escapedcat, line 279: Why do we need the `as` here? It can be undefined because i.e. via "Send" we do not have an origin, right?
Why forcing this to be a string then?
Same question for similar places in this PR.

If this is related to above...:
```ts
  if (!title) {
    title = description as string;
```
...why can't we make `title` optional and handle this in the screen here?

Just some thoughts. Happy to pair.

p.s.
The `PublisherCard` component seems to be quite crucial and used in a lot of places in different ways. Would be nice to add some unit-tests for that (later on).

escapedcat (approved):

Issue #1333, opened 2022-08-25
Error message shown twice


MoritzKa:
### Describe the bug
When payment the payment failed; status code 400 is shown twice in the pop-up screen

### Screenshots (If Any)

![image](https://user-images.githubusercontent.com/13707997/186592474-ab52e7dc-e469-45b5-8b60-ef8e9d43062a.png)

### To Reproduce

E.g. keysend payment to an offline node.

### Expected behavior

One instead of two error messages.

### Alby information

- Alby Version: 1.13
- Alby installed through [e.g. installed through the browser stores, installed manually]
- Wallet connected with Alby [e.g. LND, BlueWallet LNDhub]

### Device information [optional]:

- OS: Windows
- Browser Chrome
- Browser Version 104

### Are you working on this issue?

_(Yes/No)_

MoritzKa:
same here: 2x error message.

![image](https://user-images.githubusercontent.com/13707997/186686957-fedcb837-55dd-45ca-aeae-b6ae402e29f7.png)

bumi:
we need to check if this one can be removed: https://github.com/getAlby/lightning-browser-extension/blob/master/src/common/lib/utils.ts#L31

If we always catch the errors when we do the `api.*` or `utils.call` calls.

PR #1408, opened 2022-09-09
fix: remove error toast from utils.call

im-adithya:

Removes the toast from being called in the utils function itself as in some places we again check if there's an error and add a _second_ toast.

### Link this PR to an issue

#1333

### How has this been tested?

Manually

### Checklist


bumi:
ok, so in all instances where we use utils.call or api.[something] we catch the error and potentially show the toast if needed?

im-adithya:
Yes! And also, we can't generalize it as some connector pages use a custom toast component to show an error instead of toast

escapedcat (approved):
utACK

Issue #1379, opened 2022-09-01
UX/UI: align "Send" screens


escapedcat:
### Is your feature request related to a problem?

Currently all "Send" screens look similar but different. This is partly related to different info that is available to be displayed and partly to code-inconsistencies.

### Describe the solution you'd like

A general solution which can be used for all these screens.
This will hopefully:
- align the UI/UX
- reduce code because we can re-use the same "component" for all screens

Related to:
- https://github.com/fiatjaf/lnurl-rfc/blob/luds/09.md

### Are you working on this?

No

## Screenshots

1. Keysend (payment directly to a node)
    - Start (non-prompt)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187887989-b22596f2-8d27-4502-9c65-036bc1052d12.png" />
    - Success (non-prompt only)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187888365-188e1b39-dda6-4212-ab1d-dcb53dc93651.png" />
1. Withdraw
    - Start (prompt)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187889380-18e15fc5-1101-46e1-abf9-8e5d3a61fe6a.png" />
    - Start (non-prompt)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187889287-d532ee09-9327-4976-a21b-0afb93f0a277.png" />
    - Success (non-prompt only)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187889543-cc6e466d-a8c6-47ab-9a6d-ba7cbb46a92f.png" />
1. Channel
    - Start (prompt)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187889646-e697981a-67c1-4eb9-bac3-45f6e1717773.png" />
    - Start (non-prompt)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187889752-a0a1e10f-02df-4316-8af9-4346c4bf1c68.png" />
    - Success (non-prompt only)
        <img width="250" src="https://user-images.githubusercontent.com/1016218/187889991-0c488942-9af7-4b25-b6ed-5f8844f79e03.png" />
          - `nodeId` can be rather long
          - `k1` is "random or non-random string to identify the user's LN WALLET when using the callback URL"
1. Auth
    - Start (prompt)
[...]

Issue #1391, opened 2022-09-05
Use identicons 


bumi:
### Is your feature request related to a problem? Please describe

Currently we use one fixed default image (which looks actually like an identicon).
To make that image more useful and actually something unique per website/recipient we can use an identicon. 

Identicons provide a visual identifier for the website. 

### Describe the solution you'd like

We can use this one for example and dynamically generate the image: https://www.npmjs.com/package/identicon.js

escapedcat:
Won't do

PR #1409, opened 2022-09-09
fix: remove default image in publisher card

im-adithya:

Removes the default image whenever there isn't one.
[It still shows up when there's an error in loading the image, though]

### Link this PR to an issue

#1379 

Screenshot from 2022-09-09 13-11-27

<p>

</p>

### How has this been tested?

Manually

### Checklist

### DIFF
--- a/src/app/components/PublisherCard/index.tsx
+++ b/src/app/components/PublisherCard/index.tsx
@@ -32,25 +32,29 @@ export default function PublisherCard({
       className={classNames(
         isSmall ? "p-2" : "flex-col justify-center p-4",
         isCard && "drop-shadow rounded-lg mt-4",
+        !image && "py-8",
         "flex items-center bg-white dark:bg-surface-02dp"
       )}
     >
-      <img
-        className={`m-2 shrink-0 bg-white border-solid border-2 border-white object-cover rounded-lg shadow-2xl ${
-          isSmall ? "w-14 h-14 mr-4" : "w-20 h-20"
-        }`}
-        src={image || DEFAULT_IMAGE}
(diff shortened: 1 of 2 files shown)


bumi:
is this decided by @dvoroneca 
because we also have the #1391 idea, don't we?

bumi:
ok, good if it's a design decision +1

We can load the image from the LNURL-pay metadata response (similar to the recipient).

With the new tipping page for Alby users where they also can upload images this could fit.

https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md#wallet-to-service-interaction-flow

that should be something like: 

```js
  function getImage {
    if (!details?.metadata) return;

    try {
      const metadata = JSON.parse(details.metadata);
      const image = metadata.find(
        ([type]: [string]) => type === "image/png;base64" || type === "image/jpeg;base64"
       );

      if (image) return image[1];
    } catch (e) {
      console.error(e);
    }
  }

// or DRY with the identifier:

function findLnurlMetadata(key) {
  if (!details?.metadata) return;

    try {
      const metadata = JSON.parse(details.metadata);
      return metadata.find(
        ([type]: [string]) => type === key
       );
    } catch (e) { 
      console.error(e);
    }
}

function getImage {
  const image = findLnurlMetadata("image/png;base64") || findLnurlMetadata("image/jpeg;base64");
  if (image) return image[1];
}
```

escapedcat (approved):
utACK

escapedcat:
Tested, please rebase and I merge

im-adithya:
Done!

PR #1424, opened 2022-09-12
test(e2e): lnurl-auth

escapedcat:
- add lnurl-auth test
- use unlock for helper for send to ln-address

Add e2e test for LNURL-auth link

### DIFF
--- a/tests/e2e/002-walletFeatures.spec.ts
+++ b/tests/e2e/002-walletFeatures.spec.ts
@@ -3,7 +3,13 @@ import { getDocument, queries } from "pptr-testing-library";
 
 import { loadExtension } from "./helpers/loadExtension";
 
-const { getByText, getByLabelText, findByText, getByPlaceholderText } = queries;
+const {
+  getByText,
+  getByLabelText,
+  findByText,
+  getByPlaceholderText,
+  findAllByText,
+} = queries;
 
 const unlockExtension = async ({ page, extensionId }) => {
   await page.waitForTimeout(1000);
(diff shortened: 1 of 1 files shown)


tests/e2e/walletFeatures.spec.ts
```diff
@@ -94,4 +88,31 @@ test.describe("Wallet features", () => {
...
+    });
+
+    // go to send
+    await (await findByText($optionsdocument, "Send")).click();
+    const invoiceInput = await findByText(
+      $optionsdocument,
+      "Invoice, Lightning Address or LNURL"
+    );
+    await invoiceInput.type(
+      "lightning:LNURL1DP68GURN8GHJ7MRWW4EXCTNXD9SHG6NPVCHXXMMD9AKXUATJDSKKCMM8D9HR7ARPVU7KCMM8D9HZV6E384NXVVEKXSEXYWRYX3SN2EPCX5EKYV35VVEK2VNPXESKZVN9XVCX2VNYX5CXYCFHXUCRJV3JX5UNJV3J8Q6N2WTYVS6NGVE48Q6X2VP3YCHUNC"
```
bumi, line 106: this is problematic because it is doing an HTTP request to an external source.
maybe just put the some JSON in a gist that we control and create an encoded lnurl for that url then.

bumi (approved):

tests/e2e/walletFeatures.spec.ts
```diff
@@ -94,4 +88,31 @@ test.describe("Wallet features", () => {
...
+    await invoiceInput.type(
+      "lightning:lnurl1dp68gurn8ghj7emfwd6zuemfw3582cn4wdjhycm0de6x2mn59e3k7mf0v4ekxctsv4jxxct59a3rqvnyx4jnvenr8ycxzvfk8qex2vp5xdnrjvty8pjnjvpjv56nvtmjv9mj7v3c8pjxywp3xd3ryvnyxy6rydmyxumxve3exccnxvpcxpnxyvnyx43k2wrpx4skvtmvde6hymrpw46xstt9xfjj6ar9wd6zu6nndahqts4jlx"
+    );
+    await (await getByText($optionsdocument, "Continue")).click();
+
+    await findByText(
+      $optionsdocument,
+      "Do you want to login to gist.githubusercontent.com?"
+    );
+    await findByText($optionsdocument, "Login");
```
bumi, line 114:
```suggestion
    await findByText($optionsdocument, "lnurl.getalby.com");
    await findByText($optionsdocument, "Login");
```

I actually think we should also test that the actual login url is shown.

bumi (changes requested):
hhhmm... I think I was wrong. LNURL-auth works differently: 
https://github.com/fiatjaf/lnurl-rfc/blob/legacy/lnurl-auth.md

what we do here works (as it seems)(and actually imo it should also have been implemented like this) but in the lnurl-auth spec it says that the `tag`, `action`, and `k1` are query params on the URL. (and not returned as JSON as you do it here)

so behind that lnurl must be a URL as described [here](https://github.com/fiatjaf/lnurl-rfc/blob/luds/04.md#server-side-generation-of-auth-url-and-signature-verification) 

Also I think no LNURL spec returns the `domain` and `url` that's what we extract and add to be used. (you really really need to learn these specs and also when we write those tests we need to check the specs and use those)

tests/e2e/walletFeatures.spec.ts
escapedcat: Done via `findAllByText` because we have 2 occurrences

bumi (approved):
I think the LNURL is updated. then I guess we're good here?

yay for more tests! 🚀

PR #1430, opened 2022-09-13
Feat/1381 notifications fiat

### LINKED ISSUE #1381: Show fiat rates in notifications
### Is your feature request related to a problem? Please describe

We have now the option to show fiat amounts when we display satoshi amounts.
So far we do not do this in the notifications. we should also show the same amount information.

### Additional context

we need to convert the amount here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/events/notifications.js#L13-L23

Probably it would be helpful if we refactor the currency converter first to do the api call and caching in a background script action?

escapedcat:

- show fiat in notifications
- refactor rate to be sats- instead of a btc-rate to work directly with sats everywhere

### Link this PR to an issue

Fixes #1381

- with fiat
  image
- no fiat
  image

### How has this been tested?

- manually (see screenshots)
- unit-test


im-adithya (approved):

src/common/utils/currencyConvert.ts
```diff
@@ -3,25 +3,18 @@
...
-
-const satoshisToBitcoin = (amountInSatoshis: number | string) => {
-  return Number(amountInSatoshis) / numSatsInBtc;
+const satsToFiat = async (amount: number | string, rate: number) => {
+  return Number(amount) * rate;
 };
 
 const satoshisToFiat = async ({
-  amountInSats,
+  amount,
```
bumi, line 11: I kinda like names that describe what is in there.

bumi (approved):
utACK

did you test it with currencies where the values are different? (e.g. different amount of decimal points or values?

escapedcat:
> did you test it with currencies where the values are different? (e.g. different amount of decimal points or values?

No, thanks!
Currently this would be cut off:
image


We could move Fiat to next line:
image

src/common/utils/currencyConvert.ts
escapedcat: I thought we do not need it anymore, because we got rid of the "btc value". But I can keep `amountInSats`

im-adithya:
Maybe something like this:
`✅ Payment [to XYZ] Successful`
`Amount: 321 sats (25.73 NGN)`
`Fee: 0 sats`

escapedcat:
> Maybe something like this: `✅ Payment [to XYZ] Successful` `Amount: 321 sats (25.73 NGN)` `Fee: 0 sats`

Good idea, let me try if simple line-breaks work in there

src/common/utils/currencyConvert.ts
bumi: I don't care, was just a nit...if one reads it in a year one might not know it. - even though the function is called satoshisInFiat.

PR #1440, opened 2022-09-13
Nostr integration

### LINKED ISSUE #1233: Nostr Integration
### Intro

More and more sites continue to build on nostr. Nostr is 

> https://github.com/nostr-protocol/nostr
> The simplest open protocol that is able to create a censorship-resistant global "social" network once and for all.
>  ... it is based on cryptographic keys and signatures, so it is tamperproof...

which means that it requires users to have one or multiple sets of public-private keypairs. (one for each identity)

Since the browser extension is already connected to your node we could make use of the same cryptographic keypair to sign nostr messages. That would remove the need to separately backup nostr keys. 

### Technical analysis

Lot's of the UI is already there (to enable sites and sign messages) and could potentially be reused to handle `window.nostr` API calls in the same way we currently handle `window.webln` calls. 

The following API specs would need to be implemented in the provider:
```
async window.nostr.getPublicKey(): string // returns your public key as hex
async window.nostr.signEvent(event): Event // returns the full event object signed
async window.nostr.getRelays(): { [url: string]: RelayPolicy } // returns a map of relays
async window.nostr.nip04.encrypt(pubkey, plaintext): string // returns ciphertext+iv as specified in nip04
async window.nostr.nip04.decrypt(pubkey, ciphertext): string // takes ciphertext+iv as specified in nip04
```

There already is an implementation available here:

https://github.com/fiatjaf/nos2x

### Questions
 - How would we handle that for custodial solutions? (Alby Accounts, LNBits, etc)
 - How can we derive nostr keys from the nodes private key? 

### Links
 - [nostr](https://github.com/nostr-protocol/nostr)
 - [nostr Browser Extension](https://github.com/fiatjaf/nos2x)
 - [nostr NIP07](https://github.com/nostr-protocol/nips/blob/master/07.md)

reneaaron:

Fixes #1233 


src/extension/background-script/actions/nostr/getPublicKey.ts
```diff
@@ -0,0 +1,9 @@
+import { Message } from "~/types";
+
+const getPublicKey = async (message: Message) => {
+  return {
+    data: "4a8c13dd19bff29de83e26ab0b73441d18e39cc3838e323eecc12502c4698beb",
```
Cameri, line 5: Hmmm

reneaaron: It's WIP, key will be derived from your node's key material.

reneaaron:
Implemented a first working version.  

Some notes: 
 - nip04.encrypt / nip04.decrypt will be added in a separate step (maybe it makes sense to create some more generic confirm pages that can be reused across the application)
 - We might want to add a "remember" function sooner than later to prevent popups for every action on the site
 - Blocking functionality is also implemented for nostr (if you block a site it's blocked extension-wide)
 - Do we need separate keys on an account level? (currently they are on a global level which means you can have only one private key for all of your accounts)

reneaaron:
Review
 - [ ] Save private key in encrypted storage (`saveEncryptedSetting`)
 - [ ] Check for `window.postMessage()`
 - [ ] Check handling for multiple calls in parallel (while still a window is open)
 - [ ] Prefix `webln` actions
 - [ ] Check if `nostr-tools` are necessary

src/extension/background-script/actions/nostr/helpers.ts
```diff
@@ -0,0 +1,44 @@
+import * as secp256k1 from "@noble/secp256k1";
+import createHash from "create-hash";
```
bumi, line 2: can we use crypto-js for the hash?  
so far `create-hash` is an indirect dependency I think. crypto-js is a direct dependency.

escapedcat:
nit: button color on dark-mode looks off:  
image

Following your video, using this branch and Chromium I get:
image
Do I need to user another browser?
Copying & pasting the key manually successfully creates an account for me though.

src/extension/content-script/onstart.ts
```diff
@@ -9,6 +10,14 @@ async function onstart() {
     return;
   }
 
-  injectScript(browser.runtime.getURL("js/inpageScriptWebLN.bundle.js")); // injects the webln object
+  // window.webln
+  injectScript(browser.runtime.getURL("js/inpageScriptWebLN.bundle.js"));
+
+  // window.nostr
+  const nostrEnabled = (await api.getSettings()).nostrPrivateKey;
```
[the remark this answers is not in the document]
reneaaron: Changed to nostrEnabled.

src/app/screens/Settings.tsx
```diff
@@ -29,13 +29,47 @@ function Settings() {
...
+  // TODO: Load setting from encrypted storage
+  const [nostrPrivateKey, setNostrPrivateKey] = useState("");
+
+  const getPrivateKeyFromStorage = async () => {
+    const priv = (await utils.call("nostr/getPrivateKey")) as string;
+    setNostrPrivateKey(priv ?? "");
+  };
+
+  useEffect(() => {
+    getPrivateKeyFromStorage().catch(console.error);
```
escapedcat, line 41: Will this automatically console an error?

reneaaron: yes, if for instance the private key could not be decoded with the password, etc

src/extension/background-script/actions/nostr/helpers.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Event } from "~/extension/ln/nostr/types";
+
+export function validateEvent(event: Event) {
+  if (event.id !== getEventHash(event)) return false;
+  if (typeof event.content !== "string") return false;
+  if (typeof event.created_at !== "number") return false;
+
+  if (!Array.isArray(event.tags)) return false;
+  for (let i = 0; i < event.tags.length; i++) {
```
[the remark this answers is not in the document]
escapedcat: Ah, got it! Yes.  
Does this work?: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of  
It's a bit easier to read and we could get rid of the `i++`?  
Not blocking though.

escapedcat (approved):

bumi:
When I have my extension locked and go to nostr.chat I get this error:

src/app/screens/Settings.tsx
```diff
@@ -385,6 +417,60 @@ function Settings() {
...
+        <Setting
+          title={t("nostr.private_key.title")}
+          subtitle={t("nostr.private_key.subtitle")}
+        >
+          {!isLoading && (
+            <>
+              <div className="w-96 flex justify-end">
+                <div className="w-96 flex-auto">
+                  <Input
+                    type="text"
```
bumi, line 445: should we make this a type password to hide it? people will otherwise share sceenshots of it.

src/extension/background-script/actions/nostr/generatePrivateKey.ts
```diff
@@ -0,0 +1,31 @@
...
+  try {
+    const response = await connector.signMessage({
+      message:
+        "DO NOT EVER SIGN THIS TEXT WITH YOUR PRIVATE KEYS! IT IS ONLY USED FOR DERIVATION OF NOSTR KEY PAIRS, DISCLOSING THIS SIGNATURE WILL COMPROMISE YOUR NOSTR IDENTITY!",
+      key_loc: {
+        key_family: 0,
+        key_index: 0,
+      },
+    });
+    const keymaterial = response.data.signature;
```
bumi, line 16: we probably should add some assertion here that we don't have a  blank signature.

src/extension/background-script/actions/nostr/generatePrivateKey.ts
```diff
@@ -0,0 +1,31 @@
...
+    const uint8 = Uint8Array.from(
+      keymaterial.split("").map((x) => x.charCodeAt(0))
+    );
+    const hex = Buffer.from(secp256k1.utils.hashToPrivateKey(uint8));
```
bumi, line 20: the variable name says `hex` but the content is not hex here. 

I don't think you should do this `keymaterial.split` thing. Are you sure this is needed? (looks a bit like a stack overflow answer ;) )

can the `secp256k1.utils.hexToBytes()` help here? that takes a hex string and converts it into bytes that can be passed to `hashToPrivateKey`.

src/extension/background-script/actions/nostr/getPrivateKey.ts
```diff
@@ -0,0 +1,13 @@
+import { decryptData } from "~/common/lib/crypto";
+
+import state from "../../state";
+
+const getPrivateKey = async () => {
```
bumi, line 5: actually can we avoid this action? 
"no code is better than no code" :) any code can break and if we don't even have a function to return the private key then this can not go wrong. 

If I see this correctly, then we only need this action to prefill the input field in the settings and to warn the user that it will be overwritten.

bumi: even though we might want to have an "export key" function (but do we need this now?)

src/extension/background-script/actions/nostr/getPublicKey.ts
```diff
@@ -0,0 +1,22 @@
...
+  const encryptedPrivateKey = state.getState().nostrPrivateKey;
+  if (!encryptedPrivateKey) {
+    return;
+  }
+  const decryptedPrivateKey = decryptData(
+    encryptedPrivateKey,
+    state.getState().password as string
+  );
+  const buffer = Buffer.from(decryptedPrivateKey, "hex");
+  const publicKey = Buffer.from(
```
bumi, line 15: I think instead of the `Buffer` you can use `secp.utils.bytesToHex` 
(Buffer is some polyfill and thus I'd avoid it as much as possible?)

src/extension/background-script/actions/nostr/helpers.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (typeof tag[j] === "object") return false;
+    }
+  }
+
+  return true;
+}
+
+export async function signEvent(event: Event, key: string) {
+  return Buffer.from(
+    await secp256k1.schnorr.sign(getEventHash(event), key)
```
bumi, line 25: I am scared of 🐉 with this syntax that uses await in a function parameter.

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -0,0 +1,55 @@
...
+  // Set the message as the user needs to see the event details
+  message.args.message = JSON.stringify(message.args.event);
+
+  try {
+    const response = await utils.openPrompt({
+      ...message,
+      action: "confirmSignMessage",
+    });
+
+    const pw = state.getState().password;
```
bumi, line 31: was wondering if we might want to move the signing (which includes the lookup and decryption of the private key) out of the action.

maybe we can do this similarly as with the connector which is an object we keep in the `state`.
then we don't need to handle the password and keys in different actions

src/extension/content-script/onend.js
```diff
@@ -8,16 +8,16 @@ import shouldInject from "./shouldInject";
...
-  "keysendOrPrompt",
-  "makeInvoice",
-  "signMessageOrPrompt",
+  "webln/enable",
+  "webln/getInfo",
+  "webln/lnurl",
+  "webln/sendPaymentOrPrompt",
+  "webln/keysendOrPrompt",
+  "webln/makeInvoice",
+  "webln/signMessageOrPrompt",
```
bumi, line 17: we'll likely get conflicts here with the Webbtc PR

src/i18n/locales/en/translation.json
```diff
@@ -443,6 +443,17 @@
...
         "success": "Password changed successfully"
+      },
+      "nostr": {
+        "title": "Nostr",
+        "hint": "is a simple and open protocol that aims to create censorship-resistant social networks.",
+        "private_key": {
+          "title": "Private key",
+          "subtitle": "If you don't have a key already you can generate one based on the current account you use.",
+          "generate": "Generate",
+          "warning": "You are about to override your current private key. Are you sure?",
```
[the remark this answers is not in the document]
reneaaron: Didn't like the two terms `current key` and `old private key` referencing the same thing. 

`This will delete your old private key. Are you sure?`

src/app/screens/Settings.tsx
reneaaron: Good point, thanks for raising that. I'll change it.

src/extension/background-script/actions/nostr/getPrivateKey.ts
reneaaron: I generally share your concerns. Anyway I think it would be a good think for users to also be able to copy their private key, also after the initial setup. 

Some clients don't support `window.nostr` and you will need to plug your private key into those clients directly. 🤔

src/app/screens/Nostr/ConfirmGetPublicKey.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  async function block(event: React.MouseEvent<HTMLAnchorElement>) {
+    event.preventDefault();
+    await utils.call("addBlocklist", {
+      domain: origin.domain,
+      host: origin.host,
+    });
+    msg.error(
+      `User added site to blocklist domain, host
```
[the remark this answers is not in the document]
reneaaron: Yes, fixed that in the default confirm screen as well.

src/extension/background-script/actions/nostr/helpers.ts
reneaaron: why?

src/extension/background-script/actions/nostr/getPublicKey.ts
reneaaron: good point, thanks for the hint

src/extension/background-script/actions/nostr/helpers.ts
bumi: maybe because my mind still thinks in polyfills and old JS... and I translate async/await in Promise.resolve chains.
is this reliable in all the different potential environments?

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
bumi: generally just because this is sensitive code and data. having this in a single place sounds to me less error prone.

src/extension/background-script/actions/nostr/helpers.ts
reneaaron: the good thing about `await` is that exceptions bubble up naturally as they are not captured in `then()` and `catch()` functions. if this call fails the calling function will capture the exception and return an error to the `nostr.signEvent()` call.

bumi: I am not saying not to use await. but rather: 
```
const signature =  await secp256k1.schnorr.sign(getEventHash(event), key);
return Buffer.from(signature);
``

(but would be nice to get rid of the Buffer)

File: src/extension/background-script/actions/nostr/signEventOrPrompt.ts (hunk 12 above)
Comment (reneaaron) [reply src/extension/background-script/actions/nostr/signEventOrPrompt.ts]: moved everything to a Nostr object in state.

File: src/extension/background-script/actions/nostr/getPublicKey.ts [hunk 16]
```diff
@@ -0,0 +1,8 @@
+import state from "~/extension/background-script/state";
+
+const getPublicKey = async () => {
+  const publicKey = state.getState().getNostr().getPublicKey();
+  return publicKey;
```
bumi, line 5: don't we need to always return an object with a `data` (or `error`) key?

src/extension/background-script/router.ts
```diff
@@ -59,6 +50,29 @@ const routes = {
...
   lnurlAuth: auth,
   getCurrencyRate: cache.getCurrencyRate,
+  nostr: {
+    generatePrivateKey: nostr.generatePrivateKey,
+    getPrivateKey: nostr.getPrivateKey,
+    setPrivateKey: nostr.setPrivateKey,
+  },
+
+  // Public calls are accessible from inpage scripts
+  public: {
```
bumi, line 60: as we have this new nesting here... and just to review things... can somebody review the router function that resolves those routes?

src/extension/background-script/actions/nostr/getPrivateKey.ts
bumi: hmm. yes. I see. maybe it's OK here. 

how else could we allow people to export it? 🤔

src/app/screens/Settings.tsx
```diff
@@ -331,9 +368,74 @@ function Settings() {
...
+                  </button>
+                }
+              />
+            </div>
+            {!nostrPrivateKey && (
+              <div className="flex-none ml-2 flex-end">
+                <Button
+                  label={t("nostr.private_key.generate")}
+                  onClick={async () => {
+                    const result = await utils.call("nostr/generatePrivateKey");
```
bumi, line 425: we should tell the user here what happens and that the currently selected account is used to generate the key. 

and also that they should make a backup of this key to not loose it.

src/extension/background-script/actions/nostr/generatePrivateKey.ts
```diff
@@ -0,0 +1,34 @@
...
+        key_family: 0,
+        key_index: 0,
+      },
+    });
+    const keymaterial = response.data.signature;
+    if (!keymaterial.length) {
+      throw Error("No key material available.");
+    }
+    const uint8 = Uint8Array.from(
+      keymaterial.split("").map((x) => x.charCodeAt(0))
```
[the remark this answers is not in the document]
reneaaron: Changed to `hashToPrivateKey` (which can take 40 or more bytes of uniform input)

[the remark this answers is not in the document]
reneaaron: ℹ️ This will lead to a change in the keypairs being generated from your account.

src/app/screens/Settings.tsx
reneaaron: good point, will add that to the setting description

src/extension/background-script/actions/nostr/getPublicKey.ts
escapedcat: Good point. I think this is necessary for actions that are [called/used by utils/call](https://github.com/reneaaron/lightning-browser-extension/blob/feature/nostr/src/common/lib/utils.ts#L34). This is expecting a `data` prop.  
In this case it's only just called directly by `getPublicKeyOrPrompt`. So if `getPublicKey` is never called via a `message` does it need it own action-file? Would it be ok to move it into `getPublicKeyOrPrompt`?

reneaaron: Usually yes, but this function was only used internally. Changed usages to use state.getPublicKey() directly and removed this function completely.

src/extension/background-script/actions/nostr/getPrivateKey.ts
reneaaron: an option would be to let the user enter the password (before revealing the key). i guess that's a bit much to add for an experimental feature though.

bumi: yes, and it would still require the getprivatekey function. 
it's fine. I think we have this for the accounts also.

PR #1446, opened 2022-09-14
feat: success screen component (LNURLPay)

im-adithya:

Adds a new success/failure card component

### Link this PR to an issue

#1379

<p>

</p>

### How has this been tested?

Have to check the Failure screen (only checked it by throwing a hardcoded error)

### Checklist


src/app/screens/LNURLPay/index.tsx
```diff
@@ -263,53 +270,25 @@ function LNURLPay() {
...
-      descriptionList = [["Message", successAction.message]];
+  function renderResultAction() {
+    if (!result) return;
+
+    if (result === "success" && !successMessage) {
+      setSuccessMessage(
+        `${valueSat} (SATS) ${showFiat ? "(" + fiatValue + ")" : ""} ${t(
+          "were_sent_to"
+        )} ${navState.origin?.name || getRecipient()}`
+      );
```
im-adithya, line 281: Not sure what to do here regarding the message to show exactly.
//cc @escapedcat @bumi

escapedcat: Hm, what do you mean? Should we have a look together?

src/app/components/ResultCard/index.tsx
```diff
@@ -0,0 +1,17 @@
+export type Props = {
+  success: boolean;
```
escapedcat, line 2: Maybe make this `isSuccess` to indicate it's a boolean

escapedcat (commented):
- Would be nice to get the icons as SVGs to avoid future scaling issues and make reusability easier
- Will the `ResultCard` be applied to all "send" screens in further PRs?

src/app/screens/LNURLPay/index.tsx
```diff
@@ -318,8 +297,16 @@ function LNURLPay() {
   return (
     <>
       <div className="flex flex-col grow overflow-hidden">
-        <ScreenHeader title={tCommon("actions.send")} />
-        {!successAction ? (
+        <ScreenHeader
+          title={
+            !result
```
escapedcat, line 302: Is this a ternary if/else-if? Might be easier to read to move this into a function?

src/app/components/ResultCard/index.tsx
```diff
@@ -0,0 +1,17 @@
+export type Props = {
```
escapedcat, line 1: Simple render unit-test would be nice

escapedcat (approved):

bumi:
when does this error happen? and from the screenshot it seems the user is in a dead end there? typically a user would want to retry it? also why do we show errors here differently than in the other screens where the error message is shown under the input?

escapedcat:
> when does this error happen?

I think this was just an example-message to show how that screen looks like, right @im-adithya ?

> and from the screenshot it seems the user is in a dead end there?
> typically a user would want to retry it?

You mean because before a toast was used and the user could try again? If yes, then instead of "close" on the error-screen, let's use "Retry" and the user will see the form with the input again?

> also why do we show errors here differently than in the other screens where the error message is shown under the input?

Let me talk to @dvoroneca

bumi:
:) that does not answer my question of when this error happens. 

I question the need of this whole screen. before it was a toast() and now it is much more code, we need to navigate with some new logic on a rather hidden button, etc. 
@dvoroneca is this really better? and is this really so much better that it is worth spending the time here?

dvoroneca:
A valid concern. Would need to find out real cases when the transaction doesn't go through and other errors to answer this question.

escapedcat:
Maybe errors in general do not need to be blocking. Errors can be shown below i.e. the form.
Currently we use a mix of toast and "show below form".
We can still align the success-screens for all "send"-types.

bumi:
> A valid concern. Would need to find out real cases when the transaction doesn't go through and other errors to answer this question.

why do the cases when a transaction goes wrong matter?

dvoroneca:
The designs of these screens had a CTA specific for cases when an error occurs. [File here](https://www.figma.com/file/HXIxAFG9OUCcr2EJMJ6JMm/1-Click-Pay?node-id=321%3A8575)
Screenshot 2022-09-29 at 13 05 32

There could be cases when the connection is lost, the node is down or other technical issues that prevent from sending funds. It is true that we already have a popup error that could be displayed on the screen before this one (where you have inputs and a call to action to pay) but that popup doesn't have a CTA to direct the user to some resource.... help for example.

For us, it's also a way to know what kind of errors users facing. When they contact support and say there was an error when sending funds. 

Please let me know if I'm missing something.

dvoroneca:
How about we roll back the error screen for now and leave the success screen? @escapedcat @escapedcat @bumi

escapedcat:
> How about we roll back the error screen for now and leave the success screen?

Yeah, that would be great. At least in this case the design isn't wasted and we finally can align the send-screens in a way.

Issue #1451, opened 2022-09-15
Don't close prompts if there is an error


bumi:
### Describe the bug

Currently we always close the prompts. This means users can not see potential error messages. 
if there is an error we should not close the prompt.

Example:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/LNURLWithdraw/index.tsx#L72

but this probably applies to the other prompts, too.

### To Reproduce

Try to withdraw from lnmarkets less than 1000sats

### Alby information

- Alby Version: [e.g. 1.15.1]

PR #1452, opened 2022-09-15
fix(lnurlwithdraw): display error on prompt #1451

escapedcat:
#### Other LNURLActions

- LNURLAuth: aligning error-handling with other LNURLActions
- LNURLChannel: is already taking care of this
- LNURLPay: is already taking care of this

### Link this PR to an issue

#1451

### How has this been tested?

Tested with LNMarkets (see screenshot)

### DIFF
--- a/src/app/screens/LNURLAuth/index.tsx
+++ b/src/app/screens/LNURLAuth/index.tsx
@@ -7,7 +7,6 @@ import { useState } from "react";
 import type { MouseEvent } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
-import { toast } from "react-toastify";
 import ScreenHeader from "~/app/components/ScreenHeader";
 import { useNavigationState } from "~/app/hooks/useNavigationState";
 import { USER_REJECTED_ERROR } from "~/common/constants";
@@ -29,6 +28,7 @@ function LNURLAuth() {
   });
 
   const [successMessage, setSuccessMessage] = useState("");
+  const [errorMessage, setErrorMessage] = useState("");
 
   async function confirm() {
     try {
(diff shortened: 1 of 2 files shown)


src/app/screens/LNURLWithdraw/index.tsx
```diff
@@ -62,15 +62,14 @@ function LNURLWithdraw() {
 
       if (response.data.status.toUpperCase() === "OK") {
         setSuccessMessage("Withdraw request sent successfully.");
+        // ATTENTION: if this LNURL is called through `webln.lnurl` then we immediately return and return the response. This closes the window which means the user will NOT see the above successAction.
```
escapedcat, line 65: @bumi this seems to do the trick. Do you see any issues with this?  
I could apply this to the other prompts.

bumi: yep, I guess something like that 👍 
either moving the reply up. or extending the `if (navState.isPrompt) {` around the reply.

bumi (approved):

src/app/screens/LNURLAuth/index.tsx
```diff
@@ -50,19 +51,18 @@ function LNURLAuth() {
...
-      // We assume this is OK when it is called through webln.
-      if (navState.isPrompt) {
-        msg.reply(response);
+        toast.error(`Error: Auth status is not ok`);
       }
     } catch (e) {
       console.error(e);
       if (e instanceof Error) {
-        toast.error(`Error: ${e.message}`);
+        setErrorMessage(`Error: ${e.message}`);
```
bumi, line 65: above we're using toast and here we set an error message. should those be in sync?

bumi (approved):

src/app/screens/LNURLAuth/index.tsx
escapedcat: Yes, changed it. I think currently we use the toast if i.e. a call fails. For other errors we can use the error-message.

PR #1485, opened 2022-09-20
chore(verifymessage): remove support #1345

### LINKED ISSUE #1345: WebLN: VerifyMessage
### Describe the bug

Signing and verifying a message isn't creating valid signatures, currently they are `{ valid: false }`.

```js
await webln.enable();
var result =  await webln.signMessage(message);
console.log(result);
console.log(await webln.verifyMessage(result.signature, message));
```

➡️ [Demo](https://codepen.io/reneaaron/pen/QWmRzyg?editors=1111)
(see console output)

### Expected behavior

 - Signing and verifying should produce valid signatures. 
 - Shouldn't there be a popup that allows users to verify signatures in Alby? (users would need to see the message, signature and the result to verify the integrity)
 - What is the use case of `verifyMessage`?

escapedcat:

Removes related code and throws an error that Alby does not support this method

### Link this PR to an issue

Fixes #1345

chore

### How has this been tested?

Tested with Renes codepen:
image


reneaaron (approved):
utACK 👍

Issue #1457, opened 2022-09-16
Refactor: use `navState` for all route actions (instead of props/searchParams)


escapedcat:
### Is your feature request related to a problem?

We aligned all LNURL-actions to use `navState` with i.e. #1269 

Some actions are still using the old pattern, i.e.:
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Prompt/Prompt.tsx#L75
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Prompt/Prompt.tsx#L92
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Prompt/Prompt.tsx#L108
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/index.tsx#L260-L275

### Describe the solution you'd like

These components should not use props/searchParams but `navState` like the LNURL-action-components i.e.:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/LNURLPay/index.tsx

To keep this easy to review multiple PRs should be created for each route-action.

### Are you working on this?

No

PR #1486, opened 2022-09-20
chore: align confirmkeysend to use navstate

im-adithya:

Uses navstate in confirmkeysend

### Link this PR to an issue

#1457

### Checklist

### DIFF
--- a/src/app/screens/ConfirmKeysend/index.test.tsx
+++ b/src/app/screens/ConfirmKeysend/index.test.tsx
@@ -1,0 +1,95 @@
+import { render, screen, act } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { MemoryRouter } from "react-router-dom";
+import { settingsFixture as mockSettings } from "~/../tests/fixtures/settings";
+import * as SettingsContext from "~/app/context/SettingsContext";
+import type { OriginData } from "~/types";
+
+import ConfirmKeysend from "./index";
+
+jest.spyOn(SettingsContext, "useSettings").mockReturnValue({
+  settings: mockSettings,
+  isLoading: false,
+  updateSetting: jest.fn(),
+  getFiatValue: jest
+    .fn()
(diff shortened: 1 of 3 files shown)


src/app/screens/ConfirmKeysend/index.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    expect(screen.getByText("21 sats")).toBeInTheDocument();
+    expect(screen.getByText("(~$0.01)")).toBeInTheDocument();
+    expect(screen.queryByText("~$0.05")).not.toBeInTheDocument();
+
+    await act(() => {
+      user.click(screen.getByText("Remember and set a budget"));
+    });
+
+    const input = await screen.findByLabelText("Budget");
+    expect(input).toHaveValue(210);
```
escapedcat, line 92: Might be better to use the amount form above `*10` like we do here:
https://github.com/getAlby/lightning-browser-extension/blob/task-navstate/confirmkeysend/src/app/screens/ConfirmPayment/index.test.tsx#L96  
To make it more obvious what relation the initial budget has with the amount.

src/app/screens/ConfirmKeysend/index.tsx
```diff
@@ -6,25 +6,25 @@ import PublisherCard from "@components/PublisherCard";
...
-  destination?: string;
-  customRecords?: Record<string, string>;
-  valueSat?: string;
-};
+function ConfirmKeysend() {
+  const navState = useNavigationState();
+  const destination = navState.args?.destination as string;
+  const amount = navState.args?.amount as string;
+  const customRecords = navState.args?.customRecords as Record<string, string>;
+  const origin = navState.origin;
```
escapedcat, line 25: Do we know why we use `origin` and `originRef` here? Wondering if that's needed. Looks like this happened by accident earlier? What do you think?

escapedcat (approved):
ack

PR #1494, opened 2022-09-21
feat(allowance): enabled/disable #1142

### LINKED ISSUE #1142: Disable allowance - should be able to disable Alby from a service inside Alby without delete it

> I see that now we have green Alby icon when connected to a service. I think could be great to be able to disconnect from the service inside Alby.  

As a user I want to be able to "disable" an allowance/budget to stop auto-payment for a webpage if I want to.
Without resetting or deleting the budget.

Can disable/enable budgets via the settings-cog icon

### Describe alternatives you've considered

none

### Additional context

none

### Are you working on this?

No

escapedcat:

- enabling/disabling an allowance via allowance-menu (cog-icon)
- "disabled" batch show if allowance is disabled

### Link this PR to an issue

Fixes #1142

#### Multiple allowances with different status
image
#### Allowance is disabled
image
#### Allowance is enabled
image

### How has this been tested?

- manually via nakaphoto

### DIFF
--- a/src/app/components/AllowanceMenu/AllowanceMenu.stories.tsx
+++ b/src/app/components/AllowanceMenu/AllowanceMenu.stories.tsx
@@ -2,7 +2,7 @@ import AllowanceMenu from ".";
 
 export const Default = () => (
   <div className="max-w-xs flex justify-end">
-    <AllowanceMenu allowance={{ id: 4, totalBudget: 1000 }} />
+    <AllowanceMenu allowance={{ id: 4, totalBudget: 1000, enabled: true }} />
   </div>
 );
 
--- a/src/app/components/AllowanceMenu/index.test.tsx
+++ b/src/app/components/AllowanceMenu/index.test.tsx
@@ -30,6 +30,7 @@ const props: Props = {
   allowance: {
     id: 1,
     totalBudget: 2000,
+    enabled: true,
   },
   onEdit: mock,
 };
@@ -81,5 +82,13 @@ describe("AllowanceMenu", () => {
     });
 
     await waitFor(() => expect(mock).toHaveBeenCalled());
+  });
+
+  test.skip("enable/disable", async () => {
+    render(
+      <MemoryRouter>
+        <AllowanceMenu {...props} />
+      </MemoryRouter>
+    );
   });
 });
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -15,7 +15,7 @@ import Menu from "../Menu";
 import DualCurrencyField from "../form/DualCurrencyField/index";
 
 export type Props = {
-  allowance: Pick<Allowance, "id" | "totalBudget">;
+  allowance: Pick<Allowance, "id" | "totalBudget" | "enabled">;
   onEdit?: () => void;
   onDelete?: () => void;
 };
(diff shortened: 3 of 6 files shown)


bitcoinuser:
Hi,

In my opinion should have these categories (tags):

1. Instead of ACTIVE, should be CONNECTED. Because sometimes I'm connected but with no budget setup.
2. Should have a new tag called BUDGET, to cases where users specified a budget.
3. Should have a new tag called LIGHTNING LOGIN, to cases where users logged in in the site.
4. Should have have the tag DISCONNECTED, to cases where I'm not connected.

obs. One site could have more than one tag.

cc: @bumi

bitcoinuser:
I think something's are confusing:

1. Disable allowance is the same thing of edit allowance and add 0 value, right?
2. Disable allowance is the same thing of the block and ignore?
3. When I disable allowance is like I can't use Alby to directly interact with a page?

escapedcat:
>  - Instead of ACTIVE, should be CONNECTED. Because sometimes users are connected but with no budget setup.

Currently this will only make a website show up in "Websites". I don't think this needs an extra tag/badge.

>  - Should have a new tag called BUDGET, to cases where users specified a budget.

This could make sense

>  - Should have a new tag called LIGHTNING LOGIN, to cases where users logged in the site using Alby.

Not sure if this is possible or wanted to the right place to show it. Would need to check with the team.

>  - Should have the tag DISABLED, to cases where users DISABLED Alby on a specific site.

This is what this PR introduces, no?

>    - Disable allowance is the same thing of edit allowance and add 0 value, right?

No, I think in that case an allowance would still be active but you need to confirm every payment. But it is still "connected (enabled)"

>    - Disable allowance is the same thing of the block and ignore?

No, a disabled allowance will ask you to "connect" on a website you visit. Of that website is blocked it will not even ask you. You need to manually unblock it here:
image


>    - When I disable allowance is like I can't use Alby to directly interact with a page?

Define "directly interact". But I think yes. You need to "connect" again first to enable the allowance.

bitcoinuser:
Hi,

Let me try to clarify:

1. For me as an user, it's hard to understand the meaning of word ACTIVE on websites page. I can see now that ACTIVE means that the user marked the option remember and set a budget. But I think ACTIVE word could be changed to be more clear what really means. I think BUDGET ACTIVATED word reflect much better what it is.

2. The DISABLED as I can see, is to pause the option of (to not be asked for confirmation on payments until it is exhausted). So maybe the word could be: BUDGET DISABLED

> Currently this will only make a website show up in "Websites". I don't think this needs an extra tag/badge.

Understood. Thanks

> Not sure if this is possible or wanted to the right place to show it. Would need to check with the team.

I think would be great and useful. 

> This is what this PR introduces, no?

Yes I was confusing. But it's like pause and resume.

> No, I think in that case an allowance would still be active but you need to confirm every payment. But it is still "connected (enabled)"

Understood. Thanks

> No, a disabled allowance will ask you to "connect" on a website you visit. Of that website is blocked it will not even ask you. You need to manually unblock it here:

Understood. Thanks

bitcoinuser:
- Let me try to clarify:

For me as an user, it's hard to understand the meaning of word ACTIVE on websites page. I can see now that ACTIVE means that the user marked the option remember and set a budget. But I think ACTIVE word could be changed to be more clear what really means. I think BUDGET ACTIVATED word reflect much better what it is.

The DISABLED as I can see, is to pause the option of (to not be asked for confirmation on payments until it is exhausted). So maybe the word could be: BUDGET DISABLED

But ok, on Portuguese Brazilian I will use a word that make more clear.

bumi:
adding @dvoroneca here

bumi:
why is this closed?

escapedcat:
> why is this closed?

https://github.com/getAlby/lightning-browser-extension/issues/1142#issuecomment-1259510180

Talked to Leo about the background/context of this because also Diana and Rene were wondering what the usecase here is. Turns out we do not need this. But Leo created related tickets which we will tackle.

bumi:
@dvoroneca @reneaaron is there really no usecase in disabling the allowance?

@escapedcat you see a use case there I guess, as you have started this PR?

escapedcat:
> @escapedcat you see a use case there I guess, as you have started this PR?

I looked into it because it was part of the milestone I think. It's not lost code/experience. This will help for further "badge" adjustments along the way.

Issue #1490, opened 2022-09-20
Balance on home screen


reneaaron:
### Is your feature request related to a problem? Please describe

As a user I want to see the total balance on the home screen. Currently there is a PublisherCard:

![image](https://user-images.githubusercontent.com/100827540/191240947-d46b97f0-136c-4ebe-b595-5a7e9c75bb7e.png)

Replace the `PublisherCard`with the total balance of the current account on the Home Screen

Maybe we could also remove the balance from the dropdown then:

![image](https://user-images.githubusercontent.com/100827540/191241667-14fa7ecc-cfd6-451f-98ab-4e62b86a0c39.png)

### Are you working on this?

No.

escapedcat:
@dvoroneca assigning this to you till the design is done. Hope that's ok.

escapedcat:
Further notes:
When we talk about the design we should think about splitting `PublisherCard` (for the allwoance view) from "found lightning-address info". This tends to be highly confusing and causes misunderstandings and bugs. If we can not tackle this in this issue we can open a new one.

reneaaron:
Closing in favor of #2236.

PR #1496, opened 2022-09-21
fix(home): prefer lndata over allowance in publishercard

### LINKED ISSUE #1483: When an allowance is set the origin data from current tab is being shown in the publisher card on the home screen
### Steps to reproduce

1. Go to a [youtube channel](https://www.youtube.com/watch?v=-WAZ2UCabZs#) with a Lightning Address configured in their channel (with existing allowance for YT)
2. Open the extension popup
    image
3. You will see metadata from the page (and not from the creator) like on the homescreen:
    image

### Expected outcome

See the data from the creator on the `PublisherCard` instead of `Youtube`

escapedcat:

Prefer matching ln-address info over general origin info

### Link this PR to an issue

Fixes #1483

### How has this been tested?

Tested on YT with monvol

### DIFF
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -207,12 +207,12 @@ function Home() {
 
   function renderPublisherCard() {
     let title, image;
-    if (allowance) {
-      title = allowance.name;
-      image = allowance.imageURL;
-    } else if (lnData.length > 0) {
+    if (lnData.length > 0) {
       title = lnData[0].name;
       image = lnData[0].icon;
+    } else if (allowance) {
+      title = allowance.name;
+      image = allowance.imageURL;
     } else {
       return;
     }


bumi:
cc @reneaaron

reneaaron (approved):
utACK

im-adithya:
Two concerns:
- But what if the user presses the back button? [They would then see the MONVOL publisher card again]
- Wouldn't this completely remove the allowance information from the popup, what if there's a site like BOLT.FUN (where users also require allowance information) which also has an ln address within it?

reneaaron:
ad @im-adithya 

> But what if the user presses the back button? [They would then see the MONVOL publisher card again]
I think this is addressed with #1490 

> Wouldn't this completely remove the allowance information from the popup, what if there's a site like BOLT.FUN (where users also require allowance information) which also has an ln address within it?

It should just prefer the current pages data (picked up by a battery) over the data from the allowance. If there is no data available on the current page the information from the allowance should still be displayed. Also the allowance is only relevant for WebLN calls on the page (not for payments to creators picked up by a battery).

@escapedcat I just tested the PR and noticed some behavior (probably not related to the PR though):

Currently there is no more back button and it seems to start in the HomeScreen: (instead of the PubsliherScreen) 🤔

image
https://www.youtube.com/watch?v=-WAZ2UCabZs#

escapedcat:
@im-adithya 
> But what if the user presses the back button? [They would then see the MONVOL publisher card again]

Yes, but that's just default behaviour, no? If there's no allowance but a ln-address is found you see the publisherCard on the home-screen anyways.

> Wouldn't this completely remove the allowance information from the popup, what if there's a site like BOLT.FUN (where users also require allowance information) which also has an ln address within it?

Not completely. It will still show the url, i.e.:
image


@reneaaron 
> Currently there is no more back button and it seems to start in the HomeScreen: (instead of the PubsliherScreen) 🤔

Which color does the Alby favicon have? See my screenshot in this comment. This looks correct to me, no? For me it takes soem time till YT loaded everythign and only then the icon is green and only then I get the allowance view inclduign the back-button.

im-adithya (approved):

PR #1503, opened 2022-09-22
feat: check ln address in location

### LINKED ISSUE #1056: Alby does not pickup LN address from Github when entered in location
Alby does pickup LN address from Twitter when it is entered in location - see https://twitter.com/pavolrusnak as an example

The same is not true for Github where Alby only checks the bio field - see https://github.com/prusnak as an example

It would be nice if Alby would scan also other fields not just the bio

The Twitter location scan was added in https://github.com/getAlby/lightning-browser-extension/pull/816

As can be seen here, on Github Alby only scans the bio text (and the main article):

https://github.com/getAlby/lightning-browser-extension/blob/0d71999d03e02d7ab920b7b4e8474a6662f7d636/src/extension/content-script/batteries/GitHub.ts#L34-L36

im-adithya:

Ln Address from location in GitHub

### Link this PR to an issue

Closes #1056

### How has this been tested?

Manually

### Checklist

### DIFF
--- a/src/extension/content-script/batteries/GitHub.ts
+++ b/src/extension/content-script/batteries/GitHub.ts
@@ -42,7 +42,7 @@ function handleProfilePage() {
   if (!shortBioElement) return;
 
   const address =
-    parseElement(".Layout-sidebar .h-card [data-bio-text]") ||
+    parseElement(".Layout-sidebar .h-card") ||
     parseElement(".Layout-main article");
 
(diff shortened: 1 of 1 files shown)


escapedcat:
Doesn't seem to work:
image
Tested with my profile. Did it work for you?

escapedcat (approved):
Works!

PR #1506, opened 2022-09-22
fix: did youtube update the UI?

bumi:
Looks like (some) users getting a new UI:
image

Some might still get the old UI.


im-adithya:
Mine still looks old but the structure is the new one
i.e.
```JS
document.querySelector<HTMLImageElement>(
      "#columns #primary #primary-inner #owner #avatar img" // support maybe new UI being rolled out 2022/09
)
```
fetches the image

im-adithya (approved):

bumi:
what do we do with this?

im-adithya:
I think we can merge this as we're using `||` anyways

PR #1522, opened 2022-09-26
fix(settings): avoid race-condition with default value (usd) #1521

### LINKED ISSUE #1521: AccountMenu: shows curremcy from settings but fiat value is calculated from default currency (USD)
### Describe the bug

On every second reload of i.e. the popup the fiat value is incorrect.  
Looks like a race-condition.

### To Reproduce

1. Open popup
1. Check fiat value
1. Reload or open/close popup
1. Fiat value is wrong

### Expected behavior

Fiat value should always be calculated from currency set in settings

### Alby information

- Alby Version: 1.16.0

escapedcat:

Avoid race-condition for useEffect

### Link this PR to an issue

Fixes #1521

### How has this been tested?

Manually

### DIFF
--- a/src/app/context/SettingsContext.tsx
+++ b/src/app/context/SettingsContext.tsx
@@ -57,7 +57,7 @@ export const SettingsProvider = ({
 
   // update rate
   useEffect(() => {
-    api.getCurrencyRate({ currency: settings.currency }).then((response) => {
+    api.getCurrencyRate().then((response) => {
       setCurrencyRate(response.rate);
     });
   }, [settings.currency]);
--- a/src/common/lib/api.ts
+++ b/src/common/lib/api.ts
@@ -12,7 +12,6 @@ import type {
   SettingsStorage,
   MessageInvoices,
   DbPayment,
-  MessageCurrencyRateGet,
   MessageLnurlAuth,
   MessageSettingsSet,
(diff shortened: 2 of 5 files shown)


bumi:
can you explain this? 
this actually looks a bit like an ugly hack, can't this be otherwise? that call should only happen once also, no?

escapedcat:
> can you explain this? 
> that call should only happen once also, no?

The effect runs once with the default settings ("USD") and then again with the loaded settings.

> this actually looks a bit like an ugly hack, can't this be otherwise? 

https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect#useeffect-clean-up-function-with-boolean-flag
I've seen this before in other videos about useEffect.
With this the response will be used in the correct order.

bumi:
why is it even USD first? where is the default coming from?

escapedcat:
> why is it even USD first? where is the default coming from?

Default is being set here:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L27-L28

Defaults:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/state.ts#L48

bumi:
ok, why do we need the defaults there? those always need to come from the background script/state.

isn't that the root cause?

PR #1530, opened 2022-09-27
fix(settings): make sure new fiat-rate is being used on currency switch

escapedcat:
## Issue

When changing currency in settings the fiat-rate in the account-menu is not being updated

- make sure fiat-rate is being renewed when changing currency in settings (re-add dependency to effect)
- remove `fetachAccountInfo` call because it's not needed anymore

### How has this been tested?

Manually

### DIFF
--- a/src/app/context/SettingsContext.tsx
+++ b/src/app/context/SettingsContext.tsx
@@ -3,6 +3,7 @@ import i18n from "i18next";
 import { useState, useEffect, createContext, useContext } from "react";
 import { toast } from "react-toastify";
 import { getTheme } from "~/app/utils";
+import { CURRENCIES } from "~/common/constants";
 import api from "~/common/lib/api";
 import { getFiatValue as getFiatValueFunc } from "~/common/utils/currencyConvert";
 import { DEFAULT_SETTINGS } from "~/extension/background-script/state";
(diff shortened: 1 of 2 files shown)


src/app/context/SettingsContext.tsx
```diff
@@ -57,37 +63,39 @@ export const SettingsProvider = ({
...
+      });
+    }
+  }, [settings?.currency]);
 
   const getFiatValue = async (amount: number | string) =>
     await getFiatValueFunc({
       amount,
       rate: currencyRate,
-      currency: settings.currency,
+      currency: settings?.currency || CURRENCIES["USD"],
```
bumi, line 77: if there is a rate there is also a currency, isn't it? why the default here?

src/app/context/SettingsContext.tsx
```diff
@@ -24,17 +25,22 @@ export const SettingsProvider = ({
...
 
   // call this to trigger a re-render on all occassions
   const updateSetting = async (setting: Setting) => {
-    setSettings((prevState) => ({
-      ...prevState,
-      ...setting,
-    }));
+    if (!setting) return;
+    setSettings((prevState) => {
+      if (prevState)
```
bumi, line 38: if the default is `{}` (same type of the actual value) then this would not be needed?

escapedcat:
Adithya and me decided against removing the defaults. The function expects a settings-type value. Using undefined or `{}` or `{} as Settings` makes it more confusing.

bumi:
if you two know what you're doing then merge it, I don't understand that the implications of the dependencies right now.

PR #1531, opened 2022-09-27
Fix/1492 handle ln address server error response

### LINKED ISSUE #1492: Send to ln-address: lightning address server is online but responses with an error alby should show an error
### Describe the bug

I tried to send sats to a page that have a lightning address but the lightning server was offline.
When trying to send sats nothing happens.

### To Reproduce

1. Try to send sats using Alby on a page that have a lightning address but the lightning address server is offline.

### Expected behavior

I expected Alby show an error that the lightning server is offline or not working. But for now, nothing happens, I click on Send sats button and nothing happens.

### Alby information

- Alby Version: 1.15.1

escapedcat:

- `lnurlLib.getDetails` can return LNURLDetails or LNURLError [as defined here](https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md#pay-to-static-qrnfclink).
- Implementations need to take care of this

### Link this PR to an issue

Fixes #1492

### Screenshots
- image
- image

### How has this been tested?

Created a error test response and used it: `lnurl1dp68gurn8ghj7emfwd6zuemfw3582cn4wdjhycm0de6x2mn59e3k7mf0v4ekxctsv4jxxct59uerxv3sxsexyctrxgekgwp3xp3njcnpvvckzctrvgurwve38pnxgtmjv9mj7dp5vy6njwfs8yersd33vfjnwdfjxsunxer9x5en2vmzvsukyc35xg6rwwr98ymrvtmvde6hymr9wfex7unjv4ehqmmwwdjju6nndahq9ujger`

### DIFF
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -248,7 +248,7 @@ function Home() {
 
                   if (lnData[0].method === "lnurl") {
                     const lnurl = lnData[0].address;
-                    const lnurlDetails = await lnurlLib.getDetails(lnurl); // throws if invalid.
+                    const lnurlDetails = await lnurlLib.getDetails(lnurl);
                     if (!("tag" in lnurlDetails)) {
                       toast.error(lnurlDetails.reason);
                       return;
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -24,6 +24,7 @@ import { classNames } from "~/app/utils/index";
 import api from "~/common/lib/api";
 import lnurlLib from "~/common/lib/lnurl";
 import utils from "~/common/lib/utils";
+import { isLNURLDetailsError } from "~/common/utils/typeHelpers";
 import type { Allowance, Battery, Transaction } from "~/types";
 
 dayjs.extend(relativeTime);
@@ -249,7 +250,7 @@ function Home() {
                   if (lnData[0].method === "lnurl") {
                     const lnurl = lnData[0].address;
                     const lnurlDetails = await lnurlLib.getDetails(lnurl);
-                    if (!("tag" in lnurlDetails)) {
+                    if (isLNURLDetailsError(lnurlDetails)) {
                       toast.error(lnurlDetails.reason);
                       return;
                     }
--- a/src/app/screens/Home/index.tsx
+++ b/src/app/screens/Home/index.tsx
@@ -249,6 +249,10 @@ function Home() {
                   if (lnData[0].method === "lnurl") {
                     const lnurl = lnData[0].address;
                     const lnurlDetails = await lnurlLib.getDetails(lnurl); // throws if invalid.
+                    if (!("tag" in lnurlDetails)) {
+                      toast.error(lnurlDetails.reason);
+                      return;
+                    }
 
                     if (lnurlDetails.tag === "payRequest") {
                       navigate("/lnurlPay", {
--- a/src/app/screens/Send/index.tsx
+++ b/src/app/screens/Send/index.tsx
@@ -41,6 +41,10 @@ function Send() {
 
       if (lnurl) {
         const lnurlDetails = await lnurlLib.getDetails(lnurl);
+        if (!("tag" in lnurlDetails)) {
+          toast.error(lnurlDetails.reason);
+          return;
+        }
 
         if (lnurlDetails.tag === "channelRequest") {
           navigate("/lnurlChannel", {
--- a/src/app/screens/Send/index.tsx
+++ b/src/app/screens/Send/index.tsx
@@ -15,6 +15,7 @@ import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import { toast } from "react-toastify";
 import lnurlLib from "~/common/lib/lnurl";
+import { isLNURLDetailsError } from "~/common/utils/typeHelpers";
 
 function Send() {
   const { t } = useTranslation("translation", { keyPrefix: "send" });
@@ -41,7 +42,7 @@ function Send() {
 
       if (lnurl) {
         const lnurlDetails = await lnurlLib.getDetails(lnurl);
-        if (!("tag" in lnurlDetails)) {
+        if (isLNURLDetailsError(lnurlDetails)) {
           toast.error(lnurlDetails.reason);
           return;
         }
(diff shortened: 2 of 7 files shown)


bumi:
can all the places where `getDetails` is used handle this change?

escapedcat:
> can all the places where `getDetails` is used handle this change?

I believe so. What do you think? Is it possible to test this with a real failing request? Otherwise I could try to mock the response maybe.

bumi:
did you check all the places where `getDetails` is used? 

also when is it an axios error? 

I don't think this solves #1492 
because the server might return a HTTP 200 and the JSON like the one mentioned in the issue: `{"status": "ERROR", "reason": "couldn't find user"}`

So I don't really understand what is different with this change. 
and if you have not tested it, then it needs to be tested.

bumi:
shouldn't we also check if it is status error? 

(and the response could be completely broken, e.g when a service is down. there is no tag and no reason in the responsee)

bumi:
I guess the global catch (of a network error or such) is already handled, isn't it? 

I just say that if there is no tag it does not mean that there is a `reason` in the response. You take something from the first success JSON and assume something from the second error JSON. it likely is.
I just would have checked only if it's the second error case. if status is error and there is a reason then we show that, otherwise we need to show something generic and we don't know what's up.

escapedcat:
> I guess the global catch (of a network error or such) is already handled, isn't it?

Right, this:
https://github.com/getAlby/lightning-browser-extension/blob/fix/1492_hande-ln-address-server-error-response/src/common/lib/lnurl.ts#L76-L78

So for now this would also silently fail for the user. But I'm happy to tackle this after this PR.

> I just say that if there is no tag it does not mean that there is a `reason` in the response. You take something from the first success JSON and assume something from the second error JSON. it likely is. I just would have checked only if it's the second error case. if status is error and there is a reason then we show that, otherwise we need to show something generic and we don't know what's up.

Updated it to make sure we show `reason` or a generic message.

im-adithya (approved):

src/app/screens/Send/index.tsx
```diff
@@ -41,6 +41,14 @@ function Send() {
 
       if (lnurl) {
         const lnurlDetails = await lnurlLib.getDetails(lnurl);
+        if (!("tag" in lnurlDetails)) {
```
bumi, line 44: considering that we check for the tag further down, do we really need this if?

the issue was also only about handling the error case of the spec. 

so I'd do `if(lnurlDetails.satus == 'error' && lnurlDetails.reason) { `

src/app/screens/Send/index.tsx
```diff
@@ -41,6 +41,14 @@ function Send() {
...
+          if (lnurlDetails.reason) {
+            toast.error(lnurlDetails.reason);
+          } else {
+            toast.error("Connection problem or invalid lnurl");
```
bumi, line 48: wouldn't connection problems throw in `getDetails` ? so those would not end up here in this toast

bumi: actually the error message is wrong, isn't it? 

this is a successful response but an invalid lnurl server response. 

so it is not a connection problem and not an invalid lnurl.

bumi (commented):
generally I think any `if` is a code-smell. 
any if always makes method more complex (it's doing multiple things) and thus also more complicated to test. 

so I am not sooo happy with the double if here. do we really need that here? when is this happening?

src/app/screens/Home/index.tsx
```diff
@@ -248,7 +248,11 @@ function Home() {
 
                   if (lnData[0].method === "lnurl") {
                     const lnurl = lnData[0].address;
-                    const lnurlDetails = await lnurlLib.getDetails(lnurl); // throws if invalid.
+                    const lnurlDetails = await lnurlLib.getDetails(lnurl);
+                    if ("status" in lnurlDetails && "reason" in lnurlDetails) {
```
bumi, line 252: I would check for `status == error` because what `status` can also be `ok`.
So to me this check does not make sooo much sense. 
and the `reason` property also sounds indirect.

but I don't know why we talk so much about this :)

src/extension/background-script/actions/lnurl/index.ts
```diff
@@ -16,6 +16,13 @@ async function lnurl(message: MessageWebLnLnurl) {
   let lnurlDetails;
   try {
     lnurlDetails = await lnurlLib.getDetails(message.args.lnurlEncoded);
+    if ("status" in lnurlDetails && "reason" in lnurlDetails) {
+      if (lnurlDetails.reason) {
```
bumi, line 20: honestly this is still very strange code to me. 

you check if the status is in lnurlDetails and if reason is in lnurlDetails and then in another if you check if reason is truthy. 
and for that the error message in the else case is imo also wrong. 
connection errors would throw afik, and the lnurl is also not invalid because we got some error - so the lnurl itself is correct.

escapedcat:
@bumi I switched it to a helper function to make this more clear.

In general we should be able to trust our types.

Currently these types are expected to be `LNURLDetails`:
https://github.com/getAlby/lightning-browser-extension/blob/fix/1492_hande-ln-address-server-error-response/src/types.ts/#L326-L371
None of these have a `status` prop unless it's an error: 
- https://github.com/fiatjaf/lnurl-rfc/blob/luds/02.md
- https://github.com/fiatjaf/lnurl-rfc/blob/luds/03.md
- https://github.com/fiatjaf/lnurl-rfc/blob/luds/06.md
- https://github.com/fiatjaf/lnurl-rfc/blob/luds/04.md
    This can return `{"status": "OK"}` but not in the case of our `getDetails` function because: 
    > tag with value set to login which means no GET should be made yet.

Does this make sense? If we can't trust this we can adjust the types further. In a later PR.

bumi:
look at this: https://getalby.com/.well-known/lnurlp/bumi
I guess we implement the spec wrong then...

(btw. imo a fail of the spec then: they don't respect HTTP codes, and then selectively only use a status code)

src/common/utils/typeHelpers.ts
```diff
@@ -0,0 +1,7 @@
+import type { LNURLDetails, LNURLError } from "~/types";
+
+export const isLNURLDetailsError = (
+  res: LNURLError | LNURLDetails
+): res is LNURLError => {
+  return "status" in res && res.status === "ERROR";
```
bumi, line 6:
```suggestion
  return "status" in res && res.status.toUpperCase() === "ERROR";
```

escapedcat:
> look at this: https://getalby.com/.well-known/lnurlp/bumi I guess we implement the spec wrong then...

Guess we did :P   
Will make a note and update our types and add a comment. Thanks!

> (btw. imo a fail of the spec then: they don't respect HTTP codes, and then selectively only use a status code)

Yes, agreed it's weird.

Issue #1534, opened 2022-09-27
Add a tag called LIGHTNING LOGIN in front of websites I logged in using Lightning


bitcoinuser:
### Describe the solution you'd like

For now we have the tag ACTIVE when the user specified a budget to a website. I think should have too, a tag called LIGHTNING LOGIN, when the user logged in a website using lightning login.
I think is important, because make easy to users track in which sites they used lightning to login.

bumi:
nice, good idea to show this as a badge. 
related: https://github.com/getAlby/lightning-browser-extension/issues/148

PR #1539, opened 2022-09-28
feat(allowance): add option to change lnurlauth

### LINKED ISSUE #148: Display auto login option for publishers

im-adithya:

Adds option to set lnurlAuth for allowances from Publishers screen.

### Link this PR to an issue

Closes #148

https://user-images.githubusercontent.com/64399555/192765505-d4d6c4d0-ca98-4a7e-97c8-10ff59f317d5.mp4

### How has this been tested?

Manually

### Checklist

### DIFF
--- a/src/app/components/AllowanceMenu/index.tsx
+++ b/src/app/components/AllowanceMenu/index.tsx
@@ -1,5 +1,6 @@
 import { GearIcon } from "@bitcoin-design/bitcoin-icons-react/filled";
 import { CrossIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
+import Checkbox from "@components/form/Checkbox";
 import { useState, useEffect } from "react";
 import type { FormEvent } from "react";
 import { useTranslation } from "react-i18next";
@@ -14,7 +15,7 @@ import Menu from "../Menu";
 import DualCurrencyField from "../form/DualCurrencyField/index";
 
 export type Props = {
-  allowance: Pick<Allowance, "id" | "totalBudget">;
+  allowance: Pick<Allowance, "id" | "totalBudget" | "lnurlAuth">;
   onEdit?: () => void;
   onDelete?: () => void;
 };
@@ -29,6 +30,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
 
   const [modalIsOpen, setIsOpen] = useState(false);
   const [budget, setBudget] = useState("0");
+  const [login, setLogin] = useState(false);
   const [fiatAmount, setFiatAmount] = useState("");
   const { t } = useTranslation("components", { keyPrefix: "allowance_menu" });
   const { t: tCommon } = useTranslation("common");
@@ -44,6 +46,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
 
   function openModal() {
     setBudget(allowance.totalBudget.toString());
+    setLogin(allowance.lnurlAuth);
     /**
      * @HACK
      * @headless-ui/menu restores focus after closing a menu, to the button that opened it.
@@ -63,6 +66,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
     await utils.call("updateAllowance", {
       id: allowance.id,
       totalBudget: parseInt(budget),
+      lnurlAuth: login,
     });
 
     onEdit && onEdit();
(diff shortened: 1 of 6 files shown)


escapedcat:
> When enabled, will automatically login on a site? If not enabled I will always be asked to confirm login? 

Yes

> The way it works now, I think I'm no asked to confirm login if I logout and login again.

Correct, currently the default is "enabled" after you logged in once.
This PR will change this so a user can chose to logged in automatically or not.

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -138,6 +142,20 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
...
               />
             </div>
+            <div className="flex items-center mt-6">
+              <Checkbox
+                id="enable_login"
+                name="enable_login"
+                checked={login}
+                onChange={(e) => setLogin(!login)}
+              />
+              <label
```
escapedcat, line 152: This is the first time we use a checkbox. For everything else we use `Toggle`. Should we use toggle here as well?

im-adithya: We also use it in Remember and Set Budget right?

im-adithya: But yeah, I think it's not being highlighted much right now, may be we can use toggle and also add hint below it (like we do in settings) to let people know what it is about

bitcoinuser:
> Correct, currently the default is "enabled" after you logged in once.
This PR will change this so a user can chose to logged in automatically or not.

I really don't see a reason to have this feature. I think more confuse than helps. What do you think @bumi?

im-adithya:
I think this is more about giving control to the user to toggle the lnurlAuth option and would make more sense once the follow-up PR to this comes up which solves #1534 where we show the websites where we used lightning login and now if the user wants to disable it in any website then he can choose to edit the allowance

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -29,6 +30,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
 
   const [modalIsOpen, setIsOpen] = useState(false);
   const [budget, setBudget] = useState("0");
+  const [login, setLogin] = useState(false);
```
bumi, line 33: maybe we also call this lnurlAuth? like the attribute that we set?

src/app/components/AllowanceMenu/index.tsx
bumi: +1

escapedcat (approved):

src/i18n/locales/en/translation.json
```diff
@@ -570,8 +570,12 @@
       "new_budget": {
         "label": "New budget"
       },
-      "set_budget": {
-        "title": "Set a new budget",
+      "enable_login": {
+        "title": "Enable website login",
+        "subtitle": "Authenticate using <0>lnurlAuth</0>"
```
bumi, line 575: the text needs some update. 

```suggestion
        "subtitle": "Automatically login without confirmation when the website requests."
```

bumi (approved):

src/i18n/locales/en/translation.json
im-adithya: Done!

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -138,14 +143,35 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
                 onChange={(e) => setBudget(e.target.value)}
               />
             </div>
+            <Setting
+              title={t("enable_login.title")}
+              subtitle={
+                <Trans
```
escapedcat, line 149: Now that the string was changed do we still need the `Trans` here?

im-adithya: Oops! Changing it

im-adithya: Done!

MoritzKa:
a bit late to the party. But it works for me on LNMarkets and bolt.fun.

Issue #1510, opened 2022-09-23
Send screen from popup should not show "Remember and set a budget" when i.e. pasting a lightning invoice (or other screens)


bitcoinuser:
### Describe the bug

If I create a ln invoice on a wallet like Phoenix and copy it and paste on Alby, Alby show the option Remember and set a budget. But doesn't make sense, because it's not a service or a page, so it's not saved.

### Screenshots

![Captura de tela de 2022-09-23 10-59-01](https://user-images.githubusercontent.com/1178856/191978498-cc2a3621-98aa-4c10-8a4b-13d5aa3ac6f3.png)

### To Reproduce

1. Generate a ln invoice on a ln wallet like Phoenix and copy the invoice.
2. Click on Alby send button, then paste the invoice and click continue button

### Expected behavior

Remember and set a budget option not be showed.

### Alby information

- Alby Version: 1.15.1

vaibhavgarg237:
I'm working on this

escapedcat:
> I'm working on this

Thanks! Let me know if you get stuck or need more info

escapedcat:
Heyhey @vaibhavgarg237 , did you already start with this? Do you need any support?

vaibhavgarg237:
> Heyhey @vaibhavgarg237 , did you already start with this? Do you need any support?

It's almost complete, I got occupied in some office work...
Although I was facing some issues related to running tests locally. So I was thinking to first run locally and then do the PR. Will create the draft PR today...

escapedcat:
> Will create the draft PR today

Perfect, yes! No worries if it's not "ready ready" yet. Just seeing some progress is already helpful. 
Thanks for your quick reply!

PR #1541, opened 2022-09-28
fix: remove Remember & set budget from send #1510

vaibhavgarg237:

_Remove "Remember and set a budget" option from send screen when opting for lightning invoice_

### Link this PR to an issue

_Fixes #1510_

- _Previous behaviour_ 
image

- _After removal of "Remember and set a budget"_
image
image

### How has this been tested?

_Manual Testing_
- I used ln invoice from Phoenix and tapped Send on popup and website, which shows "Remember and set a budget" but with the changes it has been removed
- Lightning invoice used:  lnbc1m1p3ngewjpp578s6uh2gvmsuldc7khppc57spggzjuy9aj0cg4633put5x20p6cqdqqxqyjw5q9q7sqqqqqqqqqqqqqqqqqqqqqqqqq9qsqsp5e5a6h39y85fnvlenly6dmfx97y32tru4hkn84x0wcgjf8egyfk4qrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glclll0mxh6jgh2xuqqqqlgqqqqqeqqjqfs7gjqntggjnpgxp9rjh78gx2cczjuha65wdz9w2462chm8a7d78s2gwpdndsra4lf7940dh03dw67xcr8dd605lm68j6pusjj7n9tspm0mtde

### Checklist

### DIFF
--- a/src/app/screens/ConfirmPayment/index.test.tsx
+++ b/src/app/screens/ConfirmPayment/index.test.tsx
@@ -30,14 +30,11 @@ const mockOrigin: OriginData = {
 const paymentRequest =
   "lnbc250n1p3qzycupp58uc2wa29470f98wrxmy4xwuqt8cywjygf5t2cp0s376y7nwdyq3sdqhf35kw6r5de5kueeqg3jk6mccqzpgxqyz5vqsp5wfdmwtv5rmru00ajsnn3f8lzpxa4snug2tmqvc8zj8semr4kjjts9qyyssq83h74pte8nrkqs8sr2hscv5zcdmhwunwnd6xr3mskeayh96pu7ksswa6p7trknlpp6t3js4k6uytxutv5ecgcwaxz7fj4zfy5khjcjcpf66muy";
 
+let parameters = {};
+
 jest.mock("~/app/hooks/useNavigationState", () => {
   return {
-    useNavigationState: jest.fn(() => ({
-      origin: mockOrigin,
-      args: {
-        paymentRequest,
-      },
-    })),
+    useNavigationState: jest.fn(() => parameters),
   };
 });
 
@@ -46,7 +43,14 @@ describe("ConfirmPayment", () => {
     jest.clearAllMocks();
   });
 
-  test("renders with fiat", async () => {
+  test("prompt: renders with fiat", async () => {
+    parameters = {
+      origin: mockOrigin,
+      args: {
+        paymentRequest,
+      },
+    };
+
     jest.spyOn(SettingsContext, "useSettings").mockReturnValue({
       settings: { ...mockSettings },
       isLoading: false,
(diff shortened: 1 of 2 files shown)


bitcoinuser:
Does it still maintain the Remember & set budget for the other cases like, right?

vaibhavgarg237:
> Does it still maintain the Remember & set budget for the other cases, right?

I noticed it removed the set budget for other cases _like subscribing on some project on geyser.fund_. I am rechecking the codebase and cases where it should and should not show the set budget option.

escapedcat:
The issue should have more specific, sorry.  
The important part here is, that we do not just want to remove this.

For example in the prompt case we will always need it:
1. Go to https://nakaphoto.vercel.app/accounts/albydev@getalby.com
2. Click on a picture
3. Prompt shows up, click "connect"
4. Here we need this functionality:   
    image

Try if this works:
- This block needs to be hidden if the extension-popup is being used but shown if the prompt is being used:
    https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/ConfirmPayment/index.tsx/#L135
- Try moving the `!navState.origin` out and put this with an `if` around the block. Because only using the prompt we have an `origin`    
    https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/ConfirmPayment/index.tsx/#L104
    
Does this help? Let me know if you have any questions.

vaibhavgarg237:
> The issue should have more specific, sorry. The important part here is, that we do not just want to remove this.
> 
> For example in the prompt case we will always need it:
> 
> 1. Go to https://nakaphoto.vercel.app/accounts/albydev@getalby.com
> 2. Click on a picture
> 3. Prompt shows up, click "connect"
> 4. Here we need this functionality:
>    image
> 
> Try if this works:
> 
> * This block needs to be hidden if the extension-popup is being used but shown if the prompt is being used:
>   https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/ConfirmPayment/index.tsx/#L135
> * Try moving the `!navState.origin` out and put this with an `if` around the block. Because only using the prompt we have an `origin`
>   https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/ConfirmPayment/index.tsx/#L104
> 
> Does this help? Let me know if you have any questions.

Yeah, I got that one. I was still figuring out for it's tests ... That's why I did some rough changes in a separate branch to hide when someone is using send, and show when in case as described above 
https://github.com/vaibhavgarg237/lightning-browser-extension/commit/e685ce8af043d37840b811bb15b9c60e07410380

escapedcat:
@vaibhavgarg237 tests are green here 🚀

vaibhavgarg237:
> @vaibhavgarg237 tests are green here 🚀 Might be good to add a test for your change. I can help with that.

I think tests are passing because it's only checking for [demo](https://getalby.com/demo) page as it is creating a OriginData in starting itself. I think we need to add another test that only runs when in condition when the origin is not present and the current one should only run when in presence of origin.

vaibhavgarg237:
I was facing some issues related to running original tests locally. I have dm'ed you on slack regarding the same, kindly lemme know whenever you are available.
Thanks

escapedcat (approved):

escapedcat:
ack

reneaaron (approved):
utACK

PR #1545, opened 2022-09-29
refactor(lnurl): getDetails - avoid typecasting

escapedcat:
Mainly trying to avoid `let lnurlDetails = {} as LNURLDetails;` and be more explicit about types. So instead of forcing the type we let TS figure it out by defining possible types in other places.

### DIFF
--- a/src/common/lib/lnurl.ts
+++ b/src/common/lib/lnurl.ts
@@ -2,7 +2,13 @@ import axios from "axios";
 import lightningPayReq from "bolt11";
 import Hex from "crypto-js/enc-hex";
 import sha256 from "crypto-js/sha256";
-import type { LNURLDetails, LNURLError, LNURLPaymentInfo } from "~/types";
+import { isLNURLDetailsError } from "~/common/utils/typeHelpers";
+import {
+  LNURLDetails,
+  LNURLError,
+  LNURLAuthServiceResponse,
+  LNURLPaymentInfo,
+} from "~/types";
 
 import { bech32Decode } from "../utils/helpers";
 
@@ -45,6 +51,7 @@ const lnurl = {
   isLightningAddress(address: string) {
     return Boolean(fromInternetIdentifier(address));
   },
+
   findLnurl(text: string) {
     const stringToText = text.trim();
     let match;
@@ -61,27 +68,51 @@ const lnurl = {
 
     return null;
   },
+
   async getDetails(lnurlString: string): Promise<LNURLError | LNURLDetails> {
     const url = normalizeLnurl(lnurlString);
-    let lnurlDetails = {} as LNURLDetails;
-    lnurlDetails.tag = url.searchParams.get("tag") as LNURLDetails["tag"];
-    if (lnurlDetails.tag === "login") {
-      lnurlDetails.k1 = url.searchParams.get("k1") || "";
-      lnurlDetails.action = url.searchParams.get("action") || undefined;
+    const searchParamstag = url.searchParams.get("tag") as LNURLDetails["tag"];
+
+    if (searchParamstag === "login") {
+      const searchParamsK1 = url.searchParams.get(
+        "k1"
+      ) as LNURLAuthServiceResponse["tag"];
+      const searchParamsAction = url.searchParams.get(
+        "action"
+      ) as LNURLAuthServiceResponse["tag"];
+
+      const lnurlAuthDetails: LNURLAuthServiceResponse = {
+        action: searchParamsAction,
+        domain: url.hostname,
+        k1: searchParamsK1,
+        tag: searchParamstag,
+        url: url.toString(),
+      };
+
+      return lnurlAuthDetails;
     } else {
       try {
-        const res = await axios.get(url.toString());
-        lnurlDetails = res.data;
+        const { data }: { data: LNURLDetails | LNURLError } = await axios.get(
+          url.toString()
+        );
+        const lnurlDetails = data;
+
+        if (!isLNURLDetailsError(lnurlDetails)) {
+          lnurlDetails.domain = url.hostname;
+          lnurlDetails.url = url.toString();
+        }
+
+        return lnurlDetails;
       } catch (e) {
         throw new Error(
-          "Connection problem or invalid lnurl / lightning address."
+          `Connection problem or invalid lnurl / lightning address: ${
+            e instanceof Error ? e.message : ""
+          }`
         );
       }
     }
-    lnurlDetails.domain = url.hostname;
-    lnurlDetails.url = url.toString();
-    return lnurlDetails;
   },
+
   verifyInvoice({
     paymentInfo,
     payerdata,
--- a/src/common/lib/lnurl.ts
+++ b/src/common/lib/lnurl.ts
@@ -71,25 +71,25 @@ const lnurl = {
 
   async getDetails(lnurlString: string): Promise<LNURLError | LNURLDetails> {
     const url = normalizeLnurl(lnurlString);
-    const searchParamstag = url.searchParams.get("tag") as LNURLDetails["tag"];
+    const searchParamsTag = url.searchParams.get("tag");
 
-    if (searchParamstag === "login") {
-      const searchParamsK1 = url.searchParams.get(
-        "k1"
-      ) as LNURLAuthServiceResponse["tag"];
-      const searchParamsAction = url.searchParams.get(
-        "action"
-      ) as LNURLAuthServiceResponse["tag"];
+    if (searchParamsTag && searchParamsTag === "login") {
+      const searchParamsK1 = url.searchParams.get("k1");
+      const searchParamsAction = url.searchParams.get("action");
 
-      const lnurlAuthDetails: LNURLAuthServiceResponse = {
-        action: searchParamsAction,
-        domain: url.hostname,
-        k1: searchParamsK1,
-        tag: searchParamstag,
-        url: url.toString(),
-      };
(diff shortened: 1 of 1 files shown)


src/common/lib/lnurl.ts
```diff
@@ -61,27 +67,49 @@ const lnurl = {
...
     } else {
       try {
-        const res = await axios.get(url.toString());
-        lnurlDetails = res.data;
+        const { data }: { data: LNURLDetails | LNURLError } = await axios.get(
+          url.toString()
+        );
+        const lnurlDetails = data;
+
+        if ("tag" in lnurlDetails) {
```
bumi, line 99: maybe we should add some comments here. 
why do we do that?

src/common/lib/lnurl.ts
```diff
@@ -61,27 +67,49 @@ const lnurl = {
...
+          lnurlDetails.domain = url.hostname;
+          lnurlDetails.url = url.toString();
+        }
+
+        return lnurlDetails;
       } catch (e) {
         throw new Error(
           "Connection problem or invalid lnurl / lightning address."
```
bumi, line 107: should we add the e.message to that error?

src/app/screens/Home/index.tsx
```diff
@@ -248,7 +248,7 @@ function Home() {
 
                   if (lnData[0].method === "lnurl") {
                     const lnurl = lnData[0].address;
-                    const lnurlDetails = await lnurlLib.getDetails(lnurl); // throws if invalid.
+                    const lnurlDetails = await lnurlLib.getDetails(lnurl);
```
bumi, line 251: it stills throws if it's invalid. 
does that comment hurt?

src/common/lib/lnurl.ts
escapedcat: Using the `isError` function here now is well which makes this case clearer I hope

src/common/lib/lnurl.ts
escapedcat: Added possible message

src/app/screens/Home/index.tsx
escapedcat: Before we didn't even handle the error case. Now that we do that I think it's clearer what to expect. Yes, it can still throw but we don't need to wrote this as an extra comment I guess. Can re-add it though.

src/common/lib/lnurl.ts
```diff
@@ -61,27 +68,51 @@ const lnurl = {
...
   },
+
   async getDetails(lnurlString: string): Promise<LNURLError | LNURLDetails> {
     const url = normalizeLnurl(lnurlString);
-    let lnurlDetails = {} as LNURLDetails;
-    lnurlDetails.tag = url.searchParams.get("tag") as LNURLDetails["tag"];
-    if (lnurlDetails.tag === "login") {
-      lnurlDetails.k1 = url.searchParams.get("k1") || "";
-      lnurlDetails.action = url.searchParams.get("action") || undefined;
+    const searchParamstag = url.searchParams.get("tag") as LNURLDetails["tag"];
```
lisabaut, line 74: I would remove the assertion here.
 `searchParams.get` returns either a string or `null`.
 
 Let's rather test against that.

src/common/lib/lnurl.ts
```diff
@@ -61,27 +68,51 @@ const lnurl = {
...
+
+      const lnurlAuthDetails: LNURLAuthServiceResponse = {
+        action: searchParamsAction,
+        domain: url.hostname,
+        k1: searchParamsK1,
+        tag: searchParamstag,
+        url: url.toString(),
+      };
+
+      return lnurlAuthDetails;
```
lisabaut, line 92: Same here: Rather remove the assertion as the searchparam ca be `null`
Which is actually the case when copying the LNURL Auth from https://lnurl.fiatjaf.com/
=> `action` is `null`

But `k1` has to be defined, right? Otherwise the URL is broken, correct?
Let´s test this and throw otherwise.

here is my suggestion:

```TS
    if (searchParamstag === "login") {
      const searchParamsK1 = url.searchParams.get("k1");
      const searchParamsAction = url.searchParams.get("action");

      if (searchParamsK1) {
        const lnurlAuthDetails: LNURLAuthServiceResponse = {
          action: searchParamsAction,
          domain: url.hostname,
          k1: searchParamsK1,
          tag: searchParamstag,
          url: url.toString(),
        };

        return lnurlAuthDetails;
      }

      throw new Error(
        "LNURL AUTH Login is missing search params: searchParamsK1"
      );
     } else {
     ...
```

lisabaut: This would then require to change these types too:

https://github.com/getAlby/lightning-browser-extension/blob/master/src/types.ts#L347
=> `action: string | null:`

https://github.com/getAlby/lightning-browser-extension/blob/master/src/types.ts#L293-L296
=> add `action: string | null`

lisabaut (commented):
I have a suggestion to make. See my comments below.

Also I could not find the place where the `action` string from LNURL Auth is used.
Can you point me to that again? 🙇

src/common/lib/lnurl.ts
escapedcat: Yes, makes sense. `action` is defined as optional in the spec:  
> optional action enum which can be one of four strings: register | login | link | auth.

https://github.com/fiatjaf/lnurl-rfc/blob/luds/04.md#wallet-to-service-interaction-flow

escapedcat: Updated it with your suggestion. Do we need to add ` | null` to the types if they are already optional (`action?: ...`)?

escapedcat:
> Also I could not find the place where the `action` string from LNURL Auth is used. Can you point me to that again? 🙇

Uhm, I'm also not sure tbh. @reneaaron do you know of a real-world usecase?

src/common/lib/lnurl.ts
lisabaut: Either decide for:

- action is _not_ optional and can be a string or null
- or action is optional and can be a string (or undefined)

The first case happens when you just pass on the data from the LNURL url.

The latter is when you decide what to do (your current code [here](https://github.com/getAlby/lightning-browser-extension/pull/1545/files#diff-146ca15116557d76017757b844192427543e8137310612ff1d0b5e7221270187R82)).

As the lnurlDetails for LNURLAuth are always constructed from SearchParams of an URL (correct?), 
I would go for the latter case
`action: string | null;` and paste the link to the fjatfjat docs next to the types.

reneaaron:
Currently geyser.fund.

lisabaut (approved):

src/common/lib/lnurl.ts
escapedcat: Got it! Corrected it.

src/common/lib/lnurl.ts
```diff
@@ -61,27 +68,51 @@ const lnurl = {
...
   async getDetails(lnurlString: string): Promise<LNURLError | LNURLDetails> {
     const url = normalizeLnurl(lnurlString);
-    let lnurlDetails = {} as LNURLDetails;
-    lnurlDetails.tag = url.searchParams.get("tag") as LNURLDetails["tag"];
-    if (lnurlDetails.tag === "login") {
-      lnurlDetails.k1 = url.searchParams.get("k1") || "";
-      lnurlDetails.action = url.searchParams.get("action") || undefined;
+    const searchParamsTag = url.searchParams.get("tag");
+
+    if (searchParamsTag && searchParamsTag === "login") {
```
bumi, line 76: we have many such checks. there should be some way to validate an object against a schema without such nested `if`

src/common/lib/lnurl.ts
```diff
@@ -61,27 +68,46 @@ const lnurl = {
...
     } else {
       try {
-        const res = await axios.get(url.toString());
-        lnurlDetails = res.data;
+        const { data }: { data: LNURLDetails | LNURLError } = await axios.get(
+          url.toString()
+        );
+        const lnurlDetails = data;
+
+        if (!isLNURLDetailsError(lnurlDetails)) {
```
bumi, line 95: hmmm, how do we deal with errors when it is an LNURLDetailsError?

escapedcat: My understanding is that this is just a typeguard to avoid TS complaining. If the axios request fails it should fall into the catch (like the code before).  

@lisabaut wdyt?

bumi: this is not for when the HTTP request fails. this checking the lnurl response: 
https://github.com/getAlby/lightning-browser-extension/blob/6c3cd1012e63f6ad3c0a206f56e90f04c5d204f0/src/common/utils/typeHelpers.ts#L6

the HTTP request would already fail in line 90 and end up in the catch block.

escapedcat: Got it! Adjusted it

src/common/lib/lnurl.ts
```diff
@@ -61,27 +68,48 @@ const lnurl = {
...
       try {
-        const res = await axios.get(url.toString());
-        lnurlDetails = res.data;
+        const { data }: { data: LNURLDetails | LNURLError } = await axios.get(
+          url.toString()
+        );
+        const lnurlDetails = data;
+
+        if (isLNURLDetailsError(lnurlDetails)) {
+          throw new Error(`LNURL Details Error: ${lnurlDetails.reason}`);
```
bumi, line 96:
```suggestion
          throw new Error(`LNURL Error: ${lnurlDetails.reason}`);
```

PR #1565, opened 2022-10-04
test(e2e): remove outdated bluewallet test

escapedcat:


bumi:
why remove it and not just use a different lndhub URI?

escapedcat:
> why remove it and not just use a different lndhub URI?

Ah, I didn't know about this. Was just wondering why it started to fail. Thanks, updated. All good.

Issue #1572, opened 2022-10-06
Forms: disable fom/fieldset/inputs after submit


bitcoinuser:
### Describe the bug

I'm able to type text on fields after I click on confirm button

### Screenshots of the changes [optional]

[Gravação de tela de 05-10-2022 22:52:10.webm](https://user-images.githubusercontent.com/1178856/194196838-80351ada-3f92-464d-b7cb-196a99fb4b3c.webm)

### To Reproduce

See the video

### Expected behavior

Not be able to type any information on the fields of send window when payment is processing.

### Alby information

- Alby Version: 1.17

bitcoinuser:
I think even the cancel button should be disabled, because we can't cancel a payment that is already being processed. Right?

vaibhavgarg237:
I'm working on this

vaibhavgarg237:
I have disabled the Amount, Comment, Name, Email, sats Buttons, and Cancel button after tapping on confirm

PR #1576, opened 2022-10-06
fix(lnurlpay): disable input fields buttons #1572

vaibhavgarg237:

_On tapping Confirm, while loading **Amount, Comment, Name, Email, Buttons, Cancel button** should be disabled_

Fixes _#1572_

### Type of change

On tapping Confirm all these options get disabled as they should be, as details have already been specified.
image
image

### How has this been tested?

_Open [this ](https://www.youtube.com/c/KoreacomK) on your browser, tap on getAlby extension, click Send Satoshis, now on the next window after adding Amount, Comment, Name, and Email, and click on Confirm all these fields should be disabled. It's working as it should be as specified in issue._

### Checklist

### DIFF
--- a/src/app/components/ConfirmOrCancel/index.tsx
+++ b/src/app/components/ConfirmOrCancel/index.tsx
@@ -9,7 +9,7 @@ export type Props = {
   disabled?: boolean;
   loading?: boolean;
   label?: string;
-  onConfirm: MouseEventHandler;
+  onConfirm?: MouseEventHandler;
   onCancel: MouseEventHandler;
 };
 
@@ -29,8 +29,10 @@ export default function ConfirmOrCancel({
           onClick={onCancel}
           label={tCommon("actions.cancel")}
           halfWidth
+          disabled={loading}
         />
         <Button
+          type="submit"
           onClick={onConfirm}
           label={label}
           primary
--- a/src/app/screens/LNURLPay/index.tsx
+++ b/src/app/screens/LNURLPay/index.tsx
@@ -326,6 +326,11 @@ function LNURLPay() {
     );
   }
 
+  function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+    confirm();
+  }
+
   return (
     <>
       <div className="flex flex-col grow overflow-hidden">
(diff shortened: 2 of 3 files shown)


src/app/screens/LNURLPay/index.tsx
```diff
@@ -101,6 +102,7 @@ function LNURLPay() {
 
     try {
       setLoadingConfirm(true);
+      setDisabled(true);
```
escapedcat, line 105: Instead of adding a new `disabled`-state-prop could we i.e. do `disabled={loadingConfirm}`?

escapedcat (commented):
Instead of adding multiple disabled props to all kinds of components could we wrap the "formy" parts in a `form` like we do here?:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Send/index.tsx

This could help improve these things I hope:
- a user could send the form via "enter" (if we also move the confirm-function to the form submit)
- we could add a `fieldset` around the "input-components" and could try to only disable that, which should automatically disable all fields within: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attr-disabled

You wanna give that try?

src/app/screens/LNURLPay/index.tsx
vaibhavgarg237: Yeah, that will work.

vaibhavgarg237:
> Instead of adding multiple disabled props to all kinds of components could we wrap the "formy" parts in a `form` like we do here?: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Send/index.tsx
> 
> This could help improve these things I hope:
> 
> * a user could send the form via "enter" (if we also move the confirm-function to the form submit)
> * we could add a `fieldset` around the "input-components" and could try to only disable that, which should automatically disable all fields within: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attr-disabled

I updated the changes.
After wrapping all inputs & others with form and fieldset, the user is able to send the form also using Enter. I realised as a user, pressing enter is much user friendly then using mouse click.

src/app/screens/LNURLPay/index.tsx
```diff
@@ -341,92 +341,98 @@ function LNURLPay() {
...
-                      {formattedMetadata(details.metadata).map(
-                        ([dt, dd], i) => (
-                          <Fragment key={`element-${i}`}>
-                            <Dt>{dt}</Dt>
-                            <Dd>{dd}</Dd>
-                          </Fragment>
-                        )
-                      )}
-                      {details.minSendable === details.maxSendable && (
+                <form>
```
escapedcat, line 344: Can the confirm-action be moved to the form-submit like we do it here or in most form-situations?:  https://github.com/vaibhavgarg237/lightning-browser-extension/blob/disableButtons/src/app/screens/Unlock/index.tsx#L63
Would be nice to align this.

vaibhavgarg237: Actually  because of [this](https://github.com/getAlby/lightning-browser-extension/blob/a7c2332aeb8bb887e1c8b54af5611580d0c3cc52/src/app/components/ConfirmOrCancel/index.tsx#L35), it works same as expected, like on pressing enter it triggers Confirm-action( user can send the form using enter ).

`
        <Button
          type="submit"
          onClick={onConfirm}
`

But for aligning it like rest of code base we can move it to handleSubmit like implementation. Will do that...

escapedcat: Yeah, you're right, but aligning it makes it more clear and put's the action where it should be I believe.

vaibhavgarg237: Okay, will do that.

vaibhavgarg237: @escapedcat I have updated the changes.

escapedcat (approved):

src/app/components/ConfirmOrCancel/index.tsx
```diff
@@ -9,7 +9,7 @@ export type Props = {
   disabled?: boolean;
   loading?: boolean;
   label?: string;
-  onConfirm: MouseEventHandler;
```
lisabaut, line 12: I would still pass the confirm function to the component.
This component is used in so many places and in my eyes `onConfirm` should never be forgotten.

What do you think?
Why make it optional?

src/app/components/ConfirmOrCancel/index.tsx
```diff
@@ -29,8 +29,10 @@ export default function ConfirmOrCancel({
           onClick={onCancel}
           label={tCommon("actions.cancel")}
           halfWidth
+          disabled={loading}
         />
         <Button
+          type="submit"
```
lisabaut, line 35: Just a sidenote:
Type "submit" makes only sense when wrapped in a form element, right?
In this PR it is the first time that it is wrapped in a form.
In all other places where this component is used it is not wrapped in a form.
I think it is ok when it remains "submit" for all other occurrences too, but just want to point to that.

lisabaut (approved):
Thanks.

Just some minor points below.

src/app/components/ConfirmOrCancel/index.tsx
vaibhavgarg237: Actually passing `onConfirm` prop to the component & using `handleSubmit()` in form at the same time leads to calling `Confirm` fxn twice which I feel should not be called twice. I logged both events when Confirm button is clicked / Enter is pressed in that form. Below mentioned are the logs for the same.

image

Here,
first log is printed through  
image

https://github.com/getAlby/lightning-browser-extension/blob/f657162c68548f7832c16fd166083ff5e402b907/src/app/components/ConfirmOrCancel/index.tsx#L34

second log is through 

image

image

https://github.com/vaibhavgarg237/lightning-browser-extension/blob/7bfd5ae2dbd4a558d6a38bd688b7d461dab559ac/src/app/screens/LNURLPay/index.tsx#L349




For this, I think we need to make `onConfirm` prop optional and not pass it as prop so that it's not called twice as done in the current commit. Another workaround would be to remove `onSubmit` from `<form onSubmit={handleSubmit}>` through which we can pass the prop without making it optional but @escapedcat  suggested that it should be like [other form situations](https://github.com/getAlby/lightning-browser-extension/pull/1576#discussion_r990764381)

src/app/components/ConfirmOrCancel/index.tsx
escapedcat: Yeah, good point. I thought about if it's an issue in other occurrences but I think we're save.  
image
Looking at the current usage we could also create a follow-up issue to switch all screens to use forms/submit as well. wdyt?

vaibhavgarg237: Yeah that would be great. Actually I feel most users are tempted to press enter after filling form rather than clicking confirm. I feel it will improve user experience and also code readability.

escapedcat: @lisabaut in this case I would vote for `optional` `onConfirm`. This will support current usages and provides the flexibility in the future to also use it in a non-form context. wdyt?

lisabaut: Fine for me

src/app/components/ConfirmOrCancel/index.tsx
lisabaut: Yes, this is happening because two elements are attached to the same handler.
You can avoid this with `e.stopImmediatePropagation()` 
=> https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation

But as discussed below, keep the `onConfirm` method optional, declare the button as type "submit"
and refactor all other occurrences of `ConfirmOrCancel` to ensure that all are used in a form element.
Otherwise this is not consistent.

src/app/components/ConfirmOrCancel/index.tsx
escapedcat: We reached an agreement! 🎉

vaibhavgarg237: 🎉

src/app/components/ConfirmOrCancel/index.tsx
vaibhavgarg237: Okay, I will update the commit with these changes.

lisabaut: Just one thing: As @escapedcat said below, the refactoring of the other occurrences where ConfirmOrCancel is used can be done in another extra PR. Sorry for the confusion.

vaibhavgarg237: Okay. Thanks for the clarification. I will create another issue & PR for this.

PR #1577, opened 2022-10-06
feat: commando connector

### LINKED ISSUE #1390: Commando (CLN) connector
Alby doest not yet support Core Lightning. Since the [latest release](https://github.com/ElementsProject/lightning/releases/tag/v0.12.0) of Core LN, the plugin `commando` is shipped by default with Core Lightning. This plugin  is used by the [LNSocket](https://github.com/jb55/lnsocket) package, which allows one to connect from the browser to a Core LN node over an untrusted [websocket bridge server](https://github.com/jb55/ln-ws-proxy). The connection is trustless and e2e-encrypted between the browser and the node. Tor nodes can also easily be supported by this bridge server mechanism. Alby would host the bridge server (we can also use proxy.lnlink.org for development).

Authentication is done using so-called `runes`, which are like macaroons and can have fine-grained permissions.

We could even expand the webLN methods to support all commando commands, allowing one to execute all possible c-lightning commands from webLN.

kiwiidb:
Fixes #1390 

Commando is implemented through the [lnmessage](https://github.com/aaronbarnardsound/lnmessage) library, thanks to @aaronbarnardsound. This library uses a websocket proxy (lnproxy.getalby.com) to communicate with a remote CLN node over the p2p port.

Features out of scope for this PR (but coming later):
- Tor support
- WebLN support for arbitrary CLN (+plugin) RPC calls
- Finer grained permissions for WebLN calls
- Stream network gossip to website over WebLN
- I'm probably forgetting more awesome stuff

To test:
- [download and install the development build of this PR](https://github.com/getAlby/lightning-browser-extension/pull/1577#issuecomment-1270646987)
- Add a new connector, pick Core Lightning.
- Go to [here](https://github.com/getAlby/lightning-browser-extension/wiki/Test-setup#c-lightning-1 ) for the connection details of our testnet CLN node, or connect your own (mainnet or testnet) node.
- Fill in the connection details
- Try to spend and receive some [sats on testnet](https://github.com/getAlby/lightning-browser-extension/wiki/Test-setup#where-to-spend-testnet-coins).

Instruction video to test on mainnet: https://www.loom.com/share/359f2cf7e4764950983859301b6e9ecc


src/extension/background-script/connectors/commando.ts
```diff
@@ -0,0 +1,151 @@
...
+      port: this.config.port || 9735,
+      privateKey:
+        this.config.privateKey ||
+        "d6a2eba36168cc31e97396a781a4dd46dd3648c001d3f4fde221d256e41715ea",
+    });
+  }
+
+  async init() {
+    // initiate the connection to the remote node
+    await this.ln.connect();
```
bumi, line 70: as this is a promise I'd remove the async and await and simple do return `this.ln.connect()`

kiwiidb: Schermafbeelding 2022-10-18 om 09 29 24

bumi: ahhhh. I see. do we need to check the response there? 
```
const connected =  await this.ln.connect();
if (!connected) { 
   console.error("Failed to connect");
}
```

(currently we assume the connection works (init() does not have any error handling - which it should have)

escapedcat:
bumi noticed: `package-lock.json` can be removed. We use yarn and `yarn.lock`.

src/extension/background-script/connectors/commando.ts
kiwiidb: I don't think we need to check the response

.storybook/main.js
```diff
@@ -32,6 +32,12 @@ module.exports = {
       ],
       include: path.resolve(__dirname, "../"),
     });
+    config.module.rules.push({
```
reneaaron, line 35: What is that for?

kiwiidb: To solve this: 
```
ERROR in ./node_modules/lnmessage/dist/index.js 4:0-50
Module not found: Error: Can't resolve './crypto' in '/Users/kwinten/Alby/lightning-browser-extension/node_modules/lnmessage/dist'
Did you mean 'crypto.js'?
BREAKING CHANGE: The request './crypto' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

I don't understand the details, maybe @escapedcat remembers.

kiwiidb:
Boostagrams are currently not supported. As pointed out by Aaron over e-mail:

```
Actually it looks like the extratlvs​ field was only merged in to master on the 22nd of September - https://github.com/ElementsProject/lightning/pull/5619

I just checked with the CoreLn team and they said that it is going in to the next release.

So you can test it by installing CoreLn from master rather than a tagged release for the moment.

Aaron
```

So if we want to test this, we have to wait for the next Core LN release or build from master.

src/extension/background-script/connectors/commando.ts
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  async getBalance(): Promise<GetBalanceResponse> {
+    const response = (await this.ln.commando({
+      method: "listfunds",
+      params: {},
+      rune: this.config.rune,
+    })) as CommandoListFundsResponse;
+    let lnBalance = 0;
+    for (let i = 0; i < response.channels.length; i++) {
```
escapedcat, line 181: nit: Something like this might be easier to read:
```js
response.channels.forEach(channel => lnBalance += channel.channel_sat);
```

.storybook/main.js
bumi: hmm... in alby-tools this worked fine: https://github.com/getAlby/alby-tools/blob/master/package.json#L5

also is storybook even using our build system? if the build system can not handle it then we would need some webpack config update I guess?

@kiwiidb where do you see that error that you posted? Can try to remove this and run it again?

src/app/screens/connectors/ConnectCommando/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    try {
+      let validation;
+      // TODO: for native connectors we currently skip the validation because it is too slow (booting up Tor etc.)
+      if (account.connector === "nativelndhub") {
+        validation = { valid: true, error: "" };
+      } else {
+        validation = await utils.call("validateAccount", account);
+      }
+
```
bumi, line 78: please review this. this contains c&p code. 
can we validate the connection for commando before we save it (can it be initialized easily and fast enough and then unloaded again?

src/extension/background-script/connectors/commando.ts
```diff
@@ -0,0 +1,334 @@
...
+};
+export default class Commando implements Connector {
+  config: Config;
+  ln: LnMessage;
+
+  constructor(config: Config) {
+    this.config = config;
+    this.ln = new LnMessage({
+      remoteNodePublicKey: this.config.pubkey,
+      wsProxy: this.config.wsProxy || "wss://lnwsproxy.regtest.getalby.com",
```
bumi, line 89: guess we should not default to the regtest one here, but rather the production one.

src/extension/background-script/connectors/commando.ts
bumi: could probably be also nicely written with `reduce()`

```
lnBalance = response.channels.reduce((balance, channel) => balance + channel.channel_sat, 0)
```

src/extension/background-script/connectors/commando.ts
```diff
@@ -0,0 +1,334 @@
...
+      .commando({
+        method: "pay",
+        params: {
+          bolt11: args.paymentRequest,
+        },
+        rune: this.config.rune,
+      })
+      .then((resp) => {
+        const parsed = resp as CommandoPayInvoiceResponse;
+        const parsedTotalAmtStr = parsed.amount_msat.replace("msat", "");
```
bumi, line 202: puh, don't the return some proper number values without the currency string?

src/extension/background-script/connectors/commando.ts
```diff
@@ -0,0 +1,334 @@
...
+        const parsedSentAmtPaidStr = parsed.amount_sent_msat.replace(
+          "msat",
+          ""
+        );
+        const parsedTotalAmt = (parsedTotalAmtStr as unknown as number) / 1000;
```
bumi, line 207: the variable name indicates it's a string (and then there is some TypeScript fuuu as unknown as number) ? 

when dividing by 1000 we probably also want to round the value?

src/extension/background-script/connectors/commando.ts
```diff
@@ -0,0 +1,334 @@
...
+        const parsedTotalFee =
+          ((parsedSentAmtPaidStr as unknown as number) -
+            (parsedTotalAmtStr as unknown as number)) /
```
bumi, line 210: I think @escapedcat should look at this TypeScript

src/extension/background-script/connectors/commando.ts
```diff
@@ -0,0 +1,334 @@
...
+        if (parsed.invoices.length != 1) {
+          return {
+            data: {
+              paid: false,
+            },
+          };
+        }
+        return {
+          data: {
+            paid: parsed.invoices[0].status == "paid",
```
bumi, line 290: do we really need that `if (parsed.invoices.length != 1)` check? 

isn't `parsed.invoices[0]?.status == "paid"` enough? 

if there is no invoice it will be undefined and undefined is not equal to paid and such false

.storybook/main.js
bumi: oh I missed it's in the webpack config also. 
but somewhat I would like to get rid of that and the npm package should be in a way that it also works for us - because I think we don't have such a strange build system. 
maybe we can make a PR there?

src/extension/background-script/connectors/commando.ts
kiwiidb: Actually they do, I don't know why I missed this. Will update PR

kiwiidb: It's also just the wrong field, I made several brainfarts here.

src/extension/background-script/connectors/commando.ts
kiwiidb: I think we can even delete this default, it's already being set here to production: https://github.com/getAlby/lightning-browser-extension/blob/feature/commando-connector/src/app/screens/connectors/ConnectCommando/index.tsx#L26

src/extension/background-script/connectors/commando.ts
kiwiidb: In this call, they don't (ignore my deleted comments)..
I don't know why actually. Maybe we should ask them about it.

src/extension/background-script/connectors/commando.ts
kiwiidb: The fee can be msat, so I'm not sure if we want to round it, maybe?

src/extension/background-script/connectors/commando.ts
escapedcat: :D reduce was my first thought, but then... I hate to look at reduce :P

src/extension/background-script/connectors/commando.ts
escapedcat: We could do it later.  
@reneaaron mentioned that the rounding should probably not happen in the connector(s) but on UI level. I think he's right.

src/extension/background-script/connectors/commando.ts
escapedcat: nit: Try to use `===` or `!==` which also compares the type. Wondering why eslint isn't even showing a warning.

src/extension/background-script/connectors/commando.ts
bumi: +1

src/extension/background-script/connectors/commando.ts
bumi: strange that a node does not return numbers in an API response. typically you want to calculate with numbers they should not be formatted by default. 
but they have some strange things in their API (like the description hash thing)

src/extension/background-script/connectors/commando.ts
bumi: not sure about that. we need to make sure to return consistent values and types there. 
and that we can deal with those numbers everywhere.

src/extension/background-script/connectors/commando.ts
escapedcat: Updated

.storybook/main.js
escapedcat: Background: https://webpack.js.org/configuration/module/#resolvefullyspecified  

Removing `"type": "module",` from the "lnmessage" `package.json` already makes it work without this special webpack option.  
`"type": "module",` is letting node know that this is an [esm package](https://nodejs.org/api/esm.html#introduction) but this shouldn't be needed for a package that should run in the browser (only).  

I can create a PR for this and ask for feedback.

bumi: what's with the following package.json keys:
`source`, `main`, `unpkg`, `exports` ? 

packages that can run in different environments should publish the package valid for the different environments. 

Does the build step need to be improved? https://github.com/aaronbarnardsound/lnmessage/blob/master/package.json#L14 
`tsc` is not enough?

I don't know enough about the build setups in 2022, but I used `microbundle` recently in this package: https://github.com/getAlby/alby-tools/blob/master/package.json#L28

src/app/screens/connectors/ConnectCommando/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+        wsProxy,
+        privateKey,
+      },
+      connector: getConnectorType(),
+    };
+
+    try {
+      let validation;
+      // TODO: for native connectors we currently skip the validation because it is too slow (booting up Tor etc.)
+      if (account.connector === "nativelndhub") {
```
bumi, line 73: needs to be fixed.

src/app/screens/connectors/ConnectCommando/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+          id="port"
+          label={t("port.label")}
+          type="number"
+          required
+          title="port"
+          value={formData.port}
+          onChange={handleChange}
+        />
+      </div>
+      <div className="mb-6">
```
bumi, line 161: can we put proxy/privateKey behind some "advanced" link? most users probably won't edit this and shorter forms are nicer and simpler

src/i18n/locales/en/translation.json
```diff
@@ -288,6 +288,39 @@
         "errors": {
           "connection_failed": "Connection failed. Is the BTCPay connection URL correct and accessible?"
         }
+      },
+      "commando": {
+        "title": "Core Lightning",
+        "description": "Connect to your Core Lightning node",
+        "page": {
+          "title": "Connect to your Core Lightning node",
+          "instructions": "Make sure you have Core Lightning version 0.12.0 or newer, the commando plugin is running and your node is accessible over the Lightning Network. Create a rune by running 'lightning-cli commando-rune'."
```
bumi, line 297: can we add a guide/help entry on how this is done exactly?

src/app/router/connectorRoutes.tsx
```diff
@@ -13,9 +13,11 @@ import NewWallet from "@screens/connectors/NewWallet";
 import i18n from "~/i18n/i18nConfig";
 import { translationI18nNamespace } from "~/i18n/namespaces";
 
+import ConnectCommando from "../screens/connectors/ConnectCommando";
 import alby from "/static/assets/icons/alby.png";
 import btcpay from "/static/assets/icons/btcpay.svg";
 import citadel from "/static/assets/icons/citadel.png";
+import core_ln from "/static/assets/icons/core_ln.svg";
```
bumi, line 20: should we mark this with some `beta` badge?

kiwiidb: did we ever do this with any other connector? Why should we do it here then?

bumi: that's what I am asking. 
we did not have any other connector added at this stage that Alby is in now.  

LND for example was the first one used for months before fore even release Alby. 
Eclair would have been one, but that was also added very early on. so the sate of Alby was a very different. 

And this connector relies on a very new library a new websocket proxy, etc. so it is also different and not only a simple HTTP call.

kiwiidb: I personally don't feel a need to add it.

src/extension/background-script/connectors/commando.ts
kiwiidb: Turns out the problem was with me

src/app/router/connectorRoutes.tsx
bumi: So you're confident enough that this (and lnmessage) works reliable for the various CLN installations?

src/i18n/locales/en/translation.json
kiwiidb: In my opinion mostly technical people will use it anyway so I would do this later. My video on how to set it up on Digital Ocean can be published as a guide though.

src/app/router/connectorRoutes.tsx
kiwiidb: Yes. And to add to that I am also confident that there is nobody that will care if there is a "beta" badge here or not.

.storybook/main.js
escapedcat: We started a discussion with Aaron on this and will solve this later on.

src/extension/background-script/connectors/commando.ts
```diff
@@ -127,38 +129,36 @@ export default class Commando implements Connector {
...
+        return {
+          data: {
+            invoices: parsed.invoices
+              .map(
+                (invoice, index): ConnectorInvoice => ({
+                  id: invoice.label,
+                  memo: invoice.description,
+                  settled: invoice.status === "paid",
+                  preimage: invoice.payment_preimage,
+                  settleDate: invoice.paid_at,
```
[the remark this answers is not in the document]
bumi: ok, so its a timestamp.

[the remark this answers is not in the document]
bumi: 👍 thanks

.storybook/main.js
bumi: I think this can be removed no. 
Actually if not 1000% needed I would like to resolve this now. because otherwise we will forget it and not take care of it and it will stay in there.

src/i18n/locales/en/translation.json
bumi: ok, though raspiblitz people will do that for example

webpack.config.js
```diff
@@ -129,6 +129,12 @@ var options = {
         test: /\.(woff|woff2|eot|ttf|otf)$/i,
         type: "asset/resource",
       },
+      {
+        test: /\.m?js/,
+        resolve: {
+          fullySpecified: false
+        }
```
bumi, line 136: @escapedcat what's the status here? 
this still can not be removed?

escapedcat: Unfortunately not: https://github.com/aaronbarnardsound/lnmessage/issues/3#issuecomment-1305425021

src/app/screens/connectors/ConnectCommando/index.tsx
kiwiidb: Done btw

PR #1584, opened 2022-10-07
feat: use resultcard in lnurlchannel

im-adithya:

Align the success screen of LNURLChannel

#1379

### Type of change

Screenshot from 2022-10-07 22-34-17

### How has this been tested?

Manually

### Checklist

### DIFF
--- a/src/app/screens/LNURLChannel/index.tsx
+++ b/src/app/screens/LNURLChannel/index.tsx
@@ -1,8 +1,9 @@
+import Button from "@components/Button";
 import ConfirmOrCancel from "@components/ConfirmOrCancel";
 import Container from "@components/Container";
 import ContentMessage from "@components/ContentMessage";
 import PublisherCard from "@components/PublisherCard";
-import SuccessMessage from "@components/SuccessMessage";
+import ResultCard from "@components/ResultCard";
 import axios from "axios";
 import { useState, MouseEvent } from "react";
 import { useTranslation } from "react-i18next";
@@ -20,6 +21,7 @@ function LNURLChannel() {
   const { t: tComponents } = useTranslation("components", {
     keyPrefix: "confirm_or_cancel",
   });
+  const { t: tCommon } = useTranslation("common");
 
   const navigate = useNavigate();
   const navState = useNavigationState();
@@ -30,6 +32,7 @@ function LNURLChannel() {
 
   const [loadingConfirm, setLoadingConfirm] = useState(false);
   const [successMessage, setSuccessMessage] = useState("");
+  const [response, setResponse] = useState("");
   const [errorMessage, setErrorMessage] = useState("");
 
   async function confirm() {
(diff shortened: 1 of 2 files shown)


escapedcat:
image

Should we take care of that `undefined` in the response? Like say "No nodeId available" or just leave it out if it's undefined?

im-adithya:
image

escapedcat (approved):
ack, tested with fiatjefs playground

src/app/screens/LNURLChannel/index.tsx
```diff
@@ -55,7 +58,10 @@ function LNURLChannel() {
         throw new Error(`Failed to call callback: ${callbackResponse.message}`);
       }
 
-      setSuccessMessage(`${t("success")}. ${details.k1} ${nodeId}`);
+      setSuccessMessage(
+        t("success", { name: origin ? origin.name : details.domain })
+      );
+      setResponse(`${details.k1} ${nodeId ? nodeId : `[no nodeId available]`}`);
```
reneaaron, line 64: Is the nodeId supposed to be null here?

bumi: when a node id is not available this channel thing does not work. 
so I don't think that this message should ever happen and thus should be needed.

escapedcat: In this case it might just be related to testing it with fiatjef?

bumi: so nodeId comes from line 47. where we do the getInfo call. 
we request the information of the user's connected node and get the node ID from that.

that node id is sent to the callback URL - the service send opens a channel with the user's node (that's why the nodeId is needed).

this whole channel thing works only if there is a node ID available. 
it must never be null. so this `setResponse` does not make sense. 


you could add a check above that if nodeId is falsy then it shows an error and does not even try to do the lnurl-channel call.


I can only imagine that you use an Alby account or some custodial account when you run this. That is not your node so you can not open a a channel to it, or rather you would not. 

the getinfo call likely also does not return the nodeid.

bumi (changes requested):

escapedcat:
lgtm, we see the k1 and the node-id
image

src/app/screens/LNURLChannel/index.tsx
```diff
@@ -55,7 +63,10 @@ function LNURLChannel() {
         throw new Error(`Failed to call callback: ${callbackResponse.message}`);
       }
 
-      setSuccessMessage(`${t("success")}. ${details.k1} ${nodeId}`);
+      setSuccessMessage(
+        t("success", { name: origin ? origin.name : details.domain })
+      );
+      setResponse(`${details.k1} ${nodeId}`);
```
bumi, line 69: Those are some internal information I don't thinks this is useful for the user.

Let's only show the success message and no response details.

(it's enough to say "Channel request sent successfully to {{name}}" )

im-adithya: Makes sense, done!

escapedcat:
I assume bumi wanted to merge

bumi:
You had approved it. (though a bit early)

escapedcat:
Yeah, checked again today

PR #1594, opened 2022-10-09
feat: auto-focus first form field

reneaaron:

Auto-focuses the first input field in Send / Receive screens to reduce total amount of clicks necessary and improve keyboard usage.

### Type of change

### DIFF
--- a/src/app/screens/Send/index.tsx
+++ b/src/app/screens/Send/index.tsx
@@ -173,6 +173,7 @@ function Send() {
               placeholder={t("input.placeholder")}
               value={invoice}
               disabled={loading}
+              autoFocus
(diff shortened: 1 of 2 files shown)


escapedcat (approved):

src/app/screens/Send/index.tsx
```diff
@@ -187,6 +187,7 @@ function Send() {
                   <QrCodeIcon className="h-6 w-6 text-blue-500" />
                 </button>
               }
+              autoFocus
```
bumi, line 190: can we move this one up? the `onChange` and `endAdornment` are super long and this might be missed.
maybe above onchange?

PR #1595, opened 2022-10-09
fix: amount-input changes value on scroll

### LINKED ISSUE #1593: Amount-input changes value on scroll (mouse/touchpad)
### Describe the bug

Scrolling the mouse/via touchpad within an amount-input will change the value. Issue is that a user might not notice that the wheel was used and the amount was changed.

image

### To Reproduce

1. Visit 'Send'
1. Input an ln-adress i.e. `bumi@getalby.com` and click 'continue"
1. Click into the amount-input (so it's focused)
1. Scroll mouse down (or up)
1. Value changes while scrolling

### Expected behavior

Value should mot change on scroll

### Additional context [optional]

These amount-fields use our [`DualCurrencyField`-component](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/form/DualCurrencyField/index.tsx#L40) which is defined as a [`number`-input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number). This is helpful because it's easier for the input to deal with numbers and min/max values.  
We already [hide the special arrow-buttons via css](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/styles/index.css/#L54-L66) but the `steps`-functionality is still supported. We should disable/remove this as well.

g1mishra:

I have added a "wheel" event listener on input and on callback of the event just preventing the fired event.

Fixes #1593 

### Type of change

(Remove other not matching types)

### How has this been tested?

I opened up the extension in Chrom and after login went to Receive tab, In amount-input entered some numbers and tried to scroll up & down to change the value, but it not changing. 

### Checklist

### DIFF
--- a/src/app/components/form/DualCurrencyField/index.tsx
+++ b/src/app/components/form/DualCurrencyField/index.tsx
@@ -66,7 +66,11 @@ export default function DualCurrencyField({
     const ignoreScroll = (evt: globalThis.WheelEvent) => {
       evt.preventDefault();
     };
-    inputEl.current && inputEl.current.addEventListener("wheel", ignoreScroll);
+    const elem = inputEl.current;
+    elem && elem.addEventListener("wheel", ignoreScroll);
+    return () => {
+      elem && elem.removeEventListener("wheel", ignoreScroll);
+    };
   }, [inputEl]);
 
   return (
--- a/src/app/components/form/DualCurrencyField/index.tsx
+++ b/src/app/components/form/DualCurrencyField/index.tsx
@@ -61,13 +61,12 @@ export default function DualCurrencyField({
     />
   );
 
-  // run effect on input mount to ignoreScroll event
+  // run effect on input mount to ignore wheel/scroll event
   useEffect(() => {
-    if (inputEl.current == null) return;
     const ignoreScroll = (evt: globalThis.WheelEvent) => {
       evt.preventDefault();
     };
-    inputEl.current.addEventListener("wheel", ignoreScroll);
+    inputEl.current && inputEl.current.addEventListener("wheel", ignoreScroll);
   }, [inputEl]);
 
   return (
--- a/src/app/components/form/DualCurrencyField/index.tsx
+++ b/src/app/components/form/DualCurrencyField/index.tsx
@@ -1,4 +1,4 @@
-import { useRef } from "react";
+import { useEffect, useRef } from "react";
 import { classNames } from "~/app/utils";
 
 export type Props = {
@@ -60,6 +60,15 @@ export default function DualCurrencyField({
       max={max}
     />
   );
+
+  // run effect on input mount to ignoreScroll event
+  useEffect(() => {
+    if (inputEl.current == null) return;
+    const ignoreScroll = (evt: globalThis.WheelEvent) => {
+      evt.preventDefault();
+    };
+    inputEl.current.addEventListener("wheel", ignoreScroll);
+  }, [inputEl]);
 
   return (
     <div className="relative block m-0">


src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -61,6 +61,15 @@ export default function DualCurrencyField({
     />
   );
 
+  // run effect on input mount to ignoreScroll event
+  useEffect(() => {
+    if (inputEl.current == null) return;
+    const ignoreScroll = (evt: globalThis.WheelEvent) => {
+      evt.preventDefault();
+    };
+    inputEl.current.addEventListener("wheel", ignoreScroll);
```
escapedcat, line 70: Could the `if` above be avoided by doing this here?:
```ts
inputEl.current && inputEl.current.addEventListener("wheel", ignoreScroll);
```

escapedcat:
Thanks @g1mishra ! Minor comment. e2e will pass after a rebase. It's not related to this PR.

src/app/components/form/DualCurrencyField/index.tsx
g1mishra: Yes @escapedcat It could be. 
Should I do that?

escapedcat: Yes, that would be nice, thanks!

escapedcat (approved):
ack

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -61,6 +61,14 @@ export default function DualCurrencyField({
     />
   );
 
+  // run effect on input mount to ignore wheel/scroll event
+  useEffect(() => {
+    const ignoreScroll = (evt: globalThis.WheelEvent) => {
+      evt.preventDefault();
+    };
+    inputEl.current && inputEl.current.addEventListener("wheel", ignoreScroll);
```
lisabaut, line 69: I think, the event listener should be removed on unmount.

I would add:
```TS
  return () => inputEl.current && inputEl.current.addEventListener("wheel", ignoreScroll);
```

to the hook.

What do you think?

lisabaut (approved):
Thanks, just one suggestion below.

src/app/components/form/DualCurrencyField/index.tsx
g1mishra: @lisabaut At first I thought the same, but this event is on input ref and on unmount that will become null.

lisabaut: Yes, I think you are right. Thanks for the pointer.

g1mishra: @lisabaut There's a way to do that, we can assign ref to another variable, and then we can use that variable to remove the event listener.

```
const elem = inputEl.current;
return () => elem?.removeEventListener("wheel", ignoreScroll);
```
Should I do that?

bumi:
thanks for tackling this.
what do you think? can we do this by simply using a text field? -> no code is better than no code :)

src/app/components/form/DualCurrencyField/index.tsx
lisabaut: I am also still thinking about this ref and am reading about it again.

Yes, let´s rather remove the event listener even though it is highly likely that the ref is `null` on unmount. 👍 
Even though I guess this is only the case when using `useLayoutEffect` and not with `useEffect`

So like:

```TS
  useEffect(() => {
    const inputRef = inputEl.current;
    const ignoreScroll = (evt: globalThis.WheelEvent) => {
      evt.preventDefault();
    };

    inputRef && inputRef.addEventListener("wheel", ignoreScroll);

    return () => {
      inputRef && inputRef.removeEventListener("wheel", ignoreScroll);
    };
  }, [inputEl]);
```

Same code with storing the `ref.current` in a local var inside the `useEffect`:
https://bobbyhadz.com/blog/react-ref-add-event-listener

In depth reading again:
https://blog.thoughtspile.tech/2021/05/17/everything-about-react-refs/

lisabaut:
> thanks for tackling this. what do you think? can we do this by simply using a text field? -> no code is better than no code :)

IMO converting the input field from "number" to "text" would imply a bigger refactoring.
This would require a custom validation for the input value (only numbers are allowed, testing against min and max values).
Which is not necessarily less code in the end.

What do you say @g1mishra and @bumi ?

g1mishra:
> > thanks for tackling this. what do you think? can we do this by simply using a text field? -> no code is better than no code :)
> 
> IMO converting the input field from "number" to "text" would imply a bigger refactoring. This would require a custom validation for the input value (only numbers are allowed, testing against min and max values). Which is not necessarily less code in the end.
> 
> What do you say @g1mishra and @bumi ?

Agree with you @lisabaut.
@bumi In the raised issue, It was mentioned that we are using input-number because it's easier for the input to deal with numbers and min/max values. Otherwise, I could have used text with pattern props

src/app/components/form/DualCurrencyField/index.tsx
g1mishra: @lisabaut I have added removeEventListener on unmount

escapedcat:
> Agree with you @lisabaut. @bumi In the raised issue, It was mentioned that we are using input-number because it's easier for the input to deal with numbers and min/max values. Otherwise, I could have used text with pattern props

Yeah, if we switch it to normal text I'm worried we have to deal with all kind of unexpected input. So far we got around adding a input-validation lib. Let's keep it that way for now.

escapedcat:
Thanks @g1mishra !

Issue #1597, opened 2022-10-09
Settings: Text is displayed with HTML-tags


bitcoinuser:
### Describe the bug

Text with strange symbols like <br /> , </strong>...

### Screenshots of the changes [optional]

![Captura de tela de 2022-10-09 11-34-59](https://user-images.githubusercontent.com/1178856/194762750-dc3ba111-b2f8-45f1-afd0-d94996b28da4.png)

### To Reproduce

1. Go to Settings
2. LNURL-Auth

### Expected behavior

Text without strange symbols

### Alby information

- Alby Version: 1.17

### Are you working on this issue?

No

vaibhavgarg237:
@escapedcat I have opened PR #1604 to fix this.

The hint text contains HTML tags that didn't parse, so it appeared as it is in the app. For this we need to use [**html-react-parser** npm package](https://www.npmjs.com/package/html-react-parser) to parse the string. Now it's showing the text as expected with the relevant Br and Strong tags.

PR #1604, opened 2022-10-10
fix: text displayed with tags #1597

vaibhavgarg237:

__LNURL_Auth's hint text in the settings tab is displayed with HTML tags. They need to be parsed so that hint text is shown as expected with relevant break and strong tags_

Fixes _#1597_

### Type of change

#**Updated Image**
image

#**Old Image**
image

### How has this been tested?

_✅ Unit tests_

### Checklist

### DIFF
--- a/src/app/screens/Settings.tsx
+++ b/src/app/screens/Settings.tsx
@@ -343,7 +343,12 @@ function Settings() {
         >
           {t("lnurl_auth.title")}
         </a>{" "}
-        {t("lnurl_auth.hint")}
+        <Trans
+          i18nKey={"lnurl_auth.hint"}
+          t={t}
+          // eslint-disable-next-line react/jsx-key
+          components={[<strong></strong>]}
+        />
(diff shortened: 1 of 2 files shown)


escapedcat:
The strong tag should be handled like this:
- https://github.com/getAlby/lightning-browser-extension/blob/refactor/manifest-v3-support--october/src/i18n/locales/en/translation.json#L160
- https://github.com/getAlby/lightning-browser-extension/blob/refactor/manifest-v3-support--october/src/app/screens/Home/index.tsx/#L329-L337

The `br`s can be removed. I don't think we reallyt need those in the hint-texts. If we would we could use something like this maybe: https://stackoverflow.com/a/54747308/1667461  
But for now we're good without.

vaibhavgarg237:
@escapedcat I have updated the commit with the required changes.

escapedcat (approved):
🚀

PR #1614, opened 2022-10-13
fix: fiat value 0 due to slow connection

### LINKED ISSUE #1608: Sometimes fiat value appear as 0 but should show none info or other kind of rendering
### Describe the bug

Sometimes I open Alby and I look to see how much the sats I have is representing in fiat. But sometimes maybe because it take some seconds to load or maybe because the fiat quote provider service is offline, the value appear as 0.

This screenshot is working ok, I couldn't take a screenshot reproducing the problem

Captura de tela de 2022-10-11 14-05-12

### To Reproduce

The problem is hard to reproduce, maybe when fiat quote provider is slow or offline

### Expected behavior

Not show 0 but instead show none information or some other kind of rendering. 

### Alby information

- Alby Version: 1.17

vaibhavgarg237:

_During loading due to a slow connection, fiatBalance is shown ~$0, so "..." will be shown when the user has some satsBalance but due to a network issue, fiatBalance is not fetched properly._

Fixes #1608

### Type of change

### How has this been tested?

_As mentioned in the issue issue only reproduces due to [slow connection](https://github.com/getAlby/lightning-browser-extension/issues/1608#issuecomment-1276286733), so I tested by hard coding values._

### Checklist

### DIFF
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -32,6 +32,9 @@ function AccountMenu({ showOptions = true }: Props) {
   const navigate = useNavigate();
   const { accounts, getAccounts } = useAccounts();
   const [loading, setLoading] = useState(false);
+  const fiatBalValue = Number(
+    balancesDecorated.fiatBalance.replace(/[^0-9.]+/g, "")
+  );
 
   // update title
   const title =
@@ -93,7 +96,11 @@ function AccountMenu({ showOptions = true }: Props) {
             </span>
             {!!balancesDecorated.fiatBalance && (
               <span className="text-xs text-gray-600 dark:text-neutral-400">
-                ~{balancesDecorated.fiatBalance}
+                ~
+                {balancesDecorated.satsBalance !== "0 sats" &&
+                fiatBalValue === Number("0.00")
+                  ? "..."
+                  : balancesDecorated.fiatBalance}
               </span>
             )}
           </p>


src/app/components/AccountMenu/index.tsx
```diff
@@ -32,6 +32,9 @@ function AccountMenu({ showOptions = true }: Props) {
   const navigate = useNavigate();
   const { accounts, getAccounts } = useAccounts();
   const [loading, setLoading] = useState(false);
+  const fiatBalValue = Number(
+    balancesDecorated.fiatBalance.replace(/[^0-9.]+/g, "")
```
bumi, line 36: hmmm. what is this doing? the fiatBalance is the formatted string and we get the numbers out again? this sounds like it will break. 
we have numbers and should work with numbers

bumi (changes requested):

src/app/components/AccountMenu/index.tsx
escapedcat: Uhm, yeah, we should find a clean way to handle this if we really want to.  
@vaibhavgarg237 maybe you can check where the "waiting" is happening and maybe we could get a loading-state from there or send back "..." till we have a value or something else.

vaibhavgarg237: Actually, for comparison, we need numbers to match it with 0.00, so I tried to use regex for extracting numbers to match with 0.00
image

https://github.com/getAlby/lightning-browser-extension/pull/1614/files#diff-66c0c361b98e12d7a076d9ae7eb73d31816d5b749f1819ed1af6c787008127eeR101

This successfully extracts numbers, and if it doesn't match with 0, and in that case our satsBal is also not 0 then it will print the original value as usual...
_For $_
image
_On similar lines for AED & other currencies_
image

bumi: the app should know if it is still loading the balances and if it is loading we can show something. 
maybe @lisabaut could also help here with a quick thought, the has worked on the fiat amount.

But I don't think we should compare such strings or parse the numbers again. this is complex and sounds error prone.

if this is complicated we should leave it for now.

lisabaut: The component [`AccountMenu`](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/AccountMenu/index.tsx) behaves correctly:
=> `balancesDecorated.fiatBalance` is either an empty string or a formatted value.

So no need to change the code in component `AccountMenu` as it contains already the check if `balancesDecorated.fiatBalance` is an empty string or not [here](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/AccountMenu/index.tsx#L97).

As @vaibhavgarg237 [already mentioned here](https://github.com/getAlby/lightning-browser-extension/issues/1608#issuecomment-1275077227) the root cause of the fiatValue being a zero value (`$0,00`) is the value of `currencyRate` [within the local state of the SettingsContext](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L30) which got introduced with [this PR](https://github.com/getAlby/lightning-browser-extension/pull/1354/files). 

The `currencyRate` is initially `0` and gets updated after the settings are completely loaded [here](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L58-L63)
=> when `getFiatValue` is called and currencyRate is still `0`, the `fiatValue` is always formatted as a zero value (`$0,00`) 

This results in a similar problem when changing the currency on the Settings Page:
=> the fiatValue within the AccountMenu needs several steps to show the correct new value because the `currencyRate` gets updated after `getFiatValue` is called
=> `"$38.76"` (initial value) => change to CNY => "`CN¥38.76"` (correct currency, wrong value) => `"CN¥79.65"` (correct currency, correct value)

lisabaut: Here is my suggestion for a fix:

1. Revert all changes in this PR from the component `AccountMenu`
2. SettingsContext: Delete this [line](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L30) which removes `currencyRate` from the local state
3. SettingsContext: Delete this [effect](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L58-L63) accordingly
4. SettingsContext: Extend `getFiatValue` [here](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L65-L70) like this:
```TS
  const getFiatValue = async (amount: number | string) => {
    const response = await api.getCurrencyRate(); // gets rate from API or browser storage
    const value = await getFiatValueFunc({
      amount,
      rate: response.rate,
      currency: settings.currency,
    });
    return value;
  };
```

**Explanation:**
`currencyRate` is only needed for the correct result of `getFiatValue`.
`currencyRate` is not needed elsewhere => no need to store the value in the local state (also this causes a re-render on every update).
Calling `api.getCurrencyRate` ensures that the currency rate is always present and always the correct value.


What do you think?

bumi: thanks for that detailed write up

what we want to avoid is that we call the background script for each conversion. E.g. when we render a transaction list I don't think we need to go through the complexity of calling the background script to get the rate for each value that we want to convert to fiat. 
So there imo the rate should somehow be only fetched once. 

I am not sure if this is done in that `state`? ...just to make sure you're aware of this.

lisabaut: I thought about it again.

It is possible to store the latest rate (and currency for comparison) in a component ref.
Updating values in a ref does not re-render the component. Here this is useful as only the local function `getFiatValue` needs the correct rate.

Now `getFiatValue` always checks first if the latest rate stored in the ref does match with the currently selected currency.

As I tested the code myself I opened another PR for that.

Sorry @vaibhavgarg237 for high jacking your Pull Request ! 🌻 

Please have a look here
https://github.com/getAlby/lightning-browser-extension/pull/1645

escapedcat:
Closing this in favour of #1645   
@vaibhavgarg237 thanks for starting this!

src/app/components/AccountMenu/index.tsx
vaibhavgarg237: Sorry for the delayed response, actually I had to handle some medical emergency. 
@lisabaut Thanks for tackling this issue.

PR #1626, opened 2022-10-17
fix(connectors): lndhub incoming tx order #1625

### LINKED ISSUE #1625: LNDhub connection - Incoming sorting bug (oldest-newest)
There's a bug with the LNDHub connection (connected with BlueWallet) and the incoming (received) sorting right now received transactions are sorted oldest-newest.
Alby

escapedcat:

Add reverse to incoming payments to correct wrong order

Fixes #1625

### Type of change

### How has this been tested?

Manually with a bluewallet account

### DIFF
--- a/src/extension/background-script/connectors/lnd.ts
+++ b/src/extension/background-script/connectors/lnd.ts
@@ -302,7 +302,9 @@ class Lnd implements Connector {
           type: "received",
         };
       })
-      .reverse();
+      .sort((a, b) => {
+        return b.settleDate - a.settleDate;
+      });
 
     return {
       data: {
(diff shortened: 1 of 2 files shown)


bumi:
I actually think that this breaks other lndhub implementations like the Alby accounts. 
those return the invoices in the useful order. Did you test that? 

instead of doing the reverse it should be probably ordered by `settleDate`.

escapedcat:
Thanks, yes! Fixed it. Also switched it for LND.

bumi:
lnd is at least reliable, because it is specifically that LND. and they have an parameter that we use there for the sorting.

escapedcat:
> lnd is at least reliable, because it is specifically that LND. and they have an parameter that we use there for the sorting.

Yeah but I think that's just for the "whole" list in terms of "last page". We still need to "reverse" that last page. afaik that's why we added the reverse there in the first place.

PR #1633, opened 2022-10-18
feat: add support for lsat

bumi:

This exposes lsat functions to web apps. Making it easy to do the lsat handshake.

```js
window.alby.fetchWithLsat(url, fetchOptions, options);
```
fetchOptions: options passed to `fetch`
options: 
  store: a storage object to save the lsat tokens e.g. window.localStorage
  webln: a webln object if window.webln should not be used (not relevant in the Alby case)

#9 

### Type of change

### Checklist

### DIFF
--- a/src/extension/inpage-script/index.js
+++ b/src/extension/inpage-script/index.js
@@ -8,9 +8,7 @@ if (document) {
   // this is just to make double sure we load it
   if (!window.webln) {
     window.webln = new WebLNProvider();
-  }
-  if (!window.alby) {
-    window.alby = alby;
+    window.alby = alby; // nomally also loaded onstart
   }
 
   const readyEvent = new Event("webln:ready");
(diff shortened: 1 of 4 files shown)


src/extension/inpage-script/index.js
```diff
@@ -8,6 +9,9 @@ if (document) {
   if (!window.webln) {
     window.webln = new WebLNProvider();
   }
+  if (!window.alby) {
```
reneaaron, line 12: Other inpage scripts are loaded in a separate file. Is there a reason `window.alby` is handled differently?

bumi: no. thanks. did not think about that. 
changed it.

reneaaron (approved):
Concept ACK

AaronDewes:
I just opened #1916, which adds window.alby APIs that require some other code in the background script,  which would conflict with this PR. #1916 couldn't be integrated into alby-tools. In my opinion, that makes more sense than exposing a library to websites.

Websites could just embed `alby-tools`, which would also make more sense for type definitions (So they can have TypeScript integration similar to what's possible with webln).

rolznz:
I agree, lsat in https://github.com/getAlby/js-lightning-tools already use window.webln to make the payment

PR #1641, opened 2022-10-19
feat(confirmorcancel): focus the submit button

escapedcat:

Focus CTA in prompt to make confirmation faster.  
Especially for usecases like nostr-chats where you need to confirm multiple times.

### Type of change

image
### How has this been tested?

Manually

### DIFF
--- a/src/app/components/Button/index.tsx
+++ b/src/app/components/Button/index.tsx
@@ -1,6 +1,7 @@
+import { forwardRef } from "react";
+import type { Ref } from "react";
+import Loading from "~/app/components/Loading";
 import { classNames } from "~/app/utils/index";
-
-import Loading from "../Loading";
 
 export type Props = React.ButtonHTMLAttributes<HTMLButtonElement> & {
   fullWidth?: boolean;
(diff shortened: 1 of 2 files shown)


lisabaut (approved):
Thanks, code looks straight forward

reneaaron (approved):
Looking good! 💯

utACK

bumi:
this change now means that if the button is below the visible area the page scrolls down. 
you can see this for example when you send to a getalby.com lightning address.
What do you think about this?

escapedcat:
Sorry I can't follow. Can you give more details or a screenshot?

image

reneaaron:
@escapedcat This is how it looks now:

image

that is indeed a problem. however, i think we should clean up the send screen for that. a send screen should fit into the extension popup without scrolling. (amount, comment, name are very lengthy in our forms)

escapedcat:
> that is indeed a problem. however, i think we should clean up the send screen for that. a send screen should fit into the extension popup without scrolling. (amount, comment, name are very lengthy in our forms)

Thanks for the clarification! There's not ticket for that yet, right?

We can also turn of the focus for this longer lnurlpay screen for now.

bumi:
Click "Send" and enter a lightning address. we need to use the app more... 
looking at one flow is not enough.

@reneaaron do you think we can always get the button above the fold? 

anyway this thing is in master, so how do we deal with it? are we OK with how it is right now? 
talking about cleaning up the send screen does not help and will likely not happen anytime soon.

reneaaron:
> Click "Send" and enter a lightning address. we need to use the app more...
looking at one flow is not enough.

This is on me. I noticed some strange behavior but didn't investigate further. 

> @reneaaron do you think we can always get the button above the fold?

I think that this should be possible, yeah. 

I suggest we:
 - Remove the autoFocus for LNURL like @escapedcat suggested
 - Tackle the send screen in #1662

escapedcat:
See: https://github.com/getAlby/lightning-browser-extension/pull/1663

bumi:
great!

PR #1643, opened 2022-10-20
refactor: remove redundent code

escapedcat:
### Describe the changes you have made in this PR

During the refactor I didn't notice that this was redundent and doesn't need to be async anymore.

### Type of change

(Remove other not matching type)

- refactor

### How has this been tested?

Manually

PR #1645, opened 2022-10-20
refactor: store currency rate in settings ref

### LINKED ISSUE #1608: Sometimes fiat value appear as 0 but should show none info or other kind of rendering

lisabaut:

This is a refactoring which resulted from out the discussion in PR https://github.com/getAlby/lightning-browser-extension/pull/1614

- store latest currency rate in SettingContext ref instead of state to prevent a re-render
- when `getFiatValue` is called => always check if the currency in the state matches with the currency in the ref, and update otherwise
- => this ensures that the fiat value is calculated with the correct rate
- => this ensures that the latest currency rate is called from the background script only once when the currency changed 
- Also wrap `getFiatValue` in a try/catch block and display an error when a sat to fiat conversion failed

Unfortunately this reverted a few commits from [this PR](https://github.com/getAlby/lightning-browser-extension/pull/1643/files) again (sorry @escapedcat 🌷 ) 

### FIXES
- fixes https://github.com/getAlby/lightning-browser-extension/issues/1608
- and fixes bug from [this comment](https://github.com/getAlby/lightning-browser-extension/pull/1614#discussion_r997733183):

>  This results in a similar problem when changing the currency on the Settings Page:
> => the fiatValue within the AccountMenu needs several steps to show the correct new value because the currencyRate gets updated after getFiatValue is called
> => "$38.76" (initial value) => change to CNY => "CN¥38.76" (correct currency, wrong value) => "CN¥79.65" (correct currency, correct value)

### Screenshots
Try/Catch Toast Error

toast

Fixes #1608

### Type of change

### How has this been tested?

- checking how the `AccountMenu` renders the fiatValue

### Checklist

### DIFF
--- a/src/app/context/SettingsContext.tsx
+++ b/src/app/context/SettingsContext.tsx
@@ -13,7 +13,7 @@ interface SettingsContextType {
   settings: SettingsStorage;
   updateSetting: (setting: Setting) => void;
   isLoading: boolean;
-  getFiatValue: (amount: number | string) => string;
+  getFiatValue: (amount: number | string) => Promise<string>;
 }
 
 type Setting = Partial<SettingsStorage>;
--- a/src/app/context/SettingsContext.tsx
+++ b/src/app/context/SettingsContext.tsx
@@ -92,7 +92,7 @@ export const SettingsProvider = ({
           `SettingsProvider: getFiatValue with currency ${settings.currency} failed. (${e.message})`
         );
 
-      return "Error";
+      return "Fiat Error";
     }
   };
 
--- a/src/app/context/SettingsContext.tsx
+++ b/src/app/context/SettingsContext.tsx
@@ -1,8 +1,9 @@
 import dayjs from "dayjs";
 import i18n from "i18next";
-import { useState, useEffect, createContext, useContext } from "react";
+import { useState, useEffect, createContext, useContext, useRef } from "react";
 import { toast } from "react-toastify";
 import { getTheme } from "~/app/utils";
+import { CURRENCIES } from "~/common/constants";
 import api from "~/common/lib/api";
 import { getFiatValue as getFiatValueFunc } from "~/common/utils/currencyConvert";
 import { DEFAULT_SETTINGS } from "~/extension/background-script/state";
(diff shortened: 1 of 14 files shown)


bumi:
ups merged this other PR. sorry for the conflict

escapedcat (approved):
Awesome, thanks Lisa! Kinda missed the promises :P

lisabaut:
@escapedcat Please review again. I added a try/catch block too and opened a ticket for a found bug
https://github.com/getAlby/lightning-browser-extension/issues/1649

escapedcat (approved):
Right, some do not support certain currencies? Thanks!

src/app/context/SettingsContext.tsx
```diff
@@ -55,19 +60,41 @@ export const SettingsProvider = ({
...
+        amount,
+        rate,
+        currency: settings.currency,
+      });
+
+      return value;
+    } catch (e) {
+      console.error(e);
+      if (e instanceof Error)
+        toast.error(
```
bumi, line 91: will this lead to an overload of toasts if something fails?  for each payment one toast? 


Also is that message helpful for the user? the user does not know what the `SettingsProvider` and what `getFiatValue` is. 
so isn't `console.error` enough?

src/app/context/SettingsContext.tsx
```diff
@@ -55,19 +60,41 @@ export const SettingsProvider = ({
...
+          `SettingsProvider: getFiatValue with currency ${settings.currency} failed. (${e.message})`
+        );
+
+      return "Fiat Error";
```
bumi, line 95: maybe we only return an empty string? otherwise we print "Fiat Error" everywhere - which is also not very descriptive for the user.

src/app/context/SettingsContext.tsx
lisabaut: Removed the toasts

src/app/context/SettingsContext.tsx
lisabaut: Fine for me. But when the toast are removed now, at least something should be shown to the user that there was an error.

Changed it now to "??"

bumi: that error should be an exception, so I'd not optimize for that error. and for the user it's obvious that there is something wrong when the fiat amount is not show.

lisabaut:
As discussed. This can be merged

PR #1665, opened 2022-10-25
feat: focus currency-input

escapedcat:

Focus currency field

### Type of change

### How has this been tested?

Manually

### DIFF
--- a/src/app/screens/MakeInvoice/index.tsx
+++ b/src/app/screens/MakeInvoice/index.tsx
@@ -94,6 +94,10 @@ function MakeInvoice() {
             <div>
               {amountEditable ? (
                 <div className="mb-4">
+                  {/* 
+                    TODO: https://github.com/getAlby/lightning-browser-extension/issues/1666
+                    [Feature] MakeInvoice - switch currency TextField to DualCurrency field to support Fiat #1666 
+                   */}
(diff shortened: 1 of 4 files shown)


src/app/screens/MakeInvoice/index.tsx
```diff
@@ -94,6 +94,10 @@ function MakeInvoice() {
             <div>
               {amountEditable ? (
                 <div className="mb-4">
+                  {/*
```
reneaaron, line 97: I don't think that is the right place to track todos. Move the issue to the current / next sprint if you think this is important.

reneaaron (commented):
Other than the comment: utACK

src/app/screens/MakeInvoice/index.tsx
escapedcat: Hm, I kinda like doing this because if a dev is doing something in that file they also see this even if they don't know about the issue. Also better than todos without an issue, of which we have quite some.

reneaaron (approved):

src/app/screens/MakeInvoice/index.tsx
escapedcat: Sure, happy to talk

PR #1688, opened 2022-10-28
fix(nostr): return fully signed nostr event

### LINKED ISSUE #1687: nostr.signEvent is not following the spec

### Describe the bug

nostr.signEvent is not following the spec 

it should return the full event.

https://github.com/nostr-protocol/nips/blob/master/07.md

### Screenshots [optional]

_No response_

### Steps To Reproduce

required fields are annoying

### Expected behavior

it should follow the spec and return an Event and not only the signature

### Alby information

current version

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

_No response_

bumi:
The [spec NIP07](https://github.com/nostr-protocol/nips/blob/master/07.md) says the full event with signature must be returned. So far we had only returned the signature.

We now also have a special ConfirmSignMessage screen for nostr events because the default ConfirmMessage screen still has some signing specific code which we don't need and want for nostr.

fixes #1687

### DIFF
--- a/src/app/screens/Nostr/ConfirmSignMessage.tsx
+++ b/src/app/screens/Nostr/ConfirmSignMessage.tsx
@@ -12,6 +12,7 @@ import ScreenHeader from "~/app/components/ScreenHeader";
 import { useNavigationState } from "~/app/hooks/useNavigationState";
 import { USER_REJECTED_ERROR } from "~/common/constants";
 import msg from "~/common/lib/msg";
+import { Event } from "~/extension/ln/nostr/types";
 import type { OriginData } from "~/types";
 
 function ConfirmSignMessage() {
@@ -22,9 +23,8 @@ function ConfirmSignMessage() {
   });
   const navigate = useNavigate();
 
-  const message = navState.args?.message as string;
+  const event = navState.args?.event as Event;
   const origin = navState.origin as OriginData;
-  //const [rememberMe, setRememberMe] = useState(false);
   const [loading, setLoading] = useState(false);
   const [successMessage, setSuccessMessage] = useState("");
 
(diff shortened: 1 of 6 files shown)


src/app/screens/Nostr/ConfirmSignMessage.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const origin = navState.origin as OriginData;
+  //const [rememberMe, setRememberMe] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [successMessage, setSuccessMessage] = useState("");
+
+  // TODO: refactor: the success message and loading will not be displayed because after the reply the prompt is closed.
+  async function confirm() {
+    try {
+      setLoading(true);
+      msg.reply({
```
bumi, line 35: following the other/newer conventions we should do the signing call here and return the response then back.

reneaaron: really? if we move to more generic prompts & permissions we would need to move those things out again, no?

reneaaron: i feel like those prompts are more like javascript confirms: 

```
if(confirm("do you really want to do x?")) {
  // do x
}
```

src/types.ts
```diff
@@ -135,6 +135,7 @@ export type NavigationState = {
     amount?: string;
     customRecords?: Record<string, string>;
     message?: string;
+    event?: Event;
```
reneaaron, line 138: I added this to this type, but IMHO that feels very quite wrong as this type has to hold ALL types of all calls. Don't know if we want to tackle that within this PR though...

reneaaron:
Since we now have a customizable UI for this prompt I removed the JSON display from the confirm dialog, you can now see a proper message when signing: 

image

reneaaron (approved):
tACK

src/types.ts
bumi: cc @escapedcat @lisabaut

src/app/screens/Nostr/ConfirmSignMessage.tsx
bumi: I don't know what's the best approach here. 

yes, it is like `confirm()` but much more complexity can be behind that. 
one approach is to simply return true/false (which works for those simple confirms) 

or to do the actual call in the prompt and then return the response of that one. 

In here I do it also like in this PR here: https://github.com/getAlby/lightning-browser-extension/pull/1707 (simple return true/false and have the logic in the calling action)

bumi (commented):
whoa nice!

feel free to merge when you think it's good.

Issue #1692, opened 2022-10-31
Extend DB with "permissions"


lisabaut:
### Feature description

This is part of the meta ticket https://github.com/getAlby/meta/issues/230

In order to let WebLN pass generic methods of a connector to their API, the extension needs to check the user's permissions for this connector methods first.

### TODO

- [x] extend the DB with "permissions" with schema: `allowanceId, host, method, enabled, blocked, createdAt `(`allowanceId` and `host` are the same of the relational allowance)
- [x] extend allowance schema with "permissions" which is an array of permission `method`

### Describe alternatives

**Questions:**

Shouldn´t the permissions do have an `id` or `key` too?
What is the difference between "enabled"  and "blocked"? Shouldn't be one boolean enough to check if this permission is avtive?

### Are you working on this?

Yes

bumi:
what would be the `id` or `key` of a permission? 

re enabled/blocked: 
if there is no permission for a call then we prompt the user.
if a permission is set and "enabled" we execute the request automatically without prompt
if a permission is set and "blocked" we reject the request, do not prompt the user and do not execute. - but we can ignore this for now. As said I just like to avoid data migrations so I thought this could be useful. (=> how can a user give a website to list invoices but reject requesting the balance? -> that would be "list invoices" enabled, but "request balance" is blocked.)

lisabaut:
> what would be the `id` or `key` of a permission?

I meant the unique primary key within the DB table. I added this ID now.
https://dexie.org/docs/API-Reference#quick-reference

> 
> re enabled/blocked: if there is no permission for a call then we prompt the user. if a permission is set and "enabled" we execute the request automatically without prompt if a permission is set and "blocked" we reject the request, do not prompt the user and do not execute. - but we can ignore this for now. As said I just like to avoid data migrations so I thought this could be useful. (=> how can a user give a website to list invoices but reject requesting the balance? -> that would be "list invoices" enabled, but "request balance" is blocked.)

Thanks, I added this explanation to the meta ticket description https://github.com/getAlby/meta/issues/230

bumi:
> I meant the unique primary key within the DB table. I added this ID now.

ah ja, +1 for an ID

PR #1695, opened 2022-11-01
feat(db): add permissions for allowances

lisabaut:

This is the first step of the meta ticket https://github.com/getAlby/meta/issues/230
which extends the DB with a new table "permissions"  
and extends allowances with an "permissions" array.

The DB needs to have a new version for these changes.
See https://dexie.org/docs/Tutorial/Design#database-versioning
and https://dexie.org/docs/Dexie/Dexie.version()

### Notes

The permissions for an allowance are a MultiEntry Index (array).
This comes with some browser limitation.
https://dexie.org/docs/MultiEntry-Index

> Not working in:
>     Internet Explorer 10, 11
>     Non-chromium based Microsoft Edge browsers
>     Safari 8, 9.

**Is that a problem ?**

Closes https://github.com/getAlby/lightning-browser-extension/issues/1692

### Type of change

### How has this been tested?

Used the new permission actions and logged the DB changes
Also added a few basic unit tests.

### Checklist

### DIFF
--- a/src/extension/background-script/actions/permissions/__tests__/delete.test.ts
+++ b/src/extension/background-script/actions/permissions/__tests__/delete.test.ts
@@ -59,10 +59,6 @@ afterEach(() => {
 });
 
 describe("delete permission", () => {
-  afterEach(() => {
-    jest.clearAllMocks();
-  });
-
   test("removes permission method from allowance and deletes permission", async () => {
     const message: MessagePermissionDelete = {
       application: "LBE",
--- a/src/extension/background-script/actions/permissions/__tests__/delete.test.ts
+++ b/src/extension/background-script/actions/permissions/__tests__/delete.test.ts
@@ -1,0 +1,94 @@
+import db from "~/extension/background-script/db";
+import type { DbAllowance, MessagePermissionDelete } from "~/types";
+
+import deletePermission from "../delete";
+
+const mockNow = 1487076708000;
+Date.now = jest.fn(() => mockNow);
+
+const stackerNews = {
+  host: "stacker.news",
+  imageURL: "https://stacker.news/apple-touch-icon.png",
+  name: "Stacker News",
+};
+
+const mockAllowances: DbAllowance[] = [
+  {
+    ...stackerNews,
+    enabled: true,
+    id: 3,
+    lastPaymentAt: 1667291232423,
+    lnurlAuth: true,
+    remainingBudget: 500,
+    totalBudget: 500,
+    createdAt: "1667291216372",
+    tag: "",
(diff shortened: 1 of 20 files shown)


src/extension/background-script/actions/permissions/__tests__/delete.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    ...stackerNews,
+    enabled: true,
+    id: 3,
+    lastPaymentAt: 1667291232423,
+    lnurlAuth: true,
+    remainingBudget: 500,
+    totalBudget: 500,
+    createdAt: "1667291216372",
+    tag: "",
+    permissions: ["the-request-method-1", "the-request-method-2"],
```
bumi, line 26: just in case you are wondering what is a real method:
https://lightning.readthedocs.io/ or https://api.lightning.community/

bumi (commented):
* Why do you store the permissions in a permissions table and in an array in the allowance? 
* when do we need the `get` action?

lisabaut:
> * Why do you store the permissions in a permissions table and in an array in the allowance?
> 
🤔 You designed that yourself? I do not understand the question.
See the [meta ticket ](https://github.com/getAlby/meta/issues/230)again

> This can be a new table (in db.ts) with: allowanceId, host, method, enabled, blocked, createdAt
> and in the action we look up that permission.

>     * when do we need the `get` action?

I don't understand the question either. See your own description in the meta ticket
This is the whole flow of the permissions system we discussed.
1. check host
2. get allowance
3. check if request method is in allowance's permission (string in permissions array)
4. Yes => get the corresponding permission from DB and check if enabled or blocked
5. No => prompt the user

Please tell me what is in your mind. Otherwise I am just guessing why you asked me these questions.

lisabaut:
> * Why do you store the permissions in a permissions table and in an array in the allowance?
> 
>     * when do we need the `get` action?

Or do you mean we do not need to store the permissions in allowance?
😄 You see? I am already guessing again.

bumi:
where does it say that the permissions should be stored in the allowance table and in an array? and with which explanation? 
Was there a reason why you did it like that? 
I think we don't need to store that twice. a new table should be fine and also allows us to easily use the dexie API to query it.

where would you use the get action? 

Let's approach this from the perspective of what we want to achieve and implement that flow that you described and not the individual pieces (that in the end might not fit together)

bumi:
does this help: https://github.com/getAlby/lightning-browser-extension/pull/1707/files#diff-f1d7208ffb952063384e6b24f149fabfeeb17e90094c473b6c49deba9615bba6R29-R49

lisabaut:
Closing as the feature will be implemented in single PR (and not in individual steps).

PR #1713, opened 2022-11-05
chore: unify connector functions

### LINKED ISSUE #1712: Unify connector interface

### Describe the bug

The connector functions either return the response or throw an error. 
Except `connectPeer` and `getInvoices` those return an error. 

I think we should have this unified. @escapedcat why do you not throw an error in those functions like in the other functions? 

https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/connector.interface.ts#L120-L122

### Screenshots [optional]

_No response_

### Steps To Reproduce

..

### Expected behavior

unify all the things

### Alby information

...

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

_No response_

bumi:
We throw an error on the error case

fixes #1712 

### Type of change

### How has this been tested?

needs testing

### DIFF
--- a/src/app/screens/Home/DefaultView/index.tsx
+++ b/src/app/screens/Home/DefaultView/index.tsx
@@ -132,7 +132,16 @@ const DefaultView: FC<Props> = (props) => {
 
   const loadInvoices = async () => {
     setIsLoadingInvoices(true);
-    const result = await api.getInvoices({ isSettled: true });
+    let result;
+    try {
+      result = await api.getInvoices({ isSettled: true });
+    } catch (e) {
+      if (e instanceof Error) {
+        toast.error(e.message);
+      }
+      setIsLoadingInvoices(false);
+      return;
+    }
 
     const invoices: Transaction[] = result.invoices.map((invoice) => ({
       ...invoice,
--- a/src/extension/background-script/actions/ln/invoices.ts
+++ b/src/extension/background-script/actions/ln/invoices.ts
@@ -6,11 +6,8 @@ const invoices = async (message: MessageInvoices) => {
   const isSettled = message.args.isSettled;
 
   const connector = await state.getState().getConnector();
-  const result = await connector.getInvoices();
-
-  if (result instanceof Error) {
-    return { error: result.message };
-  } else {
+  try {
+    const result = await connector.getInvoices();
     const invoices: Invoice[] = result.data.invoices
       .filter((invoice) => (isSettled ? invoice.settled : !invoice.settled))
       .map((invoice: Invoice) => {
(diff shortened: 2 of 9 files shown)


escapedcat (approved):
Thanks! This makes a lot of sence

escapedcat:
Uhm, at one point in time the returned Error was used to trigger and show an error-toast. But this doesn't seem to work anymore (on master) so this change makes the current state better.  
Not sure if we should to let the user know that their connector doesn't support this via a toast.

bumi:
I don't fully get your last comment. what's the suggestion?

escapedcat:
No suggestion. Your PR improves the current situation. This is good.

Question:  
At one point in time the returned Error was used to trigger a toast within the extension to give the user feedback.  
Currently the user doesn't get any feedback in the UI, right? Do we want to give feedback?

bumi:
why does the user not get feedback? 

https://github.com/getAlby/lightning-browser-extension/pull/1713/files#diff-d07eb0245fd06ccaf878d056614cc4a944a918bbdd4267619cf3eb264bdd7eb3 -> here for example: both return error: message in the error case.

escapedcat:
image
Might a bug then? I don't see anything inside the UI currently.

bumi:
hmm. your error is strange. because getInvoices should be implemented on Lndhub

escapedcat:
> hmm. your error is strange. because getInvoices should be implemented on Lndhub

All good, that was just a test by replacing the function

bumi:
hmmm. doesn't seem we do this? https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/DefaultView/index.tsx#L133-L138

escapedcat:
When #1080 was merged it looked like this:

But now that we support most of the important connectors we're good.

bumi:
naja, can you explain that? that PR is from Jul 26. so I don't know what you try to say. 
is this related to this PR? 

> But now that we support most of the important connectors we're good.
we still don't support it in all. and if we broke something we should not say 🤷 we do most ...

escapedcat:
:D So, this is related to this PR because by looking at this I noticed that apparently "showing the toast to the user" does not work anymore. We need to decide we want to show something to the user if their connector doesn't support this yet.

- If yes, close this PR and we open an issue to fix the toast
- If no, merge this PR and keep it the current way

It's a decision of priorization

bumi:
this seems broken in master, no idea when/how it got broken (that could use investigation, because I have no idea how it could have worked before) 

this PR actually fixes this I think for the invoices. at least it should. https://github.com/getAlby/lightning-browser-extension/pull/1713/files#diff-167ef8f8c43cebe91d5e71f3724e953a69277ab2c9b30af8dfc47a0707613544

what do we need to prioritize?

escapedcat (approved):

PR #1714, opened 2022-11-06
chore: remove node version lock

### LINKED ISSUE #1314: Update node version to 18 to stay up-to-date

We're currently using v14 which is maintenance is running out April next year.
See: https://nodejs.org/en/about/releases/

- Update node version to latest v18 in i.e. 
    - `node-version`
    - `.nvmrc`
- Install it
- Test if dev task still work
- Test if build still works

#### Unit-test setup
Looks like some unit-tests fail after an update to 16/18.
This helps partly:
- Updating runcitadel #1280
- `transformIgnorePatterns: ["node_modules/(?!(dexie))/"],`
- Update `jest` (and related)to v29
- In other cases more mocking helps, i.e.:
    ```ts
    jest.mock("~/common/lib/utils");
    jest.mock("~/common/utils/currencyConvert");
    ```

#### Failing npm deps
- https://github.com/getAlby/lightning-browser-extension/pull/1549
    Latest [wext-manifest-loader](https://github.com/abhijithvijayan/wext-manifest-loader) needs node v16.15.0 as a minimum  
    Update will be paused via depfu for now

bumi:
Fixes #1314

we support probably all node versions newer than 14.x I don't think we need to lock the user to a specific version

Since we have removed the runcitadel dependency we now should be able to support all node versions again. 
Can we leave this up to the developer?


escapedcat:
In general I think it's good to indicate which latest version we support for sure. This gives us the change to update that version after we made sure it works and also helps other devs who use these npm-tools.

bumi:
so what is the latest version that we support? 
and do you mean "indicate" or "force" some version (down to the patch version)?

im-adithya:
We even support v19 which is the latest (or v18 if we have to specify the latest _LTS version_)

bumi:
@im-adithya what's your suggestion here? what do you think of this PR?

escapedcat:
[We only have specific version because of nodenv](https://github.com/nodenv/nodenv/issues/53#issuecomment-247510650). You can install a plugin to make that work but now sure how practical that is.

In general I used nvm in my past which is flexible and seems to be well known. nvm accepts also i.e. "LTS", "14", etc.

bumi:
then we add a default major version to a .nvmrc but remove the .node-version?

bumi:
@escapedcat please review

escapedcat:
```sh
node-version > nvm use
Found '/lightning-browser-extension/.nvmrc' with version <lts/*>
Now using node v18.12.1 (npm v8.19.2)
```

Nice, we should also adjust all node version in our Github actions to align this, i.e.: https://github.com/getAlby/lightning-browser-extension/blob/master/.github/workflows/build.yml#L26

bumi:
done

escapedcat (approved):

PR #1715, opened 2022-11-07
fix: properly encode the LNURLp payerdata JSON

bumi:
Maybe this worked because the old axios version did that automatically (which was wrong)

I think the updated axios version #1702 behaves differently. 
We had passed in the query params here wrong and did not encode the JSON manually. 
Thus the payerdata was sent wrong to the server

### Type of change

### How has this been tested?

send a payment to a LN address that supports payerdata.


src/app/screens/LNURLPay/index.tsx
```diff
@@ -106,8 +106,8 @@ function LNURLPay() {
       // Get the invoice
       const params = {
         amount: parseInt(valueSat) * 1000, // user specified sum in MilliSatoshi
-        comment, // https://github.com/fiatjaf/lnurl-rfc/blob/luds/12.md
-        payerdata, // https://github.com/fiatjaf/lnurl-rfc/blob/luds/18.md
+        comment: comment ? comment : null, // https://github.com/fiatjaf/lnurl-rfc/blob/luds/12.md
+        payerdata: payerdata ? JSON.stringify(payerdata) : null, // https://github.com/fiatjaf/lnurl-rfc/blob/luds/18.md
```
escapedcat, line 110: You might be able to write it like this:
```js
      const params = {
        amount: parseInt(valueSat) * 1000, // user specified sum in MilliSatoshi
        ...(comment && { comment }), // https://github.com/fiatjaf/lnurl-rfc/blob/luds/12.md
        ...(payerdata && { payerdata: JSON.stringify(payerdata) }), // https://github.com/fiatjaf/lnurl-rfc/blob/luds/18.md
      };
```
Can `params` be typed?

bumi: how, where, what? 
those params are only used in that call. why/how would I type it?

escapedcat: > why/how would I type it?

To indicate what's required and what's optional for example. To make it easier for other devs. Just a thought. No need block this.

bumi: yes, sure. but how/why would that work here? 

I am passing an amount, comment and payerdata... do you have code?

escapedcat: Simple way might be just this?:

```ts
      const params: {
        amount: number;
        comment?: string;
        payerdata?: string;
      } = {
        amount: parseInt(valueSat) * 1000, // user specified sum in MilliSatoshi
        ...(comment && { comment }), // https://github.com/fiatjaf/lnurl-rfc/blob/luds/12.md
        ...(payerdata && { payerdata: JSON.stringify(payerdata) }), // https://github.com/fiatjaf/lnurl-rfc/blob/luds/18.md
      };
```

bumi: what does this do?
for me that's writing the same thing twice.

escapedcat:
```ts
        amount: number;
        comment?: string;
        payerdata?: string;
```

You mean this? It's an "inline" type indicating what `params` accepts, how this needs to look like, what's required and what's optional.

Issue #1696, opened 2022-11-01
Extend WebLN with generic request method


lisabaut:
### Feature description

This is part of the meta ticket https://github.com/getAlby/meta/issues/230

To allow web apps to call other functionality on their connector which the WebLN API does not cover explicitly
the idea is to extend WebLN with a generic "request" method.

### Describe the solution

The generic method takes the two arguments ("method: string", "params?: Record<string, any>").

#### Example

```TS
await webln.enable();
const channels = await webln.request("listChannels", {channelId: string, otherKey: string}); 
```

See similar Pull Requests:
- https://github.com/getAlby/lightning-browser-extension/pull/1629
- https://github.com/getAlby/lightning-browser-extension/pull/1694

### Describe alternatives

See more information in the meta ticket:
- https://github.com/getAlby/meta/issues/230

### Are you working on this?

Yes

bumi:
https://replit.com/@bumi17/webln-channels-demo

PR #1724, opened 2022-11-07
feat(webln): generic request method and permissions

lisabaut:

This PR does:
- extend DB with a new table "permissions"  
- extend webln with a method "request" which checks first if a permission exists for the request method
- => YES, immediately send request
- => NO, prompt user, check if user enabled the Checkbox "Always Allow" and save this method in the DB, send request
- add a new screen "ConfirmRequestPermission"

### Follow Up
- https://github.com/getAlby/lightning-browser-extension/issues/1725
- https://github.com/getAlby/lightning-browser-extension/issues/1726

Closes https://github.com/getAlby/meta/issues/230
Closes https://github.com/getAlby/lightning-browser-extension/issues/1692
Closes https://github.com/getAlby/lightning-browser-extension/issues/1696

### Type of change

pp

### How has this been tested?

- added a few unit tests
- used browser console for window.webln
- used ripl https://replit.com/@bumi17/webln-channels-demo#src/App.js

### Checklist


src/i18n/locales/en/translation.json
```diff
@@ -594,6 +594,12 @@
         "pay_now": "Pay now"
       }
     },
+    "confirm_request_permission": {
+      "title": "Approve Request",
+      "allow": "Allow {{host}} to:",
+      "enable_method": "Send request {{method}}",
```
reneaaron, line 600: Is `Send request` the right wording here? @bumi 
Maybe `Call` is better? I don't like the word send` in there because sending in this context is associated with sending money. (`send request `sendPayment`)

reneaaron (changes requested):
Nice work @lisabaut! I still want to this PR, but added a first round of feedback.

src/extension/background-script/actions/ln/request.ts
```diff
@@ -0,0 +1,82 @@
...
+  const { origin, args } = message;
+
+  try {
+    if (!connector.requestMethod) {
+      throw new Error(
+        `${message.args.method} is not supported by your account`
+      );
+    }
+
+    const allowance = await db.allowances
```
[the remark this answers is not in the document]
lisabaut: > Why do we check for the allowance here again? This check seems to be duplicate to the general `enabled` checks that happen before even getting into this method?

This is a TypeGuard for Typescript in the first place, because `allowance.id` is needed later to save the permission in the DB (and can be undefined). 
I will keep it for now, as this check doesn't hurt either.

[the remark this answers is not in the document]
lisabaut: > If a user currently deletes an allowance, what happens with the permissions? Should they they be deleted as well? 
> I'd expect to be prompted for permissions again after I deleted an allowance.

Yes, I asked this question is the former PR [here](https://github.com/getAlby/lightning-browser-extension/pull/1695/files#diff-536b7f7cb4ac5b988af6ca70ed282bed5b563cb942b9d242da9f19061ae84d79R10-R12) too 
which got lost on its way to here. 
Thanks for bringing this up again.

I added the corresponding deletion now.

src/extension/ln/webln/index.ts
```diff
@@ -90,6 +90,17 @@ export default class WebLNProvider {
     throw new Error("Alby does not support `verifyMessage`");
   }
 
+  request(method: string, params: Record<string, unknown>) {
+    if (!this.enabled) {
+      throw new Error("Provider must be enabled before calling verifyMessage");
```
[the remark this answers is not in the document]
lisabaut: outdated

src/i18n/locales/en/translation.json
lisabaut: Changed string to "call". 
And removed `{{method}}` from the "always allow" string as there is no string mapping yet. 
A mapping can be only relevant when this follow-up ticket is addressed: https://github.com/getAlby/lightning-browser-extension/issues/1725

Screenshot is updated too.

src/extension/background-script/actions/ln/request.ts
```diff
@@ -0,0 +1,82 @@
...
+      return { error: "Could not find an allowance for this host" };
+    }
+
+    const permission = await db.permissions
+      .where("host")
+      .equalsIgnoreCase(origin.host)
+      .and((p) => p.method === args.method)
+      .first();
+
+    if (permission && permission.enabled) {
```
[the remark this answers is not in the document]
lisabaut: Yesy, right now there is either a permission saved in the DB or there is none.
Which makes the booleans "enabled"  and "blocked"  obsolete at the moment.

As @bumi said, this was to think ahead because removing entries from the DB is easier than removing when migrating.

When this follow-up ticket is addressed: https://github.com/getAlby/lightning-browser-extension/issues/1726
=> it will take another concepting again what the user can do with the permission (change "enabled" to false? Do something with "blocked"? Or only delete a permission?)

src/extension/background-script/actions/ln/request.ts
```diff
@@ -0,0 +1,82 @@
...
+      const response = await connector.requestMethod(args.method, args.params);
```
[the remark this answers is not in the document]
lisabaut: Not sure how this will improve the readability?

Right now, there is an early exit if the request method can be called immediately.
If not, do this and that (long code block).

src/extension/background-script/actions/ln/request.ts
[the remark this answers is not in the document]
reneaaron: Thanks for clarifying. 👍

src/extension/background-script/actions/ln/request.ts
```diff
@@ -0,0 +1,83 @@
...
+        enabled: boolean;
+        blocked: boolean;
+      }>({
+        args: {
+          requestPermission: {
+            method: args.method,
+          },
+        },
+        origin,
+        action: "public/webln/confirmRequestPermission",
```
reneaaron, line 51: Should this be in `webln` at all? 

Would that mean that if we want to use this popup for nostr we would need to duplicate all the code here? (for storing the permission, etc) Does it make sense to have the different actions implement this or could this be handled by the prompt in the background?

src/extension/background-script/actions/ln/request.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      const response = await connector.requestMethod(args.method, args.params);
+
+      // add permission to db only if user decided to always allow this request
+      if (promptResponse.data.enabled) {
+        const permissionIsAdded = await db.permissions.add({
+          createdAt: Date.now().toString(),
+          allowanceId: allowance.id,
+          host: origin.host,
+          method: args.method,
```
reneaaron, line 62: Should we prefix that with "webln.request"? Otherwise they could potentially collide. (e.g. nostr uses "getPublicKey" which is also available for WebLN maybe in future)

lisabaut: I added the prefix.

src/extension/background-script/actions/ln/request.ts
lisabaut: No strong opinion on the path to the prompt.
I removed "webln" now.

lisabaut: > Would that mean that if we want to use this popup for nostr we would need to duplicate all the code here? (for storing the permission, etc) Does it make sense to have the different actions implement this or could this be handled by the prompt in the background?

My initial thinking was that all the checks are done in the prompt.
This was my start in the other PR in which Bumi then pointed me to code where all is handled in the background script [in this comment](https://github.com/getAlby/lightning-browser-extension/pull/1695#issuecomment-1303298272)

Not sure about the future Nostr integration, but if this would be then more or less the same as for the webln request it would require a re-thinking, yes. 
Either extract the handling into another function which can be re-imported again for Nostr. 
Or move the checks to the prompt.
I would suggest discussing this with Bumi again and open a follow-up PR for this.

reneaaron: I guess this is exactly the reason why @bumi wanted to have practical examples implemented within this PR as only with that we are learning about how they actually will be used by different consumers and which parts of the permission system make sense to be shared between those. 

>  I would suggest discussing this with Bumi again and open a follow-up PR for this.

👍

PR #1735, opened 2022-11-08
chore: remove storybook

bumi:
we currently do not maintain it :(

related: #1711


README.md
```diff
@@ -130,8 +130,7 @@ If this is not reachable please let us know.
 
 #### Storybook.js
 
-We have a working [Storybook](https://storybook.js.org)-setup and some components have stories.
-You can find the deployed Storybook here: https://lbe-stories.netlify.app
+We have a [Storybook](https://storybook.js.org)-setup and some components have stories. But we haven't actively maintained the storybook and we're looking for support to work on the storybook setup.
```
escapedcat, line 133: Are we though? Who would  be using it?

bumi: generally storybook and reusable components are a good idea. having a design system and a library of components is nice, isn't it?

escapedcat: Sure, but we remove Storybook because nobody is using it and we do not want to maintain it. This sounds like an invite to help but for what reason? To end up in the same situation after someone created a PR to re-add it?  

I would be all for leaving this if there's a bunch of people using it and we would like some support to maintain it. But currently I don't see this.

You know what I mean?

bumi: I don't care about that text, change it if you want.

escapedcat:
- removing storybook also removed our lodash types
- instead of continuing adding all of lodash and all types to the project I switched to individual packages

PR #1739, opened 2022-11-09
build(jest): compile lnmessage to avoid module issue

escapedcat:


jest.config.js
```diff
@@ -23,10 +23,11 @@ module.exports = {
   transform: {
     "^.+\\.(t|j)sx?$": ["@swc/jest", swcConfig],
   },
+  transformIgnorePatterns: ["node_modules/(?!(lnmessage))/"],
```
bumi, line 26: "An array of regexp pattern strings that are matched against all source file paths before transformation. If the file path matches any of the patterns, it will not be transformed."

what means transformation here? and why should it not be transformed?

Issue #1725, opened 2022-11-08
Extend LND connector with requestMethod


lisabaut:
### Feature description

With this pull request https://github.com/getAlby/lightning-browser-extension/pull/1724
it is possible to extend a connector with the method `requestMethod` which enables any Client to pass methods to the Connector's API other than explicitly provided by WebLN.

Right now it is only possible to extend the LND connector as a detailed REST API exists which can be easily integrated.
=> see their docs https://api.lightning.community/#lnd-rest-api-reference

## EXAMPLE

1. Extend the LND connector with

```TS
 async requestMethod(
    method: string,
    params?: Record<string, unknown>
  ): Promise<{ data: unknown }> {
    return this.request("GET", `/v1/${method}`, params, {}).then((response) => {
      return { data: response };
    });
  }
```

2. Connect with your LND account
3. In browser console type
```TS
 await window.webln.enable();
 await window.webln.request("getinfo");
```
4. Log the response

## Problem

From the code example above it would be only possible to send a GET request to the version1 of the lightning API.
The idea is to support not only GET but also POST requests and support different API versions.

## TODO

- [ ] go over the API docs and make a list of the first methods which should definitely be supported by WebLN
- [ ] discuss with the team this list
- [ ] in the connector create a mapping for these methods (see example code below)

### EXAMPLE

```ts
const apiMethodMap = {
  getinfo: {
    method: "GET",
    version: "v1"
  },
  "chainnotifier/register/blocks": {
    method: "POST",
    version: "v2"
  },
  .....
}

 // ....

 async requestMethod(
    method: string,
    params?: Record<string, unknown>
  ): Promise<{ data: unknown }> {
    return this.request(apiMethodMap[method].method, `/${apiMethodMap[method].version}/${method}`, params, {}).then((response) => {
      return { data: response };
    });
  }
```

### Describe alternatives

- Support only GET request for the first implementation?
- Support only API v1 for the first implementation?

### Additional context

Related to ticket https://github.com/getAlby/lightning-browser-extension/issues/1717

PR #1752, opened 2022-11-14
feat: add request method functionality to lnd and commando

bumi:
This implements the requestMethod for the lnd and commando connector exposing certain functions to webln.
This allows web apps to do much more with lighthing than payments.

issue: #1725


kiwiidb:
We still have this issue of the error not being propageted back to the website eg. if `experimental-dual-fund` is not enabled.

reneaaron:
> We still have this issue of the error not being propagated back to the website eg. if `experimental-dual-fund` is not enabled.

Is that something outside of our control? Can we catch those errors somehow?

reneaaron:
Simplified the permission dialog for a bit: 
 - Remove "host" (has been in this screen 3 times, now 1 time)
 - Unified "Remember my choice and don't ask again"
 - Minor style fixes

image

reneaaron (approved):
Just tested everything with LND and worked out fine so far👌 

 - added translations for permissions for commando as well as LND
 - minor adjustments to permission screen

src/extension/background-script/connectors/commando.ts
```diff
@@ -107,6 +139,32 @@ export default class Commando implements Connector {
     await this.ln.disconnect();
   }
 
+  get supportedMethods() {
+    return supportedMethods;
+  }
+
+  methodDescription(method: string) {
+    return `commando.${method}`;
```
bumi, line 147: as we now have the description in the translation file and only use the method here I am wondering if we still need this method or if we move this convention to the `request` action. 

we can set the `description` there to ``${connector.constructor.name.toLowerCase()}.${method}`` 
then we don't need the `methodDescription` function in the connector. 

we only need to check if there is a translation for that - though as long as we have a whitelist we can make it a requirement to have a translation. 
it only will no longer work if we allow everything (and not only whitelisted ones) - though I'd for now still stick with this whitelist.

reneaaron: Agree. Will refactor this part 👍

PR #1754, opened 2022-11-14
feat!: Update Citadel SDK

AaronDewes:

this updates the Citadel SDK to the next version, which

- Works in tests again
- Doesn't have a hard dependency on Node 18
- Supports the latest API (Citadel 0.1.4 will make breaking API changes)

### Type of change

### How has this been tested?

Connected to my local node.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,9 @@
   },
   "dependencies": {
     "@bitcoin-design/bitcoin-icons-react": "^0.1.9",
-    "@noble/secp256k1": "^1.7.0",
     "@headlessui/react": "^1.7.4",
+    "@noble/secp256k1": "^1.7.0",
+    "@runcitadel/sdk-next": "^0.1.2",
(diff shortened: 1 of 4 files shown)


bumi:
Hey @AaronDewes thanks for the PR. 
what are your thoughts there? 
We had some issues with the hard dependencies in the npm package because we are less strict and have different runtimes environments (webpack browser + jest). 
also I try to reduce the external dependencies as much as possible, implementing the Citadel API seemed easy enough to do directly in the connector. 

What changes will you make? Is Citadel always recommending to use the SDK?

AaronDewes:
The API changes are only minor (`/manager-api/v1` routes moved to `/api-v2/v2`). Generally, the SDK allows me to make API changes and get them rolled out faster, but I'll recreate this PR without the SDK. I assumed the reason for the removal of the SDK was only the issues in the SDK, which are fixed now. But if you prefer it this way, I'll create a cleaner PR.

PR #1756, opened 2022-11-14
Refactor/axios

escapedcat:
prepare vor MV3
- use fetch adapter for axios

TODO:

- ~fix unit-tests~

Context:  
- Axios' default adpater doesn't work in SW. Switching to a fetch-adapter we can keep the advantages of Axios
- Currently this adpater only work in Axios v0.x not in v1.x
- I'll keep checking the reo and also the new upcoming "official" fetch adapter PR.


bumi:
can we make a PR also with the axios fetch adapter PR? (and thus also show what fails in that PR?
as this seems to get into axios it seems to me a bit more promising?

escapedcat:
> can we make a PR also with the axios fetch adapter PR? (and thus also show what fails in that PR? as this seems to get into axios it seems to me a bit more promising?

Can do but to unblock MV3 can we merge this first?

bumi:
hm, no need to close this, just because I think it is worth exploring that official(?) axios PR.

bumi:
I thought that's an alternative to it?

escapedcat:
> I thought that's an alternative to it?

It would be if the Axios one would be merged & released. For now we're better with this approach, I think. At least this works.

bumi:
the PR sounded promising to me. The maintainer said it looks good and it was actively worked on.

escapedcat:
> the PR sounded promising to me. The maintainer said it looks good and it was actively worked on.

Sure but it isn't ready and it not related to this PR. Or do you want to switch to that fork in this PR here? Not sure what your expectation is I guess.

bumi:
Why is it not related to this PR? This PR changes axios to use a fetch adapter by adding another dependency. 
An alternative to this is to use the branch from axios which has support signaled to get merged. 
so it's the same just the axios branch seems to become the official way of doing this.
(maybe you want to help me with my replit to show how it is used and then we can also help getting this in the offical axios lib?)

escapedcat:
Findings here: https://github.com/getAlby/lightning-browser-extension/pull/1825  
It might work if we use a build version of the forked branch. Not sure.

So far I would go with this PR in the current state and switch to the "official" axios adapter as soon as it's merged.

im-adithya:
Closing this as we use fetchAdapter

Issue #1022, opened 2022-06-10
Format displayed SATS amounts with seperators


escapedcat:
### Is your feature request related to a problem?

Long numbers are hard to read, i.e.: `123456789`  

### Describe the solution you'd like

![image](https://user-images.githubusercontent.com/1016218/172995800-e1b78de7-a624-4365-b1a0-702b4d86eaf8.png)

Would be nice to format in a localized way, i.e. `123,456,789`
Possible solution: [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat#basic_usage)

### Describe alternatives you've considered

none

### Additional context

Related to: #1000 

### Are you working on this?

No

secondl1ght:
Is this still blocked?

escapedcat:
No, tried to indicate it by removing the "later" label. Made this more clear now.

Issue #1742, opened 2022-11-10
💵 Explore fiat-denominated accounts


reneaaron:
### Feature description

More and more services are offering fiat denominated accounts via synthetic assets:
 - [kollider](https://kollider.xyz)
 - [galoy](https://galoy.io/)

![image](https://user-images.githubusercontent.com/100827540/201083946-13680767-6821-480d-bd6e-a6ab02d1e452.png)

Those services often offer multiple balances to their users. In Kollider's case it's: 
 - Bitcoin
 - synthetic USD 
 - synthetic EUR

We want to explore how this can be integrated into Alby by creating a connector for Kollider. (see WIP branch linked to this issue)

### Context
When connecting a Kollider account () a user would need to choose which balance should be used within Alby, it's currently out of scope to support multiple balances.

A connector would need to signal the current currency to the extension.

### Questions
 - What changes are needed in order for Alby to work with fiat-denominated accounts?
   - Receive & send screens
   - Balance in Account selector
   - Fiat currency conversion in transaction lists
   - Fiat inputs

### Useful links
 - [Kollider LNDHubx](https://lndhubx.kollider.xyz/)

### Out of scope
 - Having multiple balances in one account

### ToDo

- [ ] Link to registration page (where the user gets the credentials)

bumi:
In the first step what we imo need to look at is how we can let the connector decide what currency the return amounts have. 
E.g. when we show the account balance we have a hard coded "sats" there. this would need to be the currency of the connector. 
Same in the incoming invoices. 
I'd also say if the currency is not btc/sats then we do not show any other fiat amount (or we show the sats amount, whatever is easier) 

in the outgoing payments list I'd still try to store sats amounts.

lisabaut:
## First Implementation

- [ ] Create new connector "Kollider"
- [ ] add new connector to Websites screen
- [ ] when user creates Kollider => prompt user with "Which balance/account you want to add"
- [ ] either store this base currency in Settings (TBD)
- [ ] and/or add method to connector "getBaseCurrency" (TBD)
- [ ] Decision: No convertion will happen at any place, just do not show fiat at all

bumi:
just for completeness: 
Here is the code where I explored Kollider: https://github.com/getAlby/lightning-browser-extension/pull/1770 

there does not need to be a special prompt, just like other connectors need some information this needs a "currency": 
[...]

PR #1757, opened 2022-11-14
feat: format sats and fiat with current locale

lisabaut:

This PR does:
- passes current locale to `getFormattedFiat` 
- passes current locale to `getFormattedNumber` 
- passes current locale to `getFormattedSats` 
- moves these helpers to SettingsContext
- renders components in tests with SettingsProvider where possible to test the correct number formatting
- add jest global matchMedia config to render tests with SettingsProvider

**Please follow the commit history if you have problems comparing the diff.**

Please checkout this PR yourself and check if I missed a spot. 🙏 

## NOTEs

~I am still thinking of moving `getFormattedNumber` and `getFormattedSats` to the SettingsContext.
(Same as `getFormattedFiat`) **WDYT?**~
=> done ✅ 

Closes https://github.com/getAlby/lightning-browser-extension/issues/1022

### Type of change

fomatted

### How has this been tested?
- added more tests
- checked UI 

### Checklist

### DIFF
--- a/src/app/screens/ConfirmPayment/index.tsx
+++ b/src/app/screens/ConfirmPayment/index.tsx
@@ -17,13 +17,13 @@ import { useNavigationState } from "~/app/hooks/useNavigationState";
 import { USER_REJECTED_ERROR } from "~/common/constants";
 import msg from "~/common/lib/msg";
 import utils from "~/common/lib/utils";
-import { getFormattedSats } from "~/common/utils/currencyConvert";
 
 function ConfirmPayment() {
   const {
     isLoading: isLoadingSettings,
     settings,
     getFormattedFiat,
+    getFormattedSats,
   } = useSettings();
 
   const showFiat = !isLoadingSettings && settings.showFiat;
@@ -49,10 +49,7 @@ function ConfirmPayment() {
   const [fiatAmount, setFiatAmount] = useState("");
   const [fiatBudgetAmount, setFiatBudgetAmount] = useState("");
 
-  const formattedInvoiceSats = getFormattedSats({
-    amount: invoice.satoshis || 0,
-    locale: settings.locale,
-  });
+  const formattedInvoiceSats = getFormattedSats(invoice.satoshis || 0);
 
   useEffect(() => {
     (async () => {
--- a/src/app/screens/ConfirmPayment/index.tsx
+++ b/src/app/screens/ConfirmPayment/index.tsx
@@ -23,7 +23,7 @@ function ConfirmPayment() {
   const {
     isLoading: isLoadingSettings,
     settings,
-    getFiatValue,
+    getFormattedFiat,
   } = useSettings();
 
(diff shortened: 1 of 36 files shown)


escapedcat:
ack

> I am still thinking of moving getFormattedNumber and getFormattedSats to the SettingsContext.

Curious about the reason. To have all this in the same place. No need to have it in a currency-file?

Just for my understanding:
- Currency is formatted according to the locale-money formatting
- Sats are formatted to the locale-number formatting

Yes? Asking mainly because for a moment I was confused why both amounts are differently formatted. But I guess that makes sense because sats is just a number?

lisabaut:
> Curious about the reason. To have all this in the same place. No need to have it in a currency-file?
> 
The actual function would still be in the currency-file.
But the component import it from the context.
Therefore, the components do not need to know about the locale no more.
Same implementation as for getFiat => https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L8


> Just for my understanding:
> 
>     * Currency is formatted according to the locale-money formatting
> 
>     * Sats are formatted to the locale-number formatting
> 
> 
> Yes? Asking mainly because for a moment I was confused why both amounts are differently formatted. But I guess that makes sense because sats is just a number?

This is how INTL formats numbers with or without floats according to the language:
- 3190505 => 3,190,505 (Swedish => 3 190 505), this is a number without decimals
- 910.10 => 910.10 (Swedish 910,10), a number with decimals

Maybe the commas for big numbers confused you? 
UK, USA ... they use commas to separate large number (3 digits) and a point for decimals.
In most Europe countries it is the other way round.
Nice read: https://www.languageediting.com/number-formatting-europe-vs-us/

escapedcat (approved):

lisabaut:
> 

As discussed:

I moved now all formatting helpers to SettingsContext.
And adjusted tests to render with SettingsProvider to test that the formatting still works.

src/app/screens/ConfirmPayment/index.tsx
```diff
@@ -143,7 +149,7 @@ function ConfirmPayment() {
             <div className="my-4">
               <div className="mb-4 p-4 shadow bg-white dark:bg-surface-02dp rounded-lg">
                 <PaymentSummary
-                  amount={invoice.satoshis}
+                  amount={invoice.satoshis || "0"} // how come that sathoshis can be undefined, bolt11?
```
bumi, line 152: likely those are zero-amount invoices? which bolt11 invoice did you use?

bumi (commented):
right now we assume the balances are in sats, so we can do `getFormattedFiat(balance)` 
What do we need to change when we work on the task to allow the account/connector decide which currency it is in?

is the settings context then still the right place?

We probably then check in the account the currency of the account and then decide which formatting function we call to display it?

bumi:
Re fiat amounts: https://github.com/getAlby/lightning-browser-extension/issues/1742#issuecomment-1310269008 
we tackle that next.

PR #1759, opened 2022-11-16
fix(battery): youtube live stream

escapedcat:
apply twitter solution to YT battery

### DIFF
--- a/src/extension/content-script/batteries/YouTubeVideo.ts
+++ b/src/extension/content-script/batteries/YouTubeVideo.ts
@@ -2,9 +2,15 @@ import getOriginData from "../originData";
 import { findLnurlFromYouTubeAboutPage } from "./YouTubeChannel";
 import { findLightningAddressInText, setLightningData } from "./helpers";
 
+declare global {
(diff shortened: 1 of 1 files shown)


bumi:
Why is this needed? Is this only for the live stream videos, why?

I am actually not sure if this is not too much and is it worth it?

escapedcat:
Dunno, this came up during the conference live-stream. @reneaaron asked me to have a look, so I did. This seems to solve it. Not sure how much this is needed. Was just a first draft to test it.

bumi:
hmm, ok, then I'd say it should be checked if it can be done differently. 
I'd assume that the DOM is just slightly different on a live stream? or why did you add the observer that fires whenever the DOM changes?

escapedcat:
It's not different. Somehow the live-stream makes the dom "slower" 90% of cases.  
In chromoum (which does not support the live-streams out of the box Alby always was finding the address and turned blue.  
In chrome it only sometimes found it or only after clicking the extension-icon.  
Applying Dylans twitter code helped with this.

PR #1760, opened 2022-11-17
Allow _ in Lightning addresses

AaronDewes:
This allows things like example_subdomain.example.com to be used as LN address.


src/common/lib/lnurl.ts
```diff
@@ -16,7 +16,7 @@ const fromInternetIdentifier = (address: string) => {
   // email regex: https://emailregex.com/
```
escapedcat, line 16: I guess the comment should be updated as well then to reflect the improvement?

reneaaron (approved):
utACK

escapedcat (approved):
Thanks @AaronDewes !

src/common/lib/lnurl.ts
```diff
@@ -14,9 +14,10 @@ import { bech32Decode } from "../utils/helpers";
 
 const fromInternetIdentifier = (address: string) => {
   // email regex: https://emailregex.com/
+  // modified to allow _ in subdomains
   if (
     address.match(
-      /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+      /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-_0-9]+\.)+[a-zA-Z]{2,}))$/
```
bumi, line 20: Is a `_` in the spec for domains? I thought that's only letters, digits, and hyphen?

AaronDewes: It's allowed in subdomains, I think it may not be allowed in second level domains though 🤷‍♂️

bumi: that regex is too complicated :) 
image

we now match also `_` in the domain. but that's fine :)
thanks for the PR.

PR #1773, opened 2022-11-21
fix(home): handle undefined settle date #1403

### LINKED ISSUE #1403: Transaction appear as received 53 years ago
### Describe the bug

- Transaction appear as received 53 years ago
- Node used: Umbrel

escapedcat:

Handle empty `settleDate`

### Link this PR to an issue

Fixes #1403

### Type of change

### How has this been tested?

Manually

### DIFF
--- a/src/app/screens/Home/DefaultView/index.tsx
+++ b/src/app/screens/Home/DefaultView/index.tsx
@@ -147,7 +147,7 @@ const DefaultView: FC<Props> = (props) => {
       ...invoice,
       title: invoice.memo,
       description: invoice.memo,
-      date: dayjs(invoice.settleDate).fromNow(),
+      date: invoice.settleDate ? dayjs(invoice.settleDate).fromNow() : "",
     }));
 
     for (const invoice of invoices) {


bumi:
I assume this problem is in all the places where we show the date? or is this only on the home screen?

bumi:
Thinking about it I think that date can actually not be 0 and the error is somewhere else. 
can you explain why it is blank and why we need this check?

escapedcat:
> I assume this problem is in all the places where we show the date? or is this only on the home screen?

It's the only place incoming txs show up. I thought about it but I don't think we need to handle this for outgoing txs, do we?

> Thinking about it I think that date can actually not be 0 and the error is somewhere else.
can you explain why it is blank and why we need this check?

See slack thread.

kiwiidb:
It seems a good idea to me to just not show the settle date if the settle date is empty or 0. If LND or Kollider return a settle date 0 somewhere then that is their issue, but we should just hide it if we get bogus data. And it  also seems to me that the error is exactly here and not somewhere else.

bumi:
I don't approve it, yet. :) I still would like to know when this happens and why. And I am cautious saying that LND returns invalid settle_date responses. I did not have time to look into that.

PR #1775, opened 2022-11-22
feat(nostr): support permission for signMessage #1727

### LINKED ISSUE #1727: Nostr: Read and write request permission from DB
### Feature description

This pull request https://github.com/getAlby/lightning-browser-extension/pull/1724
extended the DB with the table "permissions".

Now, it is possible to get, add, delete a permission for a specific host/allowance.
The integration of Nostr enables two methods at the moment:
- ConfirmSignMessage
- ConfirmGetPublicKey

Both methods can now be saved in the DB (if the user decides to save them in order to be not asked again).

### TODO
- [ ] extend ConfirmSignMessage with checking first if the permission for this request exists, save this permission otherwise
- [ ] do the same for ConfirmGetPublicKey
- [ ] decide if this flow should happen within the Prompt or within the background script

### Describe alternatives

...

### Additional context

_No response_

### Are you working on this?

_No response_

escapedcat:

Currently this is PoC on how to use a `signMessage`-permission to make posts on i.e. [branle](https://branle.netlify.app) without the need to confirm every post via Alby.

- we need to enable screen to make sure we always unlock the extension when needed, especially when permission is already stored but extension locked
- using an enum for permissions, need to see if the place out well with future connector-permissions

Fixes #1727 (partly)

### Type of change

### How has this been tested?

Manually


reneaaron (commented):
The connect screen currently explicitly mentions lightning, should we maybe have a more general "connect" dialog? 

image

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +23,60 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
 
+  const enableMessage: MessageAllowanceEnable = {
+    origin: message.origin,
+    args: {
+      host: message.origin.host,
+    },
+    action: "public/webln/enable",
+  };
+
+  await enable(enableMessage);
```
bumi, line 34: can we move this into the content script? 
and call in [the nostr provider](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/ln/nostr/index.ts#L6-L16) `enable()` before the actual call? 

then we would not need to do this "hack" to use the action in another action and such. 
The flow would then be the same as in webln.

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +23,60 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+
+  const dbPermissions = await db.permissions
+    .where("host")
+    .equalsIgnoreCase(message.origin.host)
+    .toArray();
+
+  const hasPermission = dbPermissions
+    .filter((permission) => permission.enabled)
+    .map(({ method }) => method)
```
bumi, line 43: I think we should use the dexy API for this. there should be a method to query for entries where enabled=true and method=x

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +23,60 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+        ...message,
+        action: "public/nostr/confirmSignMessage",
+      });
+
+      if (!response.data.confirm) {
+        throw new Error("User rejected");
+      }
+
+      if (response.data.rememberPermission) {
+        const matchingAllowance = await db.allowances
```
bumi, line 61: as we now require those calls to have an allowance I'd move this to the permission check and also move the TS  guard there if the allowance is not there as Lisa did in the request action I think.

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +23,60 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+          .where("host")
+          .equalsIgnoreCase(message.origin.host)
+          .first();
+        const allowanceId = matchingAllowance?.id;
+
+        if (!allowanceId) return; // type-guard only
+
+        const permission: DbPermission = {
+          createdAt: Date.now().toString(),
```
bumi, line 70: why do we have those as string again? this sucks because we always need to do useless typecastings actually.

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +23,60 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+          allowanceId,
+          host: message.origin.host,
+          method: "signMessage",
```
bumi, line 73: I think this is wrong, this mus be scoped to nostr/ as we also scope the webln calls. 
cc @reneaaron

src/types.ts
```diff
@@ -497,21 +497,20 @@ export interface DbPayment {
...
+  id: number;
+}
+
 export interface DbPermission {
   id?: number;
   createdAt: string;
   allowanceId: number;
   host: string;
-  method: string;
+  method: "signMessage";
```
bumi, line 509: why is this a fixed value? there are different string values in that.

src/app/screens/Nostr/ConfirmSignMessage.tsx
```diff
@@ -73,6 +75,23 @@ function ConfirmSignMessage() {
...
+                onChange={(event) => {
+                  setRememberPermission(event.target.checked);
+                }}
+              />
+              <label
+                htmlFor="remember_permission"
+                className="cursor-pointer ml-2 block text-sm text-gray-900 font-medium dark:text-white"
+              >
+                {/* {t("remember.label")} */}
+                Remember this decission
```
bumi, line 92: why no i18n?

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: Done that, I hope, have a look please

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: Not such good API, but adjusted to what I could find, i.e.:
- https://stackoverflow.com/questions/35679590/dexiejs-indexeddb-chain-multiple-where-clauses
- https://groups.google.com/g/dexiejs/c/G3_W5PssCGA

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: Done, much better!

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: Happy to have a look after this PR

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: I think we should create types for these and combine those with i.e. this: https://github.com/getAlby/lightning-browser-extension/pull/1752/files#diff-82b11e63afd4e5a1aeac4c0023524f3d17f0aafecac1a8017236bb8f210082d0R82
Mentioned this in the PR description

src/types.ts
escapedcat: Yes, see comment above, should be types.

src/app/screens/Nostr/ConfirmSignMessage.tsx
escapedcat: Just a draft, will fix.

escapedcat:
> The connect screen currently explicitly mentions lightning, should we maybe have a more general "connect" dialog?
> 
> image

What could be a general replacement for the two static request items?  
I.e. "Allow [host] to connect to Alby"?

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: We should type this, wdyt?: https://github.com/getAlby/lightning-browser-extension/blob/feat/1727_nostr-permission-sign-message/src/extension/background-script/actions/ln/request.ts/#L24

src/types.ts
bumi: then why not do it directly? list them all of leave a "string" 
otherwise it will be missed or just takes ages.

src/extension/ln/nostr/index.ts
```diff
@@ -3,11 +3,19 @@ import { Event } from "./types";
...
+    }>("enable");
+  }
+
   async getPublicKey(): Promise<string> {
     return await this.execute("getPublicKeyOrPrompt");
   }
 
-  async signEvent(event: Event): Promise<Event> {
+  async signEvent(event: Event): Promise<Event | boolean> {
+    await this.enable();
```
bumi, line 18: guess you'd need this now in all the other calls also.

src/app/screens/Nostr/ConfirmSignMessage.tsx
```diff
@@ -27,13 +28,25 @@ function ConfirmSignMessage() {
...
+          host: origin.host,
+          method: "signMessage",
+          enabled: true,
+          blocked: false,
+        });
+      }
+
       msg.reply({
         confirm: true,
+        rememberPermission,
```
bumi, line 49: why is this in here. as you above already do an `addPermission`.

bumi (commented):
any reason why we don't do it the same way as in Lisa's ln/request function?

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
bumi: you link to a post that is 7 years ago, is that still relevant? 

this looks pretty much what we need, isn't it? 
https://dexie.org/docs/Table/Table.where()#samples

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
bumi: no idea, for me those are then just looooong random list of strings. and the same values in a type as in an array.
if TypeScript wants it, do it.

src/extension/ln/nostr/index.ts
escapedcat: Would do `getPublicKey` in a separate PR

src/types.ts
escapedcat: Wanted your thoughts on this first. Will adjust.

escapedcat:
> any reason why we don't do it the same way as in Lisa's ln/request function?

You mean using `first`? Did that now

bumi:
the concept is a different one where the permission is saved. 
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/ln/request.ts#L67-L78

escapedcat:
> the concept is a different one where the permission is saved. https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/ln/request.ts#L67-L78

- You mean _where_ in case of what exactly?
- Should I save it from within [signEventOrPrompt](https://github.com/getAlby/lightning-browser-extension/blob/feat/1727_nostr-permission-sign-message/src/extension/background-script/actions/nostr/signEventOrPrompt.ts)?
- Should I not use [the API](https://github.com/getAlby/lightning-browser-extension/blob/feat/1727_nostr-permission-sign-message/src/app/screens/Nostr/ConfirmSignMessage.tsx/#L40-L45)?

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: @lisabaut wdyt?

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: Please check current code, I already adjusted it

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +19,44 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+  if (!allowance?.id) {
+    return { error: "Could not find an allowance for this host" };
+  }
+
+  const findPermission = await db.permissions.get({
+    host: message.origin.host,
+    method: PermissionMethod["NOSTR_SIGNMESSAGE"],
+  });
+
+  const hasPermission = !!findPermission?.enabled;
```
escapedcat, line 35: https://stackoverflow.com/questions/53060324/using-a-field-with-a-boolean-type-in-dexie-js
Can't use boolean in `get/where`?

src/types.ts
```diff
@@ -497,21 +497,27 @@ export interface DbPayment {
   totalFees: number;
 }
 
+export interface Payment extends Omit<DbPayment, "id"> {
+  id: number;
+}
+
+export enum PermissionMethod {
+  NOSTR_SIGNMESSAGE = "nostr/signMessage",
+  WEBLN_LISTCHANNELS = "webln/listchannels",
```
bumi, line 506: where is this from? 
we don't have listchannels so far. 

those methods will also vary from connector to connector. LND will have different ones than CLN. 

when a a new method is added to a connector then this needs to be listed in the connector and then also here? And in the future we might remove this whitelist.
the request methods are kinda generic, so we do not necessarily know what will be in there. What do we then list here?

src/types.ts
```diff
@@ -194,10 +194,10 @@ export interface MessageAccountAll extends MessageDefault {
 
 export interface MessagePermissionAdd extends MessageDefault {
   args: {
-    host: string;
-    method: string;
-    enabled: boolean;
-    blocked: boolean;
+    host: DbPermission["host"];
```
bumi, line 197: you deleted the `add.ts` action, is this message still used somewhere?

src/i18n/locales/en/translation.json
```diff
@@ -637,7 +637,12 @@
       "title": "Nostr",
       "allow": "Allow {{host}} to:",
       "read_public_key": "Read your public key",
-      "block_and_ignore": "Block and ignore {{host}}"
+      "block_and_ignore": "Block and ignore {{host}}",
+      "confirm_sign_message": {
+        "remember": {
+          "label": "Remember this decission"
```
bumi, line 643:
```suggestion
          "label": "Remember this decision"
```

or maybe "Remember and don't ask again"?

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +19,44 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+
+      // add permission to db only if user decided to always allow this request
+      if (promptResponse.data.rememberPermission) {
+        const permissionIsAdded = await db.permissions.add({
+          createdAt: Date.now().toString(),
+          allowanceId: allowance.id,
+          host: message.origin.host,
+          method: PermissionMethod["NOSTR_SIGNMESSAGE"],
+          enabled: true,
+          blocked: true,
```
bumi, line 55:
```suggestion
```

should not be blocked.

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +19,44 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
-    }>({
-      ...message,
-      action: "public/nostr/confirmSignMessage",
-    });
-    if (!response.data.confirm) {
-      throw new Error("User rejected");
+    if (!hasPermission) {
+      const promptResponse = await utils.openPrompt<{
+        confirm: boolean;
+        rememberPermission: boolean;
```
bumi, line 41: nit: in ln/request we call it `enabled` (which is probably also not a good name)
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/ln/request.ts#L67-L68

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +19,44 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+  if (!allowance?.id) {
+    return { error: "Could not find an allowance for this host" };
+  }
+
+  const findPermission = await db.permissions.get({
+    host: message.origin.host,
+    method: PermissionMethodNostr["NOSTR_SIGNMESSAGE"],
+  });
+
+  const hasPermission = !!findPermission?.enabled;
```
lisabaut, line 35: Just to repeat again:
The check for "enabled" is _actually_ not needed.
Either the permission is saved to DB (enabled was set to "true" in the promptReponse once)
or the permission does not exist in the DB.

See this comment again for clarification and why this check is still implemented:
https://github.com/getAlby/lightning-browser-extension/pull/1724#discussion_r1017642768

src/extension/background-script/router.ts
```diff
@@ -51,8 +50,6 @@ const routes = {
   lnurl: lnurl,
   lnurlAuth: auth,
   getCurrencyRate: cache.getCurrencyRate,
-  addPermission: permissions.add,
-  deletePermission: permissions.deletePermission,
```
lisabaut, line 55: byebye

lisabaut (commented):
Thanks for cleaning this up and removing the webln changes 👍 
See some comments below

src/app/screens/Nostr/ConfirmSignMessage.tsx
```diff
@@ -19,21 +19,23 @@ function ConfirmSignMessage() {
...
   const [loading, setLoading] = useState(false);
   const [successMessage, setSuccessMessage] = useState("");
+  const [rememberPermission, setRememberPermission] = useState(false);
 
   // TODO: refactor: the success message and loading will not be displayed because after the reply the prompt is closed.
   async function confirm() {
     try {
       setLoading(true);
       msg.reply({
         confirm: true,
```
lisabaut, line 37: Replace this here with `blocked: false`

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -17,15 +19,44 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
...
+
+      // add permission to db only if user decided to always allow this request
+      if (promptResponse.data.enable) {
+        const permissionIsAdded = await db.permissions.add({
+          createdAt: Date.now().toString(),
+          allowanceId: allowance.id,
+          host: message.origin.host,
+          method: PermissionMethodNostr["NOSTR_SIGNMESSAGE"],
+          enabled: true,
+          blocked: false,
```
lisabaut, line 55:
```suggestion
          enabled: promptResponse.data.enabled,
          blocked: promptResponse.data.blocked,
```

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
lisabaut: As discussed in person:
Because right now this PR deals only with "signMessage" in one single file
and more Nostr events are later to come
I think it is fine to keep it like you started it: [PermissionMethodNostr TYPE](https://github.com/getAlby/lightning-browser-extension/pull/1775/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR487)

At some point you will need a "supportedNostrMethodMap" object and map the method to its corresponding translation. 
Similar to how it was started here for LN (=> [draft PR code](https://github.com/getAlby/lightning-browser-extension/pull/1752/files#diff-82b11e63afd4e5a1aeac4c0023524f3d17f0aafecac1a8017236bb8f210082d0R82-R108))

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
escapedcat: True, I just did that for _the future_. Can remove.

src/types.ts
```diff
@@ -497,21 +479,24 @@ export interface DbPayment {
...
+  NOSTR_SIGNMESSAGE = "nostr/signMessage",
+}
+
 export interface DbPermission {
   id?: number;
   createdAt: string;
   allowanceId: number;
   host: string;
-  method: string;
+  method: string | PermissionMethodNostr;
```
bumi, line 495: hmmm... so why string and this?

src/extension/background-script/actions/allowances/__tests__/enable.test.ts
```diff
@@ -59,7 +59,7 @@ describe("enable allowance", () => {
     const message: MessageAllowanceEnable = {
       application: "LBE",
       prompt: true,
-      action: "enableAllowance",
+      action: "public/webln/enable",
```
bumi, line 62: I think this is never `public/...` is it?

src/types.ts
escapedcat: We currently do not knwo the webln-methods but we do know the nostr ones

src/extension/background-script/actions/allowances/__tests__/enable.test.ts
escapedcat: That's what it is currently?:

https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/router/Prompt/Prompt.tsx#L68

image

bumi: true, sorry. that was added with nostr.

src/types.ts
bumi: why don't we know the webln ones? 

or rather there are actually none that use the permissions.

escapedcat: Right, so, I mean, at least currently we do not know them. At least in the context of this PR.  
Eventually we should be able to know them so we can also translate them etc.

bumi: currently it is only what is in `PermissionMethodNostr` we know that and we change this in this PR.
why do we need to keep the string?

escapedcat: Otherwise TS will fail for this: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/ln/request.ts#L39

PR #1792, opened 2022-11-29
feat(nostr): add permissions to the rest of the nostr call

bumi:
expanded it to the rest of the nostr calls.


src/extension/background-script/actions/nostr/getPublicKeyOrPrompt.ts
```diff
@@ -9,26 +10,58 @@ const getPublicKeyOrPrompt = async (message: MessagePublicKeyGet) => {
...
+          method: PermissionMethod["NOSTR_GETPUBLICKEY"],
+          enabled: true,
+          blocked: false,
+        });
+
+        !!permissionIsAdded && (await db.saveToStorage());
+      }
 
-    return response;
+      if (promptResponse.data.confirm) {
```
escapedcat, line 55: But currently this is always `true`, no?:  
https://github.com/getAlby/lightning-browser-extension/pull/1792/files#diff-d5402bf744dada20fceb38fbe951f38849c76376a409fbc8b3d32be73f58b724R28

escapedcat: And if the prompt is cancelled this will never be reached?  
Or am I missing something?

bumi: yes, I had a note somewhere. guess not in all screens.

PR #1803, opened 2022-11-30
chore: load and initialize connector on unlock

bumi:

Some connectors might have problems when two instances are initialized. 
This prevents potential race conditions to initialize the connector through parallel action calls. (e.g. loading the balance and doing a webln call)
With this we make sure that once the extension is unlocked the connector is also initialized.

### Type of change


bumi:
It should probably get some good exception handling, because if something fails connecting to the account then the user is locked out of Alby and will probably never get back in.

escapedcat (approved):
ack

PR #1809, opened 2022-12-01
fix: overflow of website title

reneaaron:

Long website names / titles currently allow the publisher card to grow indefinitely: (and shift important UI elements out of the view at the bottom)

image

### Type of change

image

### DIFF
--- a/src/app/components/PublisherCard/index.tsx
+++ b/src/app/components/PublisherCard/index.tsx
@@ -59,7 +59,7 @@ export default function PublisherCard({
       >
         <h2
           className={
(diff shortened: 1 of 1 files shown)


src/app/components/PublisherCard/index.tsx
```diff
@@ -59,7 +59,7 @@ export default function PublisherCard({
       >
         <h2
           className={
```
im-adithya, line 61: should we add a title prop?

reneaaron: nice idea, will add that! 👍

PR #1814, opened 2022-12-03
Fix/vida page

bumi:
vida.page recently has added support for the lightning meta tag because we had special support for vida.page this was ignored so far. 
This PR handles the meta tag, but also the custom lightning address.


bumi:
Why does this change break some unrelated TS linting?

PR #1817, opened 2022-12-04
feat: added blocklist note

bumi:
the blocklist only takes effect after a page reload.
This might be confusing for the user as the user adds something to the blocklist but it has no effect (until a page reload)
This alert is a quick hack to give the user some feedback.

This PR also adds a check to ignore any webln/nostr calls from the website if the enable call has failed. This prevents the website from spamming the user with multiple prompts.

### DIFF
--- a/src/app/screens/Enable/index.tsx
+++ b/src/app/screens/Enable/index.tsx
@@ -48,9 +48,7 @@ function Enable(props: Props) {
       domain: props.origin.domain,
       host: props.origin.host,
     });
-    alert(
-      `Added ${props.origin.host} to the blocklist, please reload the website`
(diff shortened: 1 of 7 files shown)


src/app/screens/Enable/index.tsx
```diff
@@ -48,6 +48,9 @@ function Enable(props: Props) {
       domain: props.origin.domain,
       host: props.origin.host,
     });
+    alert(
```
reneaaron, line 51: Move to translated string and maybe add a `.` at the end.

reneaaron (approved):
Good point. 👍

utACK

PR #1818, opened 2022-12-04
feat(qr scanner): remove potential lightning: protocol

bumi:
when scanning a QR code for an invoice there might still be the "lightning:" protocol in the text.
this change removes this.


escapedcat (approved):
utACK

PR #1819, opened 2022-12-05
feat: bring back permission actions

bumi:


im-adithya:
Made changes as per https://github.com/getAlby/lightning-browser-extension/pull/1821#issuecomment-1338058502

PR #1821, opened 2022-12-05
feat: view and edit allowance permissions

### LINKED ISSUE #1820: Allow users to see and delete allowance permissions
### Feature description

Every host/allowance now can have multiple permissions.
Right now the user does not see which permission they granted to a website and there is no way to delete them.

### Describe the solution

On the allowance edit modal we can list all granted permissions and provide an option to delete those

image

we already have the delete action: https://github.com/getAlby/lightning-browser-extension/pull/1819/files#diff-74471d8257a5d4aee886e495b6cf7f047385054e8a3114cc2fedcc5ae1315175

We might still need to add an action to list all permissions for an allowance/host. 

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

im-adithya:

Title

Fixes #1820

### Type of change

Screenshot 2023-01-16 at 6 22 55 PM

### How has this been tested?

Manually using [astral.ninja](https://astral.ninja/)

### Checklist


src/extension/background-script/actions/allowances/disablePermissions.ts
```diff
@@ -0,0 +1,15 @@
+import db from "~/extension/background-script/db";
```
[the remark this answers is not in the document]
escapedcat: Do we want tests for these?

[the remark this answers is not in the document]
im-adithya: Sure, was thinking of the same (folder and tests)
Should I wait for the PR to be merged then?

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -145,15 +193,61 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
...
+                title={t("enable_login.title")}
+                subtitle={t("enable_login.subtitle")}
+              >
+                <Toggle
+                  checked={lnurlAuth}
+                  onChange={() => setLnurlAuth(!lnurlAuth)}
+                />
+              </Setting>
+            </div>
+            <h2 className="mt-4 text-lg text-gray-900 font-bold dark:text-white">
```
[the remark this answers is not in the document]
escapedcat: Right, good point. Yeah, toggle might be better.

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -65,13 +99,27 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
...
+  }
+
   async function updateAllowance() {
     await msg.request("updateAllowance", {
       id: allowance.id,
       totalBudget: parseInt(budget),
       lnurlAuth,
     });
 
+    await msg.request("disablePermissions", {
```
escapedcat, line 117: Should we only do this if `changedPermissionIds` is filled?

src/extension/background-script/actions/allowances/disablePermissions.ts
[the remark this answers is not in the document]
escapedcat: No need to wait for that I think

src/app/components/AllowanceMenu/index.tsx
im-adithya: Oops yes, changed!

im-adithya:
Just for reference: changed `disablePermissions` to `togglePermissions` as that's what we need here

bumi:
we got a permissions folder for the actions: https://github.com/getAlby/lightning-browser-extension/pull/1819/files
best follow that one and also add the list action there. (build on that other PR, or we merge that other PR first)

I'd also delete the actual permission and not set it to `enable: false` - we currently do not have much settings there. (either it exists and the allowance has that permission, or not)

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -145,26 +202,69 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
...
+                      {permission.method}
+                    </label>
+                  </div>
+                ))}
+              </div>
+            )}
+
+            {isEmptyPermissions && (
+              <p className="text-gray-500 dark:text-neutral-400">
+                {t("no_permissions")}
```
bumi, line 257: we just don't show anything if there are no permissions stored. 
most websites will initially have no such specific permissions.

src/extension/background-script/actions/nostr/helpers.ts
```diff
@@ -33,6 +33,24 @@ export async function addPermissionFor(method: string, host: string) {
   if (!allowance?.id) {
     return false;
   }
+
+  const findPermission = await db.permissions.get({
```
bumi, line 37: where is this used?

im-adithya: This would be used whenever the user goes back to the website after _disabling_ a permission, we have to check whether to add the permission again or just toggle enable if it already exists

im-adithya: Wouldn't require this change if we proceed by deleting the permission

escapedcat:
> we got a permissions folder for the actions: https://github.com/getAlby/lightning-browser-extension/pull/1819/files best follow that one and also add the list action there. (build on that other PR, or we merge that other PR first)
> 
> I'd also delete the actual permission and not set it to `enable: false` - we currently do not have much settings there. (either it exists and the allowance has that permission, or not)

So that means that the switch-input in that list will add or delete a permission? And not just set enabled true/false?

src/extension/background-script/actions/nostr/helpers.ts
bumi: I don't understand. 
we check permissions whenever a call is executed and if there is no permission the user will be prompted. 

ideally we should not add functions that we don't use. then add them when they are needed and used.

escapedcat: Are these used though?: https://github.com/getAlby/lightning-browser-extension/pull/1819  
What's the reason to bring these back?

im-adithya: > we check permissions whenever a call is executed and if there is no permission the user will be prompted.

Yes, we do, but here we're not deleting the permission (instead, we set it to `enabled: false`) when we disable it from the UI, which would return the enabled state. Now when we prompt the user, we just have to set it to `enabled: true` instead of adding the same permission again...

But this confusion will be removed once we delete the permissions in the first step.

I made these changes in #1819 (deleting instead of disabling as you suggested) so I'm waiting for it to be merged so that I can rebase things here to avoid confusion

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -65,13 +104,31 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
...
 
+    const changedIds = changedPermissionIds();
+
+    if (changedIds.length) {
+      await msg.request("togglePermissions", {
+        ids: changedIds,
+      });
+    }
+
+    setOriginalPermissions(permissions);
```
lisabaut, line 130: I would add a small comment here:
`// DB is updated, let´s update the original permissions to the updated permission in local state too`

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -65,13 +104,31 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
...
+  function changedPermissionIds(): number[] {
+    if (!permissions || !originalPermissions) return [];
+    const ids = permissions
+      .filter((prm, i) => prm.enabled !== originalPermissions[i].enabled)
+      .map((prm) => prm.id);
+    return ids;
+  }
+
   async function updateAllowance() {
     await msg.request("updateAllowance", {
```
lisabaut, line 116: Would add a try/catch around this whole block here

src/app/components/AllowanceMenu/index.tsx
lisabaut: Agree, no need to show a default on empty permissions (yet)

src/extension/background-script/actions/allowances/listPermissions.ts
```diff
@@ -0,0 +1,18 @@
+import db from "~/extension/background-script/db";
+import type { MessagePermissionsList } from "~/types";
+
+const listPermissions = async (message: MessagePermissionsList) => {
+  const { id } = message.args;
+  const permissions = await db.permissions
+    .toCollection()
+    .filter((permission) => permission.allowanceId === id)
+    .toArray();
```
lisabaut, line 9: I _think_ you can simplify this query with

```
const permissions = await db.permissions.where({
    allowanceId: id
}).toArray();
```

src/i18n/locales/en/translation.json
```diff
@@ -710,7 +710,9 @@
       "edit_allowance": {
         "title": "Edit Allowance",
         "screen_reader": "Allowance Options"
-      }
+      },
+      "edit_permissions": "Edit Permissions",
+      "no_permissions": "No permissions."
```
lisabaut, line 715: No need for a fallback. See comment above

lisabaut (commented):
Thanks Adithya ⭐ 
I know another PR needs to be merged first, just wanted to add some nitpicks comments already. See below.

im-adithya:
Thanks for the suggestions, @lisabaut 🙌
Made all the necessary changes, ready for another review!

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -145,26 +208,65 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
...
+
+            {hasPermissions && (
+              <>
+                <h2 className="mt-4 text-lg text-gray-900 font-bold dark:text-white">
+                  {t("edit_permissions")}
+                </h2>
+                <div>
+                  {permissions.map((permission, index) => (
+                    <div key={index} className="flex items-center my-2">
+                      <Checkbox
```
escapedcat, line 237: You still wanted to switch this to be a toggle?

im-adithya: We're deleting them now, right? Wdyt? Should we change this?

escapedcat: Hm, don't think what happens inside is important or needs to be reflected in the UI.  
Eventually we might use enable/disable anyways, right?  
Also we currently do not use checkboxes in this kind of situation.

im-adithya: Makes sense. Toggle it is then!

im-adithya:
Everything is ready now, just have to decide the copy for the permissions
Screenshot 2022-12-13 at 4 44 51 PM

escapedcat:
image

1. Go to https://branle.netlify.app
2. Enable and agree and save to public-key permission
3. Post a message and save permission
4. Disable permission in Alby
5. Go to https://branle.netlify.app
6. Post a message and save permission
7. -> Permission doubled

escapedcat:
image

Turning off a permission and "save" is throwing "route not found"

escapedcat (dismissed):
I guess it works as expected.  
I find it a bit confusing that after "disabling" a permission and reloading the tab the permission is gone. But I guess that's because we do not use enable/disable yet, right?

(Sorry for my previous - now deleted - comments. I was on an outdated branch)

im-adithya:
> I find it a bit confusing that after "disabling" a permission and reloading the tab the permission is gone. But I guess that's because we do not use enable/disable yet, right?

Yup +1 on that but it's a temporary change as per what @bumi said; it's just because of deleting permissions instead of disabling them.

bumi:
@escapedcat why did you approve this? I don't think "random subtitle" should be merged. 
btw. @im-adithya from my experience it is never good to use dummy text in an app (like lorem ipsum or random subtitle) this will somehow end up in production. 


having an `enable` flag in the DB was a mistake, I should have not proposed this. it seems to come up too often and cause confusion.

escapedcat:
> why did you approve this? I don't think "random subtitle" should be merged.

Totally didn't see this... thanks

im-adithya:
Still have to decide:
- Migrations for webln methods (lnd and commando)
- Copy for nostr methods

Went with `webln/connectorName/method` and `nostr/method`

(found a way to avoid `nostr/nostr/method`) see https://github.com/getAlby/lightning-browser-extension/pull/1821/commits/b384c6e0291ed2c88123f90dc9ebb76e2546625f#diff-1b329da99ce24a934532ca1ee401d135a5d5917619b62db9711da859ce47f987R243-R253

im-adithya:
Summary of the discussion I just had with René:
- Unify the "nostr" permissions being used in Alby Prompts and in the Permission Popup to use the same descriptions (except for `signMessage`)
- Delete the `webln/methods` as discussed before in a migration, allowing connectorName to be part of the methods from now on `webln/connectorName/methods`

(Modifying tests to follow the second step would fix the unit tests)

src/i18n/locales/en/translation.json
```diff
@@ -832,6 +831,12 @@
     }
   },
   "permissions": {
+    "nostr": {
+      "getpublickey": "",
```
bumi, line 835: we need some text here.

src/extension/background-script/actions/ln/request.ts
```diff
@@ -43,13 +43,15 @@ const request = async (
       throw new Error("Could not find an allowance for this host");
     }
 
+    const connectorName = connector.constructor.name.toLowerCase();
     // prefix method with webln to prevent potential naming conflicts (e.g. with nostr calls that also use the permissions)
-    const weblnMethod = `${WEBLN_PREFIX}${methodInLowerCase}`;
+    const legacyWeblnMethod = `${WEBLN_PREFIX}${methodInLowerCase}`;
```
bumi, line 48: maybe we just drop the legacy method. 
Users will have to approve the permission again which I think is OK. and the code is easier.

bumi (commented):
let's finish this. and then move this also to the new account page of #1934

PR #1830, opened 2022-12-07
Chore/cleanup

bumi:


src/extension/background-script/db.ts
```diff
@@ -1,5 +1,4 @@
 import Dexie from "dexie";
-import "fake-indexeddb/auto";
```
bumi, line 2: @escapedcat do you know why this was loaded here? this should only be loaded in the tests, shouldn't it?

escapedcat: I assume this was a mistake

escapedcat (approved):

PR #1836, opened 2022-12-07
Add 25k option for the default amounts

bumi:
This removes the default 100sats option and adds a 25k option. 
100 is easy to type and a rather small amount.


escapedcat (approved):

src/app/components/SatButtons/index.tsx
```diff
@@ -38,6 +31,13 @@ function SatButtons({ onClick, disabled }: Props) {
         fullWidth
         disabled={disabled}
       />
+      <Button
+        icon={<SatoshiV2Icon className="w-4 h-4" />}
+        label="25K ⚡"
```
jankoegel, line 36: Why is it a capital "K" here?
https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes :)

PR #1838, opened 2022-12-07
refactor: getOS: don't use deprecated appVersion.

jankoegel:

+ From MDN:
  https://developer.mozilla.org/en-US/docs/Web/API/Navigator/appVersion

  > Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. ... Be aware that this feature may cease to work at any time.

  There's no such warning for `userAgent` and here's a source on how to use it instead:
  https://www.educative.io/answers/how-to-detect-the-operating-system-of-the-client-machine-using-js

### Screenshots of the problem [optional]

SCR-20221207-v0z

### How has this been tested?

+ only on my local system. Trusting the linked blog post that user agents are generally well-behaving and as reliable as `appVersion`.

### Checklist

### DIFF
--- a/src/app/components/CompanionDownloadInfo.tsx
+++ b/src/app/components/CompanionDownloadInfo.tsx
@@ -17,7 +17,7 @@ function CompanionDownloadInfo({ hasTorCallback }: Props) {
     if (userAgent.indexOf("Win") !== -1) return "Windows";
     if (userAgent.indexOf("Mac") !== -1) return "MacOS";
     if (userAgent.indexOf("X11") !== -1) return "UNIX";
-    if (userAgent.indexOf("Linux") !== -1) return "Linux"
+    if (userAgent.indexOf("Linux") !== -1) return "Linux";
   }
 
   // TODO: check if the companion app is already installed
--- a/src/app/components/CompanionDownloadInfo.tsx
+++ b/src/app/components/CompanionDownloadInfo.tsx
@@ -13,10 +13,11 @@ function CompanionDownloadInfo({ hasTorCallback }: Props) {
   });
 
   function getOS() {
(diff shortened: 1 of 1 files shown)


src/app/components/CompanionDownloadInfo.tsx
```diff
@@ -13,10 +13,11 @@ function CompanionDownloadInfo({ hasTorCallback }: Props) {
...
   function getOS() {
-    if (navigator.appVersion.indexOf("Win") != -1) return "Windows";
-    if (navigator.appVersion.indexOf("Mac") != -1) return "MacOS";
-    if (navigator.appVersion.indexOf("X11") != -1) return "UNIX";
-    if (navigator.appVersion.indexOf("Linux") != -1) return "Linux";
+    const userAgent = navigator.userAgent;
+    if (userAgent.indexOf("Win") !== -1) return "Windows";
+    if (userAgent.indexOf("Mac") !== -1) return "MacOS";
+    if (userAgent.indexOf("X11") !== -1) return "UNIX";
+    if (userAgent.indexOf("Linux") !== -1) return "Linux";
```
jankoegel, line 20: n.b. this returns `undefined` when no match can be found in the userAgent string (since power users can configure the userAgent themselves, this might happen).
We then take users to this page, which fails silently (as it does for all strings at the end of the URL that are not one of the OSes listed here):
https://getalby.com/install/companion/undefined

I'd suggest the following:
1. add a fallback here (e.g. "Windows" since it's the most common OS)
2. show a warning in the Rails app when `https://getalby.com/install/companion/` is called with an unknown parameter such as `undefined` at the end.

PR #1855, opened 2022-12-10
refactor: Navbar: constrain max width to same width as the content.

jankoegel:
### Related Rails app PR
+ https://github.com/getAlby/getalby.com/pull/272

### HELP NEEDED / TODOs
+ [x] review code
+ [x] depending on responsiveness needs, consider moving the navigation entries ("websites" etc.) a bit more to the right so they are more centered instead of leaning left.
+ [x] reviewer: test with an extension development setup (or pair with me / let me know how to set one up)
+ [x] reviewer: test responsiveness (current extension's header looks as broken to me on smaller breakpoints as with these changes but I'm not sure!?)

+ Navbar: constrain max width to same width as the content.
+ reason: on desktop the navigation items in the top left and right corners were easy to miss. With this change, everything is closer together and easier to notice.

### Type of change

+ UI improvement

#### Before
SCR-20221210-ugq

#### After
SCR-20221210-ug6

### How has this been tested?

+ ~~⚠️ I only tested this with my browser's inspector. I don't have a development setup for the extension yet.~~ Ok, I have the development setup for the extension now and everything looked good to me.

### DIFF
--- a/src/app/components/Navbar/Navbar.tsx
+++ b/src/app/components/Navbar/Navbar.tsx
@@ -7,17 +7,19 @@ type Props = {
 
 export default function Navbar({ children }: Props) {
   return (
-    <div className="px-4 py-2 bg-white flex justify-between items-center border-b border-gray-200 dark:bg-surface-01dp dark:border-white/10">
-      <div className="flex w-8/12 md:w-4/12 lg:w-3/12">
-        <AccountMenu />
-      </div>
-      {children && (
-        <div>
-          <nav className="flex space-x-8">{children}</nav>
+    <div className="px-4 py-2 bg-white border-b border-gray-200 dark:bg-surface-01dp dark:border-white/10">
+      <div className="max-w-screen-lg flex justify-between mx-auto w-full px-3 items-center">
+        <div className="flex w-8/12 md:w-4/12 lg:w-3/12">
+          <AccountMenu />
         </div>
-      )}
-      <div className="md:w-4/12 lg:w-3/12 flex justify-end items-center">
-        <UserMenu />
+        {children && (
+          <div>
+            <nav className="flex space-x-8">{children}</nav>
+          </div>
+        )}
+        <div className="md:w-4/12 lg:w-3/12 flex justify-end items-center">
+          <UserMenu />
+        </div>
       </div>
     </div>
   );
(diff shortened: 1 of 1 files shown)


src/app/components/Navbar/Navbar.tsx
```diff
@@ -7,17 +7,19 @@ type Props = {
...
+          </div>
+        )}
+        <div className="md:w-4/12 lg:w-3/12 flex justify-end items-center">
+          <UserMenu />
         </div>
-      )}
-      <div className="md:w-4/12 lg:w-3/12 flex justify-end items-center">
-        <UserMenu />
       </div>
     </div>
```
jankoegel, line 24: recommendation for reviewers: enable the **"hide whitespace"** option above and the diff becomes a LOT easier to read. I added an inner wrapper `<div>` and moved three classes from the existing outermost wrapper to this new inner wrapper (`.flex`, `.justify-between` and `.items-center`).

jankoegel: or install:
https://addons.mozilla.org/en-US/firefox/addon/github-whitespace-disabler/
https://chrome.google.com/webstore/detail/github-diff-whitespace-di/bppopebgnjajocjnjlginjfpjbleappc

escapedcat (approved):
I think this is a good adjustment.

This looks a bit weird, but not to problematic I think:
image

src/app/components/Navbar/Navbar.tsx
```diff
@@ -7,16 +7,10 @@ type Props = {
```
jankoegel, line 7: best reviewed in "split mode".

jankoegel, line 7: Best reviewed in "split mode". I actually added a new wrapper `div` to constrain the content's width.
Since I also removed wrapper `div`s from `<AccountMenu>`, `children` and `UserMenu` it's a bit hard to see the changes in other views.

jankoegel:
> I think this is a good adjustment.
> 
> This looks a bit weird, but not to problematic I think: image

yes, there is a weird space of about 100px display resolution around 1024px where the wallet icon has space on the left and is not flush with the content's left border. in the same space the negative margin on the `children` is a problem. Didn't investigate further since responsiveness is not a big focus. We could remove the negative margin, at full-res I felt it looks more centered with it than without.

jankoegel:
> hm, just checked again and it feels like i don't get the children that close to the wallet icon until a much smaller breakpoint on firefox. Does the following GIF match what you're seeing?

looks the same as in Firefox for me on Chrome.

PR #1857, opened 2022-12-12
fix: add favicon to extension pages (firefox)

im-adithya:

Favicons in extension pages

--

### Type of change

Screenshot 2022-12-12 at 5 15 40 PM

### How has this been tested?

Manually

### Checklist

### DIFF
--- a/static/views/options.html
+++ b/static/views/options.html
@@ -4,6 +4,7 @@
 <head>
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <link rel="icon" type="image/png" href="assets/icons/alby_icon_yellow_32x32.png">
   <title>Alby</title>
 </head>
 
--- a/static/views/welcome.html
+++ b/static/views/welcome.html
@@ -3,7 +3,8 @@
 
 <head>
   <meta charset="UTF-8" />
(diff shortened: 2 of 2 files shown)


escapedcat (approved):
ack

jankoegel:
@im-adithya looks good. should we add the same to `popup.html` and `prompt.html`!?
`popup.html` probably never appears in a tab of its own!?!
I don't know when `prompt.html` is rendered.

escapedcat:
> `popup.html` probably never appears in a tab of its own!?!

Right, I don't think so.

> I don't know when `prompt.html` is rendered.

This is being prompted if you interact with a website i.e. "connect", "do a payment", etc.  
Does never show in a tab afaik

static/views/options.html
```diff
@@ -4,6 +4,7 @@
 <head>
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <link rel="icon" type="image/png" href="assets/icons/alby_icon_yellow_32x32.png">
```
bumi, line 7: @escapedcat this path might be relevant for v3. - there was some change in the paths, wasn't it?

escapedcat: It should not matter for the metatags but will check

PR #1860, opened 2022-12-12
fix: parse the cached accounts only if it is a string

bumi:


bumi:
Issue on orion: https://orionfeedback.org/d/3898-storagelocalget-automatically-parses-json

bumi:
Can I get this in to test Orion? 
https://browser.kagi.com/ - works also maybe for iOS (iPad iPhone)

escapedcat (approved):
utACK
Maybe add a comment to the code so people will understand why we do this?

PR #1862, opened 2022-12-13
chore: replace html5-qrcode with qr-scanner

### LINKED ISSUE #1832: [Improvement] Consider replacing html5-qrcode with a smaller alternative
### Feature description

+ aim for smaller extension size. `html5-qrcode@2.3.1`, our current solution, is quite big with [350.2 kB (minified)](https://bundlephobia.com/package/html5-qrcode@2.3.1).

### Describe the solution

+ use a JS implementation to read QR codes that's smaller than our current solution.

### Describe alternatives

1. [qr-scanner](https://github.com/nimiq/qr-scanner) - `70kB` (bundle phobia reports the wrong size here [1], we need the [`legacy` package](https://www.npmjs.com/package/qr-scanner?activeTab=explore))
2. https://github.com/cozmo/jsQR - [127 kB](https://bundlephobia.com/package/jsqr@1.4.0)

---
[1] There's an even smaller version of `qr-scanner` (15kB) if the native BarcodeDetector API is installed: https://developer.mozilla.org/en-US/docs/Web/API/BarcodeDetector#browser_compatibility (they disabled it for M1/M2 macs with Ventura, so I don't think it's a good idea to rely on it :) See: https://github.com/nimiq/qr-scanner/commit/987bbb06c2c0fdbfed78bdda5e144bc29d2e880f)

### Additional context

_No response_

### Are you working on this?

No

im-adithya:

Replaces html5-qrcode with qrscanner

Fixes #1832 

### Type of change

- `chore`: Bug fix (non-breaking change which fixes an issue)

## Eariler vs Now:
<p>
Screenshot 2022-12-13 at 7 12 58 PM
Screenshot 2022-12-13 at 7 13 13 PM
</p>

### How has this been tested?

**Manually, Video:**

https://user-images.githubusercontent.com/64399555/207346421-d051bbc2-724c-4f9c-8e95-7cf37dd12e25.mov

**In Settings:**

Screenshot 2022-12-13 at 7 20 57 PM

### Checklist
- [x] find fix for Firefox (see:
  + https://github.com/nimiq/qr-scanner/issues/217
  + https://github.com/getAlby/lightning-browser-extension/pull/1899#issuecomment-1359461374
  + https://getalby.slack.com/archives/C02TD6P5XPE/p1671514296425669
  + -> fix: https://bugzilla.mozilla.org/show_bug.cgi?id=1294996#c4


src/app/screens/Settings.tsx
```diff
@@ -146,6 +156,8 @@ function Settings() {
             </p>
           )}
         </Setting>
+        {/* This is just to enable camera permissions*/}
```
jankoegel, line 159:
```suggestion
        {/* This is just to enable camera permissions */}
```

src/app/screens/Settings.tsx
```diff
@@ -133,8 +133,18 @@ function Settings() {
               label={t("camera_access.allow")}
               onClick={async () => {
                 try {
-                  await Html5Qrcode.getCameras();
-                  setCameraPermissionsGranted(true);
+                  const devices = await QrScanner.listCameras();
+                  const video =
+                    document.querySelector<HTMLVideoElement>("#reader");
```
jankoegel, line 138: We control the name of the ID here, don't we? Can we change it to a more descriptive name like `#qr-code-reader`?

src/app/screens/Settings.tsx
```diff
@@ -133,8 +133,18 @@ function Settings() {
...
+                  if (video && devices && devices.length) {
```
jankoegel, line 139:
```suggestion
                  if (video && devices && devices.length > 0) {
```

I know that `0` is falsey in JS but imo that's a language quirk and I find the code more readable for humans if we don't rely on it.

jankoegel:
Not one of your changes but was wondering about the code here:

https://github.com/getAlby/lightning-browser-extension/blob/8f4f93330bc4f70bbca23a590c4b4457c4c02eea/src/app/components/QrcodeScanner/index.tsx#L136

Can't we remove the anonymous function and just pass handleStopScanning as an onclick handler?

src/app/screens/Settings.tsx
```diff
@@ -133,8 +133,18 @@ function Settings() {
               label={t("camera_access.allow")}
               onClick={async () => {
```
jankoegel, line 134: Can we move this click handler to a separate function? This would improve readability for me and - although it might not be critical here - it's a good idea in general for rendering performance: https://marcoheine.com/blog/7-easy-ways-to-improve-your-react-performance-part-2/#avoid-anonymous-funcs

jankoegel:
@im-adithya two things I noticed during testing:
+ cannot read QR codes in Chrome when placing the QR code near the camera (when it's really big but still in focus). I always need to move it back about 15-20cm away from the camera to have any success. I didn't compare this to the QR lib on master. We should add a warning / instructions for users to make sure their scanning success rate is high.
+ cannot read QR codes in Firefox at all. I suspect this is because the "dynamic import" the qr-scanner README mentions doesn't seem to work in our setup!? Firefox does not have built-in support for the Barcode API i mentioned in my GH issue, Chrome has that support built in. What's the best way to force loading the `worker`?
image
(https://www.npmjs.com/package/qr-scanner?activeTab=explore)
+ the OLD `README` notes here could be helpful to get an idea what's happening: https://github.com/nimiq/qr-scanner/commit/60a2b3b2ced7d2e7567cfe030319df848816f1d4 I'm not sure how best to enforce the loading of `worker` in our setup / their NEW `README` confuses me and I have trouble seeing how to solve our problem!?

jankoegel:
this demo using the legacy QR code library from unkpg.com works for me in Firefox (just like their official demo [here](https://nimiq.github.io/qr-scanner/demo/) which uses the dynamic import via Webpack - I suppose we cannot use something like that in a browser extension):
https://codepen.io/jankoegel/pen/yLEmGxJ?editors=1111

tested with the first QR code from this demo page.

the `highlightScanRegion: true` option helps a lot to show users where to place their QR code.

I was able to load the legacy qr-code package into my development extension and called the `createWorker` function from it. I had no luck scanning any QR code with it so far. Not sure if we need to do more to set things up properly, perhaps i'm also calling createWorker in the wrong place!?

jankoegel:
I had no luck getting this PR to work in Firefox. @im-adithya please check with your Firefox whether you can get it running. I managed to load the `legacy` version of the script here (WIP code):
https://github.com/getAlby/lightning-browser-extension/tree/task-qr_jans-sandbox

this Codepen works for me in Firefox, so there has to be a way:
https://codepen.io/jankoegel/pen/yLEmGxJ?editors=1111

src/manifest.json
```diff
@@ -26,7 +26,7 @@
     "*://*/*"
   ],
 
-  "content_security_policy": "script-src 'self'; object-src 'self'",
```
jankoegel, line 29: reason:
https://bugzilla.mozilla.org/show_bug.cgi?id=1294996#c5

https://getalby.slack.com/archives/C02TD6P5XPE/p1671564553376349?thread_ts=1671514296.425669&cid=C02TD6P5XPE

src/app/screens/Settings.tsx
```diff
@@ -42,8 +42,18 @@ function Settings() {
     setNostrPrivateKey(priv ?? "");
   };
 
+  const checkCameraPermissions = async () => {
+    const devices = await navigator.mediaDevices.enumerateDevices();
+    const cameraDevices = devices
+      .filter((device) => device.kind === "videoinput")
+      .filter((videoDevice) => videoDevice.label !== "");
+    cameraDevices.filter((device) => device.kind === "videoinput");
```
jankoegel, line 50: Why do you filter for `device.kind === "videoinput"` twice?

And can't you simply do:

```js
devices.filter(device => device.kind === "videoinput" && device.label !== "")
```
!?

jankoegel: ah, another problem: this line actually doesn't change `cameraDevices` and you can just delete it. Leftover from debugging!?

jankoegel: on my machine I have no device where `device.kind === "videoinput" && device.label !== ""`!?
Checked on the Codepen demo I wrote and where I've given camera access permissions I get:
```js
kind: "videoinput"
label: ""
```
from `navigator.mediaDevices.enumerateDevices()`

src/manifest.json
jankoegel: 👏 great find @im-adithya :D

src/app/screens/Settings.tsx
im-adithya: Looks like firefox has an issue with iframes asking for permissions, I copied the code from your pen and ran it locally (index.html and index.js), and it worked fine:

(before allowing and after allowing:)

Screenshot 2022-12-21 at 12 17 49 PM

im-adithya: > devices.filter(device => device.kind === "videoinput" && device.label !== "")

makes sense!

> Leftover from debugging!?

Oops yes! removing it

jankoegel: ah, thanks for rebuilding the codepen locally to check that the camera is there 👍

jankoegel (approved):

src/manifest.json
reneaaron: What are the security implications of changing the CSP? Now an attacker could execute any code that he manages to get into a a blob and add to the page?

src/manifest.json
```diff
@@ -38,6 +38,7 @@
       "id": "extension@getalby.com"
     }
   },
+  "__firefox__content_security_policy": "script-src 'self' blob:; object-src 'self'",
```
escapedcat, line 41: Maybe put this right below the default entry to make clear that we have both

im-adithya: Done!

im-adithya:
Closing this because of CSP issues

PR #1872, opened 2022-12-14
feat: new ui for tipping options on publisher card

reneaaron:

Users seem to confuse the "send sats" button on the publisher card, also we want to make tipping more prominent in the app if there is tipping information available for the current site. 

### Open tasks
 - [ ] Decide about icons / emojis and their respective values
 - [ ] Implement proper buttons (currently they are a hack)
 - [ ] Take users to the tipping flow with the amount already filled in
 - [ ] Add a custom amount option 
 - [ ] Think about re-using them below the amount field (quick selection of amounts)

image

### How has this been tested?

-

### Checklist

### DIFF
--- a/src/app/screens/Home/PublisherLnData.tsx
+++ b/src/app/screens/Home/PublisherLnData.tsx
@@ -1,9 +1,9 @@
-import Button from "@components/Button";
 import PublisherCard from "@components/PublisherCard";
 import { FC, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import { toast } from "react-toastify";
+import Button from "~/app/components/Button";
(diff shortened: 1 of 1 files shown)


qqqzhch:
great try

### This is the original view
screenshot-20221218-093717
### the view  click after send sats
screenshot-20221218-093814
### After clicking coffee(1k)pizza(5k) in the new view, does the page view change?

### For the current interaction, see the text introduction. 

In the wallet, there is only the website introduction, not the text introduction or description related to tips.

It would be better to add some introductions and links related to the tip module

screenshot-20221218-094926

New users should see that the color of the wallet icon has changed, so they will click to open the wallet to take a look. If there is no introduction to the tip module here, they probably don’t know what the relationship between the tip module and the current page is.

Issue #1753, opened 2022-11-14
Add tests for the ln/request.ts 


bumi:
### Feature description

- How do we best test this action? 
- How do we best mock the `requestMethod` function on the connector?

bumi:
@lisabaut is this something you want to look at

PR #1875, opened 2022-12-15
fix: lnurl auth return value cause  error msg

qqqzhch:
### Describe the changes you have made in this PR

When the result returned by the lnurl auth api is missing a field, the prompt should be Auth: Something went wrong

### Link this PR to an issue [optional]

- Fixes #1874

### Type of change

(Remove other not matching type)

- `fix`: Bug fix (non-breaking change which fixes an issue)

### Screenshots of the changes [optional]
For example, the status field is missing
WX20221215-131744@2x

after fix 
WX20221215-134759

### How has this been tested?

_Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration_

### Checklist

- [x] My code follows the style guidelines of this project and performed a self-review of my own code
- [x] New and existing tests pass locally with my changes
- [x] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

PR #1880, opened 2022-12-15
feat: add tests for ln request

lisabaut:

Add tests for LN request

Closes https://github.com/getAlby/lightning-browser-extension/issues/1753

### Checklist

### DIFF
--- a/src/extension/background-script/actions/ln/__tests__/request.test.ts
+++ b/src/extension/background-script/actions/ln/__tests__/request.test.ts
@@ -160,6 +160,9 @@ describe("ln request", () => {
         message.args.method.toLowerCase(),
         message.args.params
       );
+
+      expect(utils.openPrompt).not.toHaveBeenCalled();
+
       expect(result).toStrictEqual(requestResponse);
     });
   });
--- a/src/extension/background-script/actions/ln/__tests__/request.test.ts
+++ b/src/extension/background-script/actions/ln/__tests__/request.test.ts
@@ -1,18 +1,28 @@
 import utils from "~/common/lib/utils";
 import type Connector from "~/extension/background-script/connectors/connector.interface";
 import db from "~/extension/background-script/db";
-import state, { State } from "~/extension/background-script/state";
 import type { MessageGenericRequest, OriginData } from "~/types";
 
 import request from "../request";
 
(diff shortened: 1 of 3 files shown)


src/extension/background-script/actions/ln/__tests__/request.test.ts
```diff
@@ -0,0 +1,326 @@
...
+  return connector;
+});
+
+jest.mock("~/extension/background-script/state", () => ({
+  getState: () => ({
+    getConnector: jest.fn(() => Promise.resolve(new ConnectorClass())),
+  }),
+}));
+
+const allowanceInDB = {
```
escapedcat, line 27: We don't type this because the function will complain if it's the wrong type?

src/extension/background-script/actions/ln/__tests__/request.test.ts
```diff
@@ -0,0 +1,326 @@
...
+    test("if connector does not support requestMethod", async () => {
+      connector = {
+        ...fullConnector,
+        supportedMethods: ["getinfo"],
+      };
+
+      const result = await request(message);
+
+      expect(console.error).toHaveBeenCalledTimes(1);
+      expect(result).toStrictEqual({
```
escapedcat, line 94: Can something like this be used to expect an error?
https://github.com/getAlby/lightning-browser-extension/pull/1875/files#diff-54f3405772ff247a2581cbc399cfb41599e415a405be1ba9c4a305e29b14ada6R50-R52

escapedcat (commented):
Just some comments/questions

src/extension/background-script/actions/ln/__tests__/request.test.ts
lisabaut: If I get you correctly:
Both the allowance and permission are typed.
If I would remove a key/value from one of them TS will complain

src/extension/background-script/actions/ln/__tests__/request.test.ts
lisabaut: I don't think so.
`request` will always return an object in the catch block.
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/ln/request.ts#L86-L92
and not a JavaScript Error

PR #1883, opened 2022-12-16
feat: always focus prompt

### LINKED ISSUE #1856: Focus the prompt
### Feature description

Sometimes it happens that the prompt looses focus and is hidden behind the actual browser window.

Ideally the prompt is always on top - or gets focus back after some time. 

### Describe the solution

Seems we can do `browser.window.update()` with `focused` and/or `drawAttention` in [here](https://github.com/getAlby/lightning-browser-extension/blob/master/src/common/lib/utils.ts#L77). Maybe with a setInterval?

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/update

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

bumi:

This makes sure that the prompt is focussed every 2 seconds. 
If the user "looses" the prompt this change brings the prompt in focus again.

Fixes #1856

### Type of change

### Video

https://share.cleanshot.com/ZmEgc6

### DIFF
--- a/src/common/lib/utils.ts
+++ b/src/common/lib/utils.ts
@@ -45,7 +45,7 @@ const utils = {
   openUrl: (url: string) => {
     browser.tabs.create({ url });
   },
-  openPrompt: <Type>(message: {
+  openPrompt: async <Type>(message: {
     args: Record<string, unknown>;
     origin: OriginData | OriginDataInternal;
     action: string;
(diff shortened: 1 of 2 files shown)


reneaaron (approved):
tACK

src/common/lib/utils.ts
```diff
@@ -45,7 +46,7 @@ const utils = {
   openUrl: (url: string) => {
     browser.tabs.create({ url });
   },
-  openPrompt: <Type>(message: {
+  openPrompt: async <Type>(message: {
```
bumi, line 49: does this need to be async? if this is async I think then line 75 should not `return new Promise`

reneaaron: We need to await the async function getWindowPosition, therefore it has to be async.

https://github.com/getAlby/lightning-browser-extension/pull/1883/files#diff-4bc1a33d21997d321e25376ba8b1376f5a25afc2b6fb9450515547c9aec95d66R73

bumi:
```
return getWindowPosition(windowWidth, windowHeight).then(dimensions => {

})
```

PR #1888, opened 2022-12-19
fix: allowances enable unit test fixture data

qqqzhch:

Edit the allowance enable test  fixture data

When checking the data, the data with id 2 is used, so there can only be one item fixture data
```
const allowance = await db.allowances.get(2);
```

### Type of change

### How has this been tested?

### Checklist

### DIFF
--- a/src/extension/background-script/actions/allowances/__tests__/enable.test.ts
+++ b/src/extension/background-script/actions/allowances/__tests__/enable.test.ts
@@ -22,7 +22,7 @@ utils.openPrompt = jest
   .fn()
   .mockReturnValue({ data: { enabled: true, remember: true } });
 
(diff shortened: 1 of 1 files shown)


escapedcat:
Ah, thanks! So the test would always have been green because there were already 2 items, right?

qqqzhch:
> Ah, thanks! So the test would always have been green because there were already 2 items, right?
yes， the second item of the old test data can be verified by the test, so the test is always green

In the previous test data, the enbale attribute of the second  item is true, which can be verified. When inserting a new data during test execution, the new  item is the third, but the second  item is still obtained for verification, test is always green,so the test is skipped

According to the test, the newly inserted item is the second item, so only one item is prepared for the test data

escapedcat (approved):

Issue #1854, opened 2022-12-10
[Feature] When the user click on Incoming tab take some seconds to load data. Should show instantly.


bitcoinuser:
### Feature description

For now, when the user click on Incoming tab take some seconds to load the transactions. I think Alby could cache the data of incoming transactions, so show instantly to the user when clicking on incoming tab. And a loading icon to the user know that Alby is still looking for new transactions.

### Describe the solution

I think Alby could cache the data of incoming transactions, so show instantly to the user when clicking on incoming tab. And a loading icon to the user know that Alby is still looking for new transactions.

### Are you working on this?

No

bitcoinuser:
Related issues: 

https://github.com/getAlby/lightning-browser-extension/issues/1536
https://github.com/getAlby/lightning-browser-extension/issues/1674

escapedcat:
We could:

1. Load incoming txs as soon as popup is being opened and cache
2. Show cached txs whn "incoming" tab is clicked, load current txs again, update view and cache these

* if account is being switched we need to behave like the popup is opened for the first time

qqqzhch:
The incoming data is a bit of slow. Although my incoming data does not contain any data, I will be familiar with this query logic

Issue #1850, opened 2022-12-09
[Feature] Split "Your ⚡️ Websites" and "Other ⚡️ Websites" into "Discover" (Other) and "Visited" (Yours)


escapedcat:
### Feature description

- Split "Websites" screen into 2
     ![image](https://user-images.githubusercontent.com/1016218/206696387-9019c074-0c93-4613-840e-fd3b6d372dee.png)

### Describe the solution

- update nav-items (tab and popup view)
- split content
- not sure yet how to handle the closable boxes section now (could be removed if all boxes closed i.e.? needs to be discussed maybe)

rolznz:
@escapedcat 

> not sure yet how to handle the closable boxes section now (could be removed if all boxes closed i.e.? needs to be discussed maybe)

I think that is a good idea. Would it be better to add these cards just on the discover view? And maybe not show the "visited" tab if you have no visited sites? Alternatively, it could have an empty state which says something like "You haven't used Alby on any sites yet. Click here to discover sites where you can use Alby"

escapedcat:
> Alternatively, it could have an empty state which says something like "You haven't used Alby on any sites yet. Click here to discover sites where you can use Alby"

I prefer this approach to keep the navigation bar simple and clear. Wdyt @reneaaron ?

reneaaron:
+1 for having an empty state there (instead of hiding navigation items)

PR #1892, opened 2022-12-19
feat(options): add closeable tab actions on Websites screen #1849

### LINKED ISSUE #1849: [Feature] Add closeable tab actions on "Websites" screen
### Feature description

- Add closable boxes if no "visited" site exists (should only show on new install)
   image

### Describe the solution

- this area (Headline, text, boxes) only shows up if list of websites is empty
- boxes can be closed
- the exact content of the boxes is TBD
    - topup-box: we only show one button to a guide for now 
       (as added here:  #1432)
    - start with basic text. icons can be added later if this is more complicated for a reason.
- the "lightning-address" box will NOT be added for now
- be aware of "out of scope" items, see #1845

jamescharlesworth:

This PR adds closeable tabs to the websites / options screen. Note: In order to persist state it is utilizing the `useSettings` hook and state by adding `setting.tips` array. I can refactor that if needed as its not exactly settings.  

Fixes #1849 

### Type of change

Screen Shot 2022-12-19 at 9 29 32 AM

### How has this been tested?

I have added a test for a new CloseableCard component.

### Checklist


escapedcat:
Thanks James! We'll have a look asap.

jamescharlesworth:
Just realized i left off the demo link on the 3rd card and pushed the [fix](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/9b64d0430fa2e2a26d5b80610048c35fefd10691).

src/i18n/locales/cs/translation.json
```diff
@@ -571,6 +571,32 @@
         "description": "",
         "no_info": ""
       },
+      "tips": {
```
escapedcat, line 574: No need to add these empty ones. [Weblate](https://hosted.weblate.org/git/getalby-lightning-browser-extension/getalby-lightning-browser-extension/) will take care of these.

escapedcat:
Dark view, just fyi:
image

src/app/screens/Publishers/index.tsx
```diff
@@ -14,7 +17,16 @@ function Publishers() {
...
 
+  const { settings, updateSetting } = useSettings();
+
   const [publishers, setPublishers] = useState<Publisher[]>([]);
+  const tips = settings.tips || [];
+
+  const hasTips = tips.length > 0;
+
+  function hasTip(id: string) {
+    return tips.indexOf(id) > -1;
```
escapedcat, line 28: Does [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) work here?

src/app/screens/Publishers/index.tsx
```diff
@@ -90,21 +102,103 @@ function Publishers() {
...
-      </p>
-
-      {publishers.length > 0 ? (
-        <PublishersTable
-          publishers={publishers}
-          navigateToPublisher={navigateToPublisher}
-        />
-      ) : (
-        <p className="dark:text-white"> {t("used.no_info")}</p>
+      {(publishers.length > 0 || !hasTips) && (
```
escapedcat, line 105: Does this mean a user needs to dismiss all tips first to see already used websites?

src/types.ts
```diff
@@ -628,6 +628,7 @@ export interface SettingsStorage {
   exchange: SupportedExchanges;
   debug: boolean;
   nostrEnabled: boolean;
+  tips: string[];
```
escapedcat, line 631: Would it be better to restrict to actual possible strings here?

escapedcat (commented):
Looks great! Thanks for this! Some minor comments.

I guess we need to make sure we're happy with the copy-text and links.
Other thoughts:
- Do we need to show the pinning-tip for Firefox?
- Should we just leave the "topup"-link even though people might use their own node? If not, how to deal with this?

Let me add some other people as reviewers to get needed feedback

src/app/screens/Publishers/index.tsx
escapedcat: This might just be irrelevant because we gonna handle this later in combination with #1905  
So no need to worry now, right @rolznz ? :)

escapedcat:
Later on we should also think about how our "secondary" button looks like:
#1905 is doing this:
image

Here it's doing:

image
image

Or are these different cases and we're good here?

src/i18n/locales/cs/translation.json
jamescharlesworth: oh, thats cool. I've removed them.

src/app/screens/Publishers/index.tsx
jamescharlesworth: yeah, that reads better. done.

src/app/screens/Publishers/index.tsx
jamescharlesworth: websites and `publishers` are the same thing, right?  Since it's an `||`, `publishers.length` (when they have websites/publishers) would trigger the block to show.

src/types.ts
jamescharlesworth: good call out. done.

jamescharlesworth:
> Later on we should also think about how our "secondary" button looks like: #1905 is doing this: image
> 
> Here it's doing:
> 
> image image
> 
> Or are these different cases and we're good here?

I noticed this too but didnt think it would be in scope for this pr (unless im implementing the button incorrectly here in some way)

src/app/screens/Publishers/index.tsx
escapedcat: Ah, thanks. Didn't read that in detail.

src/types.ts
escapedcat: Not sure if an `enum` is even needed maybe just an array of defined strings would be enough? Not exactly sure here. But thanks for adjusting!

escapedcat:
> I noticed this too but didnt think it would be in scope for this pr (unless im implementing the button incorrectly here in some way)

Right, not in scope. Just a thing we need to keep an eye on 👍

src/app/screens/Publishers/index.tsx
```diff
@@ -90,21 +103,107 @@ function Publishers() {
...
+      )}
+      {publishers.length === 0 && hasTips && (
+        <>
+          <h2 className="mt-12 mb-2 text-2xl font-bold dark:text-white">
+            {t("tips.title")}
+          </h2>
+          <p className="mb-6 text-gray-500 dark:text-neutral-500">
+            {t("tips.description")}
+          </p>
+          <div className="grid gap-4 sm:grid-cols-1 lg:grid-cols-2">
```
rolznz, line 133: It could be good to use an array and map (some things like handleClose() would not need to be duplicated)

src/app/screens/Publishers/index.tsx
```diff
@@ -90,21 +103,107 @@ function Publishers() {
...
+              <CloseableCard
+                handleClose={() =>
+                  updateSetting({
+                    tips: tips.filter((tip) => tip !== TIPS.ADDRESS),
+                  })
+                }
+                title={t("tips.address.title")}
+                description={t("tips.address.description")}
+                buttons={[
+                  <a key={1} href="https://getalby.com/demo">
```
rolznz, line 199: Is this link right? the address tip goes to the demo

escapedcat: The whole ln-address handling is unclear currently. This box is currently not activated.

src/app/screens/Publishers/index.tsx
jamescharlesworth: i thought about that but there is a lot good bit of differences in the card layouts such as showing a numbered list, different buttons in the footer, etc. I can do it in a single loop but there will likely be added complexity to determine things like which buttons to show inside the loop. Although the non-loop approach is more verbose it felt less complex and more adaptable to future changes.  Happy to do it either way.

src/app/screens/Publishers/index.tsx
jamescharlesworth: i can remove the address stuff from the PR if desired.  Just put it here as a placeholder for when its ready.

src/app/screens/Publishers/index.tsx
escapedcat: I kinda held myself back to ask for some loop and json-file because I thought it's really not necessary yet and who knows where this is going. Further thoughts where things like "markdown" and "oh, now we're building a CMS" :D

jankoegel:
idea for improvement of the "pin" box:
Add the actual icon since not all users will know what "the extension icon" is. In my experience, it's very easy to overlook.
Screenshot 2022-12-27 at 15 35 45

I don't use Chrome and recently tried to set up an adblocker on a friend's machine. Took me ages to find the pinning feature...

tests/e2e/002-walletFeatures.spec.ts
```diff
@@ -39,7 +39,7 @@ test.describe("Wallet features", () => {
       extensionId,
     });
 
-    await findByText($optionsdocument, "Your ⚡️ Websites");
+    await findByText($optionsdocument, "Your Alby wallet is ready");
```
reneaaron, line 42: Shouldn't `Your ⚡ Websites` be on this site, too?

reneaaron (changes requested):
Thanks for your contribution @jamescharlesworth, great job! 💪

I just tested the PR and noticed some flickering when changing the page: 

https://user-images.githubusercontent.com/100827540/209686158-ef34a9cb-bc4a-4cdd-879f-6e00709b9f41.mp4

Currently every user would see the same boxes, right? Did you also think about how we could conditionally display some of them? 

1) Pinning is only relevant for Chrome (default on Firefox)
2) Topup is only relevant for getalby.com users

tests/e2e/002-walletFeatures.spec.ts
escapedcat: Will be changed with this: #1905

escapedcat:
> Currently every user would see the same boxes, right? Did you also think about how we could conditionally display some of them?

@reneaaron I remember @bumi was worried about too many if/else cases in there? If that's the case, we could maybe adjust the copy-text to make this more clear?

jamescharlesworth:
> Thanks for your contribution @jamescharlesworth, great job! 💪
> 
> I just tested the PR and noticed some flickering when changing the page:

Great catch. One option to address this is to breakout the [async fetch of `listAllowances`](https://github.com/getAlby/lightning-browser-extension/blob/8eb1afe6fe2b9b25ecb32ff380d84b3e5f402ffb/src/app/screens/Publishers/index.tsx#L37-L102) to its own hook, call it from the parent `Options.tsx` and pass in publishers as a prop. This way it wont fetch on every tab change.  The drawback here (performance improvement?) is that the websites would not be fetched on every tab change.  That seems like an edge case to me and may be acceptable.  If not, any recommendations on how to solve this?

> Pinning is only relevant for Chrome (default on Firefox)

I couldnt find an existing examples in this code base were it inspects browser and does logic based on it.  Is there a best practice for identifying chrome?

> Topup is only relevant for getalby.com users

Is this the correct way to identify getAlby.com account?
```js
const {
    account,
  } = useAccount();
if (account?.alias === '🐝 getalby.com') {
 ...
}
```
^^ seems a a bit brittle/verbose

jamescharlesworth:
> Great catch. One option to address this is to breakout the [async fetch of listAllowances](https://github.com/getAlby/lightning-browser-extension/blob/8eb1afe6fe2b9b25ecb32ff380d84b3e5f402ffb/src/app/screens/Publishers/index.tsx#L37-L102) to its own hook, call it from the parent Options.tsx and pass in publishers as a prop. This way it wont fetch on every tab change. The drawback here (performance improvement?) is that the websites would not be fetched on every tab change. That seems like an edge case to me and may be acceptable. If not, any recommendations on how to solve this?

I think i have a solution to this without adding the prop from the parent.

rolznz:
> > Thanks for your contribution @jamescharlesworth, great job! 💪
> > I just tested the PR and noticed some flickering when changing the page:
> 
> Great catch. One option to address this is to breakout the [async fetch of `listAllowances`](https://github.com/getAlby/lightning-browser-extension/blob/8eb1afe6fe2b9b25ecb32ff380d84b3e5f402ffb/src/app/screens/Publishers/index.tsx#L37-L102) to its own hook, call it from the parent `Options.tsx` and pass in publishers as a prop. This way it wont fetch on every tab change. The drawback here (performance improvement?) is that the websites would not be fetched on every tab change. That seems like an edge case to me and may be acceptable. If not, any recommendations on how to solve this?
> 
> > Pinning is only relevant for Chrome (default on Firefox)
> 
> I couldnt find an existing examples in this code base were it inspects browser and does logic based on it. Is there a best practice for identifying chrome?
> 
> > Topup is only relevant for getalby.com users
> 
> Is this the correct way to identify getAlby.com account?
> 
> ```js
> const {
>     account,
>   } = useAccount();
> if (account?.alias === '🐝 getalby.com') {
>  ...
> }
> ```
> 
> ^^ seems a a bit brittle/verbose

In other projects I have used SWR which will show a cached version while fetching any changes.

It looks like there is some custom code in lib/api and lib/cache to get account info. It would be cool to consider introducing SWR, using it for all fetching and remove the custom implementation that's only used on a single call. @escapedcat what do you think? I can create an issue for this?

reneaaron:
> In other projects I have used SWR which will show a cached version while fetching any changes (as well as many other useful features).
> It looks like there is some custom code in lib/api and lib/cache to get account info. It would be cool to consider introducing SWR, using it for all fetching and remove the custom implementation that's only used on a single call. @reneaaron @escapedcat what do you think? I can create an issue for this?

Thanks for bringing that up. Sure, let's create an issue for that and discuss it in there. 👌

escapedcat:
> I just tested the PR and noticed some flickering when changing the page:

I don't see this as blocking necessarily. 

> In other projects I have used SWR which will show a cached version while fetching any changes (as well as many other useful features).

We currently use SWR here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/common/lib/api.ts/#L56-L86

This could possibly also be a soltution for #1854

>  I can create an issue for this?

👍

jamescharlesworth:
I was going down a similar caching direction and created a [PR](https://github.com/jamescharlesworth/lightning-browser-extension/pull/1/files) into my own branch with solution if you want to review? Looks like the main difference between the SWR implementation is that it would go into browser storage where mine just uses an in memory object.

jamescharlesworth:
Still need to add tests but [I broke out the tips into its own component](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/e34fa95d01aa454754f7e7a866bbbcb408a7bd59) as the business logic may grow and will keep it isolated. This commit also includes logic for [checking that the browser chrome](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/e34fa95d01aa454754f7e7a866bbbcb408a7bd59#diff-792f8b2e81810f3e99e74179e35dcfbcd623ff5d66b44c668701e4c6f2b83781R13-R15) and [checking the account is an Alby account](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/e34fa95d01aa454754f7e7a866bbbcb408a7bd59#diff-792f8b2e81810f3e99e74179e35dcfbcd623ff5d66b44c668701e4c6f2b83781R9-R11) (again not sure this is the best check).

src/app/hooks/useTips.ts
```diff
@@ -0,0 +1,37 @@
+import { useAccount } from "~/app/context/AccountContext";
```
escapedcat, line 1: Sorry what was the reasoning behind adding this hook? To encapsulate the tips-code? Asking mainly to get a better understanding. Thought we use hooks to share code among different components but the tips aren't used somewhere else, right?

jamescharlesworth: Yes, to encapsulate the tips code but was also necessary to break when I moved `<Tips />` to its own component. The additional business logic of which tips to show (chrome browser, is Alby account, etc) is needed to be run both when looping through the tips [in the Tips component](https://github.com/getAlby/lightning-browser-extension/pull/1892/files#diff-792f8b2e81810f3e99e74179e35dcfbcd623ff5d66b44c668701e4c6f2b83781R15) but also in [`Publisher/index.tsx`](https://github.com/getAlby/lightning-browser-extension/pull/1892/files#diff-abf596d81b784f3eb9e9e7be5814d0a585cfa545c1fe3142071fdbc59e1e9e33R20) to do logic on showing messaging around the titles (if has tips show x...).  The original way had all this logic in `Publisher/index.tsx` but it was really starting to grow and seemed to make more sense to break it out.  Happy to move it back into there though if you think that makes more sense.

jamescharlesworth:
I've also created a PR into this branch on my fork ([feat/issue-1849-cache-fix -> feat/issue-1849 (this pr)](https://github.com/jamescharlesworth/lightning-browser-extension/pull/1)) with swr implementation [around the publishers](https://github.com/jamescharlesworth/lightning-browser-extension/blob/feat/issue-1849-cache-fix/src/app/hooks/usePublishers.ts).  Should I merge it into this or is that something to work into the new issue?

escapedcat:
> I've also created a PR into this branch on my fork ([feat/issue-1849-cache-fix -> feat/issue-1849 (this pr)](https://github.com/jamescharlesworth/lightning-browser-extension/pull/1)) with swr implementation...

Thanks! Solution seems to work well. But I'm not sure if we need this once we separated websites and discover via #1850, right?

reneaaron:
I was just thinking of the scenario of adding some more actions in future: 

The way it's currently implemented that would require us to migrate the settings (that contain the available actions) and add the new action in there. 

I wondered if it would be easier maybe if we would just keep track of the actions that should be hidden. That way we could add new actions any time and wouldn't need to migrate anything. 

@jamescharlesworth What do you think?

jamescharlesworth:
> I was just thinking of the scenario of adding some more actions in future:
> 
> The way it's currently implemented that would require us to migrate the settings (that contain the available actions) and add the new action in there.
> 
> I wondered if it would be easier maybe if we would just keep track of the actions that should be hidden. That way we could add new actions any time and wouldn't need to migrate anything.
> 
> @jamescharlesworth What do you think?

sounds good. I made the changes but now need to to resolve the merge conflicts from a recent commit. Will take a look at that tomorrow.

jamescharlesworth:
I tried my best to make some choices between the messaging states from the merge.  Any suggestions with this messaging based on state?

No websites and tips not yet cleared
Screen Shot 2023-01-04 at 7 41 29 AM

Websites and tips not yet cleared.
Screen Shot 2023-01-04 at 7 57 11 AM


All Tips cleared and websites
Screen Shot 2023-01-04 at 7 36 39 AM

All tips cleared and no websites
Screen Shot 2023-01-04 at 7 47 06 AM

rolznz:
> idea for improvement of the "pin" box: Add the actual icon since not all users will know what "the extension icon" is. In my experience, it's very easy to overlook. Screenshot 2022-12-27 at 15 35 45
> 
> I don't use Chrome and recently tried to set up an adblocker on a friend's machine. Took me ages to find the pinning feature...

I agree it would be great to have an icon @jamescharlesworth are you able to add this?

rolznz:
Should this link open in a new tab? or are we ok for the websites tab to be replaced by the demo?

image

escapedcat:
e2e tests fail:

```sh
Error: Evaluation failed: Error: Unable to find an element with the text: Your ⚡️ Websites. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```

Should be easy-ish to fix.

rolznz:
> * Do we need to show the pinning-tip for Firefox?

> I couldnt find an existing examples in this code base were it inspects browser and does logic based on it. Is there a best practice for identifying chrome?


In case there is no existing implementation, Maybe it would be good to make a utility function (maybe in src/common/utils) so that we can change the implementation later in case it does not work in the future, maybe something like getBrowserType() === 'chrome' and just check for firefox and chrome for now? https://stackoverflow.com/a/63964511/4562693 seems like an OK answer to do this (I tested it in both FF and chrome extension consoles)

jamescharlesworth:
> In case there is no existing implementation, Maybe it would be good to make a utility function (maybe in src/common/utils) so that we can change the implementation later in case it does not work in the future, maybe something like getBrowserType() === 'chrome' and just check for firefox and chrome for now? https://stackoverflow.com/a/63964511/4562693 seems like an OK answer to do this (I tested it in both FF and chrome extension consoles)

[Updated](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/1478be244153909478381280eaf5899f44241efc#diff-8dfff597836add571f616579babc7bf3dab9bd5f2901725ffed88bb389664757R33).

> e2e tests fail:
> Error: Evaluation failed: Error: Unable to find an element with the text: Your ⚡️ Websites. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
>Should be easy-ish to fix.

Fixed.

> Should this link open in a new tab? or are we ok for the websites tab to be replaced by the demo?
[updated](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/12957d0d82acbf23768dc84db7848f4be9f1e575#diff-792f8b2e81810f3e99e74179e35dcfbcd623ff5d66b44c668701e4c6f2b83781R69).

>idea for improvement of the "pin" box: Add the actual icon since not all users will know what "the extension icon" is. In my experience, it's very easy to overlook. Screenshot 2022-12-27 at 15 35 45

@rolznz do we have an icon for the extension/plugin puzzle piece? I couldnt find one.

reneaaron:
Thanks for all your updates! 💪

I just moved the tips to the "Discover" page (that is the page users will soon be redirected to after the setup). They should see the actions from this PR + all the apps below. 

Somehow the e2e tests seem to fail now. 🤔 Any chance @escapedcat can take a look?

rolznz:
> @rolznz do we have an icon for the extension/plugin puzzle piece? I couldnt find one.

I couldn't find the exact one, but I was able to inspect this one from chrome://settings/system 

Maybe we can use it (it's close enough and SVG?) - I think it's better than screenshotting the actual icon and getting a poor quality PNG?

```
<svg viewBox="0 0 24 24" preserveAspectRatio="xMidYMid meet" focusable="false" style="pointer-events: none; display: block; width: 100%; height: 100%;"><g>
        <path d="M20.5 11H19V7c0-1.1-.9-2-2-2h-4V3.5C13 2.12 11.88 1 10.5 1S8 2.12 8 3.5V5H4c-1.1 0-1.99.9-1.99 2v3.8H3.5c1.49 0 2.7 1.21 2.7 2.7s-1.21 2.7-2.7 2.7H2V20c0 1.1.9 2 2 2h3.8v-1.5c0-1.49 1.21-2.7 2.7-2.7 1.49 0 2.7 1.21 2.7 2.7V22H17c1.1 0 2-.9 2-2v-4h1.5c1.38 0 2.5-1.12 2.5-2.5S21.88 11 20.5 11z">
        </path>
      </g></svg>
```

jamescharlesworth:
There is a lot I don't really like about [this recent commit](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/c0291182812d21489f8b03430c38d16556571574) with fix to puzzle but works to inline the image inside the translation text. Any recommendations on a cleaner approach? Thanks!

rolznz:
> There is a lot I don't really like about [this recent commit](https://github.com/getAlby/lightning-browser-extension/pull/1892/commits/c0291182812d21489f8b03430c38d16556571574) but works to inline the extension svg inside the translation text. Any recommendations on a cleaner approach? Thanks!

I think you might be able to use the Trans component https://react.i18next.com/latest/trans-component

Let me know if you can't get it working.

jamescharlesworth:
>I think you might be able to use the Trans component https://react.i18next.com/latest/trans-component
> Let me know if you can't get it working.

@rolznz I don't think the Trans component supports this type functionality but i could be wrong.

rolznz:
> > I think you might be able to use the Trans component https://react.i18next.com/latest/trans-component
> > Let me know if you can't get it working.
> 
> @rolznz I don't think the Trans component supports this type functionality but i could be wrong.

What exactly doesn't it support?

I found a codesandbox and forked it to add an img element https://codesandbox.io/s/react-i18next-trans-component-example-forked-r6jqii

Maybe the way the structure of the description in the props would need to be changed so that you can pass the icon somehow.

Would this work?

jamescharlesworth:
Thanks @rolznz that helped me figure it out. Does the [translation string](https://github.com/getAlby/lightning-browser-extension/pull/1892/files#diff-97b816b73032872750fcb1fc269ee9cb095d3c41d356dedaeb277edaf6ab57a6R627) look correct?

Also, it seems a test now is failing (on my local) not being able to find text `"Your ⚡️ Websites` that is there:
Screen Shot 2023-01-09 at 7 11 25 AM. Have you seen this before?

jamescharlesworth:
@rolznz ignore above question about test, seems to be whitespace related and have fixed it.

rolznz:
> Thanks @rolznz that helped me figure it out. Does the [translation string](https://github.com/getAlby/lightning-browser-extension/pull/1892/files#diff-97b816b73032872750fcb1fc269ee9cb095d3c41d356dedaeb277edaf6ab57a6R627) look correct?

It looks great, nice work! I am just wondering if the text has to be passed in the <Trans> component since it is already provided in the English translation file. Maybe you can pass in a span or a React fragment instead of the text? if you look at the other usages of Trans in the app, we never put text in there, only the elements that will wrap the text.

rolznz:
> ```js
> const {
>     account,
>   } = useAccount();
> if (account?.alias === '🐝 getalby.com') {
>  ...
> }
> ```
> 
> ^^ seems a a bit brittle/verbose

I asked Bumi and Rene and they agreed. They thought it might be better to compare the hostname from the connection against "ln.getalby.com". I will see if there is a way to do this.

In the meantime could you add a TODO in the utility function you added to change the implementation? it definitely is brittle as this check will break if we change the alias.

rolznz:
> > ```js
> > const {
> >     account,
> >   } = useAccount();
> > if (account?.alias === '🐝 getalby.com') {
> >  ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ^^ seems a a bit brittle/verbose
> 
> I asked Bumi and Rene and they agreed. They thought it might be better to compare the hostname from the connection against "ln.getalby.com". I will see if there is a way to do this.
> 
> In the meantime could you add a TODO in the utility function you added to change the implementation? it definitely is brittle as this check will break if we change the alias.

Edit: @jamescharlesworth I talked with @bumi and it looks like it is actually better to create a whole new connector, so that we don't have to do hacks (check the hostname or the alias, both which could change) and also this new Alby connector could talk directly to the Alby Oauth API instead of Lndhub which brings other benefits too.

This is out of the scope of this issue, so I will create a new issue and link it to this one.

jamescharlesworth:
>>Thanks @rolznz that helped me figure it out. Does the [translation string](https://github.com/getAlby/lightning-browser-extension/pull/1892/files#diff-97b816b73032872750fcb1fc269ee9cb095d3c41d356dedaeb277edaf6ab57a6R627) look correct?
>
>It looks great, nice work! I am just wondering if the text has to be passed in the component since it is already provided in the English translation file. Maybe you can pass in a span or a React fragment instead of the text? if you look at the other usages of Trans in the app, we never put text in there, only the elements that will wrap the text.

That makes sense, great feedback, thanks!

src/app/hooks/useTips.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+};
+
+export const useTips = () => {
+  const { settings, updateSetting } = useSettings();
+  const { account } = useAccount();
+
+  const tips = filterTips(settings.closedTips, account?.alias);
+
+  const filterTip = (tip: TIPS) => {
```
rolznz, line 25: I think "closeTip" is more understandable than "filterTip"

rolznz:
@reneaaron @escapedcat @jamescharlesworth  I added one more comment, but this PR looks good to me now. 

I noticed that there is no way to get the tips back if you accidentally closed one, but maybe that could be a future improvement (some link to click to re-view the tips)

src/app/hooks/useTips.ts
jamescharlesworth: sounds good, fixed!

escapedcat:
nit: Puzzle icon looks liek a cut-out in dark-mode:
image
image

escapedcat (approved):
ack

src/app/components/Tips/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+        buttons={[
+          <Button
+            key={1}
+            label={t("top_up_wallet.label1")}
+            primary
+            onClick={() => {
+              navigate("/receive");
+            }}
+          />,
+          <a key={2} href="https://getalby.com/topup">
```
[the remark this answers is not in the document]
jamescharlesworth: added

reneaaron (commented):
tACK

Great job everyone 🚀 LGTM!

reneaaron (approved):

PR #1893, opened 2022-12-19
Feature/lnd routermc

reneaaron:

Adds routermc request to LND connector.

### Type of change


src/common/lib/utils.ts
```diff
@@ -45,7 +45,7 @@ const utils = {
   openUrl: (url: string) => {
     browser.tabs.create({ url });
   },
-  openPrompt: <Type>(message: {
+  openPrompt: async <Type>(message: {
```
bumi, line 48: if you return a promise you don't need this async. So I think it's using async OR returning a Promise.

reneaaron: Some unrelated changes seemed to slip into this branch. Removed this stuff again. (also please don't review draft PRs if not explicitly requested)

PR #1899, opened 2022-12-20
hacky legacy.min

jankoegel:
⚠️ hackt test for debugging with Adithya

### DIFF
--- a/src/app/router/Options/index.tsx
+++ b/src/app/router/Options/index.tsx
@@ -9,11 +9,1130 @@ import "~/i18n/i18nConfig";
 import Options from "./Options";
 
 // Make sure to bind modal to your appElement (https://reactcommunity.org/react-modal/accessibility/)
-Modal.setAppElement("#options-root");
+// Modal.setAppElement("#options-root");
 
 // Get the active theme and apply corresponding Tailwind classes to the document
-getTheme();
+// getTheme();
 
-const container = document.getElementById("options-root") as HTMLElement;
-const root = createRoot(container);
-root.render(<Options />);
+// const container = document.getElementById("options-root") as HTMLElement;
+// const root = createRoot(container);
+// root.render(<Options />);
+
+var $jscomp = $jscomp || {};
+$jscomp.scope = {};
+$jscomp.ASSUME_ES5 = !1;
+$jscomp.ASSUME_NO_NATIVE_MAP = !1;
+$jscomp.ASSUME_NO_NATIVE_SET = !1;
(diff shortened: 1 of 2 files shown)


src/app/router/Options/index.tsx
```diff
@@ -9,11 +9,1130 @@ import "~/i18n/i18nConfig";
...
+// getTheme();
 
-const container = document.getElementById("options-root") as HTMLElement;
-const root = createRoot(container);
-root.render(<Options />);
+// const container = document.getElementById("options-root") as HTMLElement;
+// const root = createRoot(container);
+// root.render(<Options />);
+
+var $jscomp = $jscomp || {};
```
jankoegel, line 21: inlined legacy.min.js here. At the bottom of the file is the same code as in my codepen here:
https://codepen.io/jankoegel/pen/yLEmGxJ

jankoegel:
moving to a branch diff instead, sorry, didn't want to confuse anyone by opening a PR.

jankoegel:
@im-adithya my latest attempt - with inlined `legacy.min.js` (could just as well load it via an `import`) and the same code that's run in this Codepen that works in Firefox:
https://codepen.io/jankoegel/pen/yLEmGxJ

My current suspicion for why it's not working:
1. (strong favourite) `qr-scanner-worker.min.js` contains the following code:
  https://github.com/nimiq/qr-scanner/blob/abcfe1bce2703721408d8ce7ebde94a359998506/qr-scanner-worker.min.js#L1
  It creates a new worker (which you can also do with an external `my-script.js` file) via a file blob. Before making this PR I tried inlining legacy.min.js in `options.html` and paste my code from Codepen there. Interestingly, that doesn't work in Firefox. Browser extensions must not use `<script>` tags in their HTML code. This also breaks in Chrome.
  ⚠️ However, Firefox throws the same kind of error for your unmodified `task-qr` branch on the LNDHub add page:
  > Content Security Policy: The page’s settings blocked the loading of a resource at blob:moz-extension://7b1dedb6-dc50-4d1f-9fd0-c5200b45e54e/a004402a-bdaf-45f8-9c0b-fed47a7b61d5 (“script-src”).
  With no clickable link or further information where exactly the error occurs. My suspicion is that you cannot create new workers in Firefox extensions like that?

2. something with our Webpack build process breaks stuff (vs Codepen) but somehow only for the QR scanner and only for Firefox.

jankoegel:
fix for our/very similar problem:
https://stackoverflow.com/a/64621646

PR #1909, opened 2022-12-22
feat(nostr): publish a nostr ready event

bumi:

similar to the webln:ready event this publishes a nostr:ready event for web apps to know when the window.nostr object is available.


reneaaron (approved):
utACK

Probably would need to document this stuff somewhere. 🤔 Time for nostr.guide?

PR #1913, opened 2022-12-22
chore: use greater than check

bumi:
to make sure the budget is bigger than the amount

see: #1910

### DIFF
--- a/src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
+++ b/src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
@@ -29,16 +29,12 @@ const sendPaymentOrPrompt = async (message: Message) => {
 };
 
 async function checkAllowance(host: string, amount: number) {
-  // if it is a 0 amount invoice we return false and prompt
-  if (amount <= 0) {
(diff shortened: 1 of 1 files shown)


src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
```diff
@@ -29,16 +29,12 @@ const sendPaymentOrPrompt = async (message: Message) => {
...
-  if (amount <= 0) {
-    return false;
-  }
   const allowance = await db.allowances
     .where("host")
     .equalsIgnoreCase(host)
     .first();
 
-  return allowance && allowance.remainingBudget >= amount;
+  return allowance && allowance.remainingBudget > amount; // check that the budget is higher than the amount. amount can be 0
```
escapedcat, line 37: That also means if my allowance is 100 I can't send 100, right? Just checking if we're ok with that.

bumi: you can, you will just be prompted.

escapedcat (approved):

PR #1916, opened 2022-12-23
feat: Add window.alby methods

AaronDewes:

This adds a `window.alby` API and adds an initial method to it: `window.alby.addAccount` allows websites to register an account in Alby.

Fixes _#ISSUE-NUMBER_

### Type of change

grafik

### How has this been tested?

Manually calling the function from the browser dev tools. I'll create a demo app later.

### Checklist

### Things that are still missing

- Accounts should be better validated before adding them
- Better type definitions for `window.alby` (Maybe with a custom library similar to the existing webln lib on npm?)
- Documentation

I opened this PR to gather feedback on this API. I already suggested it in the Alby telegram channel.


bumi:
thanks for this PR and tackling this! looks pretty good, I will review this asap

src/app/screens/ConfirmAddAccount/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+        </MemoryRouter>
+      );
+    });
+
+    expect(
+      await screen.findByText(
+        "This website wants to add a wallet of the following type to Alby:"
+      )
+    ).toBeInTheDocument();
+    expect(await screen.findByText("Test message")).toBeInTheDocument();
```
escapedcat, line 52: [This fails in CI](https://github.com/getAlby/lightning-browser-extension/actions/runs/3766670301/jobs/6403420007), did this work locally?

AaronDewes: I haven't ran the tests yet, I'm still working on this PR (I'll make some adjustments tomorrow)

src/extension/content-script/onend.js
```diff
@@ -18,7 +19,7 @@ const weblnCalls = [
   "webln/request",
 ];
 // calls that can be executed when webln is not enabled for the current content page
-const disabledCalls = ["webln/enable"];
+const disabledCalls = ["alby/addAccount", "webln/enable"];
```
bumi, line 22: I think I would still require the enable() call to be called before prompting the user to add an account. 

just to keep the pattern and th make sure the user is aware they "connect" with the website.

bumi (commented):
For more clear separation I am thinking about also splitting up the content script code into a specil alby script. 
then we would have one for webln, nostr, and alby. 
Eventhough this means some code duplication (which we can also refactor into some helpers likely) I think this separation makes it a bit clearer and easier to follow to flow.

src/extension/content-script/onend.js
AaronDewes: Should this be shared with WebLN or have a separate enabled state?

bumi:
@AaronDewes do you still have time for this? otherwise maybe @im-adithya could you also help here?

src/extension/content-script/onend.js
rolznz: The answer here seems to be yes. In nostr we have a separate enabled state.

rolznz:
> After chatting with @reneaaron we think it's best to move the content script for this in its own file - to have this separate. Then every provider has it's own corresponding content script and we are fully flexible there and those are independent of each other.
> 
> but we don't do an enable call there as we always prompt the user anyway.
> 
> wdyt? If ok, then we get this in the next release. :shipit:

Without an enable call the user will not be able to reject further requests. Is this ok?

rolznz:
I agree it would be good to split this into a separate file. With 3 files and much duplication I think it would be best to refactor this, but I guess could be done in a follow up MR. Maybe we can also migrate these files to Typescript.

Why are these files called "onend"?

reneaaron:
> I agree it would be good to split this into a separate file. With 3 files and much duplication I think it would be best to refactor this, but I guess could be done in a follow up MR. Maybe we can also migrate these files to Typescript.

👍

> Why are these files called "onend"?

They are added via at `document_end` in the manifest.json:

image

im-adithya:
To add an account, do
`await alby.enable()` and then
`await alby.addAccount("name", "lndhub", {lnAddress: "name@getalby.com", login: "xxxxx", password: "xxxx", url: "https://ln.getalby.com"})`

src/extension/ln/alby/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Adds a wallet to the user's Alby extension
+   *
+   * @param name The name of the account
+   * @param connector The connector to use
+   * @param config The config to pass to the connector
+   * @returns Nothing, throws if user rejects
+   */
+  addAccount(name: string, connector: string, config: Record<string, unknown>) {
```
rolznz, line 28: Is the name of the account something that the user/app will always have to provide? or could it be an optional parameter?

reneaaron: I was also thinking about letting the user enter the name of the account. But that could also be added later on maybe? :thinking:

reneaaron:
@im-adithya Can you please address the review feedback in this PR. Let's get this merged! :rocket:

im-adithya:
@reneaaron Done!

src/app/screens/ConfirmAddAccount/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  async function confirm() {
+    try {
+      setLoading(true);
+      const response = await msg.request(
+        "addAccount",
+        { name, connector, config },
+        { origin }
+      );
+      msg.reply(response);
+      setSuccessMessage(tCommon("success"));
```
reneaaron, line 40: I just tested it again but this success messages is never shown? cc @im-adithya

src/extension/ln/alby/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (!this.enabled) {
+      throw new Error("Provider must be enabled before calling getInfo");
```
reneaaron, line 30:
```suggestion
      throw new Error("Provider must be enabled before calling addAccount");
```

reneaaron (commented):
Reading the code it seems there should be a success message but there is none. 

I think that adding the success message (and delaying the popup from closing) could be tricky, so I would maybe just remove it for now. 

There are efforts to introduce success messages in #2171 that we could reuse here.

src/app/screens/ConfirmAddAccount/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  async function confirm() {
+    try {
+      setLoading(true);
+      const response = await msg.request(
+        "addAccount",
+        { name, connector, config },
+        { origin }
+      );
+      msg.reply(response);
```
reneaaron, line 39: I don't think we should return the `accountId` here. Consumers of this API won't be able to do anything with this. 

I think this method doesn't necessarily need to return anything as errors will raise an exception.

src/app/screens/ConfirmAddAccount/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      setSuccessMessage(tCommon("success"));
+    } catch (e) {
+      console.error(e);
+      if (e instanceof Error) toast.error(`${tCommon("error")}: ${e.message}`);
+    } finally {
+      setLoading(false);
+    }
+  }
+
+  function reject(e: React.MouseEvent<HTMLAnchorElement>) {
```
reneaaron, line 49: Cancelling the prompt doesn't work.

src/app/screens/ConfirmAddAccount/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    e.preventDefault();
+    msg.error(USER_REJECTED_ERROR);
+  }
+
+  function close(e: React.MouseEvent<HTMLButtonElement>) {
+    if (navState.isPrompt) {
```
reneaaron, line 55: Do we need this? You can't open this popup anywhere else but via prompt.

src/extension/ln/alby/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    return this.execute("addAccount", { connector, name, config });
+  }
+
+  execute(
+    action: string,
+    args?: Record<string, unknown>
+  ): Promise<Record<string, unknown>> {
+    return new Promise((resolve, reject) => {
```
reneaaron, line 39: Now that #2218 has been merged we could use `postMessage()` here, too.

reneaaron (approved):
Pushed some last improvements and did some final testing. Published some docs here: 

https://guides.getalby.com/alby-extension-apis/

tACK

PR #1922, opened 2022-12-28
Feat/i18n zh cn

escapedcat:
Chinese is at 84%: https://hosted.weblate.org/projects/getalby-lightning-browser-extension/getalby-lightning-browser-extension/  
Other languages we already support have a similar percentage.
像差不多吧

- switched 聪 to sat/sats, we shouldn't translate these

### DIFF
--- a/src/i18n/i18nConfig.ts
+++ b/src/i18n/i18nConfig.ts
@@ -3,6 +3,7 @@ import "dayjs/locale/es";
 import "dayjs/locale/it";
 import "dayjs/locale/pt-br";
 import "dayjs/locale/sv";
+import "dayjs/locale/zh-cn";
 import i18n from "i18next";
 import LanguageDetector from "i18next-browser-languagedetector";
 import { initReactI18next } from "react-i18next";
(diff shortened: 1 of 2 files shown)


escapedcat:
@qqqzhch @ra5pvt1n you wanna give this a try? 谢！

ra5pvt1n:
Will check/update in the coming days. It's been a hectic couple of weeks. 

My opinion regarding sats - since Satoshi is a Japanses name it can be transliterated into Han characters. Not everyone can read Latin alphabets or understand basic phonology and we should not raise the barrier of entry for potential users. 

But indeed some sort of translation standard would be helpful to keep different projects consistent. Do you know of any initiatives working on standardising translations?

bumi:
wow that's amazing! maybe we can reach out on twitter and other communities about standardising translations? and what's the current preferred way?

bumi:
I guess I can merge it for now and we update the words once decided?

ra5pvt1n:
Yes do merge. 

I don't have a twitter account haha, and I don't know if there's a large Chinese speaking community there (perhaps from Hong Kong). Maybe @escapedcat @qqqzhch are more informed about this topic - where is the community most active?

The current preferred way for me is using the character 聪 for sats (also in RoboSats translations).

qqqzhch:
@escapedcat  @ra5pvt1n  
OK, there is a problem with the front-end time of my environment. I'll take a look at this after I fix it

PR #1924, opened 2022-12-29
Fix/nostr relays

bumi:
Adds some more relays and returns the relays in the proper format


reneaaron (approved):
utACK

Thanks for the fix, also stumbled upon that while using some clients. 👌

Issue #1920, opened 2022-12-27
[Nostr] support new key format


bumi:
### Feature description

The nostr key format had changed recently, we should support the new format (convert the new format to hex)

### Describe the solution

I think in nostr-tools there are some functions to convert the format to hex as we expect it.

### Are you working on this?

None

fczuardi:
nip reference https://github.com/nostr-protocol/nips/blob/master/19.md

PR #1927, opened 2022-12-31
feat(nostr): support new key format

fczuardi:

This is my attempt to address https://github.com/getAlby/lightning-browser-extension/issues/1920 the field for Nostr private keys on the settings page of the browser extension expects that the user paste a key in "hex" format. However clients like astral.ninja have adopted a new format that uses bech32 with a nsec1 prefix (see [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md)).

This patch compares the beginning of the provided Nostr private key with the known prefixes in the NIP to check if a conversion is needed, and if so, transforms the input from bech32 to hex.

### Link this PR to an issue

Fixes _#1920_

### Type of change

### How has this been tested?

I havent tested it extensively, I have compared the expected output using console log (refreshing the settings also works). And I have tried my personal `nsec` key with astral.ninja, it did seem to have signed and published new events.

### Checklist


reneaaron (approved):
@fczuardi Thanks for your PR! Just tested it and worked fine. 👌

nit: Should we maybe add a hint that private keys are always stored in HEX format (users could get confused why their private keys do not match when they revisit the settings page)

tACK

fczuardi:
It also bothers me a little that the settings page will display a different key when the user returns or refreshes the page. Maybe we should always display the bech32 version to the user, as the purpose of the "nsec" prefix is to prevent accidents like people pasting the private key on places expecting the public one, and in hex format this distinction is visually difficult.

bumi:
with more interest in nostr maybe we can move this nostr things out of the settings and labs section and give it's own section (and support keys per account)

reneaaron:
I agree, however that would give us the same problem with the value changing (but the other way around). 

Another idea would be to render the public key (npub...) below? I feel like most users will recognize their public key which would give them some indication that they've imported their key successfully.

src/common/lib/nostr.ts
```diff
@@ -0,0 +1,19 @@
+import { bech32Decode } from "../utils/helpers";
+
+const nostr = {
+  normalizeHex(str: string) {
+    const NIP19Prefixes = ["npub", "nsec", "note"];
```
bumi, line 5: are those always lowercase?

src/common/lib/nostr.ts
```diff
@@ -0,0 +1,19 @@
+import { bech32Decode } from "../utils/helpers";
```
bumi, line 1: there is a nostr/helpers.ts file - does this one also fit there?

src/common/lib/nostr.ts
pqingshuang: this so ~

PR #1930, opened 2023-01-02
fix(nostr): re-encrypt nostr key on pw change #1925

### LINKED ISSUE #1925: Re-encrypt the nostr key when changing the password

### Describe the bug

When changing the password the nostr key is not encrypted with the new password.

### Screenshots [optional]

_No response_

### Steps To Reproduce

-

### Expected behavior

-

### Alby information

master

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

None

escapedcat:

See title

Fixes #1925

### Type of change

### How has this been tested?

Manually

### DIFF
--- a/src/extension/background-script/nostr/index.ts
+++ b/src/extension/background-script/nostr/index.ts
@@ -1,7 +1,7 @@
 import * as secp256k1 from "@noble/secp256k1";
 import { Buffer } from "buffer";
-import { AES } from "crypto-js";
 import * as CryptoJS from "crypto-js";
+import { AES } from "crypto-js";
 import Base64 from "crypto-js/enc-base64";
 import Hex from "crypto-js/enc-hex";
 import Utf8 from "crypto-js/enc-utf8";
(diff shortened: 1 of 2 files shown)


reneaaron (approved):
tACK

bumi:
that is not enough as there are already broken setups out there that currently can not update the key. 
so these fail: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Settings.tsx#L40-L43

but the main problem is here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Settings.tsx#L57

src/extension/background-script/nostr/index.ts
```diff
@@ -16,7 +16,11 @@ class Nostr {
     const password = state.getState().password as string;
     const encryptedKey = state.getState().nostrPrivateKey as string;
     if (encryptedKey) {
-      return decryptData(encryptedKey, password);
+      try {
+        return decryptData(encryptedKey, password);
+      } catch (e) {
+        if (e instanceof Error) console.error("Couldn't get key");
```
bumi, line 22:
```suggestion
        if (e instanceof Error) { 
          console.error("Could not decrypt the Nostr key");
          console.error(e);
        }
```

reneaaron:
tACK (tested again with a malformed private key before the upgrade)

PR #1939, opened 2023-01-04
feat: boost button in websites

im-adithya:

This PR replaces the Blue Extension Icon on finding a Lightning Address to have a Boost Sats button in the bottom right corner.

⚠️ Styling still remains

### Type of change

Screenshot 2023-01-04 at 9 59 53 PM

### How has this been tested?

**Tested manually in each site:**

- [x] GitHub: https://github.com/im-adithya
- [x] LinkTree: https://linktr.ee/getalby
- [x] Medium: https://maxirosson.medium.com/
- [x] Reddit: https://www.reddit.com/user/derbumi
- [x] SoundCloud: https://soundcloud.com/monvol
- [x] Stack Overflow: https://stackoverflow.com/users/15290872/im-adithya
- [x] Twitter: https://twitter.com/bumi
- [x] Vida: https://vida.page/fiatjaf
- [x] Youtube: https://www.youtube.com/c/ClubMikeV

**Don't have links to check:**

- [ ] GeyserProject: 
- [ ] Mixcloud: 
- [ ] PeerTube: 
- [ ] Vimeo: 
- [ ] Monetization: 

### Checklist

### DIFF
--- a/src/extension/content-script/onend.js
+++ b/src/extension/content-script/onend.js
@@ -1,6 +1,6 @@
 import browser from "webextension-polyfill";
 
-import extractLightningData from "./batteries";
+import injectBoostButton from "./boost-button";
 import injectScript from "./injectScript";
 import getOriginData from "./originData";
 import shouldInject from "./shouldInject";
@@ -31,11 +31,12 @@ async function init() {
   }
 
   injectScript(browser.runtime.getURL("js/inpageScript.bundle.js")); // registers the DOM event listeners and checks webln again (which is also loaded onstart
+  injectBoostButton();
 
   // extract LN data from websites
   browser.runtime.onMessage.addListener((request, sender, sendResponse) => {
     if (request.action === "extractLightningData") {
-      extractLightningData();
+      // extractLightningData();
     }
   });
 
--- a/src/extension/content-script/batteries/GeyserProject.ts
+++ b/src/extension/content-script/batteries/GeyserProject.ts
@@ -1,9 +1,10 @@
+import { Battery } from "~/types";
+
 import getOriginData from "../originData";
-import { setLightningData } from "./helpers";
 
 const urlMatcher = /^https:\/\/geyser.fund\/([^/]+)(\/([^/]+))?$/;
 
-const battery = (): void => {
+const battery = (): Battery | void => {
   const urlParts = document.location.pathname.split("/");
   const project = urlParts[1];
   const name = urlParts[2];
@@ -23,16 +24,14 @@ function handleProjectPage(name: string) {
   )?.src;
 
   if (address) {
-    setLightningData([
-      {
-        method: "lnurl",
-        address: address,
-        ...getOriginData(),
-        description: "geyser.fund",
-        name: title ?? name,
-        icon: icon ?? "https://geyser.fund/logo-brand.svg",
-      },
-    ]);
+    return {
+      method: "lnurl",
+      address: address,
+      ...getOriginData(),
+      description: "geyser.fund",
+      name: title ?? name,
+      icon: icon ?? "https://geyser.fund/logo-brand.svg",
+    };
   }
 }
 
--- a/src/extension/content-script/batteries/GitHub.ts
+++ b/src/extension/content-script/batteries/GitHub.ts
@@ -1,9 +1,11 @@
+import { Battery } from "~/types";
+
 import getOriginData from "../originData";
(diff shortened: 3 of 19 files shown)


rolznz:
@im-adithya I tried this out locally
- I think there needs to be a close button, especially if it blocks other content on the page like other floating buttons
- Should the button show if you have a new install of the extension but haven't created an account yet? It does redirect to the account creation page though so it might be fine as it is.
- (Future idea) I think it would be cool to see how many tips you have sent them so far, and also how many boosts/total amount of boosts they have received in total (I guess the publisher would need to expose this publicly somehow)

pavanjoshi914:
This isn't working for youtube videos. did check out youtube videos battery where we set lightning data using ```setLightningData```, but when we try to extract lightning data from the boost button component using ```extractLightningData``` it returns ```undefined and leads to invisibility of the boost button

reneaaron:
Nice work @im-adithya, I just tested the PR and feels great already. 🚀

Some first feedback on the UI: 
 - I think we should remove the infinite loop animation of the lightning bolt but instead just play it once (after the button is initialized). It constantly draws attention to this button and distracts (especially if you are reading content or watching a video)
 - The shadows are too big and dark, also some spacing around wouldn't hurt (see @stackingsaunter's [designs](https://www.figma.com/proto/8DVcMtoCOXpPSljqtYiwQz/Boost-Button?page-id=346%3A496&node-id=377%3A974&viewport=1954%2C-4463%2C0.17&scaling=min-zoom&starting-point-node-id=377%3A974&show-proto-sidebar=1))
 - The font doesn't seem to be loaded and the text in the tooltip is *very* small: (firefox)
image
 - I think we need a loading state between the last click and displaying the payment prompt (maybe we can show a spinning wheel here?)


Can't wait to see zaps all over the web. 🙌 I'll follow up with a code review later this week.

rolznz:
@im-adithya 
I see there are two different methods to pay (sendSatsToLnurl - direct webLN payment to lightning address and generateInvoice - using Alby tools).

Is it necessary to have both? I think this will be harder to maintain and we'd need two different ways to attach metadata. The alby-tools version is better in my opinion as this seems easier to add LUD-18 data by passing payerData to generateInvoice. What do you think?

rolznz:
@im-adithya if you send a payment, is it impossible to send another without refreshing the page? 

image

pavanjoshi914:
looks great just tested out!!
few of the things to be noticed
 1. ```no WebLN``` error after payment window is closed
 2. personally i felt i had to press too long for custom payment time can be reduced
[screen-capture.webm](https://user-images.githubusercontent.com/55848322/225655740-71160926-9150-429b-be80-d6bf15d6a7fc.webm)

src/extension/content-script/onend.js
```diff
@@ -31,11 +31,12 @@ async function init() {
   }
 
   injectScript(browser.runtime.getURL("js/inpageScript.bundle.js")); // registers the DOM event listeners and checks webln again (which is also loaded onstart
+  injectBoostButton();
 
   // extract LN data from websites
   browser.runtime.onMessage.addListener((request, sender, sendResponse) => {
     if (request.action === "extractLightningData") {
-      extractLightningData();
+      // extractLightningData();
```
jankoegel, line 39: Was this intentional or a mistake? If intentional, can't you delete the entire `if` and the event listener then?

stackingsaunter:
Great work! I love the tingling animation :D

I have a few thoughts:
1.  I second @reneaaron comments regarding shadow and hover font

2. After I cancel payment, there's this and it doesn't work until I refresh. Does it have to work that way?
Screenshot 2023-03-16 at 22 03 26

3. After the popup appears with payments, I still can click and increase zap, which does nothing, this feels buggy a bit. I think the increases should be blocked then
Screenshot 2023-03-16 at 22 05 33

4. After I sent zap, it's stuck in this state and I can't do it again. I think it'd be better if the zap resteted to default state. Is it possible? However, the confirmation on the button (x sats sent) should be visible (unwinded) without hovering and then going back to defaults state. The reason for this is that people with budget set will have no visual clue they paid anything (if they don't hover)
Screenshot 2023-03-16 at 22 13 14


6. I second @pavanjoshi914  that hodl should be MUCH shorter. I think basically any holding which is not a single click on a button should trigger custom amount popup

stackingsaunter:
We can close it i think @im-adithya

pavanjoshi914:
closing as we already removed battries

PR #1940, opened 2023-01-04
Fix/UI cleanup

reneaaron:
Some smaller UI changes that boggled me for a while. 😅 

See the commits for detailed infos on what was changed.


escapedcat (approved):
utack

PR #1944, opened 2023-01-05
fix: display name instead of url as the publishercard title

reneaaron:

Currently the website url is displayed twice. Instead we want to show the website name as the title of the publisher card.


bumi:
Let's make sure we show the host somewhere prominent, too. Ultimately the host is the only thing that can not be changed and we have to be careful about phishing.

PR #1964, opened 2023-01-11
feat: move nostr key to accounts

### LINKED ISSUE #1934: Move nostr key to account
### Feature description

> with more interest in nostr maybe we can move this nostr things out of the settings and labs section and give it's own section (and support keys per account)

https://github.com/getAlby/lightning-browser-extension/pull/1927#issuecomment-1369705682

### Describe the solution
 - allow every account to have multiple "connectors" (we have to find a better name for those)
   - Lightning (what we have now)
   - nostr (which is now a global setting for all accounts)
   - liquid (maybe in future)
 - the different connectors can have settings inside of the account (e.g. nostr private key) and display information that is specific to this connector (e.g. the public key)

im-adithya:

This removes the nostr key in settings (which is being used globally across all accounts) and adds functionality for accounts to have nostr keys instead

Fixes #1934

### Type of change

Not ready yet.

### How has this been tested?

Tested manually

### Checklist

### DIFF
--- a/src/app/components/AccountMenu/index.test.tsx
+++ b/src/app/components/AccountMenu/index.test.tsx
@@ -13,8 +13,20 @@ const defaultProps = {
 };
 
 const mockAccounts: Accounts = {
-  "1": { id: "1", connector: "lnd", config: "", name: "LND account" },
-  "2": { id: "2", connector: "galoy", config: "", name: "Galoy account" },
+  "1": {
+    id: "1",
+    connector: "lnd",
+    config: "",
+    name: "LND account",
+    nostrPrivateKey: "",
+  },
+  "2": {
+    id: "2",
+    connector: "galoy",
+    config: "",
+    name: "Galoy account",
+    nostrPrivateKey: "",
+  },
 };
 
 jest.mock("~/app/context/AccountsContext", () => ({
--- a/src/extension/background-script/migrations/index.ts
+++ b/src/extension/background-script/migrations/index.ts
@@ -46,6 +46,20 @@ const migrations = {
       // state is saved with the setMigrated call
     }
   },
+  migrateisUsingGlobalNostrKey: async () => {
+    const { nostrPrivateKey, accounts } = state.getState();
+
+    if (nostrPrivateKey) {
+      Object.values(accounts).map((account) => {
+        account.nostrPrivateKey = nostrPrivateKey;
+      });
+
+      state.setState({
+        accounts,
+        nostrPrivateKey: null,
+      });
+    }
+  },
 };
 
 const migrate = async () => {
@@ -57,6 +71,11 @@ const migrate = async () => {
     );
     await migrations["migrateisUsingLegacyLnurlAuthKeySetting"]();
     await setMigrated("migrateisUsingLegacyLnurlAuthKeySetting");
+  }
+  if (shouldMigrate("migrateisUsingGlobalNostrKey")) {
+    console.info("Running migration for: migrateisUsingGlobalNostrKey");
+    await migrations["migrateisUsingGlobalNostrKey"]();
+    await setMigrated("migrateisUsingGlobalNostrKey");
   }
 };
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,5 +1,5 @@
 import { PaymentRequestObject } from "bolt11";
-import { CURRENCIES, ACCOUNT_CURRENCIES } from "~/common/constants";
+import { ACCOUNT_CURRENCIES, CURRENCIES } from "~/common/constants";
 import connectors from "~/extension/background-script/connectors";
 import {
   ConnectorInvoice,
@@ -16,6 +16,7 @@ export interface Account {
   connector: ConnectorType;
   config: string;
   name: string;
+  nostrPrivateKey: string | null;
 }
 
 export interface Accounts {
--- a/src/app/screens/Accounts/index.tsx
+++ b/src/app/screens/Accounts/index.tsx
@@ -21,7 +21,7 @@ import api from "~/common/lib/api";
 import msg from "~/common/lib/msg";
 import type { Account } from "~/types";
 
-type AccountAction = Omit<Account, "connector" | "config">;
+type AccountAction = Omit<Account, "connector" | "config" | "nostrPrivateKey">;
 
 function AccountsScreen() {
   const auth = useAccount();
--- a/src/app/screens/connectors/AlbyWallet/index.tsx
+++ b/src/app/screens/connectors/AlbyWallet/index.tsx
@@ -115,6 +115,7 @@ export default function AlbyWallet({ variant }: Props) {
         lnAddress,
       },
       connector: "lndhub",
+      nostrPrivateKey: null,
     };
 
     try {
--- a/src/app/screens/connectors/ConnectBtcpay/index.tsx
+++ b/src/app/screens/connectors/ConnectBtcpay/index.tsx
@@ -75,6 +75,7 @@ export default function ConnectBtcpay() {
         url,
       },
       connector: getConnectorType(),
+      nostrPrivateKey: null,
     };
 
     try {
--- a/src/app/screens/connectors/ConnectCitadel/index.tsx
+++ b/src/app/screens/connectors/ConnectCitadel/index.tsx
@@ -7,7 +7,7 @@ import ConnectorForm from "@components/ConnectorForm";
 import TextField from "@components/form/TextField";
 import ConnectionErrorToast from "@components/toasts/ConnectionErrorToast";
 import { useState } from "react";
-import { useTranslation, Trans } from "react-i18next";
+import { Trans, useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import { toast } from "react-toastify";
 import msg from "~/common/lib/msg";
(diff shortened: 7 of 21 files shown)


im-adithya:
I left the `nostr` in the state itself as I got to know it's using the private key from outside, and the class is not initialized with the key. This reduces the work (and also doesn't make sense to include it in all accounts if it's anyways the same class)

src/extension/background-script/actions/accounts/getById.ts
```diff
@@ -0,0 +1,35 @@
+import state from "~/extension/background-script/state";
```
escapedcat, line 1: Wanna add a test for this?

escapedcat:
Look like I can't delete the nostr key? On master I can delete and after reload the field is still empty. Doesn't seem to work here.

On master I get the "use extension option":  
image
image

Here I do not get this?:
image
image


- https://snort.social/login
- https://astral.ninja/

src/app/components/AccountMenu/index.test.tsx
```diff
@@ -13,8 +13,20 @@ const defaultProps = {
 };
 
 const mockAccounts: Accounts = {
-  "1": { id: "1", connector: "lnd", config: "", name: "LND account" },
-  "2": { id: "2", connector: "galoy", config: "", name: "Galoy account" },
+  "1": {
```
escapedcat, line 16: In another PR these could be moved to, right?:
https://github.com/getAlby/lightning-browser-extension/tree/task-nostr-account/tests/fixtures

Also these should be moved there, not sure why these neded up in `src`:  
https://github.com/getAlby/lightning-browser-extension/tree/task-nostr-account/src/fixtures

src/extension/background-script/actions/nostr/getPrivateKey.ts
```diff
@@ -1,9 +1,30 @@
...
+  if (!id) {
+    return {
+      data: state.getState().getNostr().privateKey,
+    };
+  }
+
+  const accounts = state.getState().accounts;
+  if (Object.keys(accounts).includes(id)) {
+    const password = state.getState().password as string;
+    const account = accounts[id];
```
bumi, line 18: would here `const account = state.getState().getAccount()` work? 
so we don't need to do that here

src/types.ts
```diff
@@ -16,6 +16,7 @@ export interface Account {
   connector: ConnectorType;
   config: string;
   name: string;
+  nostrPrivateKey: string | null;
```
bumi, line 19: can this be optional? `nostrPrivateKey?` ?

bumi (commented):
good progress! let's add some tests and then ship this!

reneaaron:
Just tested it, nice work! 💪

Here is some feedback on the UI: 

### Connector logo
Can we maybe use the connector logo here instead of the default avatar? 
image

### Alby lab
I think with this feature we can finally move Nostr out of the Lab status. Wdyt? 
image

### Export
The export icon isn't very intuitive. Can we at add an export label maybe? 
image

### Account name
Shouldn't the account name be reflected in the account select? 
image

src/types.ts
im-adithya: Yes, changing it!

src/extension/background-script/actions/nostr/setPrivateKey.ts
```diff
@@ -1,10 +1,32 @@
...
 const setPrivateKey = async (message: MessagePrivateKeySet) => {
-  await state.getState().getNostr().setPrivateKey(message.args.privateKey);
-  return {};
+  const id = message.args?.id || state.getState().currentAccountId;
+
+  const password = state.getState().password as string;
+  const privateKey = message.args.privateKey;
+  const accounts = state.getState().accounts;
+
+  if (id && Object.keys(accounts).includes(id)) {
```
lujakob, line 13: Just a nitpick: wouldn't it be enough to have `if (Object.keys(accounts).includes(id))` in the condition? In case `id` is `undefined` the condition is also `false`. Or do I miss something?

im-adithya: Yeah, it should work, but TypeScript (🙂) requires the `includes()` parameter to strictly be `string` and not `null` type

im-adithya:
[NEW!] Updated Nostr section

Screenshot 2023-01-20 at 4 32 03 PM

src/app/screens/Account.tsx
```diff
@@ -87,12 +102,39 @@ function AccountScreen() {
...
-    const result = await msg.request("nostr/getPrivateKey", {
-      id: account?.id,
-    });
-    const currentPrivateKey = result as unknown as string;
+  function closeNostrKeyModal() {
+    setNostrKeyModalIsOpen(false);
+  }
+
+  function generatePublicKey(priv: string) {
+    const nostr = new Nostr(priv);
```
escapedcat, line 110: Should this be done via a `msg.request...` instead of importing Nostr from the background script?

src/extension/background-script/actions/nostr/setPrivateKey.ts
```diff
@@ -1,10 +1,32 @@
...
+    const account = accounts[id];
+    account.nostrPrivateKey = privateKey
+      ? encryptData(privateKey, password)
+      : null;
```
bumi, line 17: I think we should do a special "delete" action to remove the key and not do this by setting the private key to null. I would think that having a special key would prevent the user a bit better from accidentially deleting it.

src/extension/background-script/migrations/index.ts
```diff
@@ -52,6 +52,20 @@ const migrations = {
...
+    const { nostrPrivateKey, accounts } = state.getState();
+
+    if (nostrPrivateKey) {
+      Object.values(accounts).map((account) => {
+        account.nostrPrivateKey = nostrPrivateKey;
+      });
+
+      state.setState({
+        accounts,
+        nostrPrivateKey: null,
```
bumi, line 65:
```suggestion
```

let's keep it as a backup for now and clean it up in a following PR.

src/app/screens/Account.tsx
bumi: it should, we need some good folder structure. (I dislike "utils" because those are mostly bins for everything).

bumi:
- [ ] back button (like on send/receive?)
- [ ] save button on the name - to be consistent with the save button on the nostr key? 
- [ ] can we use a copy icon instead of the big button? 
- [ ] nostr public key must displayed in the npub format
- [ ] we have an accounts/index.tsx file should the show page be something like accounts/show.tsx

reneaaron:
Since the avatars PR #1992  is now also merged, can we replace the default avatar on the detail page as well?

escapedcat:
After manually saving a pasted private-key I get a blank screen:

https://user-images.githubusercontent.com/1016218/216023858-e121a5d0-f244-420d-9ead-649e1ba84bba.mov

Can you reproduce this?

im-adithya:
> Can you reproduce this?

Whoops! Did you use a "nsec" string? It expects a hex string as a private key. Not really sure if we should also accept "nsec" directly. (cc @bumi)

But yes, will be adding a toast to specify the requirement!

escapedcat:
Should this look like this in dark-mode?:
image

src/app/screens/Accounts/Show/index.tsx
```diff
@@ -0,0 +1,592 @@
...
+                    } catch (e) {
+                      if (e instanceof Error) {
+                        toast.error(e.message);
+                      }
+                    }
+                  }}
+                  fullWidth
+                />
+              </div>
+              <div className="w-1/5 flex-none">
```
escapedcat, line 471: non-blocking: Would be nice to have this in a `form` so I can use "enter" to submit after pasting my nsec.

src/app/screens/Accounts/Show/index.tsx
```diff
@@ -0,0 +1,592 @@
+import {
```
escapedcat, line 1: nit: not sure if `Show` is the right wording for this. What about `DetailView`?

escapedcat (commented):
Added some minor comments

src/app/screens/Accounts/Show/index.tsx
im-adithya: > What about `DetailView`?

Sounds good @bumi wdyt?

reneaaron: +1 for detail. However I'd probably drop the `View` part and just have `src/app/screens/Accounts/Detail/index.tsx`

im-adithya: I've already renamed the Publisher Show Screen (the only other similar screen) in this PR itself to align things, should I change it in this itself?

reneaaron: No, I'd keep this PR as is and do the rename in a separate PR like @bumi suggested.

escapedcat (approved):

src/extension/background-script/migrations/index.ts
```diff
@@ -52,6 +52,19 @@ const migrations = {
       // state is saved with the setMigrated call
     }
   },
+  migrateisUsingGlobalNostrKey: async () => {
+    const { nostrPrivateKey, accounts } = state.getState();
+
+    if (nostrPrivateKey) {
+      Object.values(accounts).map((account) => {
+        account.nostrPrivateKey = nostrPrivateKey;
```
bumi, line 60: let's make this defensive and only overwrite it when`account.nostrPrivateKey` is blank.

PR #1970, opened 2023-01-12
feat: create LN address entry in db

### LINKED ISSUE #1965: [Feature] Contacts list
### Feature description

Adding a tab to allow users to save contacts with LN address info. Could be used for people they know, online creators, etc.

Lightning addresses make it easy to identify and pay individuals. Currently Alby stores a list of visited websites, it would be cool to extend this idea to people as well.

Screenshots from a previous hackathon project:

/contacts
Screenshot 2023-01-11 at 1 30 33 PM

/contacts/[id]
Screenshot 2023-01-11 at 1 31 23 PM

/saveContact
Screenshot 2023-01-11 at 1 28 59 PM

Extension:
Screenshot 2023-01-11 at 12 53 44 PM

Screenshot 2023-01-11 at 1 14 32 PM

### Describe the solution

The simplest version of this would likely be the contacts tab that allows you to store LN addresses under a name, and see the transaction history with that contact if you click into it (similar to websites).

**Frontend**

Screens
- ContactsHome: list of contacts
- Contact: individual contact page, transaction history, pay button
- SaveContact: Form to add or edit a contact
    - To keep it simple, perhaps only LN address and name (optional) to start (no photo, links etc.)

**Backend**
- contacts table in database (or something like this)
- associated db actions: add, delete, getById, etc.

Probably the place to start is how to save contacts in the database. Would love to hear feedback on that. In the hackathon I added a `contacts` table and added `contactId` to `payments` table, as so: https://github.com/alex-klyuev/lightning-browser-extension/blob/alby-contacts-list-hackathon/src/extension/background-script/db.ts#L32-L35. But I'm not sure if this is the best way to do it. I'll open a PR for that

### Describe alternatives

_No response_

### Additional context

A rough draft of the code was written during the hackathon here: https://github.com/alex-klyuev/lightning-browser-extension/tree/alby-contacts-list-hackathon

It includes drafts of the concepts mentioned above.

Presentation: https://vimeo.com/774674937

### Are you working on this?

Yes

alex-klyuev:

The basis is to save the LN address to the DB. This would allow for saving contact info, autofilling payment recipients, etc. One approach as done here is to create a contacts table to allow for adding further fields to contacts if desired later on, and adding a `contactId` to the payments table. Would love to hear feedback on this approach.

(Partly) Fixes #1965

### Type of change

### How has this been tested?

### Checklist

### DIFF
--- a/src/extension/background-script/db.ts
+++ b/src/extension/background-script/db.ts
@@ -2,8 +2,8 @@ import Dexie from "dexie";
 import browser from "webextension-polyfill";
 import type {
   DbAllowance,
-  DbPayment,
   DbBlocklist,
+  DbPayment,
   DbPermission,
 } from "~/types";
 
@@ -26,6 +26,11 @@ class DB extends Dexie {
     });
     this.version(3).stores({
       permissions: "++id,allowanceId,host,method,enabled,blocked,createdAt",
+    });
+    this.version(4).stores({
+      contacts: "++id, &lnAddress, accountId",
+      payments:
+        "++id,allowanceId,contactId,host,location,name,description,totalAmount,totalFees,preimage,paymentRequest,paymentHash,destination,createdAt",
     });
     this.on("ready", this.loadFromStorage.bind(this));
     this.allowances = this.table("allowances");


src/extension/background-script/db.ts
```diff
@@ -26,6 +26,9 @@ class DB extends Dexie {
     });
     this.version(3).stores({
       permissions: "++id,allowanceId,host,method,enabled,blocked,createdAt",
+      contacts: "++id, &lnAddress",
```
escapedcat, line 29: I think you need to create a "new version" every time we add new "tables" to the DB. i.e.:
```ts
    this.version(4).stores({
      contacts: "++id, &lnAddress",
    })
```

escapedcat (changes requested):
"Saving address" will follow in a new PR, right?

src/extension/background-script/db.ts
reneaaron: Should we maybe also add the current account that is being used here? I'm not sure whether a contact list should be something global or also on an account level. 🤔

alex-klyuev:
> "Saving address" will follow in a new PR, right?

Yeah, I was thinking to iron out the schema first and then add the methods to add / delete / etc.

src/extension/background-script/db.ts
alex-klyuev: Yeah, that's a good question... I suppose for myself, I use the different accounts as just different sources of funding so I would want a global account list. However I can see the case if you want to hold accounts separately.

alex-klyuev: Yeah, that's a good question... I suppose for myself, I use the different accounts as just different sources of funding so I would want a global account list. However I can see the case if you want to hold accounts separately

alex-klyuev: Would it be as simple as adding an `accountId`? I've been a bit confused on the distinction between `db.ts` and `state.ts`

alex-klyuev: > I think you need to create a "new version" every time we add new "tables" to the DB. i.e.:
> 
> ```ts
>     this.version(4).stores({
>       contacts: "++id, &lnAddress",
>     })
> ```

Looks like you need to add a new version even if you modify an existing table - would that be a problem if we want to add other fields to contacts table in the future?

reneaaron: > Would it be as simple as adding an accountId? I've been a bit confused on the distinction between db.ts and state.ts

I think so, yes. You would add the `accountId` to the contacts table and reference the account that is currently being used as you store new lightning addresses. 

> Looks like you need to add a new version even if you modify an existing table - would that be a problem if we want to add other fields to contacts table in the future?

I don't think so. It's the same situation for all other tables that already exist. But if we think the accountId could be useful it might be better to include it right away.

alex-klyuev: Sounds good, I added `accountId` to the schema

reneaaron:
The schema makes sense to me 👍 I think though that those contact-related changes should be merged as one "atomic" PR since we don't want to do client-side database migrations for features that are yet to be developed. (feel free to continue work in here or base other PR on this branch)

alex-klyuev:
Yeah, that makes sense... would it make sense to have a feature branch then?

reneaaron:
Isn't this branch already a feature branch? What's the difference between the feature branch and this one? 🤔

escapedcat:
> Isn't this branch already a feature branch? What's the difference between the feature branch and this one? 🤔

The only difference would be, that we would have already agreed on these changes here and merge those into the feature branch. So *theoretically* we wouldn't need to discuss when we want to merge the feature branch in the end.

But yeah, can be done either way.

alex-klyuev:
Perhaps I'm misunderstanding, but the feature branch on this PR is a branch on my local fork, but the new one is on the Alby repo, so I'm assuming PRs against the latter will be easier to review / have the tests run / etc.

reneaaron:
I don't think that should really matter where you develop this feature. 🤔 Do you have problems with reviews / tests in this fork?

escapedcat:
Already created https://github.com/getAlby/lightning-browser-extension/tree/feat/1965_contacts and let @alex-klyuev know yesterday. Use whatever you prefer I guess. We can collect changes here or there.

escapedcat (approved):

PR #1976, opened 2023-01-14
fix: remove window or tab based on what the browser opens

bumi:
Normally when we open a new window the prompt is the only tab in this window If that's the case we have to close the prompt using windows.remove()

Some browsers force the new window to open in a tab. In this case we have to close the prompt using tabs.remove()

Some browsers actually can close the window when tabs.remove() removes the only/last tab. But Opera fails with the following error: "Tabs cannot be edited right now (user may be dragging a tab)"

### DIFF
--- a/src/common/lib/utils.ts
+++ b/src/common/lib/utils.ts
@@ -83,6 +83,7 @@ const utils = {
           left: left,
         })
         .then((window) => {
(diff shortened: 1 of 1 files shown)


reneaaron (approved):
utACK

Issue #1972, opened 2023-01-14
[BUG] Problems caused by too long name


qqqzhch:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When logging in with the mailbox as the user name, if the name is too long, the menu will not be displayed completely

### Screenshots [optional]

![WX20230114-105623](https://user-images.githubusercontent.com/3917994/212447822-ed756641-d774-4b36-888d-d6f00e906a0a.png)

### Steps To Reproduce

1 Install a new wallet
2 Register with email and fill in the email name of alby，The name of letters in a word is greater than 5, such as hugo1234
3 Open the wallet home page to see a part of the menu

### Expected behavior

The name and menu are displayed normally

### Alby information

The latest version of master I use
alby version:1.23.0

### Device information

alby version:1.23.0

### Are you working on this?

Yes

qqqzhch:
Display is normal in full-screen view
chrome-extension://bolljnmcekcabodmdedlhhkmicblndgc/options.html#/discover

The display is abnormal only in pop mode
chrome-extension://bolljnmcekcabodmdedlhhkmicblndgc/popup.html

qqqzhch:
I need to add a new breakpoints in screens to mark the minimum size of pop page, in tailwind.config.js
The size of the pop page is 384 px
So add an xs: 300 or popview: 300

@bumi  @fczuardi 

```
screens: {
      'xspop': '300px',
      ...defaultTheme.screens,
    }
```

PR #1955, opened 2023-01-09
feat: improve navigation

reneaaron:
### Describe the changes you have made in this PR

 - Match styles from getalby.com navbar
 - Moved account dropdown to the right
 - Hide navbar on mobile devices 

image

image

image

PR #1977, opened 2023-01-15
fix: issues-1972:problems caused by too long name

qqqzhch:

- 1  add   breakpoint in tailwind.config.js
-  2 add  a small width size to AccountMenu  in pop view

- #1972 

### Type of change

WX20230115-110350

### How has this been tested?

Use a long mailbox name to test

### Checklist

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -27,6 +27,10 @@ module.exports = {
   content: ["./static/views/**/*.html", "./src/app/**/*.{js,ts,jsx,tsx}"],
   plugins: [require("@tailwindcss/forms"), require("@tailwindcss/line-clamp")],
   theme: {
+    screens: {
+      'xs': '300px',
+      ...defaultTheme.screens,
+    },
     extend: {
       animation: {
         "spin-fast": "spin 0.7s linear infinite",
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -73,7 +73,7 @@ function AccountMenu({ showOptions = true }: Props) {
   }
 
   return (
-    <div className="relative pl-2 flex bg-gray-100 rounded-md dark:bg-surface-12dp max-w-full">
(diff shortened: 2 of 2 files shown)


tailwind.config.js
```diff
@@ -27,6 +27,10 @@ module.exports = {
   content: ["./static/views/**/*.html", "./src/app/**/*.{js,ts,jsx,tsx}"],
   plugins: [require("@tailwindcss/forms"), require("@tailwindcss/line-clamp")],
   theme: {
+    screens: {
```
reneaaron, line 30: I don't think we need a new breakpoint in order to limit the width of a component.

reneaaron (commented):
Thanks for this PR, however I think there is already an upcoming improvement with #1955 that should be merged soon. 🤞

I don't think we should introduce another breakpoint just do do the layout of one particular use case. (also with *very* long account names this approach wouldn't work on higher resolutions as well) 

I suggest we revisit this PR after #1955 has been merged.

tailwind.config.js
qqqzhch: OK, I will look at the relevant code of # 1955，to see the effect of the solution

qqqzhch: WX20230116-103840
Just like mobile adaptation, pop view is more like a small-sized mobile page, and the current breakpoint cannot distinguish whether it is running on a smaller-sized page

@rolznz Do you have any ideas, I think this kind of small size problem will be quite common in the future

rolznz: @reneaaron @qqqzhch I made a new PR here https://github.com/getAlby/lightning-browser-extension/pull/1982 which doesn't introduce a breakpoint. I am not really happy with the solution (it feels a bit hacky), but it is isolated to the account menu.

reneaaron:
Closing this PR in favor of #1982 which should be merged soon. Thanks for your initiative on this problem @qqqzhch!

PR #1979, opened 2023-01-15
chore: update nostr event validation

bumi:
keep event validation in sync with nostr-tools
though we do not add nostr-tools to avoid the additional dependency.


bumi:
id should be validated

PR #1982, opened 2023-01-16
fix: make account menu width and accounts on account page responsive

rolznz:

- Make the account menu width responsive based on screen size and ensure buttons are not pushed off the screen
- Fix long account name causing a scrollbar in the accounts page

Note: For the account menu this is a bit of a hacky solution because it has to manually account for the menu icon and padding. I couldn't find a better one. Technically this is also an issue in the accounts page, but I did a simpler fix for this one as I think it's less important to have an exact fit. EDIT: it might just be better to fix the account menu width to the same size as the dropdown and remove all those checks 😅 )

Fixes _#1972_

### Type of change

small phone
image

md breakpoint
image

xl screen
image

Accounts page mobile:
image

md breakpoint
image

xl
image

### How has this been tested?

### Checklist


rolznz:
I think the dropdown menu width should match the account menu width, what do you think? @reneaaron 

image

reneaaron:
Nice! 

> I think the dropdown menu width should match the account menu width, what do you think? @reneaaron

Yes. 👍

Can we maybe add a little spacing between the menu icon and the account menu? 

image

`mr-4` should be fine 👌

rolznz:
> Nice!
> 
> > I think the dropdown menu width should match the account menu width, what do you think? @reneaaron
> 
> Yes. 👍
> 
> Can we maybe add a little spacing between the menu icon and the account menu?
> 
> image
> 
> `mr-4` should be fine 👌

Thanks for testing!

The max-width class wasn't being applied. I've fixed this now so the gap is there.

image


Also updated to make the menu width match:
image

escapedcat (approved):
ack, works great.
Maybe the select should just have a fixed width? Looks a bit weird if it jumps when switching accounts.

e2e is failing on master as well, not related to this PR

rolznz:
> ack, works great. Maybe the select should just have a fixed width? Looks a bit weird if it jumps when switching accounts.
> 
> e2e is failing on master as well, not related to this PR

I agree it's weird it changes width when you switch accounts. It would be simpler to just have a fixed width.

reneaaron:
@rolznz Do you plan to still change this to a fixed width?

rolznz:
@reneaaron I made the change
image
image

src/app/components/Menu/MenuList.tsx
```diff
@@ -18,11 +20,13 @@ function List({ position = "left", children }: Props) {
...
     >
       <Menu.Items
-        className={`${
+        className={classNames(
           position === "left"
             ? "left-0 origin-top-left"
-            : "right-0 origin-top-right"
-        } absolute z-50 mt-2 py-1 w-56 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 focus:outline-none dark:bg-surface-12dp dark:border-0`}
+            : "right-0 origin-top-right",
+          fullWidth ? "w-full" : "w-56",
```
reneaaron, line 27: Why do we need a width here?

reneaaron (commented):
Tested, feels a lot better now and involves less complexity👌

src/app/components/Menu/MenuList.tsx
rolznz: @reneaaron I think to make the menu a consistent width no matter where it is used, and make sure it does not automatically expand if one of the items is wide. I am not sure I understand the question, does this answer it?

PR #1983, opened 2023-01-16
Fix/validate nostr event

bumi:
see: https://github.com/fiatjaf/nocomment/issues/9#issuecomment-1383234383

### DIFF
--- a/src/extension/background-script/actions/nostr/helpers.ts
+++ b/src/extension/background-script/actions/nostr/helpers.ts
@@ -50,8 +50,9 @@ export async function addPermissionFor(method: string, host: string) {
 export function validateEvent(event: Event): boolean {
   if (typeof event.content !== "string") return false;
   if (typeof event.created_at !== "number") return false;
-  if (typeof event.pubkey !== "string") return false;
-  if (!event.pubkey.match(/^[a-f0-9]{64}$/)) return false;
+  // ignore these checks because if the pubkey is not set we add it to the event. same for the ID.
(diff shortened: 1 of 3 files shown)


src/extension/background-script/actions/nostr/helpers.ts
```diff
@@ -45,13 +45,18 @@ export async function addPermissionFor(method: string, host: string) {
...
+export function validateEvent(event: Event): boolean {
   if (typeof event.content !== "string") return false;
   if (typeof event.created_at !== "number") return false;
+  // ignore these checks because if the pubkey is not set we add it to the event. same for the ID.
+  // if (typeof event.pubkey !== "string") return false;
+  // if (!event.pubkey.match(/^[a-f0-9]{64}$/)) return false;
 
   if (!Array.isArray(event.tags)) return false;
-  for (const tag of event.tags) {
+  for (let i = 0; i < event.tags.length; i++) {
```
escapedcat, line 58: nit: can `forEach` bs used?: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

reneaaron (approved):
utACK

src/extension/background-script/actions/nostr/helpers.ts
bumi: this is from: https://github.com/nbd-wtf/nostr-tools/blob/master/event.ts#L61-L77
I copied it from there to be in sync.

PR #1990, opened 2023-01-17
feat: add account menu identicon avatar

lujakob:
### Describe the changes you have made in this PR

As described in the issue the changes add an Identicon avatar icon to the accounts menu. See screenshots for more infos.

### Link this PR to an issue [optional]

Fixes [_#1987_](https://github.com/getAlby/lightning-browser-extension/issues/1987)

### Type of change

- `feat`: New feature (non-breaking change which adds functionality)

### Screenshots of the changes [optional]

Bildschirmfoto 2023-01-16 um 20 10 18
Bildschirmfoto 2023-01-16 um 20 10 26
Bildschirmfoto 2023-01-16 um 20 10 40

### How has this been tested?

Tested manually by running extension.

### Checklist

- [x] My code follows the style guidelines of this project and performed a self-review of my own code
- [x] New and existing tests pass locally with my changes
- [x] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

Issue #1987, opened 2023-01-16
[Feature] Avatars for accounts


reneaaron:
### Feature description

Currently we completely miss account avatars in the application. Some reasons I'd like to add them:
 - Improve the recognizability of a specific when using Alby with multiple connectors (e.g. in the account dropdown)
 - Answer the question "do I use the right account" when you do payments
 - Making the experience feel more "social"

![image](https://user-images.githubusercontent.com/100827540/212700411-8f10ff8f-4889-44ab-96c0-81b7a0d6a575.png)

Adding [identicons](https://www.npmjs.com/package/react-identicons) could be a first step into that direction:

In future we could potentially open this up to connectors providing their own avatars or even allowing users to set their own avatar in the extension. 

### Describe the solution

Identicons should be added to the following components: 
 - `AccountMenu`
 - `Accounts` overview (introduced with #1964)
 - `Account` detail (introduced with #1964)

The entropy for the identicons should come from something that doesn't change but is specific to every account, possibly the `accountId` would make sense here.

lujakob:
I'll have a look at this.

lujakob:
@reneaaron should we merge this PR as is and add a follow up PR to add Avatar to  accounts and account once the two dependencies are merged?

PR #1992, opened 2023-01-18
feat: add account menu avatars - avvvatars lib

lujakob:

As described in the issue the changes add an Identicon avatar icon to the accounts menu. See screenshots for more infos.

This is the same change as https://github.com/getAlby/lightning-browser-extension/pull/1990 but with the https://avvvatars.com/ lib. https://github.com/nusu/avvvatars 
This lib has types, but the styles are also different. Two styles are available: character and shapes (see screenshots) I personally prefer "character".

Fixes [_#1987_](https://github.com/getAlby/lightning-browser-extension/issues/1987)

### Type of change

Bildschirmfoto 2023-01-18 um 08 06 27
Bildschirmfoto 2023-01-18 um 08 07 28

### How has this been tested?

Tested manually by running extension.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "@noble/secp256k1": "^1.7.1",
     "@tailwindcss/forms": "^0.5.3",
     "@tailwindcss/line-clamp": "^0.4.2",
+    "avvvatars-react": "^0.4.2",
     "axios": "^1.2.2",
     "bech32": "^2.0.0",
     "bolt11": "^1.4.0",
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -4,7 +4,7 @@ import {
   CheckIcon,
   PlusIcon,
 } from "@bitcoin-design/bitcoin-icons-react/filled";
-import { WalletIcon } from "@bitcoin-design/bitcoin-icons-react/outline";
+import Avvvatars from "avvvatars-react";
 import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import Skeleton from "react-loading-skeleton";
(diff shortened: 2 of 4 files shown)


lujakob:
@reneaaron this would be a PR for the avvvatars package. Two options regarding the styles. What do you think?

reneaaron:
Great, I just played around a bit with this PR. 👌Currently there seems to be a problem with the layout for longer account names: (the images seems to be squeezed)

image

I personally like the second option better as the initials (at least for how I use the account names) tend to overlap and are often not the best abbreviations. (Multiple LNBits wallets with the same first two characters "LN")

What do you think about adding those avatars to the account list as well? I think they are way better than the generic "wallet" icon currently:
image



Unrelated to that:
Since you derive the avatar from the name, this could possibly even lead to a nice effect on the account settings page (where the avatar is updated as you type the account name).

lujakob:
> Great, I just played around a bit with this PR. 👌Currently there seems to be a problem with the layout for longer account names: (the images seems to be squeezed)
> 
> image
> 
> I personally like the second option better as the initials (at least for how I use the account names) tend to overlap and are often not the best abbreviations. (Multiple LNBits wallets with the same first two characters "LN")
> 
> What do you think about adding those avatars to the account list as well? I think they are way better than the generic "wallet" icon currently: image
> 
> Unrelated to that: Since you derive the avatar from the name, this could possibly even lead to a nice effect on the account settings page (where the avatar is updated as you type the account name).

Good catch. I updated by adding "shrink-0". Nitpick: I replaced the p tag with div, since it's semantically more correct, right?
Also added the avatar to the account screen as you suggested. Also like it.

Anything else?

reneaaron:
Looks great. 🚀

We're currently in the process of finishing up #1964. After we've merged it (which should be in the next few days) we could add the avatars to the accounts detail page as well. 

Other than that I think it's fine for now!

escapedcat (approved):
ack

bumi (approved):
utACK
I think we can merge this one and then add the same also on the accounts page once that is merged?

PR #1998, opened 2023-01-19
fix(lnurlpay): handle status 500 for callback url #1995

### LINKED ISSUE #1995: [BUG] No error message is shown when payment to a lightning address fails

### Describe the bug

I have a Lightning address for my account but am incorrectly connected to my node.
If I try to pay using the extension on a different account, the send button spins and then goes back to its original state. There is no error message.

### Screenshots [optional]

image
image

### Steps To Reproduce

1. Create an Alby account at getAlby.com
2. Create a lightning address
3. Enter incorrect node credentials on the wallet tab
4. Try to send a payment to the lightning address from a different account using the Alby extension

### Expected behavior

I would expect an error message to be shown to the user - maybe to say that the destination is unreachable?

### Alby information

- Payment using Alby version 1.24.0, connected using hosted Alby account

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

None

escapedcat:

Handle status 500 for callback-url

Fixes #1995

### Type of change

image

### How has this been tested?

Manually

### DIFF
--- a/src/app/screens/LNURLPay/index.tsx
+++ b/src/app/screens/LNURLPay/index.tsx
@@ -7,7 +7,7 @@ import SatButtons from "@components/SatButtons";
 import DualCurrencyField from "@components/form/DualCurrencyField";
 import TextField from "@components/form/TextField";
 import axios from "axios";
-import React, { Fragment, useState, useEffect } from "react";
+import React, { Fragment, useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import { toast } from "react-toastify";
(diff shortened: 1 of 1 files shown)


src/app/screens/LNURLPay/index.tsx
```diff
@@ -129,6 +129,11 @@ function LNURLPay() {
           }
         );
 
+        if (response.status === 500) {
```
bumi, line 132: how about 501? :)

escapedcat: Should I just use `>== 500`?

src/app/screens/LNURLPay/index.tsx
```diff
@@ -129,6 +129,11 @@ function LNURLPay() {
           }
         );
 
+        if (response.status >= 500) {
+          toast.error("Payment aborted: Could not connect to server");
```
bumi, line 133:
```suggestion
          toast.error("Payment aborted: Recipient server error");
```

bumi (commented):
maybe we indicate that it is an error on the recipient side - not that we could not connect.

PR #2002, opened 2023-01-21
feat: include account ID in permissions

### LINKED ISSUE #1953: Store account ID to permissions
### Feature description

currently we don't associate and scope the permissions to the accounts. 
We should be able to limit the permissions for each account 

e.g. when I login (lnurl-auth) or sign a message I enable this per account

### Describe the solution

add the account Id to the permissions table

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

im-adithya:

Adds `accountId` to the DbPermission fields to link permissions to accounts. (Doesn't include a migration; ~we handle the permissions without an accountId separately everywhere~).

Fixes #1953

### Type of change

Screenshot 2023-01-21 at 6 41 38 PM

### How has this been tested?

Tested manually at various websites.

### Checklist


src/extension/background-script/db.ts
```diff
@@ -25,7 +25,8 @@ class DB extends Dexie {
       blocklist: "++id,host,name,imageURL,isBlocked,createdAt",
     });
     this.version(3).stores({
-      permissions: "++id,allowanceId,host,method,enabled,blocked,createdAt",
+      permissions:
+        "++id,accountId,allowanceId,host,method,enabled,blocked,createdAt",
```
bumi, line 29: you need to add a new version(4) for this.

look here: https://dexie.org/docs/Tutorial/Design#database-versioning

and actually if we don't query the attribute then we would not need to add it - but we want to query it. (also: https://dexie.org/docs/Version/Version.stores()#warning)


check the upgrade function: https://dexie.org/docs/Version/Version.upgrade()
I'd say we simply store the account id of the first account. (`object.keys(accounts)[0]`)

src/extension/background-script/actions/permissions/__tests__/add.test.ts
```diff
@@ -1,10 +1,20 @@
...
 
+jest.mock("~/extension/background-script/state");
+
+const defaultMockState = {
+  currentAccountId: "123456",
+};
+
+const mockState = defaultMockState;
+state.getState = jest.fn().mockReturnValue(mockState);
+
```
bumi, line 17: add another check that the currentAccountId is stored to the permission.

src/extension/background-script/actions/nostr/helpers.ts
```diff
@@ -12,29 +13,37 @@ export async function hasPermissionFor(method: string, host: string) {
...
   });
+  const accountId = state.getState().currentAccountId;
 
-  if (!allowance?.id) {
+  if (!allowance?.id || !accountId) {
     return false;
   }
 
-  const findPermission = await db.permissions.get({
+  let findPermission = await db.permissions.get({
```
bumi, line 22: I think we could directly query here the permission for that host, method, and and account.

src/extension/background-script/actions/ln/request.ts
```diff
@@ -43,14 +44,22 @@ const request = async (
...
     // prefix method with webln to prevent potential naming conflicts (e.g. with nostr calls that also use the permissions)
     const weblnMethod = `${WEBLN_PREFIX}${connectorName}/${methodInLowerCase}`;
 
     const permission = await db.permissions
       .where("host")
       .equalsIgnoreCase(origin.host)
-      .and((p) => p.method === weblnMethod)
+      .and(
+        (p) =>
+          (!p.accountId || p.accountId === accountId) &&
```
bumi, line 60: I'd say we are just not backwards compatible and simpley also query htere for the account id. 
This makes the code easier 

and please adjust and extend the test for the request method.

src/extension/background-script/actions/allowances/__tests__/delete.test.ts
```diff
@@ -85,6 +88,7 @@ describe("delete allowance", () => {
     ]);
 
     const dbPermissions = await db.permissions.toArray();
+    console.info(dbPermissions);
```
bumi, line 91: debug log should be removed .

src/extension/background-script/actions/permissions/__tests__/add.test.ts
im-adithya: Added a check by changing the permissions fixture: https://github.com/getAlby/lightning-browser-extension/pull/2002/files#diff-851f81c1aba04e58f78a0c6f5e34359aae948bc3cd375f641c0f2b65b811fe94

src/extension/background-script/actions/nostr/helpers.ts
im-adithya: But then we would be leaving the old permissions with no accountId

im-adithya: (Saw your other comment 👍)

src/extension/background-script/actions/allowances/__tests__/delete.test.ts
```diff
@@ -9,6 +9,7 @@ const mockAllowances: DbAllowance[] = allowanceFixture;
 const mockPermissions = [
   {
     id: 1,
+    accountId: "123456",
```
reneaaron, line 12: Shouldn't that be an integer?

src/extension/background-script/actions/ln/request.ts
```diff
@@ -43,14 +44,18 @@ const request = async (
       throw new Error("Could not find an allowance for this host");
     }
 
+    if (!accountId) {
```
reneaaron, line 47: Is this just for TS? When could an `accountId` be falsy?

src/extension/background-script/db.ts
reneaaron: I think this has been done in the meantime, I'll resolve this conversation.

reneaaron (changes requested):
Successfully tested 👌 the migration also worked as expected! 👍

src/extension/background-script/actions/ln/request.ts
im-adithya: Yes, just for TS

src/extension/background-script/actions/allowances/__tests__/delete.test.ts
im-adithya: `accountId`'s are strings
ex: `c2fe55fa-fe99-45da-94d5-ba0271dbabca`

escapedcat: Those are created here with `uuidv4`:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/accounts/add.ts#L19

src/extension/background-script/actions/ln/request.ts
escapedcat: We could add a comment in these cases, i.e. `// type guard`

reneaaron:
Just did a final round of testing and noticed one thing that I think we should fix. Steps to reproduce:

I have 2 accounts (account A, account B) 
 - Account A: Go to a site and give permissions to read public key
 - Go to the websites, select the website from step 1 and click on edit. You will see the permissions correctly:
image
 - On the websites page, switch to account B 
 - Edit the permissions with account B

The edit dialogue now still shows the permissions from account A. (reloading fixes the problem)

@im-adithya Can we somehow fix that? 

That being said, I think we really need to move the permissions somewhere else in another iteration, they don't fit into this "edit allowance" window very well. Also the allowance is now extension wide, while the permissions are not. (which I think is quite hard to understand)

reneaaron (changes requested):
Tested again today, left some feedback and merged `master`. 

Let's see if we can somehow fix the bug when switching accounts and get this merged! 🚀

lujakob:
> Tested again today, left some feedback and merged `master`.
> 
> Let's see if we can somehow fix the bug when switching accounts and get this merged! 🚀

I tested a bit. The bug is in AllowanceMenu line 77 `!permissions && fetchPermissions();` - the `!permissions` condition needs to be removed and `permissions` from the deps. But better double check the re-rendering performance. 

https://github.com/getAlby/lightning-browser-extension/pull/2002/files#diff-1b329da99ce24a934532ca1ee401d135a5d5917619b62db9711da859ce47f987R77

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -236,8 +239,8 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
                   {t("edit_permissions")}
                 </h2>
                 <div>
-                  {permissions.map((permission) => (
-                    <>
+                  {permissions.map((permission, i) => (
+                    <Fragment key={`element-${i}`}>
```
reneaaron, line 243: Can we use the permission id as the key here, too?

im-adithya: It's not gonna appear anyways as it's a fragment, should I change it?

reneaaron: Yeah, but since you already have a unique property `permission.id` I think there is no point in generating a new unique key based on the current index of the array. Please change if you agree or resolve the conversation if you don't.

im-adithya: Ah, got it!

src/app/components/AllowanceMenu/index.tsx
```diff
@@ -237,7 +240,7 @@ function AllowanceMenu({ allowance, onEdit, onDelete }: Props) {
                 </h2>
                 <div>
                   {permissions.map((permission) => (
-                    <>
+                    <Fragment key={`element-${permission.id}`}>
```
reneaaron, line 243: This is how I would do it: 
```suggestion
                    <Fragment key={permission.id`}>
```
<Fragment key={permission.id}>

reneaaron (approved):
Thanks for the updates @im-adithya! 🚀

tACK

PR #2005, opened 2023-01-23
feat: add liquid

### LINKED ISSUE #2074: [Feature] Signing functionality for Liquid
### Feature description

Adding Liquid support for accounts using a separate Liquid Private Key similar to Nostr.

### Describe the solution

Can be implemented similarly to the existing Nostr feature.

### Describe alternatives

--

### Additional context

--

### Are you working on this?

Yes

im-adithya:

Adds Liquid Provider and `LiquidPrivateKey` to accounts

Fixes #2074 

### Type of change

### How has this been tested?

Tested Manually

### Checklist


src/extension/background-script/actions/liquid/generatePrivateKey.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const generateRandomKey = () => {
+  return secp256k1.utils.randomPrivateKey();
+};
+
+const deriveKey = async () => {
+  const connector = await state.getState().getConnector();
+  try {
+    const response = await connector.signMessage({
```
reneaaron, line 33: Do we want to do the same derivation for liquid keys? 🤔 Unsure whether we should just generate a new set of keys. The derivation method allows you to exactly generate `1` key, if this key will ever get compromised you will have to generate a new key and backup it anyway. 

Does the advantage of not having to backup this key separately justify this additional complexity? (not necessarily from the technical perspective, but also from the user perspective, having to explain the different options, etc)

src/extension/background-script/state.ts
```diff
@@ -126,7 +148,13 @@ const state = createState<State>((set, get) => ({
     if (connector) {
       await connector.unload();
     }
-    set({ password: null, connector: null, account: null, nostr: null });
+    set({
```
reneaaron, line 151: Why do we need that here?

im-adithya: because `liquid` is initialized with the `liquidPrivateKey` and can be still used when locked

src/app/screens/Liquid/ConfirmGetPublicKey.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { useTranslation } from "react-i18next";
+import ScreenHeader from "~/app/components/ScreenHeader";
+import { useNavigationState } from "~/app/hooks/useNavigationState";
+import { USER_REJECTED_ERROR } from "~/common/constants";
+import msg from "~/common/lib/msg";
+import { OriginData } from "~/types";
+
+function LiquidConfirmGetPublicKey() {
+  const { t } = useTranslation("translation", {
+    keyPrefix: "nostr",
```
bumi, line 16: this should be liquid here - this template should not share the prefix with anything else

src/app/screens/Liquid/ConfirmSignSchnorr.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { useNavigationState } from "~/app/hooks/useNavigationState";
+import { USER_REJECTED_ERROR } from "~/common/constants";
+import msg from "~/common/lib/msg";
+import type { OriginData } from "~/types";
+
+function ConfirmSignSchnorr() {
+  const navState = useNavigationState();
+  const { t: tCommon } = useTranslation("common");
+  const { t } = useTranslation("translation", {
+    keyPrefix: "nostr",
```
bumi, line 21: this should use the liquid prefix

src/extension/background-script/actions/liquid/generatePrivateKey.ts
bumi: not sure what you mean here.

src/extension/background-script/actions/liquid/generatePrivateKey.ts
```diff
@@ -0,0 +1,55 @@
...
+      message:
+        "DO NOT EVER SIGN THIS TEXT WITH YOUR PRIVATE KEYS! IT IS ONLY USED FOR DERIVATION OF LIQUID KEY PAIRS, DISCLOSING THIS SIGNATURE WILL COMPROMISE YOUR LIQUID IDENTITY!",
+      key_loc: {
```
bumi, line 36: are those key_loc params required? if so I think those are unused.

src/extension/background-script/state.ts
reneaaron: while the extension is locked you mean?

bumi: there is an instance of the `Liquid`, `Nostr`, `Connector` class memoized in the state. Those also need to be removed from memory here.

bumi:
after some chat we thought `window.secp256k1` is the best generic name.

im-adithya:
Closing in favour of #2467

PR #2013, opened 2023-01-26
fix: avoid showing welcome screen on re-install

im-adithya:

Whenever we remove the extension and load it again, the state is persisted, but the user is redirected to welcome page to set a new password. But we can still unlock the extension using the last used password before removing the extension.

### Type of change

## Before
https://user-images.githubusercontent.com/64399555/214838447-dabe7d82-2e7b-4cc2-9cfd-3705240b827d.mov

## After
https://user-images.githubusercontent.com/64399555/214838660-307d3253-e88e-4600-a078-75aa7f07809a.mov

### How has this been tested?

Tested manually.

### Checklist

### DIFF
--- a/src/extension/background-script/index.ts
+++ b/src/extension/background-script/index.ts
@@ -158,7 +158,7 @@ browser.runtime.onInstalled.addListener(handleInstalled);
 
 console.info("Welcome to Alby");
 init().then(() => {
-  if (isFirstInstalled) {
+  if (isFirstInstalled && !state.getState().getAccount()) {
     utils.openUrl("welcome.html");
   }
   if (isRecentlyUpdated) {


src/extension/background-script/index.ts
```diff
@@ -158,7 +158,7 @@ browser.runtime.onInstalled.addListener(handleInstalled);
 
 console.info("Welcome to Alby");
 init().then(() => {
-  if (isFirstInstalled) {
+  if (isFirstInstalled && !state.getState().accounts) {
```
bumi, line 161: accounts is `{}` which is truthy, isn't it? so I'd do the same check as here: 

https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/setup/status.js#L6-L7

im-adithya: Done!

PR #2015, opened 2023-01-27
chore: refactor actions and add tests

im-adithya:

Adds tests to actions in the background-script
Migrates actions to TypeScript

### Type of change

- `chore`: Refactoring (non-breaking change to the code)

### How has this been tested?

Wrote unit tests

### Checklist


escapedcat (approved):
utack

src/extension/background-script/actions/ln/makeInvoice.ts
```diff
@@ -20,7 +21,9 @@ const makeInvoice = async (message, sender) => {
       });
       return response;
     } catch (e) {
-      return { error: e.message };
+      if (e instanceof Error) {
```
rolznz, line 24: what happens if e is not an Error? will this exception be swallowed?

I think it would be great to have a helper function we can use to get a message from an exception

rolznz (approved):

src/extension/background-script/actions/ln/makeInvoice.ts
escapedcat: Didn't you mention this recently somewhere else? Wanna create an issue for this?

im-adithya: It's because of TypeScript; it should always be an error (according to the makeInvoice function)

But yeah, we use this repeatedly and it's best to have a helper function which somehow extracts the message from exception var. Not sure how we do that, though.

PR #2017, opened 2023-01-28
Add swr caching for selected account

### LINKED ISSUE #1923: [Feature] Add SWR library to Alby extension
### Feature description

It looks like there is some custom code in lib/api and lib/cache to get account info, but it is only used for a single call and it is a custom implementation of SWR(https://github.com/vercel/swr) however with is very limited and will have to be refactored to be more general use for the other API calls.

It's best to look at the SWR project, but immediate benefits are features such as automatic retries, displaying cached data while a newer version of the data loads (removing "flicker" when switching tabs and having to wait for data to re-load), automatic debouncing, optional polling, and many more.

Introducing SWR will make the user experience better and simplify development.

### Describe the solution

Add the SWR package https://github.com/vercel/swr and eventually migrate all possible API calls to use SWR when fetching data.

### Describe alternatives

There are other libraries available such as https://github.com/TanStack/query

### Additional context

Originally from https://github.com/getAlby/lightning-browser-extension/pull/1892#issuecomment-1366346743

### Are you working on this?

None

lujakob:

Replace custom caching code in lib/api and lib/cache to get account info with SWR(https://github.com/vercel/swr)

I was interested in this issue and started debugging and testing. I got it working at some point, that's why i share it here. I'm not sure if this is how it was intended however. Feel free go give me any suggestions on how to modify or improve. In case it doesn't help at all, I'm not bothered by discarding the whole PR in favour of other implementations.

Fixes #1923

### Type of change

### How has this been tested?

Manually so far

### Checklist

### DIFF
--- a/src/app/context/AccountContext.tsx
+++ b/src/app/context/AccountContext.tsx
@@ -7,15 +7,10 @@ import {
 } from "react";
 import { toast } from "react-toastify";
 import { useSettings } from "~/app/context/SettingsContext";
-import api, { StatusRes } from "~/common/lib/api";
+import api, { StatusRes, UnlockRes } from "~/common/lib/api";
 import msg from "~/common/lib/msg";
 import utils from "~/common/lib/utils";
 import type { AccountInfo } from "~/types";
-
-interface AccountInfoData extends AccountInfo {
-  fiatBalance: string;
-  accountBalance: string;
-}
 
 interface AccountContextType {
   account: Partial<AccountInfo> | null;
@@ -31,11 +26,9 @@ interface AccountContextType {
    */
   setAccountId: (id: string) => void;
   /**
-   * Fetch the additional account info: alias/balance and update account
+   * revalidation (mark the data as expired and trigger a refetch) for the resource
    */
-  fetchAccountInfo: (options?: {
-    accountId?: string;
-  }) => Promise<AccountInfoData | undefined>;
+  updateAccountInfo: () => Promise<void>;
 }
 
 const AccountContext = createContext({} as AccountContextType);
@@ -52,17 +45,23 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
   const [loading, setLoading] = useState(true);
   const [accountBalance, setAccountBalance] = useState("");
   const [fiatBalance, setFiatBalance] = useState("");
-  const [status, setStatus] = useState<StatusRes>();
-  const [shouldFetchAccountInfo, setShouldFetchAccountInfo] = useState(false);
+  const [, setStatus] = useState<StatusRes>();
 
   const isSatsAccount = account?.currency === "BTC"; // show fiatValue only if the base currency is not already fiat
   const showFiat =
     !isLoadingSettings && settings.showFiat && !loading && isSatsAccount;
 
+  const updateFiatValue = useCallback(
+    async (balance: string | number) => {
+      const fiat = await getFormattedFiat(balance);
+      setFiatBalance(fiat);
+    },
+    [getFormattedFiat]
+  );
+
   const unlock = (password: string, callback: VoidFunction) => {
-    return api.unlock(password).then((response) => {
-      setAccountId(response.currentAccountId);
-      fetchAccountInfo({ accountId: response.currentAccountId });
+    return api.unlock(password).then(({ currentAccountId }: UnlockRes) => {
+      setAccountId(currentAccountId);
 
       // callback - e.g. navigate to the requested route after unlocking
       callback();
@@ -76,15 +75,18 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
     });
   };
 
-  const setAccountId = (id: string) => setAccount({ id });
+  const setAccountId = async (id: string) => {
+    setAccount({ id });
+  };
 
-  const updateFiatValue = useCallback(
-    async (balance: string | number) => {
-      const fiat = await getFormattedFiat(balance);
-      setFiatBalance(fiat);
-    },
-    [getFormattedFiat]
-  );
+  const updateAccountInfo = async () => {
+    const accountInfo = await api.swr.mutateAccountInfo(account?.id || null);
+    if (accountInfo) {
+      // eslint-disable-next-line
+      setAccount(accountInfo as any);
+      updateAccountBalance(accountInfo.balance, accountInfo.currency);
+    }
+  };
 
   const updateAccountBalance = (
     amount: number,
@@ -94,40 +96,23 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
     setAccountBalance(balance);
   };
 
-  // const fetchAccountInfo = async (options?: { accountId?: string }) => {
-  //   const id = options?.accountId || account?.id;
-  //   if (!id) return;
-  //
-  //   const callback = (accountRes: AccountInfo) => {
-  //     setAccount(accountRes);
-  //     updateAccountBalance(accountRes.balance, accountRes.currency);
-  //   };
-  //
-  //   const accountInfo = await api.swr.getAccountInfo(id, callback);
-  //
-  //   return { ...accountInfo, fiatBalance, accountBalance };
-  // };
+  // hook: will be called each round, pass "id = null" to skip
+  const fetchAccountInfo = async () => {
+    const accountInfo = await api.swr.useSwrGetAccountInfoTest(
+      getIdIfShouldFetch()
+    );
+    if (!accountInfo) return;
 
-  const fetchAccountInfo = async (options?: {
-    accountId?: string | null;
-  }): Promise<AccountInfoData> => {
-    const id = options?.accountId || null;
-
-    const accountInfo = await api.swr.useSwrGetAccountInfoTest(id);
-
-    if (accountInfo) {
-      // eslint-disable-next-line
-      setAccount(accountInfo as any);
-      updateAccountBalance(accountInfo.balance, accountInfo.currency);
-    }
-
-    return { ...accountInfo, fiatBalance, accountBalance } as AccountInfoData;
+    setAccount(accountInfo);
+    updateAccountBalance(accountInfo.balance, accountInfo.currency);
   };
 
-  const maybeGetAccountId = () => {
-    return !account?.name && shouldFetchAccountInfo
-      ? status?.currentAccountId || null
-      : null;
+  /**
+   * implicitly determines if data should be fetched
+   * if new account was selected, only "id" prop is available, and name is missing
+   */
+  const getIdIfShouldFetch = () => {
+    return !account?.name ? account?.id || null : null;
   };
 
   const handleGetStatus = (status: StatusRes) => {
@@ -141,27 +126,20 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
         utils.redirectPage("options.html");
       }
       setAccountId(status.currentAccountId);
-      setShouldFetchAccountInfo(true);
     } else {
       setAccount(null);
     }
   };
 
-  fetchAccountInfo({ accountId: maybeGetAccountId() });
+  fetchAccountInfo();
 
   // Invoked only on on mount.
   useEffect(() => {
     api
       .getStatus()
-      .then((status: StatusRes) => {
-        handleGetStatus(status);
-      })
-      .catch((e) => {
-        toast.error(`An unexpected error occurred (${e.message})`);
-      })
-      .finally(() => {
-        setLoading(false);
-      });
+      .then((status: StatusRes) => handleGetStatus(status))
+      .catch((e) => toast.error(`An unexpected error occurred (${e.message})`))
+      .finally(() => setLoading(false));
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
@@ -187,7 +165,7 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
       accountBalance,
       fiatBalance,
     },
-    fetchAccountInfo,
+    updateAccountInfo,
     loading,
     lock,
     setAccountId,
--- a/src/app/context/AccountContext.tsx
+++ b/src/app/context/AccountContext.tsx
@@ -7,19 +7,18 @@ import {
 } from "react";
 import { toast } from "react-toastify";
 import { useSettings } from "~/app/context/SettingsContext";
-import api from "~/common/lib/api";
+import api, { StatusRes } from "~/common/lib/api";
 import msg from "~/common/lib/msg";
 import utils from "~/common/lib/utils";
 import type { AccountInfo } from "~/types";
 
+interface AccountInfoData extends AccountInfo {
+  fiatBalance: string;
+  accountBalance: string;
+}
+
 interface AccountContextType {
-  account: {
-    id: AccountInfo["id"];
-    name?: AccountInfo["name"];
-    alias?: AccountInfo["alias"];
-    balance?: AccountInfo["balance"];
-    currency?: AccountInfo["currency"];
-  } | null;
+  account: Partial<AccountInfo> | null;
   balancesDecorated: {
     fiatBalance: string;
     accountBalance: string;
@@ -36,7 +35,7 @@ interface AccountContextType {
    */
   fetchAccountInfo: (options?: {
     accountId?: string;
-  }) => Promise<AccountInfo | undefined>;
+  }) => Promise<AccountInfoData | undefined>;
 }
 
 const AccountContext = createContext({} as AccountContextType);
@@ -53,6 +52,8 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
   const [loading, setLoading] = useState(true);
   const [accountBalance, setAccountBalance] = useState("");
   const [fiatBalance, setFiatBalance] = useState("");
+  const [status, setStatus] = useState<StatusRes>();
+  const [shouldFetchAccountInfo, setShouldFetchAccountInfo] = useState(false);
 
   const isSatsAccount = account?.currency === "BTC"; // show fiatValue only if the base currency is not already fiat
   const showFiat =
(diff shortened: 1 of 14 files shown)


lujakob:
@rolznz I will probably clean this up and might change again tomorrow. Please have a look and let me know what you think. Thanks.

src/common/lib/api.ts
```diff
@@ -33,64 +29,76 @@ export interface AccountInfoRes {
...
+
+export const getAccountInfoKey = (
+  id: string | null | undefined
+): string | null => (id ? `accountInfo/${id}` : null);
+
+// useSWR hook is expected to be called in every render, hence we expect parameter id null to skip fetching
+export const useAccountInfoCached = async (
+  id: string | null
+): Promise<AccountInfo | null> => {
+  const { data, error } = await useSWR<AccountInfoRes>(
```
rolznz, line 57: can you return the mutate function here too? unless you need to mutate an account that is not currently selected?

src/app/context/AccountContext.tsx
```diff
@@ -32,11 +26,9 @@ interface AccountContextType {
    */
   setAccountId: (id: string) => void;
   /**
-   * Fetch the additional account info: alias/balance and update account
+   * revalidation (mark the data as expired and trigger a refetch) for the resource
    */
-  fetchAccountInfo: (options?: {
-    accountId?: string;
-  }) => Promise<AccountInfo | undefined>;
+  updateAccountInfo: () => Promise<void>;
```
rolznz, line 31: I wonder if this method could be renamed to fetchAccountInfo? in my opinion it doesn't actually "update" the account

src/common/lib/api.ts
```diff
@@ -143,7 +153,8 @@ export default {
   connectPeer,
   setSetting,
   swr: {
-    getAccountInfo: swrGetAccountInfo,
+    useAccountInfoCached: useAccountInfoCached,
+    mutateAccountInfo: mutateAccountInfo,
```
rolznz, line 157: I don't think these methods should live in an "swr" object (I know it was originally like this, I don't think it's the right approach). We can use swr with any of the api methods. Also, I don't think we should introduce mutation functions here that didn't exist, especially if they only mutate the cached SWR data.

src/common/lib/api.ts
```diff
@@ -33,64 +29,76 @@ export interface AccountInfoRes {
...
+    getAccountInfoKey(id),
+    getAccountInfo
+  );
+  if (!data) return null;
+
+  return buildAccountInfo(id, data);
+};
+
+const buildAccountInfo = (
+  id: string | null,
```
rolznz, line 83: the id should exist, I don't think it should be nullable here. I think we should never use buildAccountInfo if the id or data isn't available.

src/common/lib/api.ts
```diff
@@ -33,64 +29,76 @@ export interface AccountInfoRes {
...
+  if (error) throw error;
+
+  return buildAccountInfo(id, data);
 };
+
+// revalidation (mark the data as expired and trigger a refetch) for the resource
+export const mutateAccountInfo = async (
+  id: string | null
+): Promise<AccountInfo | null> => {
+  const data = await mutate<AccountInfoRes>(
```
rolznz, line 73: Just mutating the value should be enough - SWR will return the new value in the next render. You don't need to return buildAccountInfo() as the response.

src/common/lib/api.ts
```diff
@@ -111,9 +119,11 @@ export const setSetting = (setting: MessageSettingsSet["args"]["setting"]) =>
     setting,
   });
 export const removeAccount = (id: string) =>
+  // @Todo: maybe remove Promise.all
   Promise.all([
     msg.request("removeAccount", { id }),
-    removeAccountFromCache(id),
+    // @Todo: how to remove account from swr cache?
+    // @Todo swr dows say "mutate(key)" will mark the data as expired and trigger a refetch for the resource
```
rolznz, line 126: I don't think this is really necessary to delete it from the SWR cache since it's only stored for the duration of the session

rolznz:
> @rolznz I will probably clean this up and might change again tomorrow. Please have a look and let me know what you think. Thanks.

Thanks for your changes @lujakob !

I think some parts of this PR are more complicated than necessary. I hope my comments help.

src/app/context/AccountContext.tsx
```diff
@@ -93,46 +96,50 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
     setAccountBalance(balance);
   };
 
-  const fetchAccountInfo = async (options?: { accountId?: string }) => {
-    const id = options?.accountId || account?.id;
-    if (!id) return;
+  // hook: will be called each round, pass "id = null" to skip
+  const fetchAccountInfo = async () => {
```
rolznz, line 100: This doesn't seem right. SWR hooks should be used within components or other hooks.

You can use an effect when the data from the SWR hook changes, to update the account info and balance, etc.

src/common/lib/api.ts
lujakob: By introducing swr we split the functionality of initial fetching on account change (in AccountContext) and  updating/mutating (in any other component that triggers an update/mutation). Hence there need to be two functions.
I will remove the "swr" nesting, to me it also seemed strange.

src/common/lib/api.ts
lujakob: This is actually due to a Typescript type problem: due to the id coming from the calling function "useAccountInfoCached" that calls the "useSWR" hook. By design the hooks need to be called every single render turn and the mechanics to skip it is to pass a key of null to useSWR. That's why the parent function id parameter can be null and the type here therefore as well. Do you have an idea how to solve it? If I try non-null assertion "id!" with the id argument, I get the lint error: "Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion"

lujakob: Actually I'm using "// eslint-disable-next-line @typescript-eslint/no-non-null-assertion" in this function. I can move it up to the calling function. But the issue of an id parameter of null coming in will stay there, because of what I explained above. Unless you have a better idea.

src/common/lib/api.ts
lujakob: If I remove this the AccountMenu balance of the selcted account will not be updated after having a payment success of the "Receive" page. As far as I see it, there is no render in the AccountContext when the screen of the Receive page has received the account. For this reason the mutate being called, needs to refetch the account data and update the states in AccountContext

`
  const refetchAccountInfo = async () => {
    const accountInfo = await api.refetchAccountInfo(account?.id || null);
    if (accountInfo) {
      // eslint-disable-next-line
      setAccount(accountInfo as any);
      updateAccountBalance(accountInfo.balance, accountInfo.currency);
    }
  };
`

src/common/lib/api.ts
lujakob: Yes, good point. By refactoring the `useSWR` to the component level I can easily use bound mutate. This really simplifies things.

src/app/context/AccountContext.tsx
lujakob: I refactored to call `useSWR` to the component level. This really simplifies things.

src/app/context/AccountContext.tsx
```diff
@@ -32,15 +33,37 @@ interface AccountContextType {
...
    */
-  fetchAccountInfo: (options?: {
-    accountId?: string;
-  }) => Promise<AccountInfo | undefined>;
+  refetchAccountInfo: () => Promise<void>;
 }
 
 const AccountContext = createContext({} as AccountContextType);
 
+export const getAccountInfoKey = (
```
lujakob, line 43: @rolznz `getAccountInfoKey` and `buildAccountInfo` are also used in TestConnection. Should I create a `app/utils/accounts.ts` function and move them there?

src/common/lib/api.ts
```diff
@@ -33,64 +27,23 @@ export interface AccountInfoRes {
   name: string;
 }
 
-interface StatusRes {
+export interface StatusRes {
```
lujakob, line 30: @rolznz I export the interfaces to be able to use them in AccountContext. Would it make sense to move them to "src/types.ts"?

lujakob:
@rolznz I addressed all of your comments and added three more questions for you. Do you have any further suggestions?

src/app/screens/Options/TestConnection/index.tsx
```diff
@@ -4,34 +4,39 @@ import Loading from "@components/Loading";
...
-    alias: AccountInfo["alias"];
-    name: AccountInfo["name"];
-    balance: AccountInfo["balance"];
-    currency: AccountInfo["currency"];
-  }>();
+  const [accountInfo, setAccountInfo] = useState<Partial<AccountInfo>>();
+  const [currentAccountId, setCurrentAccountId] = useState<string>("");
   const [errorMessage, setErrorMessage] = useState("");
   const [loading, setLoading] = useState(false);
+  const { data: accountInfoRes } = useSWR<AccountInfoRes>(
```
lujakob, line 28: The TestConnection is actually a one time action after adding a new account. Does it even make sense to apply `useSWR` in this case? Wouldn't it even make more sense to just apply the fetcher function directly? 

`export const getAccountInfo = (): Promise<AccountInfoRes> =>
  msg.request<AccountInfoRes>("accountInfo")`

src/common/lib/api.ts
rolznz: another option is that you can use the mutate function from the useSWR hook in the component that needs to update it. I think we should have a separation from swr and the API methods. SWR is just an optional way to retrieve data, it shouldn't be mixed in with actual API methods.

src/common/lib/api.ts
rolznz: That makes sense to me but let's leave them here for now to reduce the PR changes

src/app/screens/Options/TestConnection/index.tsx
rolznz: it will save the account in the SWR in memory cache for the next use (otherwise in this case it would have to fetch the account info twice), so it is nice to use SWR in my opinion

src/app/context/AccountContext.tsx
```diff
@@ -75,15 +111,15 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
...
 
-  const updateFiatValue = useCallback(
-    async (balance: string | number) => {
-      const fiat = await getFormattedFiat(balance);
-      setFiatBalance(fiat);
-    },
-    [getFormattedFiat]
-  );
+  const refetchAccountInfo = async () => {
+    if (typeof mutate !== "function") return;
```
rolznz, line 119: why is this needed? `if (typeof mutate !== "function") return;`

src/app/context/AccountContext.tsx
```diff
@@ -49,19 +72,32 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
...
   } = useSettings();
 
-  const [account, setAccount] = useState<AccountContextType["account"]>(null);
+  const [account, setAccount] = useState<Partial<AccountInfo> | null>(null);
   const [loading, setLoading] = useState(true);
   const [accountBalance, setAccountBalance] = useState("");
   const [fiatBalance, setFiatBalance] = useState("");
+  const [, setStatus] = useState<StatusRes>();
+
+  const { data: accountInfoRes, mutate } = useSWR<AccountInfoRes>(
```
rolznz, line 81: could you rename mutate? like `, mutate: mutateAccountInfo` so it is more obvious what it's doing?

src/app/context/AccountContext.tsx
```diff
@@ -75,15 +111,15 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
...
-    async (balance: string | number) => {
-      const fiat = await getFormattedFiat(balance);
-      setFiatBalance(fiat);
-    },
-    [getFormattedFiat]
-  );
+  const refetchAccountInfo = async () => {
+    if (typeof mutate !== "function") return;
+    const data = await mutate();
+    setAccountData(account?.id, data);
```
rolznz, line 121: I still don't think `setAccountData(account?.id, data);`  is needed. If you call mutate it will refetch the data using SWR and update the SWR cache.

src/app/context/AccountContext.tsx
```diff
@@ -93,46 +129,45 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
...
+    if (id && accountInfoRes) {
+      const accountInfo = buildAccountInfo(id, accountInfoRes);
+      setAccount(accountInfo);
+      updateAccountBalance(accountInfo.balance, accountInfo.currency);
+    }
   };
 
+  useEffect(() => {
+    setAccountData(account?.id, accountInfoRes);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rolznz, line 161: can you fix the warning rather than disabling it?

src/app/context/AccountContext.tsx
rolznz: I think it would be better to make a reusable hook e.g. src/app/hooks/useSelectedAccount and put all this logic in one place. I think it might just need an accountId parameter?

Then in the TestConnection and AccountContext you can just use the hook. This would also make AccountContext simpler too. What do you think?

lujakob: I already used to have this abstracted as "useAccountInfoCached" in "api.ts" - then you mentioned to simplify things and call "useSWR" directly in the AccountContext. I guess it was a misunderstanding then and I also think it makes sense, so I refactor to abstract it and move it to the suggested place.

src/app/context/AccountContext.tsx
lujakob: Indeed. Now the AccountMenu is updated even though I don't setAccountData. Not sure why it didn't work when I tried initially. Good you persisted.

src/app/context/AccountContext.tsx
lujakob: I would need to add setAccountData and apply useCallback() - I didn't really get it working and seems like this makes it overly complicated. Is it a problem though? This exception is used in many other effects!?

lujakob:
@rolznz I addressed all your concerns. I created the useSelectAccount hook and the AccountContext and TestConnection are even leaner now. Please have another look at it.

src/app/hooks/useSelectedAccount.ts
```diff
@@ -0,0 +1,49 @@
...
+    id,
+    name,
+    alias,
+    balance,
+    currency,
+  };
+};
+
+export function useSelectedAccount(id: string | undefined) {
+  const { data, mutate } = useSWR<AccountInfoRes>(
```
rolznz, line 33: you can just do mutate: mutateAccountInfo and then the function will be memoized. Currently the way you're doing it will create a new mutateAccountInfo function each time the hook is accessed

src/app/hooks/useSelectedAccount.ts
```diff
@@ -0,0 +1,49 @@
...
+  return {
+    id,
+    name,
+    alias,
+    balance,
+    currency,
+  };
+};
+
+export function useSelectedAccount(id: string | undefined) {
```
rolznz, line 32: Sorry, I realized this should probably be called "useAccount" not "useSelectedAccount"

rolznz: Edit: I see this would conflict with the AccountContext. hmm

src/app/screens/Options/TestConnection/index.tsx
```diff
@@ -68,6 +50,21 @@ export default function TestConnection() {
...
+   * to keep TestConnection account,
+   * even though the account menu might be changed
+   */
+  const handleSelectAccount = async () => {
+    const { currentAccountId } = await api.getStatus();
+    await msg.request("selectAccount", {
+      id: currentAccountId,
+    });
+    setAccountId(currentAccountId);
+    setCurrentAccountId(currentAccountId);
```
rolznz, line 64: I don't understand this, why do we need another variable to track the current account? when we create a new account it should be selected, and we test that one. I don't think a user would change the account while the connection is being tested.

src/app/context/AccountContext.tsx
rolznz: @lujakob sorry about that.

src/app/context/AccountContext.tsx
rolznz: It might work fine. It is just a code smell. I'm not sure how important it is to fix for now.

src/app/screens/Onboard/TestConnection/index.tsx
```diff
@@ -17,7 +17,7 @@ export default function TestConnection() {
 
   const navigate = useNavigate();
 
-  async function handleEdit(event: React.MouseEvent<HTMLButtonElement>) {
+  async function handleEdit() {
```
rolznz, line 20: Are there any changes required in this component? I see there are two different components for testing connection (onboarding vs adding an account from the options).

Please note: I do not fully understand the code in TestConnection/AccountContext. Is there someone who has more knowledge of the code who might be able to review this? @escapedcat

escapedcat: This looks like a "cleanup only", right @lujakob ? It's not needed but can be done I guess. Could be done in a different PR top keep things scoped but not a blocker really.
@rolznz is right, we have two of these. One for "adding an accoiunt via welcome/onboarding" and another one for "adding a new account after the onboarding already happened an d an accoiunt exists", which is here:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Options/TestConnection/index.tsx#L34-L37

Does this help?

src/app/screens/Options/TestConnection/index.tsx
lujakob: @rolznz what should happen when the user changes the selected account in the AccountMenu? My intention was to not change the screen data, because I thought the screen refers to the just before created Account. If you think it makes more sense for the screen info to reflect whatever account is or will be selected in in the AccountMenu, we can simplify this by just jusing accound.id directly.

src/app/screens/Onboard/TestConnection/index.tsx
lujakob: Yes that's a minor cleanup. Should I revert it, to be addressed at another point?

There are two TestConnection components. The "loadAccountInfo" handler is different however. I guess they are different.

src/app/hooks/useSelectedAccount.ts
lujakob: I updated

lujakob:
@rolznz @escapedcat thanks for reviewing. I updated accordingly.

src/app/context/AccountContext.tsx
lujakob: No worries.

src/app/screens/Options/TestConnection/index.tsx
rolznz: Sorry, I misunderstood. I see this is for the Options version of TestConnection. I don't know what should happen in this case, probably it should navigate away from this page if the user selects a different account.

I am now thinking it might be better to have these TestConnection changes in a separate PR and keep this one as just replacing the custom SWR code with the SWR library. I think your were correct in saying that it isn't that important since this is only used once in setup.

src/common/lib/api.ts
lujakob: It works now. This is resolved.

src/app/screens/Options/TestConnection/index.tsx
lujakob: @rolznz so we keep as is for now?

rolznz: Maybe, or we could revert the TestConnection changes for now and put them in a separate PR later, once this one is merged.

lujakob: I really recommend not to revert back. We would need to bring back a lot of code in AccountContext.fetchAccountInfo, api.swr.getAccountInfo and getAccountsCache that I already removed. I mean I can revert, back it is quite a lot of code and for no real reason. Right now we have a lot simplified. I suggest to keep as is. What would be the reason to revert and reimplement later?

rolznz: Sounds good, let's keep it as is then.

I just suggested it to make this PR more easier to be reviewed by other Alby team members.

lujakob:
@rolznz do you get why the e2e test doesn't succeed? Seems like problem with loading a dependency?

rolznz:
> @rolznz do you get why the e2e test doesn't succeed? Seems like problem with loading a dependency?

Yes, I think it is related to some changes that were recently merged into master. Hopefully it will be fixed soon.

escapedcat:
> @rolznz do you get why the e2e test doesn't succeed? Seems like problem with loading a dependency?

Yeah, no worries, no your fault. Restarting these helps. Did that now.

rolznz:
Some improvements in this PR:

- Instant account switching for accounts you have already connected to
- When switching account, it shows correct account icon & name
- When switching to a new account, it will show a loading component (needs work, and is only shown in the popup right now)
- Account context is a lot simpler (extracted status )

I am wondering now if we should have a loading screen for switching to a new connector rather than showing some skeleton loaders (we are not sure the connector can be connected to or not). For switching back to a connector you already connected to, it's now instant (because it uses SWR) but we should still check if the connection succeeds, otherwise show the connection failed screen.

CC @lujakob @reneaaron

lujakob:
> Some improvements in this PR:
> 
> * Instant account switching for accounts you have already connected to
> * When switching account, it shows correct account icon & name
> * When switching to a new account, it will show a loading component (needs work, and is only shown in the popup right now)
> * Account context is a lot simpler (extracted status )
> 
> I am wondering now if we should have a loading screen for switching to a new connector rather than showing some skeleton loaders (we are not sure the connector can be connected to or not). For switching back to a connector you already connected to, it's now instant (because it uses SWR) but we should still check if the connection succeeds, otherwise show the connection failed screen.
> 
> CC @lujakob @reneaaron
> 
> I wonder if it makes more sense to show a loading screen instead of skeleton loaders (#2456)

This PR is already 5 months old and interferes with a few current ones (account loading, error cases). I think we should finish the current ones and then in case re-address this one. Otherwise it gets even more confusing as it already is (we have several different issues on the topic of account loading/switch, error handling). Additionally this PR has a lot of changes and also has probably quite a few merge conflicts with master due to its age.

rolznz:
@lujakob I already addressed the merge conflicts. I don't know if the existing loading screen with skeleton loaders is the best way to do it, and does not work for the first account load (only switching accounts).

But I see what you mean - let's leave this issue for now and keep the discussion in the other tickets.

Issue #1632, opened 2022-10-18
With Alby setting page opened and try to change password with Alby locked, should show an error message


bitcoinuser:
### Describe the bug

I opened Alby settings page, then I locked the Alby, then I tried to change lock password, nothing happens and no error message appear

### To Reproduce

1. Open Alby settings page
2. Lock the extension on Alby fav icon on browser top
3. Try to change Alby lock password

### Expected behavior

Alby should show a message to unlock the extension first to change the password.

### Alby information

- Alby Version: 1.17

ShivamJhaa:
Hi I would like to work on this issue, can you please assign me?

PR #2025, opened 2023-01-30
fix: added error message for change password

ShivamJhaa:

Added error message when we try to update the password and alby account is locked.

Fixes _[#1632](https://github.com/getAlby/lightning-browser-extension/issues/1632)

### Type of change

https://user-images.githubusercontent.com/86483059/215387001-daec638b-9b00-458b-a132-8c8c65ef9c3e.mp4

### How has this been tested?

### Checklist


ShivamJhaa:
@im-adithya Can you please review this PR and start the CI test for the PR?
Thanks!

im-adithya:
Thanks for the PR @ShivamJhaa!

However, as @rolznz suggested, this page shouldn't appear whenever the user locks the extension. So instead of showing an error toast, can you modify the code to close all the extension tabs on lock?

Here's the lock function: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/state.ts#L119

You can fetch all the currently open extension tabs and close them there. See https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews

Again, thanks for letting us know about the issue; let me know if you need help!

bumi (commented):
thanks for the PR!

bumi:
@im-adithya @ShivamJhaa I'd do both. catching a potential error and showing an error message is still a good idea I think. maybe something else goes wrong, so at least the user knows.

src/app/screens/Settings.tsx
```diff
@@ -77,11 +77,17 @@ function Settings() {
...
+    try {
+      await msg.request("changePassword", {
+        password: formData.password,
+      });
+
+      toast.success(t("change_password.success"));
+      closeModal();
+    } catch (e) {
+      console.error(e);
+      if (e instanceof Error) toast.error(`An unexpected error occurred`);
```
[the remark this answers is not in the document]
ShivamJhaa: Done, thanks.

ShivamJhaa:
> Thanks for the PR @ShivamJhaa!
> 
> However, as @rolznz suggested, this page shouldn't appear whenever the user locks the extension. So instead of showing an error toast, can you modify the code to close all the extension tabs on lock?
> 
> Here's the lock function: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/state.ts#L119
> 
> You can fetch all the currently open extension tabs and close them there. See https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews
> 
> Again, thanks for letting us know about the issue; let me know if you need help!

Thanks for the help, I have made the changes as suggested. Please take a look.

https://user-images.githubusercontent.com/86483059/215647925-702ae12b-f4f5-49e3-bfaa-a87a658dd339.mp4

escapedcat (approved):
ack

escapedcat (approved):
ack

src/app/components/UserMenu/index.tsx
```diff
@@ -35,6 +35,10 @@ export default function UserMenu() {
   async function lock() {
     try {
       auth.lock(() => {
+        const allTabs = chrome.extension.getViews({ type: "tab" });
```
im-adithya, line 38: It's best to use the `browser` polyfill as we do this in Firefox as well.
You can shift this to https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/state.ts#L119 so that you won't have to import `browser` again.
And the `window.close()` would stay as the callback (to `auth.lock` fn) to close the popup

ShivamJhaa: I tried adding the same code here https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/state.ts#L119 as suggested, but this is not working. I tried adding `console.log` statements in order to find out whether this function is even getting called or not when we click on lock option on the popup, but the logging were not presents which basically confirms this function is not getting called. So can I just replace chrome with browser polyfill in this file only?

ShivamJhaa:
@im-adithya Please take a look, I have resolved the merge conflicts and need some help over this comments https://github.com/getAlby/lightning-browser-extension/pull/2025#discussion_r1094659789
Thanks!

src/app/components/UserMenu/index.tsx
im-adithya: You are checking these logs, right? (you should be)
Screenshot 2023-02-07 at 1 09 17 PM

im-adithya: This works for me 💁‍♂️ 
```
lock: async () => {
    const allTabs = browser.extension.getViews({ type: "tab" });
    for (const tab of allTabs) {
      tab.close();
    }
    const connector = get().connector;
    if (connector) {
      await connector.unload();
    }
    set({ password: null, connector: null, account: null, nostr: null });
  }

ShivamJhaa: Yes, it is working. Actually I need to restart the server in order to see the changes in effect. Sorry for the confusion.

ShivamJhaa:
@im-adithya Done the changes as suggested.

im-adithya (approved):

Issue #2027, opened 2023-01-30
"All transactions" tab


bumi:
### Feature description

Currently we show the transactions only per website and have a website list.
People have been asking to see all the transactions. 

### Describe the solution

Add an additional page "Transactions" in the top navigation next to "Websites" and show a chronological (latest fist) list of all transactions there. 

The UI and components can be reused from the website screen

### Are you working on this?

None

lujakob:
@bumi I'm interested in working on this? Can you assig it to me?

lujakob:
@bumi it would look something like this? I reused the `<TransactionsTable>` from the publisher detail page. Any comments/suggestions on this?

<img width="1071" alt="Bildschirmfoto 2023-01-30 um 22 16 00" src="https://user-images.githubusercontent.com/886152/215661635-f6817627-245c-4377-8008-a80d4dd21bdd.png">

lujakob:
@bumi I get the data from https://github.com/lujakob/lightning-browser-extension/blob/master/src/extension/background-script/actions/payments/all.ts#L5-L5

That's correct?

lujakob:
@bumi for the translations I created keys and translations in the `src/i18n/locales/en/translation.json` - how are translations handled in general. Is there an external tool and I only provide the keys? Or how is it done?

reneaaron:
@lujakob Usually we just add the English keys & terms, they are then synced to Weblate where they are being translated.

bumi:
@lujakob yes, correct. that should work. at some point we need some pagination there, but I think it's ok for now.

lujakob:
> @lujakob yes, correct. that should work. at some point we need some pagination there, but I think it's ok for now.

I also thought about pagination, considering adding it in a further step. I tested now with 1000 payments in IndexDB and it loads quite fast on screen change, as I expected to. So I think it is ok for now to implement without Pagination.

PR #2034, opened 2023-01-31
Feat/performance improved get formatted fiat

### LINKED ISSUE #2031: [Feature] Enable performance improved getFormattedFiat for transforming list of payments
### Feature description

The SettingsContext getFormattedFiat function is used to convert sats into fiat in several places of the app. The rate is fetched through messaging to the background script where an HTTP request is made. This request is cached for 10 minutes.

I found three places in the app, where this async getFormattedFiat function is called in a loop, traversing over payments. Even though the final HTTP request is cached for 10 minutes, I find it quite expensive (even though I don't have the exact number on performance costs) The three places are

https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Publisher.tsx#L54
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/DefaultView/index.tsx#L95
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/DefaultView/index.tsx#L157
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/AllowanceView/index.tsx#L63

### Describe the solution

As the rate is the same for all payments in the loop, I propose the get the rate before the loop and only transform the number into an object inside the loop.

Currently, fetching the rate and transforming the data in one function:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/context/SettingsContext.tsx#L74-L84
makes sense for all of the places where a single number is converted. 

For the places above where a potentially long array of payments needs to be transformed, I propose to provide a slight refactor and expose a function in SettingsContext, where the number and rate is provided as arguments to make it sync. To improve the loops, the rate would be fetched before the loop and inside the loop the sync function is used, being provided the rate and number.

I stumbled upon this working on the "transactions page" feature. So I could do this immediately, as I can use it in this feature.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

Yes

lujakob:

Remove blocking `getFormattedFiat` from loops to improve performance. Get the rate with `getCurrencyRate` and just do the non-blocking `getFormattedFiatFromRate` inside the loop.

The changes are: Separate `getFormattedFiatUtil` from SettingsContext `getFormattedFiat` and export `getCurrencyRate` and `getFormattedFiatFromRate`.

The relevant changes are in SettingsContext. The application of this simplification and performance improvement is in the three screen components.

This is also needed for the new https://github.com/getAlby/lightning-browser-extension/issues/2027

Fixes #2031

### Type of change

- `refactor`: Performance improvement

### How has this been tested?

Manually.

### Checklist

### DIFF
--- a/src/app/context/SettingsContext.tsx
+++ b/src/app/context/SettingsContext.tsx
@@ -19,6 +19,8 @@ interface SettingsContextType {
   updateSetting: (setting: Setting) => void;
   isLoading: boolean;
   getFormattedFiat: (amount: number | string) => Promise<string>;
+  getCurrencyRate: () => Promise<number>;
+  getFormattedFiatFromRate: (amount: number | string, rate: number) => string;
   getFormattedSats: (amount: number | string) => string;
   getFormattedNumber: (amount: number | string) => string;
   getFormattedInCurrency: (
@@ -89,19 +91,21 @@ export const SettingsProvider = ({
   const getFormattedFiat = async (amount: number | string) => {
     try {
       const rate = await getCurrencyRate();
-      const value = getFormattedFiatUtil({
-        amount,
-        rate,
-        currency: settings.currency,
-        locale: settings.locale,
-      });
-
-      return value;
+      return getFormattedFiatFromRate(amount, rate);
     } catch (e) {
       console.error(e);
 
       return "??"; // show the user that something went wrong
     }
+  };
+
+  const getFormattedFiatFromRate = (amount: number | string, rate: number) => {
+    return getFormattedFiatUtil({
+      amount,
+      rate,
+      currency: settings.currency,
+      locale: settings.locale,
+    });
   };
 
   const getFormattedSats = (amount: number | string) => {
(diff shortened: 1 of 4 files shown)


escapedcat:
lgtm

im-adithya:
Thanks for the PR @lujakob!

But I think `getFormattedFiat` is already optimized to cache:

`getCurrencyRate` only calls the `api.getCurrencyRate` on currency change (so in the for loop case, the rates are cached in the localRef)
https://github.com/getAlby/lightning-browser-extension/blob/fb3b637e2f34e22dff3d13cb81b7f0ff6d8135e0/src/app/context/SettingsContext.tsx#L74-L86

And `api.getCurrencyRate` uses the browser storage to cache currency rates (for 10 minutes) (used when the user changes currency back and forth)
https://github.com/getAlby/lightning-browser-extension/blob/fb3b637e2f34e22dff3d13cb81b7f0ff6d8135e0/src/extension/background-script/actions/cache/getCurrencyRate.ts#L62-L73

lujakob:
@im-adithya I saw the 10 minute caching. But I completely overlooked the userRef caching in SettingsContext. Then I guess my changes are unnecessary. I just close the PR then, ok?

im-adithya:
Sure, thanks!

PR #2035, opened 2023-01-31
Feat/all transactions screen

### LINKED ISSUE #2027: "All transactions" tab

lujakob:

Add "Transactions" menu item and new screen to list all transactions independent of websites. Pagination is not implemented, as the payment items are loaded fast from IndexDB. This can be a follow up once it's needed.

Fixes #2027

### Type of change

Bildschirmfoto 2023-01-31 um 14 12 50

### How has this been tested?

Manually

### Checklist

- [ x] My code follows the style guidelines of this project and performed a self-review of my own code
- [ x] New and existing tests pass locally with my changes
- [ x] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

### DIFF
--- a/src/app/screens/Transactions/index.tsx
+++ b/src/app/screens/Transactions/index.tsx
@@ -1,0 +1,37 @@
+import Container from "@components/Container";
+import TransactionsTable from "@components/TransactionsTable";
+import { useTranslation } from "react-i18next";
+import { Transaction } from "~/types";
+
+function Transactions() {
+  const { t } = useTranslation("translation");
+
+  const transactions: Transaction[] = [];
+
+  return (
+    <Container>
+      <h2 className="mt-12 mb-2 text-2xl font-bold dark:text-white">
+        {t("transactions.title")}
+      </h2>
+
+      <p className="mb-6 text-gray-500 dark:text-neutral-500">
+        {t("transactions.description")}
+      </p>
+
+      <div className="mb-12">
+        <div>
+          {transactions?.length > 0 && (
+            <TransactionsTable transactions={transactions} />
+          )}
+          {!transactions?.length && (
+            <p className="mb-6 text-gray-500 dark:text-neutral-500">
+              {t("transactions.list_empty")}
+            </p>
+          )}
+        </div>
+      </div>
+    </Container>
+  );
+}
+
+export default Transactions;
--- a/src/app/screens/Transactions/index.tsx
+++ b/src/app/screens/Transactions/index.tsx
@@ -1,9 +1,10 @@
 import Container from "@components/Container";
 import TransactionsTable from "@components/TransactionsTable";
 import dayjs from "dayjs";
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { toast } from "react-toastify";
+import { useSettings } from "~/app/context/SettingsContext";
 import msg from "~/common/lib/msg";
 import { Payment, Transaction } from "~/types";
 
@@ -12,9 +13,10 @@ function Transactions() {
     keyPrefix: "transactions",
   });
 
+  const { settings, getFormattedFiat } = useSettings();
   const [payments, setPayments] = useState<Transaction[]>([]);
 
-  async function fetchData() {
+  const fetchData = useCallback(async () => {
     try {
       const paymentsResponse = await msg.request<{
         payments: Payment[];
@@ -31,16 +33,23 @@ function Transactions() {
         })
       );
 
+      for (const transaction of transactions) {
+        const totalAmountFiat = settings.showFiat
+          ? await getFormattedFiat(transaction.totalAmount)
+          : "";
+        transaction.totalAmountFiat = totalAmountFiat;
+      }
+
       setPayments(transactions);
     } catch (e) {
       console.error(e);
       if (e instanceof Error) toast.error(`Error: ${e.message}`);
     }
-  }
+  }, [settings, getFormattedFiat]);
 
   useEffect(() => {
     fetchData();
-  }, []);
+  }, [fetchData]);
 
   return (
     <Container>
--- a/src/app/screens/Transactions/index.tsx
+++ b/src/app/screens/Transactions/index.tsx
@@ -1,34 +1,59 @@
 import Container from "@components/Container";
 import TransactionsTable from "@components/TransactionsTable";
+import dayjs from "dayjs";
+import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
-import { Transaction } from "~/types";
+import { toast } from "react-toastify";
+import msg from "~/common/lib/msg";
+import { Payment, Transaction } from "~/types";
 
 function Transactions() {
-  const { t } = useTranslation("translation");
+  const { t } = useTranslation("translation", {
+    keyPrefix: "transactions",
+  });
 
-  const transactions: Transaction[] = [];
+  const [payments, setPayments] = useState<Transaction[]>([]);
+
+  async function fetchData() {
+    try {
+      const paymentsResponse = await msg.request<{
+        payments: Payment[];
+      }>("getPayments");
+
+      const transactions: Transaction[] = paymentsResponse.payments.map(
+        (payment) => ({
+          ...payment,
+          id: `${payment.id}`,
+          type: "sent",
+          date: dayjs(payment.createdAt).fromNow(),
+          title: payment.name || payment.description,
+          publisherLink: payment.location,
+        })
+      );
+
+      setPayments(transactions);
+    } catch (e) {
+      console.error(e);
+      if (e instanceof Error) toast.error(`Error: ${e.message}`);
+    }
+  }
+
+  useEffect(() => {
+    fetchData();
+  }, []);
 
   return (
     <Container>
       <h2 className="mt-12 mb-2 text-2xl font-bold dark:text-white">
-        {t("transactions.title")}
+        {t("title")}
       </h2>
 
       <p className="mb-6 text-gray-500 dark:text-neutral-500">
-        {t("transactions.description")}
+        {t("description")}
       </p>
 
-      <div className="mb-12">
-        <div>
-          {transactions?.length > 0 && (
-            <TransactionsTable transactions={transactions} />
-          )}
-          {!transactions?.length && (
-            <p className="mb-6 text-gray-500 dark:text-neutral-500">
-              {t("transactions.list_empty")}
-            </p>
-          )}
-        </div>
+      <div>
+        {payments?.length > 0 && <TransactionsTable transactions={payments} />}
       </div>
     </Container>
   );
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -695,6 +695,11 @@
       "allow_sign": "Allow {{host}} to sign:",
       "block_and_ignore": "Block and ignore {{host}}",
       "block_added": "Added {{host}} to the blocklist, please reload the website."
+    },
+    "transactions": {
+      "title": "All transactions",
+      "description": "This page contains all transactions",
+      "list_empty": "No transactions available yet"
     }
   },
   "common": {
@@ -734,6 +739,7 @@
       "send": "Send",
       "save": "Save",
       "receive": "Receive",
+      "transactions": "Transactions",
       "close": "Close",
       "export": "Export",
       "remove": "Remove",
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -698,8 +698,8 @@
     },
     "transactions": {
       "title": "All transactions",
-      "description": "This page contains all transactions",
-      "list_empty": "No transactions available yet"
+      "description": "This page lists all transactions to websites, independent of the selected account.",
+      "list_empty": "No transactions available yet."
     }
   },
   "common": {
--- a/src/app/router/Options/Options.tsx
+++ b/src/app/router/Options/Options.tsx
@@ -13,6 +13,7 @@ import Publishers from "@screens/Publishers";
 import Receive from "@screens/Receive";
 import Send from "@screens/Send";
 import Settings from "@screens/Settings";
+import Transactions from "@screens/Transactions";
 import Unlock from "@screens/Unlock";
 import { useTranslation } from "react-i18next";
 import { HashRouter, Navigate, Outlet, Route, Routes } from "react-router-dom";
@@ -53,6 +54,7 @@ function Options() {
             <Route path="confirmPayment" element={<ConfirmPayment />} />
             <Route path="keysend" element={<Keysend />} />
             <Route path="receive" element={<Receive />} />
+            <Route path="transactions" element={<Transactions />} />
             <Route path="lnurlPay" element={<LNURLPay />} />
             <Route path="lnurlChannel" element={<LNURLChannel />} />
             <Route path="lnurlWithdraw" element={<LNURLWithdraw />} />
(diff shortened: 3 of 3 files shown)


lujakob:
This is a draft for now. I would like to add the performance improvements of https://github.com/getAlby/lightning-browser-extension/pull/2034 before merging this.

lujakob:
@bumi should the menu item "Transactions" also be included in the Popup dropdown menu? Between "Receive" and "Settings"?

 
Bildschirmfoto 2023-01-31 um 14 16 14

lujakob:
@bumi I added the Popup menu "Transactions" item, because it makes the pages consistent. A new tag with the options page is opened, same as with click on "Websites"

Bildschirmfoto 2023-02-01 um 15 13 28

lujakob:
> This is a draft for now. I would like to add the performance improvements of #2034 before merging this.

Seems like this is redundant and will not be merged. So the PR would be ready for review and suggestions.

escapedcat:
Did you already merge latest master into this? Some changes look unrelated in the diff.

lujakob:
> Did you already merge latest master into this? Some changes look unrelated in the diff.

Something went wrong with the update. I reverted. Thanks.

src/app/screens/Transactions/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { settings, getFormattedFiat } = useSettings();
+  const [payments, setPayments] = useState<Transaction[]>([]);
+
+  const fetchData = useCallback(async () => {
+    try {
+      const paymentsResponse = await msg.request<{
+        payments: Payment[];
+      }>("getPayments");
+
+      const transactions: Transaction[] = paymentsResponse.payments.map(
```
escapedcat, line 25: Wondering if it makes sense to abstract this part. We use the same/similar code in different places now.

escapedcat (commented):
ack

src/app/screens/Transactions/index.tsx
lujakob: Yes, also the mapping afterwards. I'll see

lujakob: Done

src/app/utils/payments.ts
```diff
@@ -0,0 +1,16 @@
+import dayjs from "dayjs";
```
escapedcat, line 1: Is there a way or does it make sense to use this for these as well?:
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/DefaultView/index.tsx/#L27-L34
- https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/AllowanceView/index.tsx/#L48-L57

Or maybe in a follow-up PR

lujakob: @rolznz I thought about this, but not enough, because of the different `publisherLink` - now that I think about it more, it could be handled easily as well, and would make sense imo.

I would change it to 

```
export const convertPaymentToTransaction = (
  payment: DbPayment,
  publisherLink?: string
): Transaction => ({
  ...payment,
  id: `${payment.id}`,
  type: "sent",
  date: dayjs(payment.createdAt).fromNow(),
  title: payment.name || payment.description,
  publisherLink: publisherLink || payment.location,
});
```

and apply it with

`const payments = convertPaymentsToTransactions(response.payments, "options.html#/publishers");`

I started refactoring and found this would lead into a further cleanup of the `fetchData` in DefaultView and AllowanceView. For this reason I think it would make sense to refactor in a follow up PR, as you suggest. Should I create a ticket and assign it to me, as I already know what to do and implement it after this PR is merged?

escapedcat: >  Should I create a ticket and assign it to me, as I already know what to do and implement it after this PR is merged?

❤️ please do

lujakob: I added https://github.com/getAlby/lightning-browser-extension/issues/2057

src/i18n/locales/en/translation.json
```diff
@@ -727,6 +727,11 @@
       "allow_sign": "Allow {{host}} to sign:",
       "block_and_ignore": "Block and ignore {{host}}",
       "block_added": "Added {{host}} to the blocklist, please reload the website."
+    },
+    "transactions": {
+      "title": "All transactions",
+      "description": "This page lists all transactions to websites, independent of the selected account.",
```
bitcoinuser, line 733: Only to websites? I ask because I can do transactions using Alby not only to websites, for example paying a lightning invoice.

lujakob: This requirement comes from @bumi he can explain.

lujakob:
Sorry, some reviewers got removed accidentally yesterday when I clicked on "re-request review" on one of them. I couldn't re-add them.

src/app/screens/Transactions/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const fetchData = useCallback(async () => {
+    try {
+      const { payments } = await msg.request<{ payments: Payment[] }>(
+        "getPayments"
+      );
+      const transactions: Transaction[] =
+        convertPaymentsToTransactions(payments);
+
+      for (const transaction of transactions) {
```
rolznz, line 27: this looks duplicated in the publishers screen. What about if convertPaymentsToTransactions took a boolean to calculate the fiat value?

src/app/screens/Publishers/Show/index.tsx
```diff
@@ -40,20 +41,14 @@ function Publisher() {
...
 
-        const payments: Transaction[] = response.payments.map((payment) => ({
-          ...payment,
-          id: `${payment.id}`,
-          type: "sent",
-          date: dayjs(payment.createdAt).fromNow(),
-          title: payment.name || payment.description,
-          publisherLink: payment.location,
-        }));
+        const payments: Transaction[] = convertPaymentsToTransactions(
```
rolznz, line 44: Should this be called transactions now?

src/app/screens/Transactions/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        transaction.totalAmountFiat = settings.showFiat
+          ? await getFormattedFiat(transaction.totalAmount)
+          : "";
+      }
+
+      setPayments(transactions);
```
rolznz, line 33: should this be setTransactions?

src/app/screens/Transactions/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+    } catch (e) {
+      console.error(e);
+      if (e instanceof Error) toast.error(`Error: ${e.message}`);
+    }
+  }, [settings, getFormattedFiat]);
+
+  useEffect(() => {
```
rolznz, line 40: This might be a good candidate for using SWR too, if we can get that other PR merged. Maybe could you add a TODO comment if you agree?

lujakob: I thought about it. But I wasn't too sure how and when to reset/mutate the cache. I added a Todo comment.

src/app/screens/Transactions/index.tsx
rolznz: In my opinion I think it is fine to pass getFormattedFiat to convertPaymentsToTransactions. Maybe convertPaymentsToTransactions could take an options object, with getFormattedFiat being an optional function as part of the options object, and will only be called if it exists. (Rather than a boolean as I originally suggested).

But that's just my thoughts, otherwise maybe it is fine as it is. I just noticed the duplication.

lujakob: @bitcoinuser @rolznz @im-adithya @escapedcat @reneaaron Is there anything I can do to move this PR forward? From my consideration it is done. The optimization @rolznz mentioned in the last comment here (about getFormattedFiat) is the only thing left. I can implement it, I just thought I might wait for some more opinions about it. In case you approve, I'd like to finish it that it doesn't get too stale.

src/i18n/locales/en/translation.json
escapedcat: The way it's implemented is showing "all" transactions though, right? Not only "website" ones. In this case we can just adjust the sentence here and leave out "to websites", right?

escapedcat (approved):

src/i18n/locales/en/translation.json
lujakob: I think that's correct, all payments. I get the data from https://github.com/lujakob/lightning-browser-extension/blob/master/src/extension/background-script/actions/payments/all.ts#L5-L5

I will remove this part from the translation.

lujakob: I think that's correct, all payments. I get the data from https://github.com/lujakob/lightning-browser-extension/blob/master/src/extension/background-script/actions/payments/all.ts#L5-L5

I will remove this part from the translation.

rolznz (approved):

reneaaron:
Thanks for this PR @lujakob and sorry for the long wait. I finally got my hands on it: 

To me it's very hard to understand which parts are scoped to an account currently. (Send, Receive are scoped, while Websites & Transactions are not) I tend to lean towards getting everything scoped to the account as I think that is how most of applications work and users know this pattern. (having the account as the root of ~everything, except maybe for the global settings) 

@bumi What is the use case of seeing all of your transactions on one page? Is it for statistical purposes? Should we rethink this as some kind of a "statistics" page? I've prepared a quick mockup here:

image

On this page we could have a switch that would allow this stats page to include data from multiple / all accounts and potentially even filter for time (last 30 days, etc). 

With this approach we would: 
1) Provide some interesting data to users of the extension (at least some users even like to share their numbers online and tell the world about how much they boosted / zapped / etc)
2) Avoid confusion in future as we introduce the wallet view from the popup (send, receive, transaction list) to the full-page view (this is currently missing and confuses a lot of users as they currently don't find their "wallet" view in the full-page view) and then would have two separate pages to view transactions on the account / global view (which is again confusing IMO)

What do you guys think? 

If we don't want to re-think this that is also fine for me, maybe we should look at those things though: 

## Navigation 
The new menu item breaks the layout on some screen sizes: 
image

Should we maybe just remove this menu item from the main navigation and just have it in the dropdown? 

## Naming
image

I think the page title should be the same as on the Navigation "Transactions" instead of "All transactions". Also this page does only show outgoing transactions.

lujakob:
> ## Navigation
> The new menu item breaks the layout on some screen sizes: 

I fixed this by adapting the responsive breakpoints/styles with Tailwind syntax. For small screens, the spacing is smaller now, for lg it is as before. I think it's good now. Unless you decide "Transactions" is actually not needed anyways.

> Should we maybe just remove this menu item from the main navigation and just have it in the dropdown?
> 
> ## Naming
> 
> I think the page title should be the same as on the Navigation "Transactions" instead of "All transactions". Also this page does only show outgoing transactions.

I fixed it by renaming the headline to "Transactions"

lujakob:
> Thanks for this PR @lujakob and sorry for the long wait. I finally got my hands on it:
> 
> To me it's very hard to understand which parts are scoped to an account currently. (Send, Receive are scoped, while Websites & Transactions are not) I tend to lean towards getting everything scoped to the account as I think that is how most of applications work and users know this pattern. (having the account as the root of ~everything, except maybe for the global settings)
> 
> @bumi What is the use case of seeing all of your transactions on one page? Is it for statistical purposes? Should we rethink this as some kind of a "statistics" page? I've prepared a quick mockup here:
> 
> 
> On this page we could have a switch that would allow this stats page to include data from multiple / all accounts and potentially even filter for time (last 30 days, etc).
> 
> With this approach we would:
> 
> 1. Provide some interesting data to users of the extension (at least some users even like to share their numbers online and tell the world about how much they boosted / zapped / etc)
> 2. Avoid confusion in future as we introduce the wallet view from the popup (send, receive, transaction list) to the full-page view (this is currently missing and confuses a lot of users as they currently don't find their "wallet" view in the full-page view) and then would have two separate pages to view transactions on the account / global view (which is again confusing IMO)
> 
> What do you guys think?
> 
> If we don't want to re-think this that is also fine for me, maybe we should look at those things though:
> 

Regarding scoping transactions to accounts: I think it's a fair point. Even though I see this point (keeping the consistency of scoping all data to accounts), I also see the use case of a user wanting to see all spending across accounts. However, because keeping consistency regarding scope of accounts between transactions page and websites page is important, this would stretch the scope of this PR quite a bit I guess. How about rolling out this PR as a first step, discuss scoping of data for websites/transactions and if approved implement it as a follow up PR (I would volunteer ;-) )

I think it might be a good idea to enable both (account scoped/unscoped) as well as a filter for time (last 30 days, etc). This could be a follow up PR "Implement optional account scoping and time filter for websites and transactions". What do you think?

reneaaron:
Thanks for your feedback and updating the PR. I also discussed this with @stackingsaunter today.

> Regarding scoping transactions to accounts: I think it's a fair point. Even though I see this point (keeping the consistency of scoping all data to accounts), I also see the use case of a user wanting to see all spending across accounts. However, because keeping consistency regarding scope of accounts between transactions page and websites page is important, this would stretch the scope of this PR quite a bit I guess. 

Currently we don't even record the account the payment was made with, that will be changed very soon with #2140. This will allow us to scope this page to the account soon. 

I think we could work on getting the "wallet" page into the full screen view next:

image

That would allow us to streamline our navigation (send, receive, transactions could be removed) and have all of those things available from this wallet page which I think is way more understandable for users as they already know this patter from other wallets. This "all transactions" page could then also be accessed from the wallet page by clicking on "see all". (instead of having a main navigation item) 

I would want to prevent us from having switches to toggle between "all accounts" / "current account" all over the place and I don't really understand the use case yet. (other than for statistical purposes on a separate stats page maybe) 

> How about rolling out this PR as a first step, discuss scoping of data for websites/transactions and if approved implement it as a follow up PR (I would volunteer ;-) )

That is super nice, great to have you here @lujakob! 💯 

@stackingsaunter I think we could see it as a first step into this direction, especially since we are now tackling #2140 and all users who only have 1 account won't have these UX problems. (which should probably be most of our extension users).

stackingsaunter:
@reneaaron I guess it won't hurt, the only thing I am hesitant about is if we implement something and then completely change it 2 or 3 weeks later and that it confuse users.

Just let me reiterate the the timeline of those issues as I understand now:

1. We implement Transactions page in extended extension view that shows all transactions across accounts
2. When #2140 is implemented we implement Outgoing transactions per account
3. Then we merge Send, Receive and Transactions into "Wallet" page, but allow to click "See All" in Transactions

In that scenario, Wallet shows transactions per account, but what "See All" shows? 
All transactions across accounts or also per account?

reneaaron:
> 1. We implement Transactions page in extended extension view that shows all transactions across accounts
> 2. When https://github.com/getAlby/lightning-browser-extension/pull/2140 is implemented we implement Outgoing transactions per account
> 3. Then we merge Send, Receive and Transactions into "Wallet" page, but allow to click "See All" in Transactions

That'd be the way we would do it probably, yeah. 👍 Are you okay with that? 

> In that scenario, Wallet shows transactions per account, but what "See All" shows?
All transactions across accounts or also per account?

Also per account. I don't think we should have any "global" transactions anymore. (it also doesn't fit together with the incoming transactions, they already are displayed per account). Next steps then could be to merge incoming / outgoing transactions.

stackingsaunter:
@reneaaron okay that sounds like a reasonable path from what we have now!

reneaaron (approved):
Changed some minor things in the subheading of the page, LGTM! 🚀

tACK

I'll create some issues for the follow-up tasks we identified in this PR.

PR #2036, opened 2023-02-01
Chore/msg request typings

rolznz:

Add typescript support to the bridge between the content and background script. Types no longer need to be manually passed to msg.request (or the response type casted to something else). Errors will be shown if wrong parameters or a non-existent action is used. 

### Type of change

- `fix`: Improve developer experience, and exposed quite a few issues with incorrect properties passed. Some of these now have an eslint warning on them.

image
image
image

### How has this been tested?

### Checklist


rolznz:
Note - some issues like the below have FIXMEs for now:
- Account.config is a string, but when creating an account it is not a string
- domain does not exist on blocklist, and imageUrl and name are expected but never passed

src/app/screens/Enable/index.tsx
```diff
@@ -47,7 +47,9 @@ function Enable(props: Props) {
     await msg.request("addBlocklist", {
       domain: props.origin.domain,
       host: props.origin.host,
-    });
+      // FIXME: typings
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any);
```
escapedcat, line 52: Instead of comment and `as any` you could use `as FixMe`. Did this in a another project and it worked quite well. We already use this in some places in this codebase as well.

escapedcat (commented):
Will check further once the base PR is merged

src/extension/background-script/router.ts
```diff
@@ -11,6 +11,46 @@ import * as settings from "./actions/settings";
...
+};
+
+export type ExtendedRouteArgs = RouteArgs & NostrRouteArgs;
+
+type RouteResponseReturnDataType<T> = T extends { data: infer U }
+  ? U
+  : undefined;
+
+export type RouteResponse = {
+  //[P in RouteKey]: Awaited<ReturnType<(typeof routes)[P]>>["data"];
```
lujakob, line 39: is this comment a leftover?

src/extension/background-script/router.ts
```diff
@@ -11,6 +11,46 @@ import * as settings from "./actions/settings";
...
+  [P in RouteKey]: RouteResponseReturnDataType<
+    Awaited<ReturnType<(typeof routes)[P]>>
+  >;
+};
+
+type NostrRouteResponse = {
+  //[P in NostrKeyNoPrefix as `nostr/${P}`]: Awaited<ReturnType<(typeof routes["nostr"])[P]>>["data"];
```
lujakob, line 46: is this comment a leftover?

rolznz: Yes, I have removed it now. Thanks!

src/extension/background-script/router.ts
rolznz: Yes, I have removed it now. Thanks!

src/app/screens/Enable/index.tsx
rolznz: Thanks, I have updated the 4 occurrences.

bumi:
oh. @rolznz can you merge master and fix the conflicts?

rolznz:
@bumi fixed. I also disabled adding the alias to lnd/cln on connect so there are no functional changes in this PR. I can remove those changes completely if that is better.

src/common/lib/api.ts
```diff
@@ -38,21 +30,21 @@ export interface GetAccountRes
   extends Pick<Account, "id" | "connector" | "name"> {
   nostrEnabled: boolean;
 }
-interface StatusRes {
-  configured: boolean;
-  unlocked: boolean;
-  currentAccountId: string;
-}
+// interface StatusRes {
```
escapedcat, line 33: Should this be deleted?

PR #2039, opened 2023-02-01
chore: cleanup old migrations

bumi:
Old migrations already ran and don't need to run again. We can delete them.


reneaaron:
What if someone didn't upgrade to the recent versions (for whatever reason, e.g. a notebook that hasn't been used for a while). Wouldn't they run into problems as they upgrade to the latest version?

bumi:
* migrateisUsingLegacyLnurlAuthKeySetting is 6 months old. and if it is missed then a different LNUR-auth config is set
* migratedeleteLegacyWeblnPermissions is from the last release I think. but if it is missed it would just have some old data in the permissions table.

PR #2041, opened 2023-02-02
fix: remove obsolete file

reneaaron:
Found no reference of this file so I deleted it.


escapedcat (approved):
Looks like it's from pre tailwind maybe

PR #2043, opened 2023-02-02
chore: add remove private key action

im-adithya:

- Adds remove private key action
- Renames Show to Detail in Accounts and Publishers Screen
- Trims the pasted nostr private keys to avoid accidental whitespace while pasting keys

Follow up to #1964

### Type of change

- `chore`

### How has this been tested?

Manually.

### Checklist


src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -148,20 +148,23 @@ function AccountScreen() {
...
-      nostrPrivateKey && nostrlib.hexToNip19(nostrPrivateKey, "nsec");
-
-      await msg.request("nostr/setPrivateKey", {
-        id: account.id,
-        privateKey: nostrPrivateKey,
-      });
-
       if (nostrPrivateKey) {
+        // Validate the private key before saving
+        nostrPrivateKey && generatePublicKey(nostrPrivateKey);
```
reneaaron, line 153: Why `nostrPrivateKey &&`?

src/extension/background-script/actions/nostr/removePrivateKey.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  if (id && Object.keys(accounts).includes(id)) {
+    const account = accounts[id];
+    if (account.nostrPrivateKey) delete account.nostrPrivateKey;
+    accounts[id] = account;
+    state.setState({ accounts });
+    await state.getState().saveToStorage();
+    return {
+      data: {
+        accountId: id,
```
reneaaron, line 18: Why do we return the `accountId` as a result here? That isn't used anywhere, is it?

reneaaron (changes requested):

src/app/screens/Accounts/Detail/index.tsx
escapedcat: I guess because `generatePublicKey` is expecting a value

src/extension/background-script/actions/nostr/removePrivateKey.ts
escapedcat: Maybe it can be useful for some to get the related id because it could be either the provided id or the current one? Just a thought.

escapedcat: But yeah, currently there's not need for it.

src/app/screens/Accounts/Detail/index.tsx
im-adithya: Oops, I forgot to remove that

> I guess because generatePublicKey is expecting a value

I used to do this without the if statement, so I copy-pasted it :P

src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -148,20 +148,23 @@ function AccountScreen() {
...
-      nostrPrivateKey && nostrlib.hexToNip19(nostrPrivateKey, "nsec");
-
-      await msg.request("nostr/setPrivateKey", {
-        id: account.id,
-        privateKey: nostrPrivateKey,
-      });
-
       if (nostrPrivateKey) {
+        // Validate the private key before saving
+        generatePublicKey(nostrPrivateKey);
```
reneaaron, line 153: Why do you need to generate the the public key here?

reneaaron (changes requested):

src/app/screens/Accounts/Detail/index.tsx
im-adithya: We use the useEffect hook to update the public key and private key (in npub and nsec from the hex private key string), but here we have to show the errors as this is the try block, so we do it again here, just for validation.

reneaaron (approved):
tACK

PR #2053, opened 2023-02-03
perf: incoming tab take some seconds to load data

### LINKED ISSUE #1854: [Feature] When the user click on Incoming tab take some seconds to load data. Should show instantly.

qqqzhch:

Load incoming txs as soon as popup is being opened 

if account is being switched Load incoming

- Fixes #1854

### Type of change

### Checklist

### DIFF
--- a/src/app/screens/Home/DefaultView/index.tsx
+++ b/src/app/screens/Home/DefaultView/index.tsx
@@ -156,7 +156,7 @@ const DefaultView: FC<Props> = (props) => {
 
     setIncomingTransactions(invoices);
     setIsLoadingInvoices(false);
-  }, [getFormattedFiat, settings]);
+  }, [getFormattedFiat, settings.showFiat]);
 
   useEffect(() => {
     const load = async () => {
--- a/src/app/screens/Home/DefaultView/index.tsx
+++ b/src/app/screens/Home/DefaultView/index.tsx
@@ -8,10 +8,11 @@ import TransactionsTable from "@components/TransactionsTable";
 import { Tab } from "@headlessui/react";
 import dayjs from "dayjs";
 import relativeTime from "dayjs/plugin/relativeTime";
-import { FC, useEffect, useState } from "react";
+import { FC, useCallback, useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import { toast } from "react-toastify";
+import { useAccount } from "~/app/context/AccountContext";
 import { useSettings } from "~/app/context/SettingsContext";
 import { PublisherLnData } from "~/app/screens/Home/PublisherLnData";
 import { classNames } from "~/app/utils/index";
@@ -66,6 +67,7 @@ const DefaultView: FC<Props> = (props) => {
     !isLoadingInvoices && incomingTransactions?.length === 0;
 
   const navigate = useNavigate();
+  const { account } = useAccount();
 
   const { t } = useTranslation("translation", { keyPrefix: "home" });
   const { t: tCommon } = useTranslation("common");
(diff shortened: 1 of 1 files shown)


escapedcat:
This works, data is there immediately. But it might not be the "latest" data, right?
I feel like step 2 is missing in a way: https://github.com/getAlby/lightning-browser-extension/issues/1854#issuecomment-1345523880

qqqzhch:
> This works, data is there immediately. But it might not be the "latest" data, right? I feel like step 2 is missing in a way: [#1854 (comment)](https://github.com/getAlby/lightning-browser-extension/issues/1854#issuecomment-1345523880)

If the pop page remains open for a long time, the data is not updated.
If the balance changes and the data is loaded again it should be much better, I'll try it

qqqzhch:
The relevant test code for this component is still missing and I will add the relevant test code after this pr merge.
Can the e2e tests be run only if the local Lightning Network node is linked, or the data be mocked and run without linking the node?
metamask has also made a test site for testing dapp related functions, I think we should also make a similar site
https://metamask.github.io/test-dapp/

escapedcat:
You can have a look at [our e2e tests](https://github.com/getAlby/lightning-browser-extension/tree/master/tests/e2e), we use testnet there.  
I.e. here: https://github.com/getAlby/lightning-browser-extension/blob/master/tests/e2e/002-walletFeatures.spec.ts/#L88-L109

escapedcat:
> If the balance changes and the data is loaded again it should be much better

Ah, interesting. Did you test this? Does it work?

src/app/screens/Home/DefaultView/index.tsx
```diff
@@ -161,7 +156,14 @@ const DefaultView: FC<Props> = (props) => {
 
     setIncomingTransactions(invoices);
     setIsLoadingInvoices(false);
-  };
+  }, [getFormattedFiat, settings]);
```
im-adithya, line 159: `settings.showFiat` would suffice right?

im-adithya (approved):
Nice improvement @qqqzhch! Works as expected, even on balance change!
(Left a small change)

qqqzhch:
@escapedcat @im-adithya 
I have not done enough tests on pages that are opened and stay particularly long, not yet
- 1 Now only repeated transactions between multiple accounts in one wallet, did the test
- 2 I think I still have to install two wallets in two browsers, and keep the wallets inside both browsers open for a long time, to do tests of transactions

src/app/screens/Home/DefaultView/index.tsx
qqqzhch: Good suggestion settings.showFiat is better

im-adithya:
I went with the assumption that if we're not updating the balance real-time, there's no point in showing the recent transactions (which didn't even reflect in the balance, as that might lead to confusion).

So I think as long as we have balance (`balancesDecorated?.accountBalance`) as a dependency, we're fine!

@escapedcat thoughts?

escapedcat:
Tested with sending some sats from an account to another account. The balance is only updated on "reopening" the extension anyways. So unless we do not have any "polling" I guess adding the effect-stuff is useless? In the case of "an invoice is being payed" this works though. So I guess it's still good to have it.

qqqzhch:
Yes, come to think of it, the balance does not change instantly issue, there are too many types of wallet connections, for different wallet connections my knowledge is relatively low, to read in real time, need all connect  support wss, or use setInterval

escapedcat (approved):

escapedcat:
> Yes, come to think of it, the balance does not change instantly issue, ...

I think we're good here for now.

PR #2072, opened 2023-02-05
fix: spacing on lg

reneaaron:

Removes duplicate spacing on lg screens. (container already has `px-4`)


escapedcat (approved):
utack

PR #2073, opened 2023-02-05
Fix/add host to lndhub errors

reneaaron:

Currently the host isn't included in the error message which makes it hard for users to know where things are going wrong.

### DIFF
--- a/src/extension/background-script/connectors/lndhub.ts
+++ b/src/extension/background-script/connectors/lndhub.ts
@@ -1,5 +1,5 @@
-import axios, { AxiosRequestConfig, Method } from "axios";
-import type { AxiosResponse } from "axios";
+import type { AxiosResponse, Method } from "axios";
+import axios, { AxiosRequestConfig } from "axios";
 import lightningPayReq from "bolt11";
 import Base64 from "crypto-js/enc-base64";
 import Hex from "crypto-js/enc-hex";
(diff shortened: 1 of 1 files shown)


escapedcat (approved):
utack

src/extension/background-script/connectors/lndhub.ts
```diff
@@ -330,34 +330,41 @@ export default class LndHub implements Connector {
...
       this.access_token = authData.access_token;
       this.refresh_token_created = +new Date();
       this.access_token_created = +new Date();
 
       return authData;
+    } catch (e) {
+      throw new Error(
+        `API error (${this.config.url})${
+          e instanceof Error ? `: ${e.message}` : ""
+        }`
```
im-adithya, line 366: Why are we propagating this till here instead of adding this (host) in the above line itself?

src/extension/background-script/connectors/lndhub.ts
```diff
@@ -330,34 +330,41 @@ export default class LndHub implements Connector {
...
+      );
 
-    if (authData.error || authData.errors) {
-      const error = authData.error || authData.errors;
-      const errMessage = error?.errors?.[0]?.message || error?.[0]?.message;
+      if (authData.error || authData.errors) {
+        const error = authData.error || authData.errors;
+        const errMessage = error?.errors?.[0]?.message || error?.[0]?.message;
+
+        throw new Error(errMessage);
```
im-adithya, line 353: Like this: (see below comment)
```
throw new Error(`API error (${this.config.url}) ${errMessage}`)

src/extension/background-script/connectors/lndhub.ts
reneaaron: There are 2 types of errors: 
 - 5xx errors (errors on the api, not authorized, etc) throw an exception
 - Error fields in the response (that indicate the request failed) 
 
In both cases it would make sense to include the hostname in the error message.

im-adithya: Ah, got it!

im-adithya (approved):

PR #2083, opened 2023-02-08
feat: add schnorr to nostr

im-adithya:

Adds signSchnorr to nostr

Fixes _#ISSUE-NUMBER_

### Type of change

Signed using Nostr key:

Screenshot 2023-02-08 at 1 39 41 PM

### How has this been tested?

Tested on ionio: https://github.com/im-adithya/ionio/tree/alby-liquid
(change window.liquid to window.nostr in the Alby app (https://github.com/im-adithya/ionio/tree/alby-liquid/examples/alby-app) before testing)

### Checklist

### DIFF
--- a/src/app/router/Prompt/Prompt.tsx
+++ b/src/app/router/Prompt/Prompt.tsx
@@ -12,8 +12,9 @@ import MakeInvoice from "@screens/MakeInvoice";
 import NostrConfirm from "@screens/Nostr/Confirm";
 import NostrConfirmGetPublicKey from "@screens/Nostr/ConfirmGetPublicKey";
 import NostrConfirmSignMessage from "@screens/Nostr/ConfirmSignMessage";
+import NostrConfirmSignSchnorr from "@screens/Nostr/ConfirmSignSchnorr";
 import Unlock from "@screens/Unlock";
-import { HashRouter, Outlet, Route, Routes, Navigate } from "react-router-dom";
+import { HashRouter, Navigate, Outlet, Route, Routes } from "react-router-dom";
 import { ToastContainer } from "react-toastify";
 import Providers from "~/app/context/Providers";
 import RequireAuth from "~/app/router/RequireAuth";
@@ -81,6 +82,10 @@ function Prompt() {
             <Route
               path="public/nostr/confirmSignMessage"
               element={<NostrConfirmSignMessage />}
+            />
+            <Route
+              path="public/nostr/confirmSignSchnorr"
+              element={<NostrConfirmSignSchnorr />}
             />
 
             <Route path="lnurlAuth" element={<LNURLAuth />} />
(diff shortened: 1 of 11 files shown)


dannydeezy:
@im-adithya i tested this locally and was able to successfully send a transaction with https://nosft.xyz 
https://github.com/dannydeezy/nosft/pull/3

dannydeezy:
@im-adithya one comment is that `signSchnorr` accepts and receives via hex string, whereas the popular `ecpair` javascript library has a `signSchnorr` function that accepts and receives via Buffer.

I am happy with either one, but something to consider, i'm not sure which would be better

dannydeezy:
if you want to test sending w/ nosft, here is a deploy branch that should work with this version of alby
https://deploy-preview-5--fabulous-meringue-01474b.netlify.app/

kiwiidb (changes requested):
I was able to reproduce and verify the signatures using a short Go program:

```
package main

import (
	"encoding/hex"
	"fmt"
	"log"

	"github.com/btcsuite/btcd/btcec/v2"
	"github.com/btcsuite/btcd/btcec/v2/schnorr"
)

func main() {

	//pubkey hex should be your nostr pubkey with either a 03 or a 02 prepended (not sure how to know which one)
	pubkeyHex := "03_OR_02+YOUR_NOSTR_PUBKEY_IN_HEX"
	//msgToSign should be 32 bytes (so 64 hex chars), as it should be the sha256 hash of the actual message
	msgToSign := "YOUR_HEX_MSG_OF_64_CHARS"
	//this should be the result of await nostr.signSchnorr(msgToSign) from the web console
	sigFromExtensionHex := "128_CHAR_HEX_SIG_RESULT"

	msgBytes, err := hex.DecodeString(msgToSign)
	if err != nil {
		log.Fatal(err)
	}
	pkBytes, err := hex.DecodeString(pubkeyHex)
	if err != nil {
		log.Fatal(err)
	}
	pk, err := btcec.ParsePubKey(pkBytes)
	if err != nil {
		log.Fatal(err)
	}
	sigFromExtensionBytes, _ := hex.DecodeString(sigFromExtensionHex)
	sig, _ := schnorr.ParseSignature(sigFromExtensionBytes)
	fmt.Println(sig.Verify(msgBytes, pk))
}

```

The only thing that I would remark is that we allow arbitrary-length (even-sized) hex strings to be signed, while BIP340 specifies that the message _must_ be 32 bytes, i.e. the SHA256 hash of the actual message. If we sign a nostr event, Alby takes the [sha256](https://github.com/getAlby/lightning-browser-extension/blob/task-schnorr-nostr/src/extension/background-script/actions/nostr/helpers.ts#L94) of it internally. So, should we do that as well here, or should we specify and even check that the input is a 32-byte hex encoded hash? Or should we allow arbitrary (even-sized, hex encoded) messages to be signed? It seems that @dannydeezy  is doing the hashing [here](https://github.com/dannydeezy/nosft/pull/3/files#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67R79), so it seems to be better to leave the hashing to the website developers. I think we should add a check that the input is 32 bytes and hex-encoded.

kiwiidb (approved):

dannydeezy:
looks good to me!

PR #2094, opened 2023-02-09
feat: added receive another payment button

### LINKED ISSUE #2022: [Feature] Receive screen: add "Receive another payment" button
### Feature description

Problem: Currently there is no way to get back to the receive payment form after having received a payment. Would't it make sense to have a button to get back to the form? Currently only navigating to another screen and back (or a page reload for the options page) makes this possible.

Bildschirmfoto 2023-01-28 um 07 09 31

### Describe the solution

Add orange button "Receive another payment" underneath the QR Code "Payment received" that resets the "paid" state and show the form again.

The solution is simple and I could implement it.

### Describe alternatives

Having a less prominent text link with label "Receive another payment" or "Go back" underneath the QR code?

### Additional context

_No response_

### Are you working on this?

None

manavdesai27:

I have added a Receive another payment button under the Payment Received confetti on successful payment, which on click takes the user back to Receive screen.

Fixes #2022 

### Type of change

https://user-images.githubusercontent.com/74018438/217906810-8f4ca75e-5ecf-4f32-b5de-3af106b523ad.mp4

### How has this been tested?

### Checklist

### DIFF
--- a/src/app/screens/Receive/index.tsx
+++ b/src/app/screens/Receive/index.tsx
@@ -100,6 +100,13 @@ function Receive() {
       });
   }
 
+  function receiveAnotherPayment() {
+    setPaid(false);
+    setPollingForPayment(false);
+    setInvoice(undefined);
+    navigate("/receive");
+  }
+
   async function createInvoice() {
     try {
       setLoading(true);
@@ -140,6 +147,17 @@ function Receive() {
             </div>
           )}
         </div>
+        {paid && (
+          <div className="my-4">
+            <Button
+              type="submit"
+              label={tCommon("actions.receive_again")}
+              primary
+              fullWidth
+              onClick={receiveAnotherPayment}
+            />
+          </div>
+        )}
         {!paid && (
           <>
             <div className="mt-8 mb-4 flex justify-center">
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -769,6 +769,7 @@
       "send": "Send",
       "save": "Save",
       "receive": "Receive",
+      "receive_again": "Receive another payment",
       "close": "Close",
       "export": "Export",
       "remove": "Remove",


manavdesai27:
@bumi @im-adithya Could I get a review on this? Thanks!

bumi:
@manavdesai27 great, thanks for the PR. we'll try to get to it asap, but we got a bit of a pull request backlog right now :)

src/app/screens/Receive/index.tsx
```diff
@@ -100,6 +100,13 @@ function Receive() {
       });
   }
 
+  function receiveAnotherPayment() {
+    setPaid(false);
```
im-adithya, line 104: formData should be defaulted; otherwise, we'll end up with the same amount and description on clicking "Create Invoice" the second time.

src/app/screens/Receive/index.tsx
```diff
@@ -100,6 +100,13 @@ function Receive() {
...
+    setPollingForPayment(false);
+    setInvoice(undefined);
```
im-adithya, line 106: Using undefined is not a good practice. You can change the invoice type to `{paymentRequest: ...} | null` and use `null` here.

manavdesai27:
@im-adithya I have made the needed changes.

src/app/screens/Receive/index.tsx
```diff
@@ -44,7 +44,7 @@ function Receive() {
   const [invoice, setInvoice] = useState<{
     paymentRequest: string;
     rHash: string;
-  }>();
+  } | null>();
```
bumi, line 47: as this is a change in the type - can the code handle null in the `invoice` variable? or is somewhere some `invoice.paymentRequest` or something like that in the code?

manavdesai27: Let me see into it.

im-adithya: Nope, checked it. We always check if `invoice` is null or not before proceeding!

manavdesai27: There are 3 places where we use invoice.PaymentRequest
image
Would this create an issue?

bumi: I guess `confirm` can only be called if the invoice is set? - that's how I understand @im-adithya 👍

manavdesai27: Where are we checking if the invoice is set?

im-adithya: > I guess confirm can only be called if the invoice is set?

Ah! No, we enable "Confirm" after we have the formData (i.e. amount and description/memo) and then set the invoice using that, invoice is the result with the QR information here.

```
  function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
    event.preventDefault();
    createInvoice();
  }

Comment (im-adithya) [reply src/app/screens/Receive/index.tsx:47]: > Would this create an issue?

Nope, that is unrelated to this change.

Comment (manavdesai27) [reply src/app/screens/Receive/index.tsx:47]: Well then ig null won't be an issue here.

Comment (im-adithya) [reply src/app/screens/Receive/index.tsx:47]: > Where are we checking if the invoice is set?

Because the QR you see is coming from the invoice (response):
```
<QRCode value={invoice.paymentRequest.toUpperCase()} level="M" />
```
And to show that, you need to ensure it exists :)

manavdesai27: > > I guess confirm can only be called if the invoice is set?
> 
> Ah! No, we enable "Confirm" after we have the formData (i.e. amount and description/memo) and then set the invoice using that, invoice is the result with the QR information here.
> 
> ```
>   function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
>     event.preventDefault();
>     createInvoice();
>   }
> ```

So, we generate invoice using formData everytime. So it will always overwrite null?

im-adithya: You have to see a new QR each time, so yeah, it should!

manavdesai27: > You have to see a new QR each time, so yeah, it should!

Well then null won't be an issue!

bumi (approved):
@im-adithya can you do a last check and then merge.

PR #2133, opened 2023-02-20
fix(storage): use local when sync is unavailable

im-adithya:

Uses `browser.storage.local` when `browser.storage.sync` is unavailable.

Fixes #

### Type of change

### How has this been tested?

Tested in the latest Tor Browser

### Checklist

### DIFF
--- a/src/extension/background-script/state.ts
+++ b/src/extension/background-script/state.ts
@@ -121,14 +121,11 @@ const state = createState<State>((set, get) => ({
     const syncStorage = browser.storage.sync;
 
     return syncStorage
-      .set({ test: "alby" })
-      .then(() => {
-        syncStorage.remove("test");
-        return syncStorage.get(browserStorageKeys).then((result) => {
-          // Deep merge to ensure that nested defaults are also merged instead of overwritten.
-          const data = merge(browserStorageDefaults, result as BrowserStorage);
-          set(data);
-        });
+      .get(browserStorageKeys)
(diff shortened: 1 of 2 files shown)


escapedcat (approved):
utack

src/extension/background-script/state.ts
```diff
@@ -118,23 +118,51 @@ const state = createState<State>((set, get) => ({
...
+    const syncStorage = browser.storage.sync;
     const current = get();
     const data = {
       ...browserStorageDefaults,
       ...pick(current, browserStorageKeys),
     };
-    return browser.storage.sync.set(data);
+
+    return syncStorage
+      .set({ test: "alby" })
```
bumi, line 158: we should avoid too many calls to this. afaik those are rate limited.

im-adithya:
> can we do this check in the init() call and then cache it globally?

We can't because the calls to browserStorage would be different (because we don't want to overlap with accounts, see `mockSync`), instead I used a flag and cached it to know what to use so that we don't waste calls

im-adithya:
> What is "mockSync"? Just a random string to test if we can access it?

No, it's the place we allot for (the should've been) sync storage in the local storage because both the sync storage and local storage have the accounts key-value pair (and can overlap in future too), which might get overwritten if we don't do this

reneaaron (approved):
tACK 

✅ Tested in TOR browser & Firefox

PR #2138, opened 2023-02-23
fix: redirect to password screen

### LINKED ISSUE #2137: [BUG] welcome screens are not protected by password

### Describe the bug

We can still visit the choose-path, test-connection links without setting a password

### Screenshots [optional]

-

### Steps To Reproduce

Go to .../welcome.html#/choose-path without setting a password, it opens.

### Expected behavior

It should redirect to set password screen

### Alby information

latest

### Device information

mac os

### Additional context

_No response_

### Are you working on this?

Yes

im-adithya:

Adds a password exists action and uses it before showing the page and redirects to set password if there isn't

Fixes #2137

### Type of change

### How has this been tested?

Manually

### Checklist


src/extension/background-script/actions/setup/getPassword.ts
```diff
@@ -0,0 +1,10 @@
+import { MessageGetPassword } from "~/types";
+
+import state from "../../state";
+
+const getPassword = (message: MessageGetPassword) => {
+  const passwordExists = state.getState().password ? true : false;
```
bumi, line 6: I think we have the status action for this: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/setup/status.js

bumi (commented):
how do you manage to get to those pages? Did you open those manually or did you somehow get there while using the app?

src/app/router/Welcome/Welcome.tsx
```diff
@@ -31,16 +33,15 @@ function getRoutes(
     {
       path: "/choose-path",
       name: i18n.t("translation:welcome.nav.connect"),
+      element: passwordExists ? (
```
rolznz, line 36: I think this change stopped the Alby login / create and choose connector links no longer work

rolznz:
I feel like this is complicating the routing as now we have to support both password and non-password versions on the same route. Would it be simpler to have a proper redirect to the choose password route? that way each route only renders what it is supposed to render.

rolznz:
> how do you manage to get to those pages? Did you open those manually or did you somehow get there while using the app?

I just realized this is also possible if you only have one account and then delete it.

bumi:
> I just realized this is also possible if you only have one account and then delete it.

which is fine, isn't it? then the user goes through the onboarding again and also has to set a passwor with which the new account is encrypted then.

PR #2139, opened 2023-02-23
fix: add type in publishPaymentNotification

im-adithya:
## ⚠️ Requires Input

This change starts calling ln.keysend.success for keysend payments (as desired), but that would mean we are not calling the success/failure notifications here: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/events/index.js#L13-L14
Wanted to know if that's also done on purpose or I should add notifs for keysend payments

The `ln.keysend.success` event isn't being triggered at all because we only publish the `ln.sendPayment.success` event everytime.

Fixes _#ISSUE-NUMBER_

### Type of change

### How has this been tested?

Didn't test yet, require input.

### Checklist

### DIFF
--- a/src/extension/background-script/events/index.js
+++ b/src/extension/background-script/events/index.js
@@ -2,16 +2,18 @@ import PubSub from "pubsub-js";
 
 import { updateAllowance } from "./allowances";
 import {
-  paymentSuccessNotification,
-  paymentFailedNotification,
-  lnurlAuthSuccessNotification,
   lnurlAuthFailedNotification,
+  lnurlAuthSuccessNotification,
+  paymentFailedNotification,
+  paymentSuccessNotification,
 } from "./notifications";
 import { persistSuccessfullPayment } from "./persistPayments";
 
 const subscribe = () => {
   PubSub.subscribe("ln.sendPayment.success", paymentSuccessNotification);
   PubSub.subscribe("ln.sendPayment.failed", paymentFailedNotification);
+  PubSub.subscribe("ln.keysend.success", paymentSuccessNotification);
+  PubSub.subscribe("ln.keysend.failed", paymentFailedNotification);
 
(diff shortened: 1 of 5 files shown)


bumi:
it makes sense to me, but we should add the according subscriptions also to the keysend payments for the success/error notifications

src/extension/background-script/events/index.js
```diff
@@ -2,16 +2,18 @@ import PubSub from "pubsub-js";
...
+  lnurlAuthSuccessNotification,
+  paymentFailedNotification,
+  paymentSuccessNotification,
 } from "./notifications";
 import { persistSuccessfullPayment } from "./persistPayments";
 
 const subscribe = () => {
   PubSub.subscribe("ln.sendPayment.success", paymentSuccessNotification);
   PubSub.subscribe("ln.sendPayment.failed", paymentFailedNotification);
+  PubSub.subscribe("ln.keysend.success", paymentSuccessNotification);
```
rolznz, line 15: should we have a for loop on ['sendPayment', 'keysend'] here? looks like there are now 3 subscriptions which are the same for both

rolznz:
I am just wondering now that this issue comes because of how the spec was implemented. The name "sendPayment" is a generic name for a payment (but actually just bolt11) whereas "keysend" is a named from the specific type of payment. What happens if we add other ways of payment too? what about paying using webln.lnurl()? are we going to subscribe to all these individually?

I don't have answers, just maybe things to consider. CC @bumi

rolznz (approved):

src/extension/background-script/events/index.js
im-adithya: Makes sense! Done!

bumi:
lnurl payments are `sendPayment` calls with a bolt11. 
other forms can be discussed when there will be other forms. (bolt12 🌈🦄)

Issue #1318, opened 2022-08-21
Show full tx history


secondl1ght:
### Is your feature request related to a problem? Please describe

Currently the Outgoing and Incoming transaction history screens display an unknown amount of records to the user.

### Describe the solution you'd like

It would be good to display all of the records, either with pagination or just an infinite scroll. Could also limit to a certain amount initially and then have a 'show all' button at the bottom.

### Additional context

![image](https://user-images.githubusercontent.com/85003930/185812682-02841532-2ceb-4223-af6e-57b3a6963448.png)

### Are you working on this?

No

manavdesai27:
@secondl1ght @escapedcat Which transactions are we not showing here? According to the code all the transactions must be visible here.

escapedcat:
> @secondl1ght @escapedcat Which transactions are we not showing here? According to the code all the transactions must be visible here.

(Some?) connectors don't return all (invoice) entries per default, i.e. LND:

>  By default, the first 100 invoices created will be returned.

See: https://lightning.engineering/api-docs/api/lnd/lightning/list-invoices

PR #2140, opened 2023-02-23
feat: add account id to transactions

### LINKED ISSUE #1831: Filter outgoing transactions by account
### Feature description

The outgoing transaction list currently displays all transactions from all accounts but should be scoped to the current account in use. 

### Describe the solution

Save the account ID along with the [transaction data](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/db.ts#L23) at the time of sending a payment and filter the outgoing transaction list by the current account in use.

im-adithya:

Adds account id to payments

Fixes #1831 (first step)

### Type of change

### How has this been tested?

Didn't test yet.

### Checklist


src/extension/background-script/actions/ln/keysend.ts
```diff
@@ -7,6 +7,14 @@ import state from "../../state";
 export default async function keysend(message: Message) {
   PubSub.publish(`ln.keysend.start`, message);
   const { destination, amount, customRecords } = message.args;
+
+  const accountId = await state.getState().currentAccountId;
+  if (!accountId) {
+    return {
+      error: "select an account.",
```
escapedcat, line 14: nit: below it's Capital-case

src/extension/background-script/actions/payments/listByAccount.ts
```diff
@@ -0,0 +1,23 @@
+import type { MessagePaymentListByAccount } from "~/types";
+
+import db from "../../db";
+
+const listByAccount = async (message: MessagePaymentListByAccount) => {
+  const { accountId } = message.args;
+  const limit = message?.args?.limit || 2121;
```
escapedcat, line 7: Where is 2121 coming from? Is this random or does it have context? Wanna add a comment?

escapedcat (approved):
utack

src/extension/background-script/actions/payments/listByAccount.ts
bumi: it's a nice number. I think I added this because we don't have pagination.

escapedcat: :P better add a comment otherwise people might read more into it

src/extension/background-script/actions/payments/all.ts
```diff
@@ -3,6 +3,7 @@ import type { MessagePaymentAll } from "~/types";
 import db from "../../db";
 
 const all = async (message: MessagePaymentAll) => {
+  // TODO: Add pagination instead of limiting to 2121
```
escapedcat, line 6: #1318 could be added here

src/extension/background-script/migrations/index.ts
```diff
@@ -40,6 +41,18 @@ const migrations = {
     }
   },
 
+  migratePaymentsWithoutAccountId: async () => {
+    const { accounts } = state.getState();
+    if (Object.keys(accounts).length == 1) {
```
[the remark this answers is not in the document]
bumi: hmm. the issue is that we then still have unexpected data in the DBs which we will always have to account for. 

can we associate the payments now, but not change the filtering and do that in an upcoming release. 
I assume old payments are less relevant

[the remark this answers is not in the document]
reneaaron: I thought the same. 👍

>can we associate the payments now, but not change the filtering and do that in an upcoming release.

Currently the transaction list is not filtered by account, so that should already be the case.

reneaaron (approved):
tACK

PR #2143, opened 2023-02-25
Add safari to compatible browsers

chmac:

The [macOS installer](https://github.com/getAlby/alby-installer-macos) suggests that Safari is supported. Feel free to reject this PR if I've misunderstood something.

### Type of change


bumi:
Safari is only party supported because we don't have a food update process so far. people would need to download the macOS app. 
we probably have to first get this app officially in the app store that update are then handled by the apple app store.

chmac:
@bumi I think all safari extensions have to be installed as mac apps. At least that's how it's always worked for all the extensions I've installed. But updates for the mac app don't work well, is that what you  mean?

PR #2152, opened 2023-02-28
feat: improve LNURL ux

### LINKED ISSUE #1569: Use minSendable and maxSendable values from LNURLp response to cap user input
### Feature description

Right now, when paying using LNURL-pay (either directly of through a Lightning Address), the `minSendable` and `maxSendable` fields aren't use to their full extent. From what I can see, `minSendable` only seems to be used to populate the amount field on Alby. If a user tries to pay an amount that is lower than `minSendable`, no error will display until the payment is actually tried and rejected by the recipient, with a vague "could not fetch invoice" error.

### Describe the solution

Alby should use `minSendable` et `maxSendable` fields sent by the recipient's "service" to cap user input. For example, display an error below the amount field if the amount is out of the [minSendable, maxSendable] range, and disable the "Send" button. Furthermore, the 4 prefilled amount buttons should take the min and max field into account. For example, the lowest amont could be set to minSendable, while the highest amount could be set to max(maxSendable, balance).

### Additional context

If you want to try this, you can for example get a Lightning Address from lnmarkets.com and try to pay an amount that is lower than 1000 sats to it.

### Are you working on this?

Yes

reneaaron:

Have the LNURL payment form respect the provided `min` and `max` values. 

 - Add the values to the form input 
 - Disable SatsButtons that are out of the range

Fixes #1569 

### Type of change

image

### How has this been tested?

[LNURLs used for testing](https://github.com/getAlby/lightning-browser-extension/wiki/%F0%9F%A4%A0-Alby-in-the-wild)

### Checklist

### DIFF
--- a/src/app/components/SatButtons/index.tsx
+++ b/src/app/components/SatButtons/index.tsx
@@ -5,39 +5,29 @@ import Button from "../Button";
 type Props = {
   onClick: (amount: string) => void;
   disabled?: boolean;
+  min?: number;
+  max?: number;
 };
 
-function SatButtons({ onClick, disabled }: Props) {
+function SatButtons({ onClick, disabled, min, max }: Props) {
+  const sizes = [1000, 5000, 10000, 25000];
+
   return (
     <div className="flex gap-2">
-      <Button
-        icon={<SatoshiV2Icon className="w-4 h-4" />}
-        label="1k ⚡"
-        onClick={() => onClick("1000")}
-        fullWidth
-        disabled={disabled}
-      />
-      <Button
-        icon={<SatoshiV2Icon className="w-4 h-4" />}
-        label="5k ⚡"
-        onClick={() => onClick("5000")}
-        fullWidth
-        disabled={disabled}
-      />
-      <Button
-        icon={<SatoshiV2Icon className="w-4 h-4" />}
-        label="10k ⚡"
-        onClick={() => onClick("10000")}
-        fullWidth
-        disabled={disabled}
-      />
-      <Button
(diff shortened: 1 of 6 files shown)


src/app/components/form/DualCurrencyField/range.tsx
```diff
@@ -0,0 +1,19 @@
+import { useTranslation } from "react-i18next";
```
reneaaron, line 1: Not sure if this component makes sense. If there is a more elegant way to solve that please let me know 😅

rolznz: It looks ok to me. Maybe "RangeLabel" as the name? and it might be good to throw an exception or return null instead of an empty fragment if neither field is passed.

rolznz (approved):

escapedcat:
Does this make #1697 obsolete?

reneaaron:
> Does this make #1697 obsolete?

I think so, yes. It just makes more sense to me (and is a lot less complicated I guess).

src/app/components/SatButtons/index.tsx
```diff
@@ -5,39 +5,29 @@ import Button from "../Button";
...
-        label="25k ⚡"
-        onClick={() => onClick("25000")}
-        fullWidth
-        disabled={disabled}
-      />
+      {sizes.map((size) => (
+        <Button
+          key={size}
+          icon={<SatoshiV2Icon className="w-4 h-4" />}
+          label={size / 1000 + "k"}
```
bumi, line 21: so the head of Emoji is removing the  ⚡️ emojis here?

PR #2169, opened 2023-03-04
fix: remove max amount from keysend payments

bumi:


MoritzKa:
it works here: https://rss.com/podcasts/alby-buzz-test/

src/app/screens/Keysend/index.tsx
```diff
@@ -119,8 +119,7 @@ function Keysend() {
                   <DualCurrencyField
                     id="amount"
                     label={t("amount.label")}
-                    min={+0 / 1000}
-                    max={+1000000 / 1000}
+                    min={0}
```
reneaaron, line 122: Shouldn't the min be `1`? 

```suggestion
                    min={1}
```

reneaaron (approved):
utACK

PR #2182, opened 2023-03-08
Update websites.json: add a Nostr client

BlowaterNostr:
Add https://blowater.deno.dev/ to Nostr website page

### Type of change

### DIFF
--- a/src/app/screens/Discover/websites.json
+++ b/src/app/screens/Discover/websites.json
@@ -121,7 +121,7 @@
       {
         "title": "Blowater",
         "subtitle": "A web chat client with delightful UX",
-        "logo": "https://blowater.deno.dev/logo.png",
+        "logo": "https://cdn.getalby-assets.com/alby-extension-website-screen/blowater_thumbnail.png",
         "url": "https://blowater.deno.dev"
       },
       {
--- a/src/app/screens/Discover/websites.json
+++ b/src/app/screens/Discover/websites.json
@@ -119,6 +119,12 @@
         "url": "https://blogstack.io/"
       },
       {
+        "title": "Blowater",
+        "subtitle": "A web chat client with delightful UX",
+        "logo": "https://blowater.deno.dev/logo.png",
+        "url": "https://blowater.deno.dev"
+      },
+      {
         "title": "emon.chat",
         "subtitle": "Instant messaging client with built in lightning payments",
         "logo": "https://cdn.getalby-assets.com/alby-extension-website-screen/emon_thumbnail.svg",


BlowaterNostr:
@reneaaron Will reproduce and give you a better report.

reneaaron:
@BlowaterNostr Cool, thanks! 🙌 Feel free to open a new issue for that 👍

BlowaterNostr:
Maybe it's not necessarily a bug of Alby. But the way my application is interacting with it. The direct observation is that Using Alby on Firefox with Blowater will trigger a null reference exception very deep in Preact's code with no call stack from application code. With the same code, but Chrome works.

reneaaron:
Okay. Is the code available publicly? Happy to take a quick look...

BlowaterNostr:
@reneaaron the code is not open sourced yet. I will consider it once it reaches feature stability.

BlowaterNostr:
@reneaaron I have found the reason and it's a good indicator that Alby should support concurrent API calls or at least throw errors when user concurrently calls.

The story is this, in Blowater, we concurrently call `encrypt/decrypt/signEvent/getPublicKey` in many different promise chains. The current behavior of `window.nostr` is to simply return the result of the first call in the current JS tick. For example, if
```
getPublicKey()
encrypt()
```
are called in the same tick. `await encrypt()` will return the public key. Although in the log Alby gives a warning message, but the code behavior is not a good design. I would categorize it as either a bug or a design flaw.

I worked around this problem by adding a queue in my application code to wait for the previous call to resolve before invoking the next.

But, because Blowater also use IndexedDB, therefore, wrong data before I added the queue were persisted. I happened to never delete the database on Firefox, so that only Firefox appears buggy. I deleted the database and everything works now.

BlowaterNostr:
#2218 does not block the development of my client, can we merge this first? @reneaaron

BlowaterNostr:
Since #2218 is just merged, can we merge this now?

src/app/screens/Discover/websites.json
```diff
@@ -118,6 +118,12 @@
         "logo": "https://cdn.getalby-assets.com/alby-extension-website-screen/blockstack-logo.png",
         "url": "https://blogstack.io/"
       },
+      {
+        "title": "Blowater",
+        "subtitle": "A web chat client with delightful UX",
+        "logo": "https://blowater.deno.dev/logo.png",
```
reneaaron, line 124:
```suggestion
        "logo": "https://cdn.getalby-assets.com/alby-extension-website-screen/blowater_thumbnail.png",
```

BlowaterNostr:
@reneaaron 
image

Blowater isn’t showing up. Is this because the action failed?

bumi:
@BlowaterNostr you might stil be on the old version. 
CleanShot 2023-05-07 at 11 42 38@2x

PR #2187, opened 2023-03-09
chore: call it alby account

bumi:
image

### DIFF
--- a/src/i18n/locales/en/translation.json
+++ b/src/i18n/locales/en/translation.json
@@ -39,7 +39,7 @@
       "title": "Connect",
       "description": "To start using Alby to make online payments, connect your lightning wallet to the extension.",
       "alby": {
-        "title": "Alby",
+        "title": "Alby Account",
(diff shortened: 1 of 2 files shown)


stackingsaunter:
I think we should be consistent with capitalisation, so I suggest always capitalise "Alby Account"

src/i18n/locales/en/translation.json
```diff
@@ -39,7 +39,7 @@
       "title": "Connect",
       "description": "To start using Alby to make online payments, connect your lightning wallet to the extension.",
       "alby": {
-        "title": "Alby",
+        "title": "Alby Account",
         "description": "Sign up or use your existing Alby account to get started with lightning payments in no time.",
```
stackingsaunter, line 43: capitalise Alby **A**ccount in the description as well

stackingsaunter (changes requested):
capitalise Alby **A**ccount in the description as well

reneaaron (approved):

stackingsaunter (approved):

Issue #1956, opened 2023-01-09
[Feature] Add Alby connector


rolznz:
### Feature description

Currently we are using the lndhub connector for hosted Alby accounts as well as other lndhub connections such as Bluewallet. There is no reliable way to determine whether the lndhub connection is an Alby hosted account or a connection to some other node (only option is checking the account alias or connector config url, which could both change). Also, it limits the possibilities of what we can support since we need to be compatible with the lndhub API. If we create an Alby connector, we can use the Alby OAuth API.

### Describe the solution

Create a new "Alby" connector and start using it for new user accounts. Consider whether a migration is needed in the future or we can keep existing accounts using the lndhub connector.

**TODO**
 - [ ] Create a new Alby connector (copy LNDHub) 
 - [ ] Refactor `isAlby()` to use the account type
 - [ ] Migrate all existing LNDHub accounts (check for the lndhub url (ln.getalby.com))

### Additional context

From the discussion on how to determine whether or not the connection is an Alby hosted wallet or not: https://github.com/getAlby/lightning-browser-extension/pull/1892

### What does this enable?
- Show lightning address on receive tab
- Add link to alby account in the extension connector UI
- Add extra metadata in invoices and transactions (available from the Alby backend)
- Using an Alby API rather than the LndHub API

### Are you working on this?

None

PR #2195, opened 2023-03-11
feat: avatars & account menu overhaul

reneaaron:

image


stackingsaunter:
I prepared handoff for the dropdown 

Screenshot 2023-03-17 at 23 04 52

Screenshot 2023-03-17 at 23 05 05

[Figma file here ](https://www.figma.com/file/AMzMvcwrr2IdunqMakaB8r/navigation%2C-dropdown%2C-receive-onchain-work%2C-connectors%2C-nostr-events?node-id=119-31384&t=PLf4EExSybimRo3k-0)

stackingsaunter:
Dark version uses white text and neutral300 for "balance" and fiat
Screenshot 2023-03-18 at 18 46 16

bitcoinuser:
> Dark version uses white text and neutral300 for "balance" and fiat Screenshot 2023-03-18 at 18 46 16

Hi, I think instead of (Switch Account) should be (Selected Account). And in the selected account itens could list the one that is selected with the check mark icon in front, the way works today.

reneaaron:
There are still some open todos in here, but maybe we can do a first round of review? 

**📃 Open tasks**
- [ ] Fix tests
- [ ] Fix overflow for long account names
- [ ] Design review

src/app/components/AccountMenu/index.tsx
```diff
@@ -73,49 +78,77 @@ function AccountMenu({ showOptions = true }: Props) {
...
+          </Menu.Item>
+          <Menu.ItemButton
+            onClick={() => {
+              openOptions(`accounts/${authAccount?.id}`);
+            }}
+          >
+            <WalletIcon className="h-5 w-5 mr-2 text-gray-700 dark:text-neutral-300" />
+            {t("options.account.account_settings")}
+          </Menu.ItemButton>
+          {isAlbyAccount(authAccount?.alias) && (
```
reneaaron, line 134: How could we get rid of this? Currently we don't access the initialized connector in here 🤔 (and even with the connector we would need to introduce an `Alby` connector (+ try to migrate older ones?)

rolznz:
In the new design you cannot see your balance without clicking on the account select dropdown
image

I think we should keep the balance always visible. @stackingsaunter @reneaaron thoughts?

rolznz (approved):
tACK

rolznz:
@reneaaron @stackingsaunter great improvement. The menu looks much more clean and the links to manage your Alby account should reduce a lot of confusion.

src/app/components/AccountMenu/index.tsx
```diff
@@ -73,49 +77,78 @@ function AccountMenu({ showOptions = true }: Props) {
...
+            <Menu.ItemButton
```
bumi, line 134: I am wondering if we should add this also (or only) to the account detail page. 
Right now we say "Account settings" and "manage the Alby account" 
maybe it can only be "Account settings" and in there we have a prominent link to manage the Alby account? 

potentially this could even be its own section explaining the user a bit the difference?

src/app/components/AccountMenu/index.tsx
```diff
@@ -73,49 +77,78 @@ function AccountMenu({ showOptions = true }: Props) {
...
+              onClick={() => {
+                window.open(`https://getalby.com/login`, "_blank");
```
bumi, line 136: I always link to the /user  - even though we redirect from the login page.

src/app/components/AccountMenu/index.tsx
bumi: we only have the alias in the authAccount not the host or something else? 

guess we should move asap to an individual connector to have a proper way to identify this.
right now this could be as easy as duplicating lndhub.ts into alby.ts and setting the connector to `alby`.

bumi (approved):
this looks great! nice work! 

I did not check the code in detail, but LGTM.

bumi:
CleanShot 2023-04-05 at 10 30 09@2x

is this link always going to the accounts list? because now if I open the accounts details from the new menu, then this back link does not work (there is nothing to go back to)

stackingsaunter:
> I think we should keep the balance always visible. @stackingsaunter @reneaaron thoughts?

I think it's important on the main screen and when transacting, that's why I think it's better to move account balance to the amount inputs, see here https://github.com/getAlby/lightning-browser-extension/issues/2237 and https://github.com/getAlby/lightning-browser-extension/issues/2236. It should work well if those all are implemented

stackingsaunter:
If there's no account to switch, can we not have "switch account" in the dropdown? 

Screenshot 2023-04-05 at 16 43 55

reneaaron:
> guess we should move asap to an individual connector to have a proper way to identify this.
> right now this could be as easy as duplicating lndhub.ts into alby.ts and setting the connector to alby.

Added those things and some todos to #1956 and added it to our project board, let's tackle that in a follow up PR.

Tackled the feedback from this PR, thanks for testing! 🙌

src/app/components/AccountMenu/index.tsx
reneaaron: Added some info to #1956

PR #2207, opened 2023-03-14
fix: change lndconnect host to http (tor)

reneaaron:

Cloud9 nodes connect via TOR, there is no need to use HTTPS here. (as this also requires users to confirm that they want to accept the self-signed cert)


reneaaron:
There is currently no way to fix that since browsers check certs before even a connection is established. Best way forward is probably to move towards LNC and such where self signed certs / http hosting is not required anymore.

PR #2208, opened 2023-03-14
Feature/eclair improvements

kiwiidb:

- Add `/listinvoices` function to connector to avoid error messages. However, Eclair has currently no way of listing only settled invoices, so all created invoices are displayed.
- Fixes a bug where a payment fee was displayed as 1000x too much.
- Use the [`/usablebalances`](https://acinq.github.io/eclair/#usablebalances) endpoint to fetch the balance, which is easier and more accurate than using the `/channels` endpoint. 

### Type of change

### How has this been tested?

Tested locally by using Eclair.

### Checklist

### DIFF
--- a/src/extension/background-script/connectors/eclair.ts
+++ b/src/extension/background-script/connectors/eclair.ts
@@ -2,19 +2,20 @@ import Base64 from "crypto-js/enc-base64";
 import UTF8 from "crypto-js/enc-utf8";
 
 import Connector, {
-  SendPaymentArgs,
-  SendPaymentResponse,
   CheckPaymentArgs,
   CheckPaymentResponse,
+  ConnectorInvoice,
   ConnectPeerResponse,
+  GetBalanceResponse,
   GetInfoResponse,
   GetInvoicesResponse,
-  GetBalanceResponse,
+  KeysendArgs,
   MakeInvoiceArgs,
   MakeInvoiceResponse,
+  SendPaymentArgs,
+  SendPaymentResponse,
   SignMessageArgs,
   SignMessageResponse,
-  KeysendArgs,
 } from "./connector.interface";
 
 interface Config {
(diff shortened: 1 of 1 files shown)


escapedcat (approved):
utack

src/extension/background-script/connectors/eclair.ts
```diff
@@ -61,30 +69,38 @@ class Eclair implements Connector {
...
+        }) => ({
+          id: invoice.paymentHash,
+          memo: invoice.description,
+          settled: true,
+          settleDate: invoice.timestamp * 1000,
+          totalAmount: invoice.amount / 1000,
+          type: "received",
+        })
+      )
+      .sort((a: ConnectorInvoice, b: ConnectorInvoice) => {
```
reneaaron, line 90: Should we add some filtering then here (like we do for commando?):

https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/commando.ts#L206

reneaaron: I think the API should always return the same kind of data, also across connectors.

kiwiidb: Eclair doesn't support this kind of filter at the moment.

reneaaron: Exactly, the question is whether we should filter it on the client?

kiwiidb: This call doesn't even tell you if an invoice was settled or not :) 
The alternative is to use the [audit](https://acinq.github.io/eclair/#audit) api endpoint, but that misses the memo field.

Also if we are being realistic, there are very few people that use their Eclair node in this way so I wouldn't bother with too much more effort. In fact, we haven't received  a single complaint in the 1year+ that it was in the extension while it was broken.

reneaaron: > This call doesn't even tell you if an invoice was settled or not :)

Oh, okay then. 👍

reneaaron (approved):
See comment

utACK

src/extension/background-script/connectors/eclair.ts
```diff
@@ -38,7 +39,14 @@ class Eclair implements Connector {
...
 
   get supportedMethods() {
-    return ["getInfo", "keysend", "makeInvoice", "sendPayment", "signMessage"];
+    return [
+      "getInfo",
+      "keysend",
+      "makeInvoice",
+      "sendPayment",
+      "signMessage",
+      "listInvoices",
```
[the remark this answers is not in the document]
kiwiidb: The point of this PR is not to add extra features but just to fix a few broken things. So I don't think it's necessary.

PR #2218, opened 2023-03-15
fix: race conditions

reneaaron:

First steps for fixing the issue @BlowaterNostr brought up in #2182: 

> I have found the reason and it's a good indicator that Alby should support concurrent API calls or at least throw errors when user concurrently calls.
> 
> The story is this, in Blowater, we concurrently call encrypt/decrypt/signEvent/getPublicKey in many different promise chains. The current behavior of window.nostr is to simply return the result of the first call in the current JS tick. For example, if
> 
> getPublicKey()
> encrypt()
> 
> are called in the same tick. await encrypt() will return the public key. Although in the log Alby gives a warning message, but the code behavior is not a good design. I would categorize it as either a bug or a design flaw.
> 
> I worked around this problem by adding a queue in my application code to wait for the previous call to resolve before invoking the next.
> 
> But, because Blowater also use IndexedDB, therefore, wrong data (should be an event, but was {enabled: true}) before I added the queue were persisted. I happened to never delete the database on Firefox, so that only Firefox appears buggy. I deleted the database and everything works now.
> 
> Fundamentally, TS is not type safe when it comes to serialize raw JSON to objects. Therefore, if application code always do type check during de serialization at runtime, this bug can be prevented. But this is not the norm in TS/JS code.
> 
> The communicate between extension and application is similar to serialization because it's also across program boundaries. We need better type safety here.
> 
> On Alby side, it should not return incorrect data (silent error) even if concurrent call is not allowed. Throwing exceptions is acceptable if concurrent call needs more time to design.

I've also put together a little repro case to test this: 
https://codepen.io/reneaaron/pen/zYJjZbW

So currently some of the calls will still fail as we currently only allow one call at a time, however the results always match the call as now the promises are resolved for the executing call.


src/extension/ln/nostr/index.ts
```diff
@@ -61,19 +80,46 @@ export default class NostrProvider {
...
+          !window.nostr._requests[messageEvent.data.id]
         ) {
+          if (
+            messageEvent.data &&
+            messageEvent.data.response &&
+            messageEvent.data.application === "LBE" &&
+            messageEvent.data.scope === "nostr" &&
+            !window.nostr._requests[messageEvent.data.id]
+          ) {
+            console.log(
```
reneaaron, line 93: I'm seeing events here that already have been removed from the `_requests` object:  

image

Any idea where this might come from? They seem to be called twice. 🤔

reneaaron: Seems those event handlers are not cleaned up if aborted early. (`nostr call already executing`)

image

src/extension/ln/nostr/index.ts
```diff
@@ -41,6 +56,8 @@ export default class NostrProvider {
   // NOTE: new call `action`s must be specified also in the content script
   execute<T>(action: string, args?: Record<string, unknown>): Promise<T> {
     return new Promise((resolve, reject) => {
+      const id = Math.random().toString().slice(4);
```
rolznz, line 59: Would it be better to use an incremental ID? I think there would be no chance of collisions in that case

src/extension/ln/webbtc/index.ts
```diff
@@ -6,15 +6,30 @@ type RequestInvoiceArgs = {
   defaultMemo?: string;
 };
 
+declare global {
```
rolznz, line 9: This is duplicated in 3 files and is going to be difficult to maintain, I think we should really consider extracting all the duplicated provider code

src/extension/ln/nostr/index.ts
```diff
@@ -41,6 +56,8 @@ export default class NostrProvider {
...
+      window.nostr._requests[id] = { resolve, reject };
```
rolznz, line 60: I don't believe the _requests object is actually needed. Execute creates a dynamic function for this specific request, and can compare messageEvent.data.id against the id that was passed in the postMessage.

src/extension/ln/nostr/index.ts
bumi: incrementing requires state, can we rely on the state? 
I'd say the collision is unlikely.

src/extension/ln/nostr/index.ts
reneaaron: I think it's needed because the promise isn't resolved directly but through another callback from the global `message` event posted on the window. 

[window.addEventListener("message", handleWindowMessage);](https://github.com/getAlby/lightning-browser-extension/blob/fix/race-condition/src/extension/ln/nostr/index.ts#L107)

Since multiple calls can be executed in parallel and the order of events doesn't have to match the order they were called that lead to the race conditions that could lead to certain functions returning results of other function calls. 

`window.nostr.signMessage` could potentially return the result of `window.nostr.getPublicKey` under certain conditions. By referencing the promises through this unique id that shouldn't be the case anymore.

src/extension/ln/webbtc/index.ts
```diff
@@ -102,43 +110,6 @@ export default class WebBTCProvider {
...
-          // 3. data: the actual response data
-          resolve(messageEvent.data.data.data);
-        }
-        // For some reason must happen only at the end of this function
-        window.removeEventListener("message", handleWindowMessage);
-      }
-
-      window.addEventListener("message", handleWindowMessage);
-    });
+    return postMessage(action, "webln", args);
```
rolznz, line 113: should this be "webbtc?"

src/extension/ln/nostr/index.ts
reneaaron: Thanks for your feedback, with the added check for the id in the event handler this became much simpler 👌

src/extension/ln/webbtc/index.ts
reneaaron: no, webbtc currently is handled by webln actions internally. that will change with the mnemonic / psbt PR soon, though.

src/extension/content-script/onendnostr.js
```diff
@@ -43,14 +43,16 @@ async function init() {
...
+          error:
+            "window.nostr call cancelled (rejecting further window.nostr calls until the next reload)",
+        });
         return;
       }
+
       // if a call is active we ignore the request
       if (callActive) {
-        console.error("nostr call already executing");
+        postMessage(ev, { error: "window.nostr call already executing" });
```
rolznz, line 55: is callActive still needed now that we support calls executing in parallel?

src/extension/providers/postMessage.ts
```diff
@@ -0,0 +1,51 @@
+export function postMessage<T>(
+  scope: string,
+  action: string,
+  args: T | undefined
+): Promise<T> {
+  return new Promise((resolve, reject) => {
+    const id = Math.random().toString().slice(4);
+
+    // post the request to the content script. from there it gets passed to the background script and back
+    // in page script can not directly connect to the background script
```
rolznz, line 10: It seems this might be false and doesn't really make sense to me. When you post a message to the window, anything listening to the window should receive it. Are you able to add a TODO: comment to re-investigate this?

src/extension/content-script/onendnostr.js
reneaaron: we don't support sending in parallel. there is still only one request at a time. i think the main purpose of this if statement is to prevent websites from being able to spin up a larger amount of payment windows.

PR #2222, opened 2023-03-16
feat: contentUri to alby boost button

pavanjoshi914:
_A clear and concise description of what you want to happen_
add contentUri field to battery responses of Alby boost button
support for youtube video, youtube channel, vimeo and medium

### Type of change

### Checklist

### DIFF
--- a/src/extension/content-script/batteries/VimeoVideo.ts
+++ b/src/extension/content-script/batteries/VimeoVideo.ts
@@ -16,6 +16,7 @@ const battery = (): Battery | void => {
   const text = videoDescription.textContent || "";
   let match;
   let recipient;
+  const contentUri = document.location.toString();
   // check for an lnurl
   if ((match = text.match(/(lnurlp:)(\S+)/i))) {
     recipient = match[2];
@@ -38,6 +39,7 @@ const battery = (): Battery | void => {
     ...getOriginData(),
     name,
     icon: imageUrl,
+    contentUri: contentUri,
   };
 };
 
--- a/src/extension/content-script/batteries/VimeoVideo.ts
+++ b/src/extension/content-script/batteries/VimeoVideo.ts
@@ -1,9 +1,11 @@
+import { Battery } from "~/types";
+
 import getOriginData from "../originData";
-import { findLightningAddressInText, setLightningData } from "./helpers";
+import { findLightningAddressInText } from "./helpers";
 
 const urlMatcher = /^https:\/\/vimeo.com\/.*\d{4,}\/?$/;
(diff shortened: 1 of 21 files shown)


rolznz:
From Slack:

Please review Adithya's latest changes. It now includes alby-tools
. I think we can use that to include the contentUri in the LUD18 data by passing the following to `ln.generateInvoice`:
```
payerData: {
          contentUri,
        },
```

`getPayerData` also looks a bit complicated now, could you think about simplifying it? you could create an empty object first and then check each field, if it is valid, then add it to the object
.

src/extension/content-script/batteries/VimeoVideo.ts
```diff
@@ -16,6 +16,7 @@ const battery = (): Battery | void => {
   const text = videoDescription.textContent || "";
   let match;
   let recipient;
+  const contentUri = document.location.toString();
```
rolznz, line 19: could you remove these changes for now (vimeo, medium, youtube channel)? I don't think document.location is correct (it's the full URL and can contain strange query parameters etc.). We haven't aligned on how to pass metadata yet, so I think we should focus on just one battery - the youtube video one.

src/extension/content-script/batteries/YouTubeVideo.ts
```diff
@@ -23,6 +23,7 @@ const battery = async (): Promise<Battery | void> => {
   }
   let match;
   let lnurl;
+  const contentUri = document.location.toString();
```
rolznz, line 26: I think we should extract the canonical url from the html document. The location includes query parameters and other things which can produce a different content URI.

Note: if we decide not to include contentUri and have a different schema per content type, then we could just include the video ID.

pavanjoshi914: resolved

pavanjoshi914:
did some research on ipfs providers. here are some ipfs providers we can consider

| IPFS Providers | Amount of Storage Space | Centralized/Decentralized |
| --- | --- | --- |
| Pinata | 1 GB (free) | Centralized |
| Temporal Cloud | 1 GB (free) | Centralized |
| Fleek | 1 GB (free) | Centralized |
| Infura | 5 GB (free) | Centralized |
| Textile.io | 2 GB (free) | Centralized |
| Filecoin | Varies (decentralized) | Decentralized |
| Arweave | Varies (decentralized) | Decentralized |
| Textile Threads | Varies (decentralized) | Decentralized |
| Skynet | Varies (decentralized) | Decentralized |

PR #2239, opened 2023-03-19
feat(i18n): support polish language

escapedcat:
Prepare Polish  
image

Waiting for the matching Weblate PR


stackingsaunter (approved):
I went through all the screens, the translation seems very good.

jesterhodl:
Thanks! @escapedcat when/if I update weblate to correct strings or add new ones, do you recommend I ping you some how?
Or is there an automated process that deals with updates?

PR #2270, opened 2023-03-25
feat: webln.getBalance()

### LINKED ISSUE #2242: Add getBalance() to WebLN
### Feature description

`webln.getBalance()` should return the current balance in sats. 

### Describe the solution

**Extension**
- [ ] Add to `webln`
- [ ] Create a new permission to read the account balance

**Others**
 - [ ] Add docs to webln.guide
 - [ ] Create typings in webln typescript package

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

reneaaron:

Adds `webln.getBalance()` to WebLN.

Fixes #2242 

### Screenshots

image


src/extension/background-script/connectors/commando.ts
```diff
@@ -142,7 +142,15 @@ export default class Commando implements Connector {
...
   get supportedMethods() {
-    return supportedMethods;
+    return [
+      "getInfo",
+      "keysend",
+      "makeInvoice",
+      "sendPayment",
+      "signMessage",
+      "getBalance",
+      ...supportedMethods,
```
reneaaron, line 152: Not happy about the API surface here since some of the methods are in `webln` and others are only accessible via `webln.request()` 🤔

rolznz: I agree. It is two separate APIs right? (the WebLN high level API that is the same across all node types, and then the WebLN.request low level API which are methods exposed directly from the node).

reneaaron:
**TODO**
- Remove native methods from `methods`
- Release `getBalance` in WebLN

bumi:
Do we ship it? :)

reneaaron:
@bumi After spending some more thoughts on this my proposal for the getInfo calls is this: 

`getInfo()` returns an object with keys for every method that is accessible via `window.webln`. 

```json
{
      "getInfo",
      "keysend",
      "makeInvoice",
      "sendPayment",
      "signMessage",
      "getBalance",
}
```

Connectors that support `webln.request` would return something like this: 

```json
{
      "getInfo",
      "...",
      "request": [
          "connect",
          "disconnect",
          "fundchannel",
          "...",
      ]
}
```

Consumers can then check for supported methods like this: 

```js
const methods = (await webln.getInfo()).methods;

const keysendSupported = "keysend" in methods;
const fundChannelSupported = "request" in methods && "fundchannel" in methods.request;
```

An alternative would be to flatten this object and separate them with dots: 

```json
{
      "getInfo",
      "request.connect",
      "request.fundChannel"
}
```

I tend to lean towards the latter one as it's less complex (also for consumers) and you have to check for specific methods (fundchannel, etc) anyway if you want to implement certain connector-specific functions.

src/extension/background-script/actions/webln/getBalanceOrPrompt.ts
```diff
@@ -0,0 +1,83 @@
+import utils from "~/common/lib/utils";
+import db from "~/extension/background-script/db";
+import state from "~/extension/background-script/state";
+import { MessageDefault } from "~/types";
+
+const getBalanceOrPrompt = async (message: MessageDefault) => {
```
rolznz, line 6: This prompt / allowance / permission logic would be great to be extracted - most of the lines are the same across different actions

rolznz (approved):

src/extension/background-script/connectors/commando.ts
```diff
@@ -142,7 +142,15 @@ export default class Commando implements Connector {
...
   get supportedMethods() {
-    return supportedMethods;
+    return [
+      "getInfo",
+      "keysend",
+      "makeInvoice",
+      "sendPayment",
+      "signMessage",
+      "getBalance",
+      ...supportedMethods.map((i) => `request.${i}`),
```
rolznz, line 152: there are some inconsistencies here (2 other places it is `...Object.keys(methods).map((i) => `request.${i}`),`)

I wonder if we should call them all "weblnRequestSupportedMethods" or something to make it specific, then have a helper method to make sure they all use the same flattening function

reneaaron: Good point. I added a `flattenRequestsMethods()` function that is used in all connectors.

supertestnet:
What is the status of this pull request? I too want to use the webln object to display a user's balance in my app

reneaaron:
@supertestnet Thanks for providing your feedback, glad to hear this is something you find useful. 

@bumi Anything blocking from your side? Feel free to merge any time.

Issue #2250, opened 2023-03-22
Revamp forms layout in onboarding screens


stackingsaunter:
There's currently a lot of inconsistencies regarding layout on onboarding & connector screens and this is a proposition to improve them a bit, eg:
<img width="723" alt="Screenshot 2023-03-22 at 01 17 38" src="https://user-images.githubusercontent.com/68239231/226770234-8a990874-a241-4d0c-b7d5-4edb0b6b4b75.png">

I propose changes to the design to make it more cohesive, declutter it a bit, make it a bit clearer for users to go through:
- make forms a vertical 600px wide box with 40px margins (if there are any screenshots/videos they can be placed within the layout)
- delete Alby bee from forms
- delete back button from forms (we don't use it anywhere else, users can use their browser back button, we will need another 2 buttons layout in a new onboarding screen which would force us to have 3 buttons there if we'd like to keep back button)
- all buttons under the forms should have fixed 256px size (same size as we use in settings)
- spacing between header and first paragraph should be 16px
- standard spacing between items in form should be 24px (there might be some cases where it's better to have in bigger though)
- connector forms now display a connector logo (64x64px) next to the header

You can see the changes here (or in [Figma file](https://www.figma.com/file/AMzMvcwrr2IdunqMakaB8r/navigation%2C-dropdown%2C-receive-onchain-work%2C-connectors%2C-nostr-events?node-id=38-4400&t=vmLhv4UDl7bFNrWm-0)):

<img width="756" alt="Screenshot 2023-03-22 at 01 33 27" src="https://user-images.githubusercontent.com/68239231/226771994-90fcb521-8964-41a9-9467-2a01afc6d2aa.png">

<img width="1065" alt="Screenshot 2023-03-22 at 01 31 17" src="https://user-images.githubusercontent.com/68239231/226771756-85fbe83e-04f5-4cc8-9e70-ba6862e6a13f.png">

<img width="935" alt="Screenshot 2023-03-22 at 01 31 35" src="https://user-images.githubusercontent.com/68239231/226771783-ec1ac2bd-8703-4818-bdba-91d08ce0de2d.png">

<img width="728" alt="Screenshot 2023-03-22 at 01 31 53" src="https://user-images.githubusercontent.com/68239231/226771813-feb74cf5-038b-4a65-9f8f-3ddc4c8e8056.png">
[...]

Issue #2203, opened 2023-03-13
CLN Support for BTCPay option


pavlenex:
As agreed on Telegram, opening this issue as a reminder to add ability for BTCPay Server's using CLN as their LN implementation to connect to Alby.

reneaaron:
Hey @pavlenex! Thanks for reporting this issue, I just looked into it:  

Currently it just assume it's a LND connection but since BTCPay Server supports different Lightning Node implementations we need to add support for exchanging other node backend credentials via the config URL. 

https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/connectors/ConnectBtcpay/index.tsx#L42-L44

I currently don't have access to a BTCPay Server running on CLN, could you maybe provide the response of this config requests on your CLN setup? (just open the URL it in your browser and post the redacted JSON here)

stackingsaunter:
@pavlenex I want to prepare design for this, but what should be the form in that connector? Standard CLN form or a config data analogous to LND? I'd need a copy for updated instruction and link for CLN config then  

<img width="613" alt="Screenshot 2023-03-20 at 13 10 10" src="https://user-images.githubusercontent.com/68239231/226335170-7c512fa8-fe80-435f-b9f2-e257016cb6d0.png">

<img width="528" alt="Screenshot 2023-03-20 at 13 12 12" src="https://user-images.githubusercontent.com/68239231/226335603-3cd0b106-38cf-461f-ae40-b4243bff684b.png">

pavlenex:
@stackingsaunter we expose it same way we do the LND. I think it would make sense to have it same way you do for LND.

stackingsaunter:
@pavlenex great, can you just give me specific copy in exchange of "LND Rest - See information" and an example link (the current one has lnd within)

pavlenex:
@stackingsaunter Sure, here you go. Bear in mind, for c-lightning we call it lnd.config for some reason cc @kukks
![Screenshot 2023-03-21 at 11 39 48 AM](https://user-images.githubusercontent.com/36959754/226582796-3e2fec74-0e88-4d79-a682-dbd2522b6d41.png)
![Screenshot 2023-03-21 at 11 39 41 AM](https://user-images.githubusercontent.com/36959754/226582792-5f64be7c-33e8-4d60-99c6-5621ac17106a.png)

[...]

PR #2272, opened 2023-03-26
feat: add connector multiple implementations

### LINKED ISSUE #2254: Add intermediary steps for personal nodes that can run multiple implementations
Related to https://github.com/getAlby/lightning-browser-extension/issues/1835 
Dependent on https://github.com/getAlby/lightning-browser-extension/issues/2203

Some nodes can have multiple apps/connectors and currently we assume only one (LND). Good example is Umbrel, which leads straight to LND connector, while Umbrel users run CLN on Umbrel as well, or can connect using LNC.

Those connectors with multiple apps/nodes should have intermediary step when users can choose which implementation they run. That step can utilize cards design from main connectors screen. I also included here minor changes to Umbrel display of LND and LNC. Here are steps for Umbrel, BTCPay, Citadel, RaspiBlitz, myNode and Start9

**Umbrel:**
Screenshot 2023-03-23 at 02 02 46

Changes in connectors:
- **Lightning Node** should lead to what is currently under Umbrel connector 
- **Lightning Terminal**  (LNC) should lead to what is currently "LNC with LND" 

**Citadel:**
Screenshot 2023-03-23 at 02 04 27
- **Citadel Login** should lead what is currently Citadel
- the rest can lead to standard LNC, CLN and lnbits connectors.

**BTCPay:**
Screenshot 2023-03-23 at 02 05 20

- **LND** should lead to what is currently BTCPay
- **Core Lightning** should lead to currently missing https://github.com/getAlby/lightning-browser-extension/issues/2203

**RaspiBlitz, myNode, start9**
Screenshot 2023-03-23 at 02 04 59

- LND on those screens should lead to what is currently under RaspiBlitz, myNode or Start9 connectors
- the rest can lead to standard LNC, CLN, lnbits

New images:
image

image
 
image
(Updated Core Lightning logo)

lujakob:

Add intermediary step for connectors that allow multiple apps/nodes (e.g. Umbrel)

Current state is "draft"

Fixes #2254

### Type of change

umbrel

### How has this been tested?

Manually

### Checklist


lujakob:
@reneaaron you could have a look and test if you want. So far I only implemented the UI. I did not change any logic that happens after "submitting" a connector form. So submitting a LND connector form directly and submitting one for Umbrel does the same thing, as nothing was mentioned in the ticket. Is this correct.

Also the second BTCPay is missing #2203 - but I asked for getting assigned the ticket and can add it. Should I do this in this PR or separate and then include here afterwards?

reneaaron:
@lujakob Awesome, I just tested it. 👌 Some things i noticed: 

 - [ ] LNC Logo needs to be updated, too
image

 - [ ] Center connectors
image

> Also the second BTCPay is missing https://github.com/getAlby/lightning-browser-extension/issues/2203 - but I asked for getting assigned the ticket and can add it. Should I do this in this PR or separate and then include here afterwards?

I'd prefer to do them separately since the changes are unrelated.

lujakob:
> @lujakob Awesome, I just tested it. 👌 Some things i noticed:
> 
> * [ ]  LNC Logo needs to be updated, too
>   image
> * [ ]  Center connectors
>   image
> 
> > Also the second BTCPay is missing #2203 - but I asked for getting assigned the ticket and can add it. Should I do this in this PR or separate and then include here afterwards?
> 
> I'd prefer to do them separately since the changes are unrelated.

I updated the LNC logo. Centering the connector elements for the grid layout with `justify-content: center` had no effect. TBH I'm not sure if it makes sense anyways. In case there is only one row, yes. But if we have more than one rows it might look kinda weird. We didn't center the elements of the last row in the connector overview as well so far.

lujakob:
@rolznz I have the problem of centering the connector elements. As I saw in the Git history you made the first couple of commits on the `ChooseConnectorPath` component I was wondering if you might know how to center with CSS grid. My google search attempts did not work out. If not I'll try something else. Maybe it even makes sense to replace grid with flexbox?

lujakob:
@reneaaron 

* I refactored to have connectorRoutes the "children" way, as discussed. Still I needed to abstract the children into distinct arrays, as they are applied in two locations each. (They are also passed to the `<ChooseConnector>` component)
* I could not resolve the centering of grid elements. Let's way for @rolznz 
* BTCPayserver is reverted to as before as discussed. Providing a CLN connector option is of low prio and can be a follow up

rolznz:
> @rolznz I have the problem of centering the connector elements. As I saw in the Git history you made the first couple of commits on the `ChooseConnectorPath` component I was wondering if you might know how to center with CSS grid. My google search attempts did not work out. If not I'll try something else. Maybe it even makes sense to replace grid with flexbox?

@lujakob I see, the grid is set to 5 columns so there is an invisible column on the right hand side if there are less than that many columns. I would use standard flex layout with flex-wrap rather than grid, unless there is a good reason to use grid?

image

lujakob:
> > @rolznz I have the problem of centering the connector elements. As I saw in the Git history you made the first couple of commits on the `ChooseConnectorPath` component I was wondering if you might know how to center with CSS grid. My google search attempts did not work out. If not I'll try something else. Maybe it even makes sense to replace grid with flexbox?
> 
> @lujakob I see, the grid is set to 5 columns so there is an invisible column on the right hand side if there are less than that many columns. I would use standard flex layout with flex-wrap rather than grid, unless there is a good reason to use grid?
> 
> image

I don't see a good reason, but I never used grid before so I don't really know. But if there's no objections (and solution) from anyone I would refactor to flexbox to be able to center.

reneaaron:
@lujakob Also checked back with @stackingsaunter, let's center them! 🙌

lujakob:
@reneaaron @stackingsaunter I replaced the grid styles with flex-box and centered the content. The responsive behaviour switches from 2-col to 3-col to 5 col accordingly.

Bildschirmfoto 2023-04-11 um 10 32 59

lujakob:
@reneaaron I think it's ready for review. By the way, I decided not to abstract the icon and title from the `ConnectorRoute` elements. I have the impression it hardly improves anything.

tailwind.config.js
```diff
@@ -58,6 +58,11 @@ module.exports = {
         "surface-16dp": lighten(surfaceColor, 15),
         "surface-24dp": lighten(surfaceColor, 16),
       },
+      flexBasis: {
```
reneaaron, line 61: Not sure about this implementation. :thinking: Is there maybe a better way to do it with built-in classes?

src/app/router/connectorRoutes.tsx
```diff
@@ -30,12 +33,173 @@ import raspiblitz from "/static/assets/icons/raspiblitz.png";
...
+    logo: lnbits,
+  },
+];
+
+const start9ConnectorRoutes = [
+  {
+    path: "lnd",
+    element: <ConnectStart9 />,
+    title: i18n.t("translation:choose_connector.lnd.title"),
+    logo: lnd,
```
reneaaron, line 169: The definition of different connectors is repeated quite some times. Could we define the connectors and then only assign those values to the different connector routes so we can maintain them in one place?

reneaaron (changes requested):
Thanks @lujakob :raised_hands: feels already really nice and collected some outside feedback. They seem to like it :+1:

tailwind.config.js
lujakob: I could not find a way to do this with the provided Tailwind CSS classes. That's why I went for this option. The problem was, setting the styles with the provided classes, the gap is not being considered and it breaks early having only 4 in a row and some leftover space on the right side. I also found other people asking for this problem on Stackoverflow.

src/app/router/connectorRoutes.tsx
lujakob: I DRYied this up and created a map object to only define each connector object once.

src/i18n/locales/en/translation.json
```diff
@@ -358,6 +361,30 @@
         }
       }
     },
+    "choose_umbrel_connector": {
```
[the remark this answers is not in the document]
lujakob: First I would need to translate a title by the path (e.g. "Umbrel"), then send it as a parameter to the translation for the whole sentence (e.g. "Connect to {title}"). To be able to do this I have to create a function and then overwrite the title like

```
function getDistribution() {
    const name = t()
    return {
        ...connectorMap["umbrel"],
        title: <translate title sentence and pass translated name as param>
    }
```

This would be fine, the problem is I cannot overwrite the `<ChooseConnector>` props

```
  umbrel: {
    path: "umbrel",
    element: (
      <ChooseConnector
        title={i18n.t("translation:choose_umbrel_connector.title")}
        description={i18n.t("translation:choose_umbrel_connector.description")}
        connectorRoutes={umbrelConnectorRoutes}
      />
    ),
    title: i18n.t("translation:choose_connector.umbrel.title"),
    logo: umbrel,
    children: umbrelConnectorRoutes,
  },
```

lujakob:
@reneaaron I updated one as suggested and commented on two.

tailwind.config.js
lujakob: Awesome, thanks for helping out. 👍 I was so focused on the "gap", I forgot about the negative margin trick.

src/i18n/locales/en/translation.json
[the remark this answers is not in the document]
lujakob: @reneaaron What do you think? Is it ok for you?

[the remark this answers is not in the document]
reneaaron: @lujakob I think I'll have to wrap my head around that. Particularly not happy about introducing all of those new translations that are literally the same. (except for the connector name)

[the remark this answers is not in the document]
lujakob: @reneaaron I have an idea now, I think that might work.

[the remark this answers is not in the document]
lujakob: @reneaaron I could improve it by adding a function. Please have a look and let me know what you think. Unfortunately I could not fix the type error for dynamic `i18n` keys and added a `ts-ignore` for now. Maybe you have an idea or I stumble upon something in the meantime.

Also I had to add some interfaces and unions, because the kollider route has a different structure and approach.

Another thing that I noticed, the routing config for the welcome page didn't work. I already noticed this when I started working on this issue but forgot about it. The routing config for Options and Welcome is done differently and expect a slightly different structure for nested routes. One of them had the index page `ChooseConnector` within the children's array and `index` set, the other not.

To avoid further complicating the structure of provided routes of `getConnectorRoutes` I modified the routing strategy in welcome and adapted it to the way it is in options. Let me know what you think.

lujakob:
To finalize this I want to wait for #2309 to be merged to master.

src/app/router/Welcome/Welcome.tsx
```diff
@@ -90,15 +24,74 @@ function WelcomeRouter() {
...
+                      title={i18n.t("translation:choose_connector.title")}
+                      description={i18n.t(
+                        "translation:choose_connector.description"
+                      )}
+                      connectorRoutes={connectorRoutes}
+                    />
+                  }
+                ></Route>
+                {connectorRoutes.map((connectorRoute) => {
+                  if (connectorRoute.children) {
```
rolznz, line 56: Is this if check needed? 
Couldn't you just use `{connectorRoute.children?.map(...)}` and remove the else?

src/app/router/connectorRoutes.tsx
```diff
@@ -30,121 +33,239 @@ import raspiblitz from "/static/assets/icons/raspiblitz.png";
...
+        connectorRoutes={children}
+      />
+    ),
+    title: i18n.t(normalizeKey(`translation:choose_connector.${key}.title`)),
+    logo: distributionMap[key]["logo"],
+    children,
+  };
+}
+
+const distributionMap: { [key: string]: { logo: string } } = {
```
rolznz, line 232: It seems strange here you only store the logo - you could pass the logo directly to the getDistribution function, right? or alternatively you could also store the children in this map and change getDistribution to have a single argument?

rolznz (approved):
tACK with minor feedback. I see there is a conflict, @lujakob can you fix this one? it looks just like the back button was removed (see reasoning in https://github.com/getAlby/lightning-browser-extension/issues/2250)

This is a great improvement! Thanks lujakob!

src/app/router/Welcome/Welcome.tsx
lujakob: It is an if/else condition.

src/app/router/connectorRoutes.tsx
lujakob: I refactored to the latter.

src/app/router/Welcome/Welcome.tsx
rolznz: I mean, can't you replace the condition with just this?
```
                      <Route
                        key={connectorRoute.path}
                        path={connectorRoute.path}
                      >
                        <Route index element={connectorRoute.element} />
                        {connectorRoute.children?.map((connectorRoute) => (
                          <Route
                            key={connectorRoute.path}
                            path={connectorRoute.path}
                            element={connectorRoute.element}
                          />
                        ))}
                      </Route>
```

lujakob: Now I see. I don't know if this works. I took this from Options page. Now that in the kollider PR #2309 routing got changed, I will refactor this anyways once it's in master.

lujakob: I refactored to use the `renderRoutes` introduced by im-aditiya

lujakob:
@reneaaron this is ready for merge now. I integrated the latest changes from master (Kollider changes incl renderRoutes function) so the Options.html and Welcome.html routing is the same now and removed the back button on chooseConnector. Please have another look if you think any changes needed. Otherwise, let's merge it :-)

reneaaron:
Just tested again, neat! :rocket: Thanks @rolznz for your review! 

One little thing that I noticed during testing: 

When I now select a distribution (e.g. myNode) I see this: 

image

When I click on LND the logo is not anymore the LND (but the one from the distribution):

image

Also LNC shows the LND Logo (instead of the lightning terminal logo from the overview)
image

Isn't it a bit strange that the logo changes while navigating to the specific connector form? Could we use the same Logo there? 

Also the LND connector form reads "Connect to your myNode" while that is also true for all other connectors of the distribution? 🤔

Maybe we need to tweak the hierarchy a bit to make this more consistent. Can also be done in a follow-up PR i think.  cc @stackingsaunter

reneaaron (approved):
tACK

stackingsaunter:
Tested and works great, besides stuff that @reneaaron pointed out! I think the header logo (and header name) should generally be the same as the last choice user makes, so in case of myNode -> LND it should have LND logo and "**Connect to LND**" with LND logo. In case of Umbrel -> Lightning Node it should be "Connect to Lightning Node" with Lightning Node logo and LNC should be "Connect to Lightning Terminal (LNC)" with LNC logo, etc. 

Besides that, total fire, thank you greatly for this @lujakob 🙏 🙏

PR #2188, opened 2023-03-09
chore: add tests for keysend and sendpayment

im-adithya:
### Describe the changes you have made in this PR

Adds tests for keysend and sendpayment, including the events involved in the call

### Link this PR to an issue [optional]

Follow-up to #2139 

### Type of change

(Remove other not matching type)

- `chore`

### How has this been tested?

Manually

### Checklist

- [x] My code follows the style guidelines of this project and performed a self-review of my own code
- [x] New and existing tests pass locally with my changes
- [x] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

PR #2276, opened 2023-03-27
Fix/1051 mv3 txs dates

escapedcat:
- image
    [dayjs needs a number not a string](https://day.js.org/docs/en/parse/unix-timestamp-milliseconds)
- undefined host was throwing an error

### DIFF
--- a/src/app/utils/payments.ts
+++ b/src/app/utils/payments.ts
@@ -8,7 +8,7 @@ export const convertPaymentToTransaction = (
   ...payment,
   id: `${payment.id}`,
   type: "sent",
-  date: dayjs(payment.createdAt).fromNow(),
+  date: dayjs(+payment.createdAt).fromNow(),
   title: payment.name || payment.description,
   publisherLink: publisherLink || payment.location,
 });
--- a/src/extension/background-script/events/__test__/persistPayments.test.ts
+++ b/src/extension/background-script/events/__test__/persistPayments.test.ts
@@ -36,6 +36,24 @@ const updatedPayments: DbPayment[] = [
   },
 ];
 
+const updatedPaymentsWithoutOrigin: DbPayment[] = [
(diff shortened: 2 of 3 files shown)


escapedcat:
@lujakob would you mind having a look?

lujakob:
> @lujakob would you mind having a look?

... at anything specific? To me it looks good.

escapedcat:
@im-adithya initally I thought [this was the issue](https://github.com/getAlby/lightning-browser-extension/pull/2188/files#diff-f887d69cb1f2351f8defa916d4e22966b392c8eaf356a70242d274295278a736R38) but indexDB always stored to the date as a string I think. Your change was just switching to TS and making this obvious.
I'm not sure why we didn't have this "338 years ago" issue earlier. Last time we updated dayjs was 3 months ago.

I think we also should reflect the correct types in the code [for these two](https://github.com/getAlby/lightning-browser-extension/blob/master/src/types.ts#L610-L611) because in the indexDB they are also just strings, right?

@lujakob thanks, just wanted you to look at the dayjs update because you also worked on this part before.

im-adithya:
> Your change was just switching to TS

Yes I did that because DbPayment has `createdAt: string`

> but indexDB always stored to the date as a string I think

No, indexDB can store numbers too, I think it's my change that caused this issue. Before my change, although the DbPayment was typed to have createdAt as string, we were passing a number and that was being stored in indexedDB as well.

Screenshot 2023-03-30 at 10 45 06 AM

(createdAt is a number, hence no issue before that change)


> Checking out [the current version](https://github.com/getAlby/lightning-browser-extension/releases/tag/v1.28.1) I also get that issue but on my "prod install" everything is fine.

And #2188 was after the current release so that should definitely be it!

escapedcat:
Thanks for the explanation @im-adithya !

PR #2302, opened 2023-04-04
feat: colors & gradients

reneaaron:

image


src/app/components/Hyperlink/index.tsx
```diff
@@ -1,14 +1,26 @@
...
+  children,
+  href,
+  className,
+}: Props) {
   return (
     <a
-      className="cursor-pointer text-sky-500 hover:text-sky-400"
+      className={classNames(
+        "cursor-pointer text-blue-700 hover:text-blue-500",
+        className ?? ""
```
rolznz, line 21: can't you just do `className`? if it is falsy it should be omitted, right?

rolznz (approved):
utACK, looks great!

src/app/components/Hyperlink/index.tsx
reneaaron: It's an optional param for the component that isn't compatible with the classNames() func, thus the check.

> Argument of type 'string | undefined' is not assignable to parameter of type 'string | boolean'.

PR #2310, opened 2023-04-06
fix: await decrypt

im-adithya:

--

Fixes messages not being decrypted issue in nostr

### Type of change

### How has this been tested?

On astral

### Checklist

### DIFF
--- a/src/extension/background-script/actions/nostr/decryptOrPrompt.ts
+++ b/src/extension/background-script/actions/nostr/decryptOrPrompt.ts
@@ -18,9 +18,11 @@ const decryptOrPrompt = async (message: MessageDecryptGet) => {
     );
 
     if (hasPermission) {
-      const response = await (
-        await state.getState().getNostr()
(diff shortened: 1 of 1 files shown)


src/extension/background-script/actions/nostr/decryptOrPrompt.ts
```diff
@@ -18,10 +18,9 @@ const decryptOrPrompt = async (message: MessageDecryptGet) => {
     );
 
     if (hasPermission) {
-      const response = (await state.getState().getNostr()).decrypt(
-        message.args.peer,
-        message.args.ciphertext
-      );
+      const response = await (
```
bumi, line 21: I think we should do this in two lines

const nostr = await state.getState().getNostr();
and then 
const response = await nostr.decrypt(message.args.peer, message.args.ciphertext);

this makes this easier to read and more obvious. 

and I think it prevents that some await is missed.

bumi (changes requested):

bumi (approved):

PR #2311, opened 2023-04-06
fix: removed line clamp plugin (included by default)

reneaaron:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,6 @@
     "@lightninglabs/lnc-web": "^0.2.2-alpha",
     "@noble/secp256k1": "^1.7.1",
     "@tailwindcss/forms": "^0.5.3",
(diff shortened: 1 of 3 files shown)


im-adithya:
But don't we use `line-clamp` in our files? Is it not related to that? I didn't remove it because of that reason

reneaaron:
Have a look at the screenshot I posted, you'll see the build warning there:

It's included by default (I removed it and the clamping still works).

im-adithya:
> It's included by default (I removed it and the clamping still works).

Aaah 😄 Sorry, I misunderstood it as the plugin is included by default (in addition, in default), so I thought, why are we removing this? So it must mean it is included in the main one by default, without the need to add anything; got it!

im-adithya (approved):

Issue #2290, opened 2023-03-29
[BUG] "Copy Invoice" should be made into more intuitive text


SveeJ:
### Describe the bug

When the scan QR screen opens for receiving sats, there is a QR that can be used to generate an invoice string that can be pasted in the send section of the extension. I think the 'copy' text should be replaced due to the fact that it's unintuitive to the end-user what it is that is being copied. Changing it to something along the lines of: 

Copy Invoice String or Copy Invoice Code or simply Copy Invoice (looking for community feedback) would be most optimal

<img width="383" alt="unintuitive text" src="https://user-images.githubusercontent.com/62150721/228510797-1c4443a4-0fb2-4c96-8d5c-8745e45763dd.png">

### Steps To Reproduce

1. Click on Receive
2. Generate an Invoice
3. Click on Copy

### Expected behavior

(Copy Invoice String/Code)

### Alby information

Alby 1.28.1

escapedcat:
@stackingsaunter wdyt?

stackingsaunter:
Copy Invoice suits the best. Many people don't know what string is and I don't think "code" describes it well, so just Copy Invoice

mohitvdx:
"Copy address"
being more commonly used. what do you think?

stackingsaunter:
"Address" is incorrect, this is not an address. Let's stay with Copy Invoice

SveeJ:
Alright perfect could I be assigned this issue?

PR #2313, opened 2023-04-06
fix : "Copy Invoice" should be made into more intuitive text #2290

Pratik-A-N:

I have updated the copy text in Recieve page to Copy Invoice

Fixes _#2290

### Type of change

Screenshot from 2023-04-07 03-17-24

### How has this been tested?

It can be tested by going in receive page -> Create Invoice -> Copy Invoice 

### Checklist

### DIFF
--- a/src/app/components/AccountMenu/index.test.tsx
+++ b/src/app/components/AccountMenu/index.test.tsx
@@ -68,54 +68,10 @@ describe("AccountMenu", () => {
     });
 
     await screen.findByText("Switch account");
-
+    expect(screen.getByText("Balance")).toBeInTheDocument();
     expect(screen.getByText("LND account")).toBeInTheDocument();
     expect(screen.getByText("Galoy account")).toBeInTheDocument();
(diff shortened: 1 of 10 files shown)


src/app/screens/Receive/index.tsx
```diff
@@ -174,7 +174,7 @@ function Receive() {
                     navigator.clipboard.writeText(invoice.paymentRequest);
                     setCopyLabel(tCommon("copied"));
                     setTimeout(() => {
-                      setCopyLabel(tCommon("actions.copy"));
+                      setCopyLabel(tCommon("actions.copy")+" Invoice");
```
reneaaron, line 177: That doesn't work this way. You'd need to create a new [translation](https://github.com/getAlby/lightning-browser-extension/blob/master/src/i18n/locales/en/translation.json#L1) for that instead of concatenating a string here.

Pratik-A-N: Thank you for guiding me, I will try to do by the way you suggested

Issue #1566, opened 2022-10-05
UI: update tab-style - ON HOLD


escapedcat:
### Feature description

Update tabs-style
![image](https://user-images.githubusercontent.com/1016218/193981295-682c57c4-852c-4e03-8215-4848ed1e0d7f.png)
![image](https://user-images.githubusercontent.com/1016218/193981263-39200ac4-3e9d-45cd-ac86-4461dd89d2fa.png)
![image](https://user-images.githubusercontent.com/1016218/193981310-c2feb0ab-89fe-416a-987c-c3a37d880a71.png)

If you start working on this please coordinate with @dvoroneca regarding which styling is preffered or i.e. easiest to implement.  
We also need to consider dark-mode.

### Describe the solution

We currently use the default tabs with an minimal adjusted style-set:
https://headlessui.com/react/tabs

We only use the tabs here currently:
- dark-mode:  
    https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/index.tsx/#L396-L445
![image](https://user-images.githubusercontent.com/1016218/193980978-7e5f3e61-c994-4365-b40a-3df7ef3869c9.png)
- light-mode:  
    ![image](https://user-images.githubusercontent.com/1016218/193981471-52b28769-0deb-43dc-bc43-03a693158800.png)

(If #1560 is done we will use tabs in two file screens)

### Describe alternatives

_A clear and concise description of any alternative solutions or features you've considered_

### Additional context

- This is a follow-up to #1560
- If #1560 is done, we should also move Tabs in their own component to make it more reusable (like we already do with other components). This can be done in this or in a new issue.

escapedcat:
@Nasimmhn yes! Thanks! Let us know if you need any support or get stuck.

escapedcat:
Hey @Nasimmhn you need any support? Did you already start?

escapedcat:
@Nasimmhn unassining you here to give other people a chance. I guess you're busy with other things. No worries.

whyprax:
Hello @escapedcat, is this issue still open?
If yes, can you please assign it to me?

escapedcat:
@whyprax you got it. Let ue know if you have any issues or get stuck.

whyprax:
Hey @escapedcat, I need your help in setting up the project.

I have followed these steps : 
    Ensure you have
    
        [Node.js](https://nodejs.org/) v14 [14.20.1] installed (We only support v14 currently)
[...]

PR #2322, opened 2023-04-08
Add "on-chain" tab to Receive screens when users use Alby Accounts

### LINKED ISSUE #2240: Add "on-chain" tab to Receive screens when users use Alby Accounts
Many users now [are seeking on-chain topup in the extension, not realising they can log in on getalby.com](https://github.com/getAlby/lightning-browser-extension/issues/2186) (If they're using Alby Account) and top up it there.

We should give some guidance to users that they can do that and the best place is Receive tab, where divide it into 2 tabs, one for Lightning (what we already have) and other screen which will just link to https://getalby.com/node/receive :

Screenshot 2023-03-20 at 12 03 40

However, since we only offer onchain topup for Alby Account, this should be only displayed when users are using Alby Account (with @getalby.com). For all other connectors, nothing should change in Receive tag compared to what we have now.

"Go Alby Account on getalby.com ->" button should lead to https://getalby.com/node/receive

Handoff ([Figma file](https://www.figma.com/file/AMzMvcwrr2IdunqMakaB8r/navigation%2C-dropdown%2C-receive-onchain-work%2C-connectors%2C-nostr-events?node-id=74-8425))
Screenshot 2023-03-20 at 12 04 01

Icons used:
Block
Lightning

HARI-124:

I have added the on-chain tab for alby users, as mentioned in issue #2240 . I created a state to check whether the user is in the lightning or onChain tab and rendered the elements conditionally. I have added functionality to check alby users by comparing their alias with `🐝 getalby.com`. But the text colors are not changing for the button component.

Fixes  #2240 

### Type of change

Screenshot 2023-04-08 185855

https://user-images.githubusercontent.com/97289088/230723820-64631db6-940c-4a58-8faa-b581be44ae0b.mp4

### How has this been tested?

I have viewed the changes in the extension and attached the video.

### Checklist


rolznz:
Hi @HARI-124, thanks for the PR!

rolznz:
I believe there should be no orange outline on the focused tab, nor on the text of the tab. It looks like there might be some other minor styling divergences from the original mockup. To me the text on the buttons looks too big.

image

image

src/app/screens/Receive/index.tsx
```diff
@@ -49,7 +50,8 @@ function Receive() {
   const [paid, setPaid] = useState(false);
   const [pollingForPayment, setPollingForPayment] = useState(false);
   const mounted = useRef(false);
-
+  const [isLightning, setLightning] = useState(true);
+  const isAlbyUser = auth.account?.alias === "🐝 getalby.com";
```
rolznz, line 54: We already have a utility function in the code, can you use that?

src/app/screens/Receive/index.tsx
```diff
@@ -49,7 +50,8 @@ function Receive() {
   const [paid, setPaid] = useState(false);
   const [pollingForPayment, setPollingForPayment] = useState(false);
   const mounted = useRef(false);
-
+  const [isLightning, setLightning] = useState(true);
```
rolznz, line 53: I wonder if it would be easier to understand if this was renamed to receiveType. What do you think?

type ReceiveType = 'onchain' | 'lightning';

src/app/screens/Receive/index.tsx
```diff
@@ -232,47 +234,121 @@ function Receive() {
...
-                    min={0}
-                    label={t("amount.label")}
-                    placeholder={t("amount.placeholder")}
-                    fiatValue={fiatAmount}
-                    onChange={handleChange}
-                    autoFocus
-                  />
-                </div>
 
+      {isAlbyUser ? (
```
rolznz, line 238: In my opinion it would be nice to split these into separate components - ReceiveTypeTabs, LightningReceiveTab and OnchainReceiveTab.

src/app/screens/Receive/index.tsx
```diff
@@ -232,47 +234,121 @@ function Receive() {
...
+        <Container maxWidth="sm">
+          <div className="flex mb-6  mt-4 text-xl ">
+            <Button
```
rolznz, line 241: Rather than using two buttons I think we should use a Tabs component to make sure our styling is consistent throughout the app. There is one usage already of Tab.Group in src/app/screens/Home/DefaultView/index.tsx.

@stackingsaunter could you review this issue?

https://github.com/getAlby/lightning-browser-extension/issues/1566

HARI-124:
@rolznz I have checked the comments I will try to resolve the issues

src/app/screens/Receive/index.tsx
HARI-124: So if the user is an alby user then it should show LightningReciveTab and OnChainRecieveTab else it should show RecieveTypeTabs

rolznz: @HARI-124 RecieveTypeTabs would be the TabGroup which only the Alby user sees. for non-Alby connectors, I think we only need to show the content of the LightningReceiveTab. This should also fix the account switching bug I mentioned.

HARI-124:
https://user-images.githubusercontent.com/97289088/230979817-43de9b23-051c-4f42-a439-5b95e0e48875.mp4
wallet

I have corrected the issues which you mentioned above and made the tabs according to the given figma file. Let me know if there are any further changes.

[https://www.figma.com/file/AMzMvcwrr2IdunqMakaB8r/navigation%2C-dropdown%2C-receive-onchain-work%2C-connectors%2C-nostr-events?node-id=74-8427&t=QoVeTMFe3TGSU4TP-0](url)

src/app/screens/Receive/index.tsx
stackingsaunter: I answered in the issue. Definitely we should use one tabs component. If we have tabs component in the extensions, we should use that. Any change to tabs design like discussed in this issue or the design I made for on-chain tab can be updated separately I guess. So either this PR should update tabs component in general or use current one IMHO

stackingsaunter:
Hi @HARI-124, thank you for your work on this! 

From design side, I see a few places to correct:

- There needs to be spacing between icon and copy. 6px would work best if there's no box around the icon
- Inactive state has different bg than original design. It's fine, I think it's better without the border like in my design, but I think it'd be better if you change it from gray200 to gray100 and then do gray200 as a hover state (which is now missing). See screenshot:

image

- Icons should always have the same color as text, which is not the case now:
Screenshot 2023-04-10 at 23 56 13

- I like to avoid having those hard corners:
image

Is it possible to have it done this way?:
Screenshot 2023-04-11 at 00 00 33

Perhaps the inactive tab can be 38px, active 40px, and the bg of the whole element could be same color as inactive state, so it will create an effect of active state being in front. Perhaps @rolznz can suggest how to make it nicely

image

rolznz:
@HARI-124 it looks like the code has not been extracted yet, and now the alby vs non-alby lightning code is duplicated. Could you please look into this?

HARI-124:
> @HARI-124 it looks like the code has not been extracted yet, and now the alby vs non-alby lightning code is duplicated. Could you please look into this?

I could not get exactly can you please elaborate?

HARI-124:
I have created a new tab component and used for both lighting tab and on-chain tab.

HARI-124:
@rolznz In my understanding, for non-alby users, it should show the receive tab, but if the user is an alby user, it should show both the on-chain tab and receive tab since the receive tab is common for both users I repeated the code two times. Would you like me to create a receive tab component and use it for alby and non-alby users?

src/app/components/Tab/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type Props = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+  active: boolean;
+  icon?: React.ReactNode;
+  leftTab?: boolean;
+  rightTab?: boolean;
+  label?: string;
+};
+
+export default function TabComponent({
```
rolznz, line 12: I would call this component "Tab" and rename the imported one 
`import { Tab as HeadlessUiTab } from ...`

src/app/screens/Receive/index.tsx
```diff
@@ -232,47 +238,196 @@ function Receive() {
...
+                ></TabComponent>
+
+                <TabComponent
+                  active={recieveType === "onchain"}
+                  onClick={() => {
+                    setRecieveType("onchain");
+                  }}
+                  rightTab={true}
+                  icon={
+                    <svg
```
rolznz, line 281: I don't think the svg code should be directly embedded. Can you add them as a file to static/assets/icons?

src/app/components/Tab/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { classNames } from "~/app/utils/index";
+
+type Props = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+  active: boolean;
+  icon?: React.ReactNode;
+  leftTab?: boolean;
+  rightTab?: boolean;
+  label?: string;
+};
+
```
rolznz, line 11: Maybe you could also export the other components from this tile we need for tabs, like Tab.Group and Tab.List, and we can apply any styling here if we need to.

Can you update the code in src/app/screens/Home/DefaultView/index.tsx to use the new tab components too?

src/app/screens/Receive/index.tsx
```diff
@@ -232,47 +238,196 @@ function Receive() {
...
+                      width="24"
+                      height="24"
+                      viewBox="0 0 24 24"
+                      className={
+                        recieveType === "onchain"
+                          ? "text-gray-700 dark:text-neutral-200 "
```
rolznz, line 287: selected tab styling should be done in the Tab component

rolznz:
> @rolznz In my understanding, for non-alby users, it should show the receive tab, but if the user is an alby user, it should show both the on-chain tab and receive tab since the receive tab is common for both users I repeated the code two times. Would you like me to create a receive tab component and use it for alby and non-alby users?

Hi @HARI-124 that's correct. It would be great if the code is not duplicated. Can you make a new component for that? 

Note: it is not a tab itself, but the contents of the tab. For non-alby users there is no tab group, but this content will be displayed.

rolznz:
> I have created a new tab component and used it for both the lighting tab and the on-chain tab. I have made styling according to the figma file. The SVG color is not changing can you help me to resolve that issue? @rolznz

Normally you need to use `"currentColor"` instead of a hardcoded color in the SVG in order to pass color to it. 
image

src/app/components/Tab/index.tsx
HARI-124: sure np

HARI-124:
@rolznz I have made changes as requested, but still svg is not changing color. Since we are importing svg as img element the styles are not working on svg. I cannot import it as a react component as it shows an error.

rolznz:
> @rolznz I have made changes as requested, but still svg is not changing color. Since we are importing svg as img element the styles are not working on svg. I cannot import it as a react component as it shows an error.

It looks like we don't use SVGR. There are a few examples of creating our own components for svg icons. Here is an example: src/app/icons/ConnectedSiteIcon.tsx

Are you able to create a component for the SVG icon like the above?

src/app/components/RecieveTab/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+type Props = React.FormHTMLAttributes<HTMLFormElement> & {
+  handleSubmit?: (event: React.FormEvent<HTMLFormElement>) => void;
+  loading?: boolean;
+  fiatAmount: string;
+  handleChange?: (
+    event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>
+  ) => void;
+};
+
+const RecieveTab = ({
```
rolznz, line 16: This isn't actually a tab. And it's only for lightning, right?
Should it be called `LightningReceiveForm` or something?

src/app/components/Tab/Tab.tsx
```diff
@@ -0,0 +1,41 @@
+import { Tab as HeadlessUiTab } from "@headlessui/react";
+import { classNames } from "~/app/utils/index";
+
+type Props = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+  active?: boolean;
+  icon?: React.ReactNode;
+  leftTab?: boolean;
```
rolznz, line 7: leftTab and rightTab still need to be removed.

src/app/components/Tab/TabList.tsx
```diff
@@ -0,0 +1,14 @@
+import { Tab as HeadlessUiTab } from "@headlessui/react";
+import { classNames } from "~/app/utils/index";
+
+type Props = React.HTMLAttributes<HTMLDivElement> & {
```
rolznz, line 4: This Props accepts all values from React.HTMLAttributes<HTMLDivElement> but it is only passing className to HeadlesssUiTab.List. Shouldn't you pass all the props? otherwise I would change the props just to have a className field and not inherit from the above interface.

HARI-124:
> > @rolznz I have made changes as requested, but still svg is not changing color. Since we are importing svg as img element the styles are not working on svg. I cannot import it as a react component as it shows an error.
> 
> It looks like we don't use SVGR. There are a few examples of creating our own components for svg icons. Here is an example: src/app/icons/ConnectedSiteIcon.tsx
> 
> Are you able to create a component for the SVG icon like the above?

sure I will create a component for svg

src/app/components/Tab/TabList.tsx
HARI-124: I thought in case we need to use any attributes; we could just add it as props.

src/app/components/Tab/Tab.tsx
HARI-124: I also removed active property and used `selected` props from the headlessTab className to have the same effect.

rolznz:
> @rolznz @stackingsaunter I won't be available till May 6 due to my end-semester exams so I will start the work after that

Thanks for your work so far. Good luck with your exams!

rolznz:
@HARI-124 I have picked up where you left off and polished it a bit.

stackingsaunter:
@rolznz in extended view, it's too far stretched

Screenshot 2023-04-25 at 21 49 54

Also, "Account" in the button should be with capitals for consistency with the rest of the extension and how we brand it.

It bothers me a little that the text itself is moving when switching tabs, but that's a detail and I guess to reduce it would be too hacky, so it's fine :P

rolznz:
@stackingsaunter thanks for testing! I've fixed the tab widths. 

I think the text moving in the popup is because the onchain tab has the text vertically centered. I think users won't notice the transition too much :-)

bitcoinuser:
Hi,

I think have two tabs is not good. Even to translate to other languages. Onchain is strange.

In my opinion should just have a message on receive window telling something like: If you want to deposit using a bitcoin address instead of a lightning address, please click here.

bitcoinuser:
> Hi,
> 
> I think have two tabs is not good. Even to translate to other languages. Onchain is strange.
> 
> In my opinion should just have a message on receive window telling something like: If you want to deposit using a bitcoin address instead of a lightning address, please click here.

Or a topup button on main Alby window, in the middle of send and receive buttons.

HARI-124:
any update?

rolznz:
@stackingsaunter @reneaaron could you guys review this and @bitcoinuser 's feedback? In my opinion it would be good to merge since it also has the tab component updates which need to be used in other parts of the app.

stackingsaunter (approved):
Hm, so after and iterating discussions we decided to go in a direction closer to @bitcoinuser vision, something like this:

image

In the meantime, I think this can be merged. As @rolznz said we will use tabs in other places, so thanks a lot for this contribution @HARI-124

stackingsaunter (approved):
Hm, so after and iterating discussions we decided to go in a direction closer to @bitcoinuser vision, something like this:

image

In the meantime, I think this can be merged. As @rolznz said we will use tabs in other places, so thanks a lot for this contribution @HARI-124

stackingsaunter (approved):
Hm, so after and iterating discussions we decided to go in a direction closer to @bitcoinuser vision, something like this:

image

In the meantime, I think this can be merged. As @rolznz said we will use tabs in other places, so thanks a lot for this contribution @HARI-124

rolznz:
@HARI-124 I will close this PR as unfortunately the UX has now completely changed. I have extracted your changes for the tabs into a separate PR. Thank you very much for your work!

PR #2328, opened 2023-04-09
fix: accounts info connector race condition

### LINKED ISSUE #2327: [BUG] development mode Polar LND node connect fails

### Describe the bug

Connecting the extension with a Polar LND node locally causes an error. The `getBalance` request fails with status `ERR_CERT_AUTHORITY_INVALID`

I investigated and found the `getInfo` request fails. It seems like a race condition problem, as the failing `getChannel` is first in the network panel, the succeeding `getInfo` is second.

I removed the Promise.all and it fixes the problem.

### Screenshots [optional]

Bildschirmfoto 2023-04-09 um 09 03 06

### Steps To Reproduce

connect to LND node with Polar locally

### Expected behavior

It should work.

### Alby information

latest master

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

Yes

lujakob:

Fix Polar LND node connection request race condition resulting in connector error

Fixes #2327

### Type of change

### How has this been tested?

Manually

### Checklist

### DIFF
--- a/src/extension/background-script/actions/accounts/info.ts
+++ b/src/extension/background-script/actions/accounts/info.ts
@@ -7,10 +7,9 @@ const info = async (message: MessageAccountInfo) => {
   const currentAccountId = state.getState().currentAccountId;
   const currentAccount = state.getState().getAccount();
 
-  const [info, balance] = await Promise.all([
-    connector.getInfo(),
-    connector.getBalance(),
-  ]);
+  // keep sequence to prevent race conditions in Polar/LND connection
+  const info = await connector.getInfo();
+  const balance = await connector.getBalance();
 
   if (!currentAccount || !currentAccountId) {
     return { error: "No current account set" };


rolznz:
@lujakob I feel like this does not fix the root cause, but I also don't know why there would be a race condition. Isn't this just two standard `fetch` API calls?

lujakob:
> @lujakob I feel like this does not fix the root cause, but I also don't know why there would be a race condition. Isn't this just two standard `fetch` API calls?

They should just work yes. They do if I request them independently via Postman. But I have no clue what the exact issue is. Maybe it for some reason is sending them at the exact same time with `Promise.all` Maybe some Chrome browser internals with local SSL sending two requests at the same time - I have no clue.

lujakob:
@rolznz do you also you Polar to connect to an LND connector? I really wonder why only I have this problem?! Or do you know any devs who use Polar? I would suppose most here...

rolznz:
@lujakob mostly I use the Alby regtest environment (https://github.com/getAlby/lightning-browser-extension/wiki/Test-setup) for testing unless I have a specific need to setup a local network using Polar

bumi (approved):
This loads the data probably a bit slower (no longer in parallel)
but less parallel is probably also good.

reneaaron:
I feel like this just fixes some symptoms (but not the root cause). Why would parallel requests be a problem?

bumi:
then let's leave it and it has to somehow work.

Issue #1891, opened 2022-12-19
[Feature] Use a forms library for the extension forms


rolznz:
### Feature description

Benefits: form validation, saving/restoring form states, more consistency, simpler submission, less code/boilerplate/duplication, and more.

### Describe the solution

Choose a form library such as https://github.com/react-hook-form/react-hook-form and gradually migrate forms to use the form library instead of custom code.

### Are you working on this?

None

escapedcat:
[I'm with you](https://github.com/getAlby/lightning-browser-extension/issues/679#issuecomment-1064745918) ;)  
Even though we do not have lot's of validation (yet... right?) it will align our forms and makes to code-base easier to handle in the future.

PR #2329, opened 2023-04-09
fix: changing logical validations to form data input password and con…

### LINKED ISSUE #2221: [BUG] Click on close x icon on change unlock password, doesn't close the window

### Describe the bug

Click on close x icon on change unlock password, doesn't close the window. See the video:

https://user-images.githubusercontent.com/1178856/225449964-fcee4506-78e0-4f55-a31a-a4b3a0c296d7.mp4

### Screenshots [optional]

_No response_

### Steps To Reproduce

1- Go to change unlock password
2- Click on X close icon

### Expected behavior

Close the window. But now needs too clicks to close the window.

### Alby information

Alby: 1.28.1

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

No

henriq4:

Fixing check triggered when removing focus from input in change unlock password area.

As I commented in Issue #2221, the error is that the issue occurs when the focus is automatically on the input and is removed (such as when closing the modal), triggering error checks. Although this behavior is expected and not a bug, it can still lead to a poor user experience.

in the code, I modified the checks to occur at the correct times.

### Link this PR to an issue

Fixes #2221

### Type of change

https://user-images.githubusercontent.com/62851165/230783413-9e7549f5-c1ea-4d0f-801d-6d0f9cf04ec0.mp4

### How has this been tested?

Manually. No others tests was created.

### Checklist

### DIFF
--- a/src/app/components/PasswordForm/index.tsx
+++ b/src/app/components/PasswordForm/index.tsx
@@ -77,12 +77,20 @@ export default function PasswordForm<
     let passwordErrorMessage: errorMessage = "";
     let passwordConfirmationErrorMessage: errorMessage = "";
 
-    if (!formData.password) passwordErrorMessage = "enter_password";
-    if (confirm && !formData.passwordConfirmation) {
+    if (confirm && !formData.password && formData.passwordConfirmation) {
+      passwordErrorMessage = "enter_password";
+    }
+
+    if (confirm && formData.password && !formData.passwordConfirmation) {
       passwordConfirmationErrorMessage = "confirm_password";
-    } else if (confirm && formData.password !== formData.passwordConfirmation) {
+    } else if (
+      confirm &&
+      formData.passwordConfirmation &&
+      formData.password !== formData.passwordConfirmation
+    ) {
       passwordConfirmationErrorMessage = "mismatched_password";
     }
+
     setErrors({
       passwordErrorMessage,
       passwordConfirmationErrorMessage,


rolznz:
Hi @henriq4, thanks for the PR! 

I think this is better than what it was previously, but still has some issues:
1. If you click on the confirm password input, it shows the error before you even start typing
2. If you click on the disabled "Change" button when the password doesn't match, it shows no error message. Ideally the Change button would always be enabled so that it is clickable and will show an error if the validation fails.

image


I think the best way to fix this is to not add custom logic, but use a well-tested form library that handles the validations for us.  https://github.com/getAlby/lightning-browser-extension/issues/1891

henriq4:
Hey @rolznz! It's my pleasure to be here in the Alby community.

When I started coding, I thought the same thing: the button should be enabled all the time, and the check should only happen when the form is submitted. However, since this did not happen, there was a double check for the same fields in real-time, which caused these bugs.

I agree with you about adding an external library for these situations. I think the best library is [React Hook Form](https://react-hook-form.com/) because of its performance. However, I tried to do it differently because, when searching Alby's code, I realized that they prefer to use their own means for validation. It was my mistake not to have seen the previous issues on the subject.

As a possible solution, I'd be happy to add the library and modify the component to act as an observable React Hook Form component if that's something you would like.

I would also suggest using a library specialized in defining schemas for data validation, such as [Yup](https://github.com/jquense/yup) or [Zod](https://github.com/colinhacks/zod), for these cases.

rolznz:
@henriq4 great, thanks for being willing to work on this! if you do a PR for this, could you link it to the original issue as well and please keep the changes as minimal as possible so it's easier for the Alby team to review and more likely to get merged in.

Let me know if you need any help.

henriq4:
Of course, I'll leave the original issue linked!

I will also be opening a new PR for this feature to separate the discussions.

rolznz:
closing this PR. The password confirmation will be removed with the new Alby OAuth connector (coming soon)

PR #2318, opened 2023-04-07
feat: BalanceBox component implemented

11WALTER11:
### Describe the changes you have made in this PR

_I created a new compoents inside ```src/app/components/BalanceBox``` which adds a new funtionality of displaying users account balance on the home popup screen_

### Link this PR to an issue [optional]

Fixes _#2236_

### Type of change

- `feat`: New feature (non-breaking change which adds functionality)

### Screenshots of the changes [optional]
Proposed Design

### Checklist

- [x] My code follows the style guidelines of this project and performed a self-review of my own code
- [x] New and existing tests pass locally with my changes
- [x] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

Issue #2236, opened 2023-03-18
Implement balance box on home popup screen 


stackingsaunter:
Currently, the only place we show user balance is account dropdown. As user balance is central and essential, we'd like to move it out from the dropdown ([which will be revamped as well](https://github.com/getAlby/lightning-browser-extension/pull/2195)) and place it on the home screen.

Proposed design:
<img width="828" alt="Screenshot 2023-03-18 at 19 04 29" src="https://user-images.githubusercontent.com/68239231/226125932-74e25099-b5d1-4d8a-a111-60bbbeb9719f.png">

Handoff ([Figma file here](https://www.figma.com/file/AMzMvcwrr2IdunqMakaB8r/navigation%2C-dropdown%2C-receive-onchain-work%2C-connectors%2C-nostr-events?node-id=126-22571)): 
<img width="730" alt="Screenshot 2023-03-18 at 19 05 53" src="https://user-images.githubusercontent.com/68239231/226125979-524341b1-cf57-45a2-a9a9-4395d3ec71f1.png">

Note: this issue relates to balance box only, not to the dropdowns on the screenshots (this is another issue: https://github.com/getAlby/lightning-browser-extension/pull/2195#issuecomment-1474933581)

Related issue: https://github.com/getAlby/lightning-browser-extension/issues/1490

Tejasp2003:
i would like to work on this issue

stackingsaunter:
@Tejasp2003 go ahead!

lujakob:
@Tejasp2003 I have your issue as a dependency for the #2243 - so I will reuse the component you are creating. Would be good if you can keep in mind it will be reused in the bigger options screen. You can see the designs attached in my issue.

11WALTER11:
@stackingsaunter can i work on this issue?

stackingsaunter:
@11WALTER11 sure, seems like @Tejasp2003 is not responding so go ahead!

stackingsaunter:
> @stackingsaunter can i get access to the figma file ?

Granted

lujakob:
@11WALTER11 please keep in mind this component will be reused on a normal page for #2243 when coding the HTML and CSS and ideally it would already behave responsive.

ritikraj26:
I am willing to work on this issue. Is it still open?

stackingsaunter:
@ritikraj26 there's already a PR, so I'm closing the issue

reneaaron:
[...]

PR #2332, opened 2023-04-09
feat: add wallet page

### LINKED ISSUE #2243: Merge Send, Receive and Transactions pages into "Wallet"
Currently, in extended view of the extension we have 3 seperate pages for functionalities that are accessible from the main popup screen:
Screenshot 2023-03-20 at 13 32 01

We'd like to merge them into one Wallet page:
Screenshot 2023-03-20 at 13 33 55

"Receive" and "Send" pages would be accesible from respectively  Receive & Send buttons from this Wallet page. "See all transactions" should open to the page that is currently under "Transactions".

This page includes new balance box that has to be implemented first: https://github.com/getAlby/lightning-browser-extension/issues/2236

Handoff ([Figma file](https://www.figma.com/file/AMzMvcwrr2IdunqMakaB8r/navigation%2C-dropdown%2C-receive-onchain-work%2C-connectors%2C-nostr-events?node-id=162-26570)):
Screenshot 2023-03-20 at 13 45 39

lujakob:

Add wallet page and update menu accordingly. Combine "receive", "send" and "transactions" into one single "wallet" page.

Fixes #2243

### Type of change

Bildschirmfoto 2023-04-09 um 14 00 57

### How has this been tested?

Manually

### Checklist


lujakob:
This is a draft to be able to have a first look.

I copied the `<BalanceBox>` from [PR](https://github.com/getAlby/lightning-browser-extension/pull/2318) to be able to test it. As soon as this PR is merged to master I will overwrite the component.

The "see all transactions" link is not implemented yet. It is not available on transactions page yet either. I think this should be done in a follow up PR. (actually there are issues already regarding this)

bitcoinuser:
Hi,

If in all Alby places the name is "Account", why choose "Wallet" word? Thanks.

reneaaron:
@lujakob I just merged #2236 :raised_hands: Feel free to continue with this one :wink:

lujakob:
> @lujakob I just merged #2236 🙌 Feel free to continue with this one 😉

Yesss, awesome. I just noticed and am back on it 😀

lujakob:
@reneaaron I think this is ready for review as far as I can see now. Please have a look. 

I did not change the active state color of the tabs header links "Outgoing" / "Incoming" as seen on the design, because I didn't know which color/class it is and I didn't find an example in the UI.

I don't have transactions in my Alby testnet accounts. Sending/receiving didn't work. (I wrote you a Telegram message)

lujakob:
@reneaaron I tested incoming/outgoing transactions with different lnbits accounts now and it works fine. Both are filtered by account now. From my point of view only the tab menu item active state color is missing now.

reneaaron:
Awesome, thanks very much for your PR! Just did some first round of testing and worked great so far. :+1: 

> I did not change the active state color of the tabs header links "Outgoing" / "Incoming" as seen on the design, because I didn't know which color/class it is and I didn't find an example in the UI.

I think that's fine for now. We're currently working on new tab styles in #2240.

I'd prefer to keep the balance box on top of the buttons to better align this view with the popup. Possibly we need to limit the button size (they seem very huge to me on desktop), but we can do that in a followup PR. cc @stackingsaunter 
 
image

You now removed the Transactions page. Wasn't the idea to link to this page with a `See all` link where we could display more of them (or at one day offer pagination / incremental loading)?

lujakob:
@reneaaron 

> I'd prefer to keep the balance box on top of the buttons to better align this view with the popup. Possibly we need to limit the button size (they seem very huge to me on desktop), but we can do that in a followup PR. cc @stackingsaunter

I moved the balance box on top.

> You now removed the Transactions page. Wasn't the idea to link to this page with a `See all` link where we could display more of them (or at one day offer pagination / incremental loading)?

True. I didn't have this in mind anymore. I reverted the Transactions page component. In the wallet page the list is limited to 5 items. "See all transactions" navigates to `/transactions/incoming` or `/transactions/outgoing` according to the selected tab. I abstracted the fetching of transactions and invoices into hooks, as they are used in both wallet and transactions. The components look a lot cleaner now as well.

src/app/router/Options/Options.tsx
```diff
@@ -55,7 +56,17 @@ function Options() {
             <Route path="confirmPayment" element={<ConfirmPayment />} />
             <Route path="keysend" element={<Keysend />} />
             <Route path="receive" element={<Receive />} />
-            <Route path="transactions" element={<Transactions />} />
+            <Route path="wallet" element={<Wallet />} />
+            <Route path="transactions">
```
lujakob, line 60: @reneaaron I tried to redirect the page to the first child in case of a user typing in the "/transactions" url manually to prevent showing a white page. I did not succeed. Maybe you have an idea.

src/app/screens/Transactions/index.tsx
```diff
@@ -55,17 +49,17 @@ function Transactions() {
...
-      {transactionsLoading ? (
+      {isLoading ? (
         <div className="flex justify-center mt-12">
           <Loading />
         </div>
       ) : (
         <div>
-          {transactions?.length > 0 && (
-            <TransactionsTable transactions={transactions} />
+          {Array.isArray(list) && list.length > 0 && (
```
lujakob, line 61: fyi: I had to add Array.isArray for type reasons here.

src/extension/background-script/actions/accounts/info.ts
```diff
@@ -7,10 +7,8 @@ const info = async (message: MessageAccountInfo) => {
   const currentAccountId = state.getState().currentAccountId;
   const currentAccount = state.getState().getAccount();
 
-  const [info, balance] = await Promise.all([
-    connector.getInfo(),
-    connector.getBalance(),
-  ]);
+  const info = await connector.getInfo();
```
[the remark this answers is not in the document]
lujakob: This is a debugging leftover. Good catch. I will revert it.

reneaaron (changes requested):
I think we should merge the `Wallet` and the  `DefaultView` into one screen as there is quite some code duplication which already produces bugs (eg the one page displays outgoing transactions from all accounts, while the other does only display transactions from the current account, tab styles don't match, etc). 

Clarified with @stackingsaunter to remove the "Wallet" heading and add the links "See all transactions" to the `DefaultView`. 

I think the safest way forward is to remove the new Wallet page (but keep the new hooks for loading transactions) and change the existing `DefaultView` to use the new logic and display the "see all" links.

Does that make sense to you?

reneaaron:
Yes, I think that's the idea. Very much like e.g. we also don't have 2 different receive screens for the popup / full view, they are just used on those different pages.

lujakob:
> Yes, I think that's the idea. Very much like e.g. we also don't have 2 different receive screens for the popup / full view, they are just used on those different pages.

I refactored the `DefaultView` component accordingly and removed the `Wallet` component. Please have another look.

FYI: I added the `isOptionsPage` prop to be able to show the "Wallet" title on the options page. Initially I used a Tailwind CSS condition, then I noticed it doesn't work, because we always want to show the title on the options page, also for narrow screens.

reneaaron:
Thanks very much for the update, that is a lot better now already. :+1: 

However, I think we should get rid of most (if not all) differences between the two different views. (they add quite some complexity for no good reason IMO)  

I would: 
 - Remove the heading "Wallet" (also for the options page) 
 - Have the same page size (10 is okay, also for the popup view, everything > 4 will be below the fold anyway and I don't think that's a problem) 
 - ~~Open the "see all transaction" links in the same context (don't open a new options page if you click them in the popup)~~ I see the problem now, that would require us to handle those pages differently too, as they would need to have back buttons in the popup. Let's skip that for now and open the options page like you implemented it. :+1: 

Some UI related things I noticed: 
 - Do we want to add the wallet link to the dropdown here? It's just opening the same page really. If we do, I think we should add a proper wallet icon
image
 - The "see all transactions" links appear visually too separated to me (the tabs are closer to the transactions as the links in the bottom which makes me think this is an unrelated link)
image
 - The buttons are imo too huge (never seen this big buttons anywhere before 😅)
 
 I'll check back with @stackingsaunter what we could do regarding the UI related things.

lujakob:
> * Remove the heading "Wallet" (also for the options page)

I removed it

> * Have the same page size (10 is okay, also for the popup view, everything > 4 will be below the fold anyway and I don't think that's a problem)

I don't understand what you mean by page size? I actually understand nothing of this sentence 😅
> * ~Open the "see all transaction" links in the same context (don't open a new options page if you click them in the popup)~ I see the problem now, that would require us to handle those pages differently too, as they would need to have back buttons in the popup. Let's skip that for now and open the options page like you implemented it. 👍
> 
> Some UI related things I noticed:
> 
> * Do we want to add the wallet link to the dropdown here? It's just opening the same page really. If we do, I think we should add a proper wallet icon

I removed it. I think Saunter also mentioned the same thing a while ago.

> * The "see all transactions" links appear visually too separated to me (the tabs are closer to the transactions as the links in the bottom which makes me think this is an unrelated link)
I adjusted the distance
> * The buttons are imo too huge (never seen this big buttons anywhere before 😅)
> 
> I'll check back with @stackingsaunter what we could do regarding the UI related things.

Yes let me know about it

lujakob:
@reneaaron @stackingsaunter any updates on this one? How would you like to modify the buttons?

src/app/router/Options/Options.tsx
reneaaron: I don't think we need to do that.

reneaaron:
Fixed some minor things and tested everything. Thanks very much @lujakob for the PR, great work! :muscle: 

tACK

lujakob:
@reneaaron thanks for testing and doing the final updates. I merged master.

reneaaron (approved):

lujakob:
@reneaaron can we merge this? I have two dependency tickets waiting for this. Thx.

PR #2338, opened 2023-04-11
feat: mnemonic

reneaaron:

 - Have one mnemonic to base other keys on (bitcoin, nostr, liquid, etc)
 - Signing functionality via `window.webbtc`
   - signPsbt(psbt)
   - signSchnorr()?
   
   
**Resources**
 - https://iancoleman.io/bip39/
 - https://github.com/satoshilabs/slips/blob/master/slip-0132.md
 - `signPsbt()` [demo](https://codepen.io/reneaaron/pen/JjmPGoj)

Screenshots (18/05/23)

image

image

image

image

image

image

image

image

image

image


src/extension/content-script/onendwebbtc.js
```diff
@@ -0,0 +1,105 @@
...
+
+async function init() {
+  const inject = await shouldInject();
+  if (!inject) {
+    return;
+  }
+
+  const SCOPE = "webbtc";
+
+  // message listener to listen to inpage webln calls
```
lujakob, line 24: typo: "to inpage webln calls" -> "to inpage webbtc calls"

src/app/screens/Accounts/BackupSecretKey/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+            {t("backup.description1")}
+          </p>
+          <div className="flex flex-col gap-4">
+            <ProtocolListItem
+              icon={
+                <NostrIcon className="text-gray-500 dark:text-neutral-500" />
+              }
+              title={t("backup.protocols.nostr")}
+            />
+            <ProtocolListItem
```
rolznz, line 110: @im-adithya we could add another one 

```
<ProtocolListItem
              icon={
                <LiquidIcon className="text-gray-500 dark:text-neutral-500" />
              }
              title={t("backup.protocols.liquid")}
            />
```

src/types.ts
```diff
@@ -17,6 +17,7 @@ export interface Account {
   config: string;
   name: string;
   nostrPrivateKey?: string | null;
+  mnemonic?: string | null;
```
rolznz, line 20: changePassword is not re-encrypting the mnemnoic

reneaaron: I've added that. :heavy_check_mark:

MoritzKa:
Awesome work everybody!!!

Two times I got: 
`Error: Could not establish connection. Receiving end does not exist.`
when I saved the mnemonic key. But I cannot reproduce it.

MoritzKa (approved):
- npub and nsec are reproducible
- a set Nostr key will be overwritten from a new Secret Key
Do we know how the automated update works of the extension after a release? E.g. I derived my key from an Alby account. This is not possible anymore. Will it remain in the extension after an automatic update to mnemonic seed?

kiwiidb (commented):
Tested with:
- generated mnemonic, imported it again and verified the nostr key is identical
- window.webbtc.signPsbt functionality on the codepen


One thing that I find an issue is that the nostr key and the btc key are different.
Deezy's product is built by using the nostr key as a bitcoin key. It seems like we now want to go to the ordinal marketplaces and add our `getAddress` / `signPsbt` functionality to them so people can use Alby to trade their ordinals there, am I correct in this? But the people that will come to us through Deezy will have their ordinals on their nostr key, so if they would login to these marketplaces, they won't be compatible?

MoritzKa:
> But the people that will come to us through Deezy will have their ordinals on their nostr key, so if they would login to these marketplaces, they won't be compatible?
Yeah. But they aren't compatible today either and people cannot mint with their Nostr key and use another ordinals wallet than NosFT. At least not out of the box. 
from a conversation with Danny: 
Moritz Kaminski: Can I convert my Nostr private key to a bitcoin private key to use it in a bitcoin wallet once they are ready for ordinals and inscriptions?
Danny D: you technically can. but there's one caveat that makes it difficult. the way i generated the taproot bitcoin address from the nostr pubkey uses a "pay to pubkey" script type, but the recommended way to do taproot addresses is a "tap tweak key path spend" as that allows for more privacy. i'm not fully sure what the implications are here

rolznz:
> * a set Nostr key will be overwritten from a new Secret Key
>   Do we know how the automated update works of the extension after a release? E.g. I derived my key from an Alby account. This is not possible anymore. Will it remain in the extension after an automatic update to mnemonic seed?

@MoritzKa it will never overwrite your existing nostr key. It will only generate a new nostr key if you do not have one set in your account.

If you have an existing key, it will show as `imported` in your account settings.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "ESNext",
```
reneaaron, line 3: Why do we need to change that? 

https://www.typescriptlang.org/tsconfig#target

> The special ESNext value refers to the highest version your version of TypeScript supports. This setting should be used with caution, since it doesn’t mean the same thing between different TypeScript versions and can make upgrades less predictable.

reneaaron:
Nostr event signing currently doesn't work on snort when using a mnemonic :thinking: 

image

src/app/screens/Accounts/BackupSecretKey/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+              title={t("backup.protocols.ordinals")}
+            />
+          </div>
+
+          <p className="mb-8 text-gray-500 dark:text-neutral-500">
+            {t("backup.description2")}
+          </p>
+          <MnemonicInputs mnemonic={mnemonic} disabled>
+            <>
+              {/* TODO: consider making CopyButton component */}
```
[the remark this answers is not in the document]
rolznz: @reneaaron no - `InputCopyButton` is specifically for putting inside an input field as an `endAdornment`

src/app/screens/Accounts/NostrAdvancedSettings/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        if (priv) {
+          const keyOrigin = await getNostrKeyOrigin(priv, accountMnemonic);
+          setNostrKeyOrigin(keyOrigin);
+        }
+      }
+    } catch (e) {
+      console.error(e);
+      if (e instanceof Error) toast.error(`Error: ${e.message}`);
```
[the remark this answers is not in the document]
rolznz: There shouldn't be any - this part should consistently work or there is something quite wrong and the page would be in an invalid state. I will remove the try-catch blocks

src/common/lib/btc.ts
```diff
@@ -0,0 +1,58 @@
+import * as btc from "@scure/btc-signer";
+
+export function getTaprootAddressFromPrivateKey(
+  privateKey: string,
+  networkType?: keyof typeof networks
+) {
+  const network = networkType ? networks[networkType] : undefined;
+
+  const address = btc.getAddress("tr", Buffer.from(privateKey, "hex"), network);
+  if (!address) {
```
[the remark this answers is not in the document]
rolznz: The btc.getAddress can return undefined but it should never happen since we have a valid private key from the mnemonic.

tsconfig.json
rolznz: @reneaaron thanks for catching this. I think this is another change related to getting signPsbt working. Reverted.

rolznz:
> Nostr event signing currently doesn't work on snort when using a mnemonic 🤔
> 
> image

@reneaaron that looks like a snort error. Did you log out and re-login with the new account before you posted?

I can sign events fine using my generated key. Note: I tested on iris.to because snort.social is often unresponsive for me

src/app/screens/Accounts/BackupSecretKey/index.tsx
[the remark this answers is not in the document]
rolznz: @reneaaron I think we should just remove this button anyway like Bumi suggested - see https://github.com/getAlby/lightning-browser-extension/pull/2490

src/extension/background-script/actions/webbtc/index.ts
```diff
@@ -0,0 +1,6 @@
+import getAddress from "~/extension/background-script/actions/webbtc/getAddress";
+import getAddressWithPrompt from "~/extension/background-script/actions/webbtc/getAddressWithPrompt";
```
bumi, line 2: why is this using a different import path than with `./getinfo`?

src/extension/background-script/actions/webbtc/getAddress.ts
```diff
@@ -0,0 +1,60 @@
...
+      throw new Error("No password set");
+    }
+    const account = await state.getState().getAccount();
+    if (!account) {
+      throw new Error("No account selected");
+    }
+    if (!account.mnemonic) {
+      throw new Error("No mnemonic set");
+    }
+    const mnemonic = decryptData(account.mnemonic, password);
```
bumi, line 26: maybe we do this similar and move this to the `state` that the handling of the decryption is only done in one place? I think with the nostr key we do it similarly.

src/extension/background-script/actions/webbtc/getAddress.ts
```diff
@@ -0,0 +1,60 @@
...
+    const settings = state.getState().settings;
+
+    const derivationPath =
+      settings.bitcoinNetwork === "bitcoin"
```
bumi, line 30: wondering if this should be an account setting and not a global setting?

src/extension/background-script/actions/webbtc/getAddress.ts
```diff
@@ -0,0 +1,60 @@
...
+        ? BTC_TAPROOT_DERIVATION_PATH
+        : BTC_TAPROOT_DERIVATION_PATH_REGTEST;
+
+    const privateKey = derivePrivateKey(mnemonic, derivationPath);
+    const publicKey = derivePublicKey(mnemonic, derivationPath);
+
+    const address = getTaprootAddressFromPrivateKey(
```
bumi, line 37: I have a bit of an object oriented programming mind...
so maybe we can have some key object that we load from the state which we then can ask for derived keys?
then we don't need to handle decryption of the mnemonic in the various places and also don't need to pass in the mnemonic to those functions. 

e.g. 

```
const key = state.getState().getAccountKey(); 
const derivedKey = key.deriveKey(derivationPath);
derivedKey.publicKey;
```

those methods here are also rather similar: https://github.com/getAlby/lightning-browser-extension/pull/2338/files#diff-7d7f5ebb783c099012b8a0f530ba786ceae995894f55dba8b57f0a6834592a9dR13-R30

and once we have the key derived we can get the public key from that without the need to fully derive it again?

src/app/screens/Accounts/NostrAdvancedSettings/index.tsx
```diff
@@ -0,0 +1,264 @@
...
+      throw new Error("No id set");
+    }
+
+    if (!mnemonic) {
+      throw new Error("No mnemonic exists");
+    }
+
+    const nostrPrivateKey = await deriveNostrPrivateKey(mnemonic);
+
+    await handleSaveNostrPrivateKey(nostrPrivateKey);
```
bumi, line 103: wondering if this should be a background action?

src/app/screens/Accounts/ImportSecretKey/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+          <MnemonicInputs mnemonic={mnemonic} setMnemonic={setMnemonic}>
+            <>
+              <Button
+                outline
+                icon={<CopyIcon className="w-6 h-6 mr-2" />}
+                label={importPasteLabel}
+                onClick={async () => {
+                  try {
+                    setMnemonic(await navigator.clipboard.readText());
```
bumi, line 118:
```suggestion
                    const clipboardText = await navigator.clipboard.readText();
                    setMnemonic(clipboardText);
```

src/app/screens/Accounts/ImportSecretKey/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+        if (priv) {
+          setCurrentPrivateKey(priv);
+        }
+
+        const accountMnemonic = (await msg.request("getMnemonic", {
+          id,
+        })) as string;
+        if (accountMnemonic) {
+          setHasMnemonic(true);
+        }
```
bumi, line 50: can these calls somehow be combined? to get one state from the background?

src/app/screens/Accounts/ImportSecretKey/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+    try {
+      if (id) {
+        const priv = (await msg.request("nostr/getPrivateKey", {
+          id,
+        })) as string;
+        if (priv) {
+          setCurrentPrivateKey(priv);
+        }
+
+        const accountMnemonic = (await msg.request("getMnemonic", {
```
bumi, line 45: I am scared of those awaits in brackets.  :)

src/app/components/MnemonicInputs/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <div className="border-[1px] border-gray-200 rounded-lg py-8 px-4 flex flex-col gap-8 items-center justify-center w-[520px] self-center">
+      <h3 className="font-semibold dark:text-white">{t("inputs.title")}</h3>
+      <div className="flex flex-wrap gap-4 justify-center items-center">
+        {[...new Array(12)].map((_, i) => (
+          <div key={i} className="flex justify-center items-center">
+            <span className="w-7 text-gray-500 slashed-zero dark:text-neutral-500">
+              {i + 1}.
+            </span>
+            <Input
```
bumi, line 35: we should use a `div` for sensitive information

src/app/styles/index.css
```diff
@@ -60,6 +59,10 @@ input.dual-currency-field::-webkit-inner-spin-button {
 }
 
 /* Firefox */
-input.dual-currency-field[type=number] {
+input.dual-currency-field[type="number"] {
   -moz-appearance: textfield;
 }
+
+input[type="text"]::-webkit-calendar-picker-indicator {
```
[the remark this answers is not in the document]
rolznz: Sorry, I forgot to add a comment. Removes an arrow on the datalist of word suggestions when you are typing a word in the import secret key page. https://stackoverflow.com/questions/20937475/remove-datalist-dropdown-arrow-in-chrome

src/app/utils/getNostrKeyOrigin.ts
```diff
@@ -0,0 +1,29 @@
...
+  | "legacy-account-derived"
+  | "unknown"
+  | "secret-key";
+
+export async function getNostrKeyOrigin(
+  nostrPrivateKey: string,
+  mnemonic: string | null
+): Promise<NostrKeyOrigin> {
+  if (mnemonic) {
+    const mnemonicDerivedPrivateKey = await deriveNostrPrivateKey(mnemonic);
```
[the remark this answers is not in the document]
rolznz: @bumi what do you mean here? in getNostrKeyOrigin it is trying to find where your nostr key was generated from. For example, your nostr key might not be generated from your secret key if you already have a nostr key before you generate your secret key.

src/app/utils/saveMnemonic.ts
```diff
@@ -0,0 +1,20 @@
...
+    id: accountId,
+  })) as string;
+  const hasNostrPrivateKey = !!priv;
+
+  await msg.request("setMnemonic", {
+    id: accountId,
+    mnemonic,
+  });
+
+  if (!hasNostrPrivateKey) {
```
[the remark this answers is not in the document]
rolznz: Maybe, but then we need more UI to guide the user to generate their nostr key after setting their secret key.

I think it is simpler for the user as it is currently. Is it worthwhile to make this change?

src/extension/content-script/onstart.ts
```diff
@@ -13,6 +13,10 @@ async function onstart() {
   // window.webln
   injectScript(browser.runtime.getURL("js/inpageScriptWebLN.bundle.js"));
 
+  // window.webbtc
+  // TODO: Add check if current account has keys
+  injectScript(browser.runtime.getURL("js/inpageScriptWebBTC.bundle.js"));
```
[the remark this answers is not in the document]
rolznz: @reneaaron I think we should just remove this? otherwise the user would need to restart the extension after generating a secret key?

src/extension/background-script/actions/webbtc/index.ts
rolznz: I made them all relative for now. I think actually it would be good to do the opposite though and use an eslint rule to have consistent absolute imports, and add extra aliases in tsconfig so these imports aren't so long. Maybe "@background-script/actions/webbtc/getAddressWithPrompt"

src/extension/background-script/actions/webbtc/getAddress.ts
rolznz: It makes sense to me as an account setting. @reneaaron do you have any thoughts here? I think we just did it this way to keep it as simple as possible while still being able to change it without having to build the extension locally (e.g. with an environment variable)

src/extension/background-script/actions/webbtc/getAddress.ts
rolznz: Edit: I see your suggestion further down with `state.getState().getAccountKey();`

src/extension/background-script/actions/webbtc/getAddress.ts
rolznz: @bumi sounds good to me. I will create a separate PR for this change. Maybe we should call this method `getSecretKey()` to maintain the same terminology?

src/app/screens/Accounts/ImportSecretKey/index.tsx
rolznz: I am not sure. The nostr key is not necessarily related to the mnemonic if it is imported. Would we add a new method like getKeys() or something which returns all the different private and public keys as well as the mnemonic?

src/app/components/MnemonicInputs/index.tsx
rolznz: @bumi I don't understand what you mean here. Note: the security PR has been merged now and this input will be of type "password".

src/app/screens/Accounts/ImportSecretKey/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+      if (!account || !id) {
+        // type guard
+        throw new Error("No account available");
+      }
+
+      if (
+        mnemonic.split(" ").length !== 12 ||
+        !bip39.validateMnemonic(mnemonic, wordlist)
+      ) {
+        console.error("Invalid mnemonic: '" + mnemonic + "'");
```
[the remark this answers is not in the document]
rolznz: Good point, thanks. Removed

src/extension/background-script/actions/webbtc/index.ts
rolznz: I will create a separate issue for this

src/app/utils/getNostrKeyOrigin.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    if (mnemonicDerivedPrivateKey === nostrPrivateKey) {
+      return "secret-key";
+    }
+  }
+
+  // TODO: consider removing this at some point and just returning "unknown"
+  const legacyAccountDerivedPrivateKey = (
+    await msg.request("nostr/generatePrivateKey")
+  ).privateKey;
```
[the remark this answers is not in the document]
rolznz: True. Fixed

src/app/components/MnemonicInputs/index.tsx
reneaaron: Add default password field visibility switch

reneaaron: Remove past button, but leave paste logic in the form.

src/app/components/MnemonicInputs/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+              type={revealedIndex === i ? "text" : "password"}
+              required
+              onFocus={() => setRevealedIndex(i)}
+              onBlur={() => setRevealedIndex(undefined)}
+              //disabled={disabled}
+              contentEditable={!disabled}
```
reneaaron, line 45: readonly?

src/app/components/MnemonicInputs/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const words = mnemonic?.split(" ") || [];
+  while (words.length < 12) {
+    words.push("");
+  }
+
+  return (
+    <div className="border-[1px] border-gray-200 rounded-lg py-8 px-4 flex flex-col gap-8 items-center justify-center w-[520px] self-center">
+      <h3 className="font-semibold dark:text-white">{t("inputs.title")}</h3>
+      <div className="flex flex-wrap gap-4 justify-center items-center">
+        {[...new Array(12)].map((_, i) => (
```
reneaaron, line 34: use words array

src/app/components/MnemonicInputs/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <Input
+              type={revealedIndex === i ? "text" : "password"}
+              required
+              onFocus={() => setRevealedIndex(i)}
+              onBlur={() => setRevealedIndex(undefined)}
+              //disabled={disabled}
+              contentEditable={!disabled}
+              block={false}
+              className="w-24 text-center"
+              autoComplete=""
```
reneaaron, line 48: do we really need this property to prevent password managers from saving / suggesting something? also test with firefox.

src/app/screens/Accounts/BackupSecretKey/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const [hasMnemonic, setHasMnemonic] = useState(false);
+  // TODO: useNostrPrivateKey hook
+  const [currentPrivateKey, setCurrentPrivateKey] = useState("");
+
+  const { id } = useParams();
+
+  const fetchData = useCallback(async () => {
+    try {
+      if (id) {
+        const priv = (await msg.request("nostr/getPrivateKey", {
```
bumi, line 36: use accounts/get to check if the nostr key is set.

src/app/screens/Accounts/BackupSecretKey/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+  // TODO: useMnemonic hook
+  const [hasMnemonic, setHasMnemonic] = useState(false);
+  // TODO: useNostrPrivateKey hook
+  const [currentPrivateKey, setCurrentPrivateKey] = useState("");
+
+  const { id } = useParams();
+
+  const fetchData = useCallback(async () => {
+    try {
+      if (id) {
```
rolznz, line 35: remove this check (never needed)

src/app/screens/Accounts/BackupSecretKey/index.tsx
```diff
@@ -0,0 +1,162 @@
+import Container from "@components/Container";
```
rolznz, line 1: Split this page into generate + backup

src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -94,6 +87,17 @@ function AccountDetail() {
         if (priv) {
           setCurrentPrivateKey(priv);
         }
+
+        const accountMnemonic = (await msg.request("getMnemonic", {
```
rolznz, line 91: Do not fetch the mnemonic. The account should have a hasMnemonic field in it.

src/app/screens/Settings/index.tsx
```diff
@@ -369,6 +369,40 @@ function Settings() {
...
+            <div className="w-64">
+              <Select
+                name="network"
+                value={settings.bitcoinNetwork}
+                onChange={async (event) => {
+                  await saveSetting({
+                    bitcoinNetwork: event.target.value,
+                  });
+                }}
+              >
```
reneaaron, line 393: Move this setting to the accounts. (or remove it :wink:)

src/app/components/MnemonicInputs/index.tsx
rolznz: removed `autoComplete` as it did nothing.

I also no longer set the type to password any more when the input is not selected because it had multiple issues on firefox (autocomplete not working, plus it kept suggesting passwords even using `autoComplete="off"`).

I think this is OK though because we change the value of the input to "•••••" when it is not selected, so the words cannot be cached by the browser

src/app/screens/Accounts/ImportSecretKey/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [mnemonic, setMnemonic] = useState<string>("");
+  const account = useAccount();
+  const { t: tCommon } = useTranslation("common");
+  const { t } = useTranslation("translation", {
+    keyPrefix: "accounts.account_view.mnemonic",
+  });
+
+  const [hasFetchedData, setHasFetchedData] = useState(false);
+  const [hasMnemonic, setHasMnemonic] = useState(false);
+  const [currentPrivateKey, setCurrentPrivateKey] = useState("");
```
rolznz, line 26: Do not access private key here (only public key is needed)

rolznz: + add nostr prefix

src/app/utils/saveMnemonic.ts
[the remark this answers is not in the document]
rolznz: leaving as is for now but moving this functionality to the background script

src/app/utils/getNostrKeyOrigin.ts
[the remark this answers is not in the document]
rolznz: @bumi @reneaaron I realized setting it to imported by default will not be good for users who previously derived their key from their alby account. But we could simplify this and just use "imported" for both.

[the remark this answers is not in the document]
rolznz: this is a lot simpler now without checking for the old derived key, and I also updated the nostr generatePrivateKey function to use the mnemonic

src/app/screens/Accounts/ImportSecretKey/index.tsx
rolznz: both data are now available from `await api.getAccount()` (`nostrEnabled`, `hasMnemonic`)

"getMnemonic" is only requested in one place now - the backup screen 🎉

src/extension/background-script/actions/nostr/generatePrivateKey.ts
```diff
@@ -1,54 +1,41 @@
...
+  const password = await state.getState().password();
+  if (!password) {
+    return {
+      error: "Password is missing.",
+    };
+  }
+  const accounts = state.getState().accounts;
+
+  if (id && Object.keys(accounts).includes(id)) {
+    const mnemonic = await state.getState().getMnemonic();
```
rolznz, line 17: `generatePrivateKey` supports passing an id arg but we always use the mnemonic of the currently-selected account. I think this could cause unexpected behaviour. Thoughts?

rolznz: I changed this to create a new mnemonic object for the current account

src/app/screens/Accounts/ImportSecretKey/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  function cancelImport() {
+    // go to account settings
+    navigate(`/accounts/${id}`);
+  }
+
+  async function importKey() {
+    try {
+      if (hasMnemonic) {
+        if (!window.confirm(t("import.confirm_overwrite"))) {
```
rolznz, line 57: use consistent confirmation (write account name to confirm)

src/app/components/MnemonicInputs/index.tsx
```diff
@@ -0,0 +1,99 @@
+import {
```
rolznz, line 1: move into mnemonic subfolder along with SecretKeyDescription

src/app/screens/Accounts/NostrAdvancedSettings/index.tsx
```diff
@@ -0,0 +1,263 @@
...
+import Alert from "~/app/components/Alert";
+import Button from "~/app/components/Button";
+import InputCopyButton from "~/app/components/InputCopyButton";
+import TextField from "~/app/components/form/TextField";
+import { useAccount } from "~/app/context/AccountContext";
+import { GetAccountRes } from "~/common/lib/api";
+import msg from "~/common/lib/msg";
+import { default as nostr, default as nostrlib } from "~/common/lib/nostr";
+
+function NostrAdvancedSettings() {
```
reneaaron, line 20: Nostr Settings

src/common/lib/nostr.ts
```diff
@@ -17,6 +20,13 @@ const nostr = {
   hexToNip19(hex: string, prefix = "nsec") {
     return bech32Encode(prefix, hex);
   },
+  generatePublicKey(privateKey: string) {
```
reneaaron, line 23: Generates public key from private key, rename to `derivePublicKey`.

reneaaron: Use bg script action, review usages of this function.

rolznz: maybe the method can be moved to the nostr object (to review)

src/extension/background-script/actions/mnemonic/generateMnemonic.ts
```diff
@@ -0,0 +1,11 @@
+import * as bip39 from "@scure/bip39";
+import { wordlist } from "@scure/bip39/wordlists/english";
+import type { MessageMnemonicGenerate } from "~/types";
+
+const getMnemonic = async (message: MessageMnemonicGenerate) => {
```
reneaaron, line 5: generateMnemonic

src/extension/background-script/state.ts
```diff
@@ -20,11 +22,15 @@ interface State {
   currentAccountId: string | null;
   nostrPrivateKey: string | null;
   nostr: Nostr | null;
+  mnemonic: Mnemonic | null;
```
rolznz, line 25: Add TODO to move these to the account level of state (then they can handle the decryption)

src/extension/background-script/actions/nostr/getPublicKey.ts
```diff
@@ -0,0 +1,38 @@
+import { decryptData } from "~/common/lib/crypto";
+import Nostr from "~/extension/background-script/nostr";
+import { MessageNostrPublicKeyGet } from "~/types";
+
+import state from "../../state";
+
+const getPublicKey = async (message: MessageNostrPublicKeyGet) => {
+  const id = message?.args?.id;
```
rolznz, line 8: should always pass an ID

src/extension/background-script/actions/nostr/setPrivateKey.ts
```diff
@@ -12,16 +14,30 @@ const setPrivateKey = async (message: MessagePrivateKeySet) => {
       error: "Password is missing.",
     };
   }
-  const privateKey = message.args.privateKey;
+  // make sure private key is saved in hex format
+  const privateKey = nostr.normalizeToHex(message.args.privateKey);
```
reneaaron, line 18: try / catch, return error object

reneaaron: or move to some validate() function

src/extension/background-script/bitcoin/index.ts
```diff
@@ -0,0 +1,84 @@
+import * as btc from "@scure/btc-signer";
+import Mnemonic from "~/extension/background-script/mnemonic";
+import { BitcoinAddress, BitcoinNetworkType } from "~/types";
+
+const BTC_TAPROOT_DERIVATION_PATH = "m/86'/0'/0'/0";
+const BTC_TAPROOT_DERIVATION_PATH_REGTEST = "m/86'/1'/0'/0";
+
+class Bitcoin {
+  readonly networkType: BitcoinNetworkType;
```
reneaaron, line 9: Clear memoized objects when the network setting in the account setting is changed.

src/extension/background-script/bitcoin/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  bech32: "bc",
+  bip32: {
+    public: 0x0488b21e,
+    private: 0x0488ade4,
+  },
+  pubKeyHash: 0x00,
+  scriptHash: 0x05,
+  wif: 0x80,
+};
+const regtest: Network = {
```
reneaaron, line 69: Use testnet?

src/common/lib/nostr.ts
rolznz: I don't think this can be moved yet. In the nostr settings you can set a new private key and public key will be updated without saving (I think how this works could be changed. I will add a TODO)

src/extension/background-script/mnemonic/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    const privateKeyBytes = this._hdkey.derive(path).privateKey;
+    if (!privateKeyBytes) {
+      throw new Error("invalid derived private key");
+    }
+    return secp256k1.utils.bytesToHex(privateKeyBytes);
+  }
+
+  derivePublicKey(path: string) {
+    const publicKeyBytes = this._hdkey.derive(path).publicKey;
+    if (!publicKeyBytes) {
```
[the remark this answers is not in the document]
rolznz: changed to return the hdkey object

PR #2340, opened 2023-04-11
#1960 payment confirmation

### LINKED ISSUE #1960: [Feature] Add success screen confirmation when paying webln invoices
### Feature description

Currently, when paying Webln called invoices, Alby does not show any in-extension confirmation and only relays on in-browser toast confirmations.

Unfortunately, this means all users who have notifications turned off for their browsers (or are in "focus" mode for example)  never see any visual confirmation of their payment from Alby when paying webln invoices.

This gets worse if websites when users is paying does not show any visual confirmation as well - in that case users have 0 visual clues if the payment went through at all. That requires users to open the extension and check outgoing transactions. leads to bad experience of confusion and clunkiness. Good example of such behaviour is tipping taggers on btcmap.org

### Describe the solution

I propose a quick, 2 or 3 seconds in-extension success confirmation screen. I can see two ways of implementation, depending on technical constrains: 

1. If it’s possible for Alby to detect user notifications settings, then: 
- If user has notifications on, show only browser toast messages 
- If user has notifications turned off, display in-extension success confirmation

2. If we can't detect users settings, then I suggest we show only in-extension confirmation.

The quickest implementation of this is to take current payment confirmation screen that Alby displays when we make payments directly in-extension and just a simple disappearing bar that will indicate that the window will close soon.

You can see how it could look like on [this prototype](https://www.figma.com/proto/AFU2lTs5H7UfEyfc4Nz6hk/Success-screen-for-webln-payments?page-id=0%3A1&node-id=6%3A530&scaling=min-zoom&starting-point-node-id=6%3A315).

### Describe alternatives

Nagging all places that implemented Webln to add visual confirmations when invoice is paid.

### Additional context

_No response_

### Are you working on this?

None

Ladyperpetua:

I added a success screen confirmation when paying webln invoices by adding an animated success tick, a toast message, a QRCode for extra verification.

Fixes #1960

### Type of change

feat : New feature (non-breaking change which adds functionality)

https://user-images.githubusercontent.com/70077989/231241521-d69dbea2-f051-4361-9357-feddfb0fd69e.mp4

new alby feature

### Checklist


stackingsaunter:
@Ladyperpetua thanks for tackling it!

I think you mistook stacker.news design of deposit of a desired screen we want to have in Alby, which may be my fault in not explaining well what is it about :P Sorry about that

The animation works well in the build when user initiated transactions from the extension, but it doesn't work with WebLN initiated payments (when the payment is initiated on some outside app/website).

See a video, where btcmap.org is an example and user has no visual confirmation of payment sent:

https://user-images.githubusercontent.com/68239231/232496349-6b4d2a87-f2e6-406a-8ebb-be55c62630cd.mov

You can try yourself: https://btcmap.org/leaderboard
In this case we need confirmation that autodissapears after animation ends (1800ms of total display of confirmation seems optimal).

Your implementation also does differ from the handoff from [this comment](https://github.com/getAlby/lightning-browser-extension/issues/1960#issuecomment-1413962014) and [handoff file](https://www.figma.com/file/AFU2lTs5H7UfEyfc4Nz6hk/Success-screens?node-id=84%3A1320&t=1892PHdhuORMivbi-1). Could you update it accordingly? 

We also do not need toast messages anymore, since will always be this visual confirmation and we do not need both.

To sum up what would need to be changed:
- delete toast messages
- we don't need that QR code and invoice in the field after the payment is sent
- update implementation according to this design (see working propotype here for in [pop-up payments](https://www.figma.com/proto/AFU2lTs5H7UfEyfc4Nz6hk/Success-screens?page-id=84%3A1320&node-id=87-1533&viewport=1055%2C462%2C0.37&scaling=min-zoom&starting-point-node-id=87%3A1533&show-proto-sidebar=1) and here for [webln payments](https://www.figma.com/file/AFU2lTs5H7UfEyfc4Nz6hk/Success-screens?node-id=84-1321&t=1892PHdhuORMivbi-0)). Also take note there's no green bar in this design.

If you have any questions, don't hesitate to ask

Ladyperpetua:
Alright thank you..will work on these accordingly

rolznz:
Closing this PR as it does not address the original issue.

PR #2341, opened 2023-04-12
feat(ui): implemented balance info and made styling according to the …

### LINKED ISSUE #2237: Implement balance info into transacting inputs
Because we're going to move balance from the [account dropdown to the home screen](https://github.com/getAlby/lightning-browser-extension/issues/2236), that will remove user's visibility of their balance while transacting.

I propose showing user balance under the input, whenever they need to pay. Then fiat value of sats can be moved inside the input, to the right side:

Screenshot 2023-03-18 at 19 34 23

Receive or setting allowence doesn't need to show balance:

Screenshot 2023-03-18 at 19 36 47

Handoff (Figma file):

Screenshot 2023-03-18 at 19 46 30

HARI-124:
…figma file

I have implemented the balance info below the transaction input as mentioned in the issue #2237 . I have made styling changes according to the given figma file.

Fixes #2237 

### Type of change

Screenshot 2023-04-12 200613

Screenshot 2023-04-12 200543

### How has this been tested?

I have checked the output and submitted the screenshots in the pr

### Checklist

- [] I checked if I need to make corresponding changes to the documentation (and made those changes if needed)

### DIFF
--- a/src/app/screens/LNURLPay/index.tsx
+++ b/src/app/screens/LNURLPay/index.tsx
@@ -24,9 +24,9 @@ import msg from "~/common/lib/msg";
 import utils from "~/common/lib/utils";
 import type {
   LNURLError,
+  LNURLPayServiceResponse,
   LNURLPaymentInfo,
   LNURLPaymentSuccessAction,
-  LNURLPayServiceResponse,
   PaymentResponse,
 } from "~/types";
 
@@ -61,7 +61,6 @@ function LNURLPay() {
       Math.floor(+details?.minSendable / 1000).toString()) ||
       ""
   );
-  const userBalance = auth.account?.balance;
   const [fiatValue, setFiatValue] = useState("");
   const [comment, setComment] = useState("");
   const [userName, setUserName] = useState("");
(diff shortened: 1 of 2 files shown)


src/app/screens/LNURLPay/index.tsx
```diff
@@ -418,6 +418,7 @@ function LNURLPay() {
                               value={valueSat}
                               onChange={(e) => setValueSat(e.target.value)}
                               fiatValue={fiatValue}
+                              hint={`Balance: ` + userBalance + ` sats`}
```
reneaaron, line 421: @rolznz What's the best way to make this account balance & currency available to multiple places? 

Since we now have USD / EUR accounts (kollider) and moved the balance out of the dropdown we will now need this in several places in our application. 

What's the right way to solve this? Should we create a separate hook for that?

reneaaron (commented):
Did some first testing and feels great so far, thanks for taking the initiative here! 🙌

✅ Added some style fixes to this PR
❓️ Left some review feedback

src/app/screens/LNURLPay/index.tsx
reneaaron: I just found out about `balancesDecorated` which exactly does what I need :heavy_check_mark:

reneaaron (approved):
tACK

PR #2350, opened 2023-04-13
fix: register listeners sync

### LINKED ISSUE #2305: [BUG] An unexpected error occurred (Could not establish connection. Receiving end does not exist)

### Describe the bug

Clicking on the extension icon while logged in sometimes throws and error and shows a logged-out state

### Screenshots [optional]

image

image

image

image

### Steps To Reproduce

First Issue:
1. Checkout and build the latest master branch (yarn run dev:chrome)
2. Login to the extension
3. Close the popup
4. Switch to a different application
5. Go back to your browser and click the Alby extension icon

Second issue:
1. Log into the extension with multiple accounts
2. Switch from LND to an Alby hosted account (2nd screenshot)

### Expected behavior

The popup opens and you are still logged in, with no errors. You can also switch accounts without errors.

### Alby information

- Master branch (latest)

### Device information

Chrome Windows

### Additional context

May be related to the recent manifest version 3 update

### Are you working on this?

None

reneaaron:

Possibly fixes #2305 :crossed_fingers: 

[MV3 Migration Guide](https://developer.chrome.com/docs/extensions/migrating/to-service-workers/#register-listeners)

### DIFF
--- a/src/extension/background-script/index.ts
+++ b/src/extension/background-script/index.ts
@@ -116,6 +116,21 @@ const routeCalls = (
   return call;
 };
 
+browser.runtime.onMessage.addListener(debugLogger);
+
+// this is the only handler that may and must return a Promise which resolve with the response to the content script
+browser.runtime.onMessage.addListener(routeCalls);
+
+// Update the extension icon
+browser.tabs.onUpdated.addListener(updateIcon);
+
+// Notify the content script that the tab has been updated.
+browser.tabs.onUpdated.addListener(extractLightningData);
+
+// The onInstalled event is fired directly after the code is loaded.
+// When we subscribe to that event asynchronously in the init() function it is too late and we miss the event.
+browser.runtime.onInstalled.addListener(handleInstalled);
+
 async function init() {
   console.info("Loading background script");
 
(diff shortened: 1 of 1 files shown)


rolznz:
@reneaaron nice! I think this fixes this particular issue. I am just worried about what will happen if we receive requests while the init() function is still running, which might introduce strange errors. I still think it would be good to wait for the service worker / background script to finish the init() before we start interacting with it.

CC @bumi

reneaaron:
> I am just worried about what will happen if we receive requests while the init() function is still running, which might introduce strange errors. I still think it would be good to wait for the service worker / background script to finish the init() before we start interacting with it.

What would we do with messages that are sent before init() finished then? Drop them? Couldn't that lead to other problems? (promises not being resolved, etc?)

rolznz:
> > I am just worried about what will happen if we receive requests while the init() function is still running, which might introduce strange errors. I still think it would be good to wait for the service worker / background script to finish the init() before we start interacting with it.
> 
> What would we do with messages that are sent before init() finished then? Drop them? Couldn't that lead to other problems? (promises not being resolved, etc?)

@reneaaron this fixes the issue for me and was just raising the possibility of other issues, but I can't produce them. I think we should merge this 🚀

bumi:
I don't think it is possible to easily handle messages sent before the init is done.

PR #2357, opened 2023-04-18
fix: attach observer to description element

im-adithya:

Fixes the Youtube Battery to not show the blue icon from the previous creator.

### Type of change

### How has this been tested?

Manually, https://www.youtube.com/@ClubMikeV

### Checklist

### DIFF
--- a/src/extension/content-script/batteries/YouTubeVideo.ts
+++ b/src/extension/content-script/batteries/YouTubeVideo.ts
@@ -14,6 +14,8 @@ declare global {
   }
 }
 
+let oldVideoId: string;
+
 const setData = async (): Promise<void> => {
   let text = "";
   document
@@ -21,12 +23,20 @@ const setData = async (): Promise<void> => {
       "#columns #primary #primary-inner #meta-contents #description .content"
     )
     .forEach((e) => {
-      text += ` ${e.textContent}`;
+      text += `${e.textContent} `;
     });
+
+  const searchParams = new URLSearchParams(window.location.search);
+  const videoId = searchParams.get("v");
+  if (videoId === oldVideoId) {
+    return;
+  }
+  oldVideoId = videoId as string;
   const channelLink = document.querySelector<HTMLAnchorElement>(
     "#columns #primary #primary-inner #meta-contents .ytd-channel-name a"
   );
   if (!text || !channelLink) {
+    resetLightningData();
     return;
   }
   let match;
@@ -49,7 +59,10 @@ const setData = async (): Promise<void> => {
     }
   }
 
-  if (!lnurl) return;
+  if (!lnurl) {
+    resetLightningData();
+    return;
+  }
 
   const name = channelLink.textContent || "";
   const imageUrl =
@@ -86,7 +99,6 @@ const battery = async (): Promise<void> => {
       }
 
       const observer = new MutationObserver(() => {
-        resetLightningData();
         setData();
       });
 
--- a/src/extension/content-script/batteries/YouTubeVideo.ts
+++ b/src/extension/content-script/batteries/YouTubeVideo.ts
@@ -17,6 +17,14 @@ declare global {
 let oldVideoId: string;
 
 const setData = async (): Promise<void> => {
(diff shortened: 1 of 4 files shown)


src/extension/content-script/batteries/YouTubeVideo.ts
```diff
@@ -62,6 +66,24 @@ const battery = async (): Promise<void> => {
...
 };
 
+const battery = async (): Promise<void> => {
+  setData();
+  const videoDescription = document.querySelector(
+    "div#bottom-row.style-scope.ytd-watch-metadata"
+  );
+  if (!videoDescription) return;
+
+  const observer = new MutationObserver((mutations) => {
```
bumi, line 76: let's debounce this function to not call it too often

e.g. let's debounce it for 200ms or something. @reneaaron what do you think?

https://underscorejs.org/docs/modules/debounce.html

src/extension/content-script/batteries/YouTubeVideo.ts
```diff
@@ -62,6 +66,24 @@ const battery = async (): Promise<void> => {
...
+    const latestMutation = mutations[mutations.length - 1];
+    const newDescription = latestMutation.target.textContent?.trim();
+    if (!newDescription) return;
+    resetIcon();
+    setData();
+  });
+
+  observer.observe(videoDescription, { childList: true, subtree: true });
```
bumi, line 84: can we do this on the exact element without the childList/subtree requirement?

src/extension/content-script/batteries/YouTubeVideo.ts
reneaaron: I think we can just return if the URL didn't change, I can't see a reason why you would try to fetch this info for the same URL more than once. :thinking:

rolznz:
Navigating from this user's video and back does seem to work, however the extension often seems to load the wrong state:

Sometimes I click on the alby extension icon and it shows the correct pay screen, sometimes not (on the same video)

image

image

I also reloaded the extension refreshed the page (alby icon green), unlocked the extension and saw the incorrect view again:
image

src/extension/content-script/batteries/YouTubeVideo.ts
```diff
@@ -62,6 +88,40 @@ const battery = async (): Promise<void> => {
...
+    // because the above element doesn't have complete text
+    const descriptionContent = document.querySelector(
+      "#columns #primary #primary-inner #meta-contents #description .content"
+    );
+    // we need to ensure both are present because they don't load at the same time
+    // and this causes an issue because "description" alone isn't full
+    if (description && descriptionContent) {
+      clearInterval(descriptionInterval); // Stop checking for the element
+
+      if (!window.ALBY_BATTERY) {
```
rolznz, line 106: what is ALBY_BATTERY for?

src/extension/content-script/batteries/YouTubeVideo.ts
```diff
@@ -62,6 +88,40 @@ const battery = async (): Promise<void> => {
   ]);
 };
 
+const battery = async (): Promise<void> => {
+  function waitForDescription() {
```
rolznz, line 92: It feels kind of strange that this battery returns a promise but inside sets an interval and returns immediately. Maybe it should have a standard `void` return type?

im-adithya: Changing it!

src/extension/content-script/batteries/YouTubeVideo.ts
im-adithya: It just adds it for the session, i.e. if refreshed it goes away

> what if two different batteries used the same ALBY_BATTERY variable

No, because it adds only in the youtube.com/video links; if there's any other battery which does this then that too would limited to the regex pattern of that battery

im-adithya:
> however the extension often seems to load the wrong state:

Talked to Roland to resolve the above issue^ (will be resolved after the boost button PR)

im-adithya:
#1791 can be closed if we merge this

reneaaron:

PR #2363, opened 2023-04-19
feat: balanceBox component

11WALTER11:

_Continuation from #2318 #2361_

Fixes _#2318 , #2361_


lujakob:
@11WALTER11 please consider https://github.com/getAlby/lightning-browser-extension/pull/2318#pullrequestreview-1392323154

reneaaron:
Implemented some further improvements for the loading state, was quite a ride :sweat_smile: Please have a look @lujakob. 👀

I don't see a reason why we can't just add the className in your PR (where it's actually needed).

lujakob:
> Implemented some further improvements for the loading state, was quite a ride 😅 Please have a look @lujakob. 👀

Yess, it looks good to me 👍
 
> I don't see a reason why we can't just add the className in your PR (where it's actually needed).

True 👍

src/app/context/AccountContext.tsx
```diff
@@ -77,11 +77,9 @@ export function AccountProvider({ children }: { children: React.ReactNode }) {
     });
   };
 
-  const setAccountId = (id: string, willFetch: boolean) => {
+  const setAccountId = (id: string) => {
     setAccount({ id });
-    if (willFetch) {
-      setBalanceLoading(true);
-    }
+    setBalanceLoading(true);
```
lujakob, line 82: TBH I don't get this line of code: the balanceLoading state is set to `true` in the first line of `fetchAccountInfo` and set to `false` before the return statement. Why do we need to set it here? And if so, the only places where it is set to false is within the `fetchAccountInfo` function as well. So if we set it to `true` outside, but never call it, it will never be set to `false`.

lujakob: Also why is it set to false in the callback of `fetchAccountInfo` if it is being set after the await function anyways.

lujakob: @reneaaron I don't understand this callback function. The only place in the whole repo where `await api.swr.getAccountInfo` is called is inside `fetchAccountInfo` of AccountContext. Inside of `api.swr.getAccountInfo` the callback is being called just before the `return resolve(account)`. To me it seems it would make no difference if this callback is removed entirely and the functions of this callback

``` 
    const callback = (accountRes: AccountInfo) => {
      setAccount(accountRes);
      updateAccountBalance(accountRes.balance, accountRes.currency);
      setBalanceLoading(false);
    };
```

would just be called after the `await`.

reneaaron: @lujakob Thanks for your review! 

> Edit: I think I found it. Probably you need it because if this delay of calling setAccountId and fetchAccountInfo

Exactly. I'll add a comment. 

The callback is also called here. I think it's purpose is to update the UI from the cache (while still revalidating in the background and then updating again as the new data is loaded. 
https://github.com/11WALTER11/lightning-browser-extension/blob/update/src/common/lib/api.ts#L69

static/views/options.html
```diff
@@ -1,15 +1,17 @@
...
-  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-  <link rel="icon" type="image/png" href="assets/icons/alby_icon_yellow_32x32.png">
-  <title>Alby</title>
-</head>
-
-<body class="bg-gray-50 dark:bg-surface-00dp">
-  <div id="options-root"></div>
-</body>
-
+  <body class="bg-gray-50 dark:bg-surface-00dp slashed-zero">
```
reneaaron, line 14: `slashed-zero` is the only change here (and for other html files)

src/app/components/AccountMenu/index.tsx
```diff
@@ -85,12 +85,12 @@ function AccountMenu({ showOptions = true }: Props) {
...
               className={`flex-auto mx-2 py-3 overflow-hidden max-w-[14rem] text-left`}
             >
-              <p
+              <div
                 title={title || ""}
                 className="text-sm font-medium text-gray-700 dark:text-neutral-400 text-ellipsis overflow-hidden whitespace-nowrap"
               >
-                {title || <Skeleton />}
-              </p>
+                {loading ? <SkeletonLoader className="w-20" /> : title || "⚠️"}
```
reneaaron, line 92: Currently errors while switching accounts result in an empty account name. I've added a warning icon for now: 

image

Also I created a followup issue to tackle this problem: #2377

reneaaron:
I've reverted some changes for loading states as we need to tackle problems in #2377 first. 

I think this is now merge-able, please have another look. @lujakob

lujakob (approved):
tACK

src/app/components/SkeletonLoader/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { classNames } from "~/app/utils";
+
+export type Props = {
+  className?: string;
+};
+
+function SkeletonLoader({ className }: Props) {
+  return (
+    <Skeleton
+      style={{ opacity: 0.2 }}
```
reneaaron, line 11:
```suggestion
```

rolznz (approved):

PR #2364, opened 2023-04-19
chore: await initialization before routing a msg

bumi:
This uses a deferred promise to make sure the initialization has finished before we process a message call.
This also fixes some LNC issues and makes sure that the WASM is namespaced.


src/extension/background-script/index.ts
```diff
@@ -154,13 +161,23 @@ async function init() {
...
+init()
+  .then(() => {
+    if (resolveInit) {
+      resolveInit();
+    }
+    if (isFirstInstalled && !state.getState().getAccount()) {
+      utils.openUrl("welcome.html");
+    }
+    if (isRecentlyUpdated) {
+      migrate();
```
rolznz, line 173: Should we consider moving migrate() to the bottom of the init() function?

src/extension/background-script/index.ts
```diff
@@ -154,13 +161,23 @@ async function init() {
...
-  if (isFirstInstalled && !state.getState().getAccount()) {
-    utils.openUrl("welcome.html");
-  }
-  if (isRecentlyUpdated) {
-    migrate();
-  }
-});
+init()
+  .then(() => {
+    if (resolveInit) {
```
rolznz, line 166: shouldn't resolveInit always exist here? can we throw an exception instead if it undefined?

rolznz:
@bumi the fix works well for me, for a basic test of adding a 10 second sleep to the start of the init function. Without the fix doing this does some strange things, like launching the welcome screen and asking to set a new password.

reneaaron:
Did some testing and added delays to test. Routing calls are stacked during initialization and resolved after that: 👍️

image

src/extension/background-script/index.ts
reneaaron: You probably mean to the top of the init function, right? For me it would make sense to await the migration and only then `resolveInit` to avoid calls being executed while migrations are still taking place.

rolznz: I agree `resolveInit` shouldn't be called until the migrations have run. But don't the migrations depend on `init`? that's why I suggested it being moved to the last line in the init() function.

src/extension/background-script/index.ts
reneaaron: I think it's just for TS because `resolveInit?`, can't be changed to `resolveInit` because it's only set at runtime I guess.

src/extension/background-script/state.ts
```diff
@@ -96,19 +96,25 @@ const state = createState<State>((set, get) => ({
   },
   getConnector: async () => {
     if (get().connector) {
-      return get().connector as Connector;
+      const connector = (await get().connector) as Connector;
```
rolznz, line 99: I think the `if (get().connector)` should be removed and just do this:

````
const connector = (await get().connector) as Connector;
if (connector) {
  return connector;
}
```

rolznz: Ah, I see this won't work for the first request where connector is null.

src/app/components/AccountMenu/index.tsx
```diff
@@ -52,10 +52,10 @@ function AccountMenu({ showOptions = true }: Props) {
   async function selectAccount(accountId: string) {
     setLoading(true);
     try {
-      setAccountId(accountId);
       await msg.request("selectAccount", {
         id: accountId,
       });
+      setAccountId(accountId);
```
reneaaron, line 58: Why did you change the order here? 

This leads to some side effects (e.g. the avatar changes only after the account has been loaded) :thinking:

bumi: is that a problem? 

I wanted to avoid some react dependency to do requests before the new account was switched.

PR #2393, opened 2023-05-03
fix: pngcrush images

reneaaron:


reneaaron:
Images are very well optimized :sweat_smile:

PR #2399, opened 2023-05-06
feat: Alby OAuth Connector

pavanjoshi914:
Fixes https://github.com/getAlby/lightning-browser-extension/issues/1956

```LaunchWebAuthFlow```  identity api to launch oauth flow in extension. there is still a url mismatch for ```redirect Uri ```as i am using test_client as a oauth client which has a redirectUri  pointing to```localhost:8080```. i need a new oauth client to test rest of the code.

### Type of change

image

### Checklist

### CODE REVIEW

**Comment by rolznz** [comment]
Some issues found so far:
- Hardcoded fields for OAuth clients - I am fixing this
- Client ID changes every time unpackeddev extension is loaded - I am fixing this
- Does not work in regtest - I am fixing this
- Connect with Alby button should be primary color
- Connect with Alby button needs a loading state because it takes a long time to launch the modal (why does it take so long?)
- Connect with Alby is impossible to login because it launches webln instead of allowing to enter username/password

File: src/extension/background-script/actions/accounts/edit.ts [hunk 1]
```diff
@@ -8,6 +9,21 @@ const edit = async (message: MessageAccountEdit) => {
   if (accountId in accounts) {
     accounts[accountId].name = message.args.name;
 
+    if (
```
rolznz, line 12: This feels strange, it only applies to alby and the call to edit has to pass a fake message. I would not call this method, instead, edit these fields from within the alby connector.

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,620 @@
...
+
+    await authClient.requestAccessToken(authToken);
+  }
+
+  generateHmacVerification(uri: string) {
+    const mac = hmacSHA256(uri, HMAC_VERIFY_HEADER_KEY).toString(Base64);
+    return encodeURIComponent(mac);
+  }
+
+  async request<Type>(
```
rolznz, line 543: Why not use the alby-js-sdk client? otherwise we need to reimplement all the calls here? e.g. just use `const result = await client.sendPayment({ invoice: bolt11 })`; instead of doing the http call manually.

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,620 @@
...
+      typeof data.payment_preimage === "object" &&
+      data.payment_preimage.type === "Buffer"
+    ) {
+      data.payment_preimage = Buffer.from(data.payment_preimage.data).toString(
+        "hex"
+      );
+    }
+
+    // HACK!
+    // some Lnbits extension that implement the LNDHub API do not return the route information.
```
rolznz, line 240: why is this needed? lnbits? lndhub? this is the new Alby oauth connector, it should use the Alby API via the alby js sdk?

src/extension/background-script/actions/ln/invoices.ts
```diff
@@ -15,7 +15,7 @@ const invoices = async (message: MessageInvoices) => {
           invoice.custom_records
         );
         const type = invoice.type;
-        if (type !== "received") {
+        if (type !== "received" && type !== "incoming") {
```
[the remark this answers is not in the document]
pavanjoshi914: for other connectors, type is receiving, right? don't we have to consider both here?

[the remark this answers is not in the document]
rolznz: Hmm... I think at the connector level it should only return incoming/received invoices, not outgoing invoices. But it seems like the other connectors specifically add a new field called type to the invoices with value "received". Maybe we should update the alby invoices to also have the type "received", so it is consistent

pavanjoshi914:
image

in the ```geInfo``` function, we are assigning lightning addresses to the alias. but what if the account has no lightning address? there was no fallback that is why caution symbol on the creation of a new account

pavanjoshi914:
i just opened extension. i had to authorize again. this means it is not refreshing token with refresh token as expiry of refresh token is for month right?

src/app/screens/connectors/ConnectAlby/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+            oAuthToken: validation.oAuthToken,
+          },
+        };
+
+        const addResult = await msg.request("addAccount", account);
+        if (addResult.accountId) {
+          await msg.request("selectAccount", {
+            id: addResult.accountId,
+          });
+          if (fromWelcome) {
```
pavanjoshi914, line 47: if the user already has an account and tries to add an alby account it will go to discover as expected. but the discover page is rendered under the name of the same existing account and not the new account. i had to refresh to make existing account visible under extension

pavanjoshi914:
> i just opened extension. i had to authorize again. this means it is not refreshing token with refresh token as expiry of refresh token is for month right?

logic seems fine, but call to refreshAccessToken is failing. authClient do contain valid access token and refresh token

image

src/app/screens/connectors/ConnectAlby/index.tsx
pavanjoshi914: i think this is happening because the way we persisting the account in the connector and setting the state of the account.

```
 if (this.account.id) {
        const accounts = state.getState().accounts;
        const password = (await state.getState().password()) as string;

        const configData = decryptData(
          accounts[this.account.id].config,
          password
        );
        configData.oAuthToken = this.config.oAuthToken;
        accounts[this.account.id].config = encryptData(configData, password);
        state.setState({ accounts });
        // make sure we immediately persist the updated accounts
        await state.getState().saveToStorage();
      }
```

also this part is not executed when we add account. during that time this.account.id will be undefined. is this valid behaviour @rolznz ?

rolznz: @pavanjoshi914 I think it is because it no longer uses the TestConnection code which seems to update the account list and set the selected account to the newly created account. I still don't think it makes sense to test the connection for the Alby connector because we already successfully logged in, but we need the functionality to select and update the accounts. Could you have a look at this?

pavanjoshi914: checking! but how we are persisting tokens for the first time we add account as the above code is not executed

rolznz: @pavanjoshi914 the account is added, but the accounts list and selected account is not updated in the frontend

pavanjoshi914:
> > also, shall we use the lightning address as an alias for new Alby accounts? yes we will have connector field to differentiate alby accounts from non alby accounts but we need to consider working of isAlbyAccount function as well. we shall refactor all the instances of this functioon
> 
> I think we might actually need to have two functions - isAlbyAccount and isAlbyLndhubAccount (legacy, for the original checks). I don't think we will be able to migrate the existing ones to use the new OAuth connector because it would break some things such as LNURL auth. What do you think @reneaaron ?

created two different functions to identify alby accounts and legacy lndhub alby accounts to showcase additional features

src/app/hooks/useTips.ts
```diff
@@ -1,15 +1,20 @@
 import { useAccount } from "~/app/context/AccountContext";
 import { useSettings } from "~/app/context/SettingsContext";
-import { isAlbyAccount } from "~/app/utils";
+import { isAlbyAccount, isAlbyLNDHubAccount } from "~/app/utils";
 import { TIPS } from "~/common/constants";
 
 const DEFAULT_TIPS = [TIPS.TOP_UP_WALLET, TIPS.MNEMONIC];
 
-export const filterTips = (closedTips: TIPS[], connector = "") => {
+export const filterTips = (closedTips: TIPS[], connector = "", alias = "") => {
```
rolznz, line 8: I don't think filterTips should have to know how to deal with `connector` and `alias`. We should try limit this as much as possible.

rather than doing the calculation inside I wonder if it should pass these parameters as boolean values (isAlbyAccount, isAlbyLNDHubAccount). 

Note: I think when we add the onchain address for the Alby OAuth connector, then this tip might link to a different place for the new one (like receive screen in the extension)?

pavanjoshi914: did those changes kindly review

doc/SETUP.md
```diff
@@ -4,19 +4,22 @@
...
+  `$ cp .env.example .env`
+- Set your desired .env settings to connect to either regtest or production for your specific browser.
 
 ### 💻 Load extension into browser
 
 - Start development build, which will automatically watch for file changes:
 
   - Chrome\
-    `$ yarn run dev:chrome`
+    `$ dotenv yarn run dev:chrome`
```
rolznz, line 16: Now we use dotenv does it make sense to also use cross-env? it could be removed and all env variables moved to the .env file

.github/workflows/build.yml
```diff
@@ -13,6 +13,7 @@ jobs:
 
     name: build-prod
     env:
+      ALBY_API_URL: "https://getalby.com/api"
```
rolznz, line 16: Do we still need WALLET_ROOT_URL, plus inconsistency with final "/"

src/app/hooks/useTips.test.ts
```diff
@@ -32,14 +32,14 @@ jest.mock("~/app/context/AccountContext", () => ({
...
   test("should not have top up wallet tip when using a non-alby account", async () => {
-    tmpAccount = { id: "1", name: "LND account", alias: "" };
+    tmpAccount = { id: "1", name: "LND account", connector: "" };
     const { tips } = useTips();
     expect(tips.length).toBe(1); // mnemonic
     const hasTopUpWallet = tips.some((tip) => tip === TIPS.TOP_UP_WALLET);
     expect(hasTopUpWallet).toBe(false);
   });
   test("should have top up wallet tip when having alby account", async () => {
-    tmpAccount = { id: "2", name: "Alby", alias: "🐝 getalby.com" };
```
rolznz, line 42: should we leave in tests for old Alby lndhub account? (where `isAlbyLNDHubAccount` returns true)

src/app/screens/Onboard/TestConnection/index.tsx
```diff
@@ -29,6 +29,7 @@ export default function TestConnection() {
 
     try {
       const response = await api.getAccountInfo();
+      // we shall use connector property here
```
rolznz, line 32: is this comment needed?

src/app/screens/Options/TestConnection/index.tsx
```diff
@@ -11,6 +11,7 @@ import api from "~/common/lib/api";
 import msg from "~/common/lib/msg";
 import type { AccountInfo } from "~/types";
 
+// alais should be replaced with connector
```
rolznz, line 14: is this comment needed?

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,378 @@
...
+    console.error(
+      `${this.constructor.name} does not implement the connectPeer call`
+    );
+    throw new Error("Not yet supported with the currently used account.");
+  }
+
+  async getInvoices(): Promise<GetInvoicesResponse> {
+    const incomingInvoices = (await this._request((client) =>
+      client.incomingInvoices({})
+    )) as Invoice[];
```
rolznz, line 77: Cast is not needed if we merge https://github.com/getAlby/alby-js-sdk/pull/19

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,378 @@
...
+    console.info("incomingInvoices", incomingInvoices);
+
+    const invoices: ConnectorInvoice[] = incomingInvoices.map(
+      (invoice, index): ConnectorInvoice => ({
+        custom_records: invoice.custom_records,
+        id: `${invoice.payment_request}-${index}`,
+        memo: invoice.memo,
+        preimage: "", // lndhub doesn't support preimage (yet)
```
rolznz, line 85: is preimage available now?

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,378 @@
...
+  async getAccessToken(authResult: string, authClient: auth.OAuth2User) {
+    const authToken = new URL(authResult).searchParams.get("code");
+    if (!authToken) {
+      throw new Error("Authentication failed: missing token");
+    }
+
+    await authClient.requestAccessToken(authToken);
+  }
+
+  generateHmacVerification(uri: string) {
```
rolznz, line 284: No longer used, delete

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,378 @@
...
+    const mac = hmacSHA256(uri, HMAC_VERIFY_HEADER_KEY).toString(Base64);
+    return encodeURIComponent(mac);
+  }
+
+  async request<Type>(
```
rolznz, line 289: no longer used, delete this function

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,378 @@
...
+          accounts[this.account.id].config,
+          password
+        );
+        configData.oAuthToken = this.config.oAuthToken;
+        accounts[this.account.id].config = encryptData(configData, password);
+        state.setState({ accounts });
+        // make sure we immediately persist the updated accounts
+        await state.getState().saveToStorage();
+      }
+      console.info(
```
rolznz, line 368: review and remove console logs that leak oauth / refresh token

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -13,17 +14,17 @@ interface Route {
...
+    "5482373484"?: string;
+  } & Record<string, string>;
   id: string;
   memo: string;
   preimage: string;
   settled: boolean;
   settleDate: number;
   totalAmount: string;
-  type: "received";
+  type: string;
```
rolznz, line 27: undo - this should be type: "received"

src/extension/background-script/actions/ln/invoices.ts
```diff
@@ -11,12 +11,15 @@ const invoices = async (message: MessageInvoices) => {
...
     let invoices: Invoice[] = result.data.invoices
       .filter((invoice) => (isSettled ? invoice.settled : !invoice.settled))
-
-      .map((invoice: Invoice) => {
+      .map((invoice) => {
         const boostagram = utils.getBoostagramFromInvoiceCustomRecords(
           invoice.custom_records
         );
-        return { ...invoice, boostagram };
+        const type = invoice.type;
```
rolznz, line 18: undo this change, once type is reverted to "received" in `ConnectorInvoice` in connector.interface

.github/workflows/build.yml
pavanjoshi914: it is only used in the previous walletCreate method to create alby wallet, we can remove it!

src/app/hooks/useTips.test.ts
pavanjoshi914: added one more test for legacy lndhub alby accounts. where alias being checked and card being showed

src/app/screens/Onboard/TestConnection/index.tsx
pavanjoshi914: removed

src/app/screens/Options/TestConnection/index.tsx
pavanjoshi914: considering migration of existing account. left this comment here. but no more the case as we use two separate functions. removed

src/extension/background-script/connectors/alby.ts
pavanjoshi914: resolved

src/extension/background-script/connectors/alby.ts
pavanjoshi914: resolved

src/extension/background-script/connectors/alby.ts
pavanjoshi914: removed

src/extension/background-script/connectors/connector.interface.ts
pavanjoshi914: oauth connector will be using type as received instead of incoming/outgoing. resolved

src/extension/background-script/actions/ln/invoices.ts
pavanjoshi914: resolved

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,378 @@
...
+    }
+    return result;
+  }
+
+  private _getRequestOptions(): Partial<RequestOptions> {
+    return {
+      ...(process.env.ALBY_API_URL
+        ? { base_url: process.env.ALBY_API_URL }
+        : {}),
+    };
```
rolznz, line 343: Should we pass a user_agent here? requires a new deployment of alby-js-sdk

rolznz (changes requested):

src/extension/background-script/connectors/alby.ts
rolznz: I guess we do not need preimage for incoming payments?

src/app/screens/Accounts/GenerateSecretKey/index.tsx
```diff
@@ -89,9 +98,27 @@ function GenerateSecretKey() {
                   htmlFor="has_backed_up"
                   className="cursor-pointer ml-2 block text-sm text-gray-900 font-medium dark:text-white"
                 >
-                  {t("backup.confirm")}
+                  {t("generate.confirm")}
                 </label>
               </div>
+              {!isAlbyAccount(connector) && (
+                <div className="flex items-center">
+                  <Checkbox
```
rolznz, line 106: Move outside of mnemonic inputs box

rolznz (changes requested):

src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -462,6 +459,39 @@ function AccountDetail() {
...
+                    <p className="text-gray-900 dark:text-white font-medium">
+                      {t("mnemonic.lnurl.title")}
+                    </p>
+                    <p className="text-gray-500 text-sm dark:text-neutral-500">
+                      {t("mnemonic.lnurl.use_mnemonic")}
+                    </p>
+                  </div>
+
+                  <div className="w-1/5 flex-none flex justify-center">
+                    <Toggle
```
rolznz, line 476: right align the toggle and center vertically

src/app/screens/Accounts/GenerateSecretKey/index.tsx
rolznz: Update: remove this checkbox completely

src/i18n/locales/en/translation.json
```diff
@@ -446,6 +446,10 @@
           },
           "inputs": {
             "title": "Your Secret Key"
+          },
+          "lnurl": {
+            "title": "LNURL Auth",
```
rolznz, line 451: use less technical wording here, and use "LNURL-auth" in description

src/app/screens/connectors/ConnectAlby/index.tsx
```diff
@@ -0,0 +1,84 @@
...
```
reneaaron, line 47: Remove fromWelcome and just do the test connection step (similar to what other connectors do)

src/extension/background-script/actions/lnurl/auth.ts
```diff
@@ -35,18 +35,38 @@ export async function authFunction({
...
+  if (
+    account.connector === "alby" ||
+    (account.mnemonic && account.useMnemonicForLnurlAuth)
+  ) {
+    if (!account.mnemonic) {
+      throw new Error("Please set a secret key to use LNURL auth");
+    }
+    const mnemonic = await state.getState().getMnemonic();
+
+    lnSignature = await mnemonic.signMessage(LNURLAUTH_CANONICAL_PHRASE);
```
reneaaron, line 55: keymaterialForSignature?

src/extension/background-script/actions/setup/validateAccount.js
```diff
@@ -7,9 +7,14 @@ const validateAccount = async (message, sender) => {
 
   try {
     const info = await connector.getInfo();
-    await connector.unload(); // unload the connector again, we just checked if it works but have no persistence
```
reneaaron, line 10: Remove this? Doesn't that change the behavior for adding accounts in the extension?

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,353 @@
...
+      const info = await this._request((client) =>
+        client.accountInformation({})
+      );
+      const alias = info.lightning_address
+        ? info.lightning_address
+        : "🐝 getalby.com";
+
+      return {
+        data: {
+          alias: alias,
```
reneaaron, line 105: alias: "🐝 getalby.com"
+ https://guides.getalby.com/alby-wallet-api/reference/api-reference/account

src/app/screens/connectors/ConnectAlby/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+    try {
+      const validation = await msg.request("validateAccount", initialAccount);
+      if (validation.valid) {
+        if (!validation.oAuthToken) {
+          throw new Error("No oAuthToken returned");
+        }
+
+        const account = {
+          ...initialAccount,
+          name: (validation.info as { data: WebLNNode }).data.alias,
```
rolznz, line 35: use email, fallback to username part of lightning address, finally fallback to "getalby.com"

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,353 @@
...
+    return {
+      data: {
+        paymentRequest: data.payment_request,
+        rHash: data.payment_hash,
+      },
+    };
+  }
+
+  private async authorize(): Promise<auth.OAuth2User> {
+    // console.info("this.config.oAuthToken", this.config.oAuthToken);
```
reneaaron, line 194:
```suggestion
```

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,353 @@
...
+  private _getRequestOptions(): Partial<RequestOptions> {
+    return {
+      ...(process.env.ALBY_API_URL
+        ? { base_url: process.env.ALBY_API_URL }
+        : {}),
+    };
+  }
+
+  private async _updateOAuthToken(newToken: Token) {
+    console.info("Updating account oauth token", newToken);
```
reneaaron, line 327:
```suggestion
```

src/app/screens/connectors/ConnectAlby/index.tsx
bumi: can we somehow use some less fancy syntax here?

doc/SETUP.md
```diff
@@ -4,19 +4,22 @@
 
 - Install dependencies\
   `$ yarn install`
+- Create development .env file
+  `$ cp .env.example .env`
+- Set your desired .env settings to connect to either regtest or production for your specific browser.
```
jankoegel, line 9: do we also have a setup to connect with a local development Rails app?

rolznz: We are going to undo this change, so you can use the same setup instructions as before (just set one env variable to point at your locally running API)

src/app/screens/connectors/ConnectAlby/index.tsx
rolznz: Changed to 
`name: accountInfo.email || accountInfo.lightning_address || "getalby.com",`

image

src/app/components/AccountMenu/index.tsx
```diff
@@ -130,7 +130,8 @@ function AccountMenu({ showOptions = true }: Props) {
             <WalletIcon className="h-5 w-5 mr-2 text-gray-700 dark:text-neutral-300" />
             {t("options.account.account_settings")}
           </Menu.ItemButton>
-          {isAlbyAccount(authAccount?.alias) && (
+          {(isAlbyLNDHubAccount(authAccount?.alias) ||
```
rolznz, line 133: Draft PR for taking avatar image from Alby account: https://github.com/pavanjoshi914/lightning-browser-extension/pull/4

playwright.config.ts
```diff
@@ -7,7 +7,7 @@ import { devices } from "@playwright/test";
  */
 // require('dotenv').config();
 
-if (process.env.WALLET_CREATE_URL === "https://getalby.com/api/users") {
+if (process.env.ALBY_API_URL === "https://getalby.com/api") {
```
reneaaron, line 10: Just use the BASE_URL (domain) for the value, append paths in the extension.

src/extension/background-script/connectors/alby.ts
reneaaron: Yes, please add.

webpack.config.js
```diff
@@ -12,10 +12,62 @@ const TsconfigPathsPlugin = require("tsconfig-paths-webpack-plugin");
...
+        id: "uQkyHFBkaC",
+        secret: "0agh0cKkGWQSXTGRz9oy",
+      },
+    },
+  },
+  production: {
+    mainnet: {
+      // TODO: add these
+    },
+  },
```
reneaaron, line 50:
```suggestion
```

webpack.config.js
```diff
@@ -12,10 +12,62 @@ const TsconfigPathsPlugin = require("tsconfig-paths-webpack-plugin");
...
+};
+
+const oauthBrowser =
+  process.env.TARGET_BROWSER === "firefox" ? "firefox" : "chrome";
+
+// setup ALBY_OAUTH_CLIENT_ID
+const selectedOAuthCredentials =
+  oauthCredentials[process.env.NODE_ENV][network][oauthBrowser];
```
reneaaron, line 58: Test production build with explicitly set credentials.

src/extension/background-script/actions/lnurl/auth.ts
```diff
@@ -35,25 +35,47 @@ export async function authFunction({
...
-    key_loc: {
-      key_family: 0,
-      key_index: 0,
-    },
-  });
+  // use mnemonic for LNURL auth
+  if (
+    account.connector === "alby" ||
+    (account.mnemonic && account.useMnemonicForLnurlAuth)
+  ) {
```
reneaaron, line 49: Remove this check, add a useful message to `signMessage()` exceptions.

src/extension/background-script/connectors/alby.ts
rolznz: @bumi could you let me know when you have deployed a new version of alby-js-sdk?

playwright.config.ts
rolznz: We cannot do this right now due to inconsistency of app location between getalby.com and app.regtest.getalby.com

webpack.config.js
rolznz: fixed

src/app/screens/connectors/ConnectAlby/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+          AlbyAccountInformation;
+
+        const account = {
+          ...initialAccount,
+          // legacy accounts may not have either an email address or lightning address
+          name:
+            accountInfo.email || accountInfo.lightning_address || "getalby.com",
+          config: {
+            ...initialAccount.config,
+            oAuthToken: validation.oAuthToken,
```
rolznz, line 44: we need to store the ID of the alby account too

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,352 @@
...
+      }
+      authUrl += "&webln=false"; // stop getalby.com login modal launching lnurl auth
+      const authResult = await this.launchWebAuthFlow(authUrl);
+      const code = new URL(authResult).searchParams.get("code");
+      if (!code) {
+        throw new Error("Authentication failed: missing authResult");
+      }
+
+      const token = await authClient.requestAccessToken(code);
+      await this._updateOAuthToken(token.token);
```
rolznz, line 252: We need to make sure the oAuth token is for the same account ID as it was originally created, otherwise show an error message and ask them to try again

"You logged into the wrong account. Please try again"

src/app/screens/connectors/ConnectAlby/index.tsx
rolznz: decided not do do this, instead updating the account to match the data from the api

src/extension/background-script/connectors/alby.ts
rolznz: decided not do do this, instead updating the account to match the data from the api

pavanjoshi914:
Few things I encountered during testing
1. I setup lnd test account, and alby oauth account in regtest environment. i am able to receive sats from lnd to alby oauth account but not able to send payment to lnd.
i am able to send payment from one alby oauth account to another oauth account 

2. if i add new account account menu updates the state to new account but, test connection screen still fall backs to the existing account
image


3. checkpayment function for connector is broken. i am always catching error as soon as invoice is created

image

src/app/screens/Receive/index.tsx
```diff
@@ -52,7 +52,7 @@ function Receive() {
   const [paid, setPaid] = useState(false);
   const [pollingForPayment, setPollingForPayment] = useState(false);
   const mounted = useRef(false);
-  const isAlbyUser = isAlbyAccount(auth.account?.alias);
+  const isAlbyUser = isAlbyOAuthAccount(auth.account?.alias);
```
reneaaron, line 55: oAuth || lndhub

src/app/screens/ConfirmPayment/index.tsx
```diff
@@ -79,13 +79,18 @@ function ConfirmPayment() {
...
       setLoading(true);
+      // TODO: move to api
       const response = await msg.request(
         "sendPayment",
         { paymentRequest: paymentRequest },
         {
           origin: navState.origin,
         }
       );
+      if (response.error) {
```
reneaaron, line 90: Double check if that is needed? Is that really the case for existing connectors too, that no errors would be shown?

src/app/screens/connectors/ConnectAlby/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+          ...initialAccount,
+          config: {
+            ...initialAccount.config,
+            oAuthToken: validation.oAuthToken,
+          },
+        };
+
+        const addResult = await msg.request("addAccount", account);
+        if (addResult.accountId) {
+          await msg.request("selectAccount", {
```
reneaaron, line 42: Use API or useAccounts()?

src/extension/background-script/actions/ln/sendPayment.ts
```diff
@@ -33,13 +33,18 @@ export default async function sendPayment(
     response = await connector.sendPayment({
       paymentRequest,
     });
+
+    if (!response.data.preimage) {
```
reneaaron, line 37: Remove this check? If there is an error during sendPayment an exception should be thrown on connector level.

src/extension/background-script/connectors/alby.ts
```diff
@@ -0,0 +1,377 @@
...
+      `${this.constructor.name} does not implement the connectPeer call`
+    );
+    throw new Error("Not yet supported with the currently used account.");
+  }
+
+  async getInvoices(): Promise<GetInvoicesResponse> {
+    const incomingInvoices = (await this._request((client) =>
+      client.incomingInvoices({})
+    )) as Invoice[];
+    console.info("incomingInvoices", incomingInvoices);
```
reneaaron, line 74:
```suggestion
```

src/app/screens/connectors/ConnectAlby/index.tsx
rolznz: Will add a TODO (should be updated on all connectors)

src/extension/background-script/actions/ln/sendPayment.ts
rolznz: removed

src/app/screens/ConfirmPayment/index.tsx
rolznz: There are multiple places where an error is returned, and the sendPayment action is wrapped in try/catch. I think we need this check, I don't understand how it worked before.

rolznz (approved):

PR #2436, opened 2023-05-16
fix: replace htmltotext function

reneaaron:

Replace the htmltotext function with DomParser to avoid JS injections. 

`htmlTotext("abc<svg><svg onload=alert(1)</svg></svg>123")`


im-adithya (approved):
tACK 👍

PR #2438, opened 2023-05-17
fix: speed up injection of window objects

reneaaron:

Remove unnecessary imports to minimize file size of inpage script bundle to minimize time for `window.nostr` being available on websites.


im-adithya (approved):
tACK

PR #2441, opened 2023-05-19
fix: integration with BBW

nicolasburtey:
the current Alby behavior with BBW has a message "getInvoice" is not yet supported when opening the browser extension.

yet it works properly to send/receive payment and show balance, so the message is a bit misleading (reported by one of our users)

I'm also removing the login from phone number,
as it's easier now to use a token to login with Alby

- remove the login flow from phone number 
- clarify the login flow using token
- remove error message poping up when opening alby extension following `getInvoices` trigger

Fixes _#ISSUE-NUMBER_

### Type of change

### How has this been tested?

Tested locally on Chrome

### Checklist

e2e test fail but seem unrelated to my changes


reneaaron:
Hey @nicolasburtey, thanks for your PR! I just successfully connected my BBW account to Alby :muscle: 

Do these access tokens need to be refreshed or are they valid forever? 

Regarding the transaction lists: 

I fully agree that the popup isn't the most elegant way to handle this missing feature, however it's one of the last connectors that doesn't implement this method. I was wondering if there maybe is a way for us to load the transaction history here as well? Do you see a problem fetching this information via a GQL query like that? 

```graphql
query getinfo {
	me {
		defaultAccount {
			defaultWalletId
			wallets {
				id
				balance,
				transactions {
					edges { node { memo, settlementAmount, direction, status}}
					pageInfo { hasNextPage }
				}
			}
		}
	}
}
```

src/app/screens/connectors/ConnectGaloy/index.tsx
```diff
@@ -52,163 +52,17 @@ export default function ConnectGaloy(props: Props) {
...
-      console.error(e);
-      if (e instanceof Error) {
-        toast.error(`${t("galoy.errors.setup_failed")}: ${e.message}`);
-      }
-    } finally {
-      setLoading(false);
+  async function loginWithJwt(event: React.FormEvent<HTMLFormElement>) {
+    if (!jwt) {
+      return null;
     }
```
reneaaron, line 64: There is an error thrown if jwt is falsy below. (see line 80) This should trigger the validation message for the user. 

```suggestion
```

reneaaron (approved):
Only a little nitpick, other than that: let's get this merged! 

Looking forward to the PR for the invoice list! :rocket:

Issue #2435, opened 2023-05-16
Delete all data when resetting the extension


reneaaron:
### Describe the bug

When you currently reset the extension (because you have forgotten the password, etc) only the connectors data is deleted, but not the payment data recorded in the extension. 

Therefore `src/extension/background-script/actions/setup/reset.ts` should be enhanced and also reset all data saved in other places (mainly `browser.storage` comes to mind here).

### Steps To Reproduce

1. Send an outgoing transaction
1. Reset the extension by typing a wrong unlock password and click the reset link
1. Walk through the onboarding flow again
1. Outgoing transactions are still being displayed

### Expected behavior

After resetting the extension all data should be gone.

taboca:
May I work on this one?

fczuardi:
Yes, I was able to reproduce it, the accounts I have to setup again, but Connected Websites and Outgoing Transactions are preserved from before the resetting.

taboca:
Yes https://github.com/getAlby/lightning-browser-extension/issues/2435#ref-pullrequest-1721025257

fczuardi:
hi @reneaaron , we ( @taboca, @mquasar and me) submitted a patch that includes the indexed db clears, so the Outgoing transactions should clear now.

taboca:
When we approached this part of the code [with the 3 heads team session 33 yesterday] we celebrated a bit of advancement which does not clears out the need to understand the archictecture. 

Aside from the actual fix, I think it is worth to understand the logic and the strategy in place and what could happen in the future. To start the line of learning, indeed the browser.storage appears as a sort of "logical" one place or source of truth to make a developer-user think something like "Wait, I can clear my entire browser storage by hand and that is it". 

But then, as the attempt didn't work, I caught myself learning about how the browser's IndexDB infrastructure was also used with an extra layer on top, Dexie, which triggered the mind to think "Ah, maybe this code might need to do lot of DB-like operations, queries for example". 

Which brings me to throw this line of inquiry just wondering if anyone that touched more this part of the code can add thoughts from a more strategic or architectural point of view. Some ideas that crossed my mind when looking at parts of the code: 

* It seems that DB-like operations, queries, delete, insertions and such, are needed because UI-originated actions (perhaps other server actions too?) create need for quick DB-like operations. 
* It seems that the DB schema for the DB records are needed to be versioned because app changes over time. 
[...]

PR #2445, opened 2023-05-23
fix(indexed-db): reset now clears db tables

fczuardi:
Co-authored-by: @taboca
Co-authored-by: @mquasar

_A clear and concise description of what you want to happen_

Fix: #2435

### Type of change

### How has this been tested?

### Checklist


src/extension/background-script/db.ts
```diff
@@ -86,6 +86,21 @@ export class DB extends Dexie {
     return true;
   }
 
+  async clearAllTables() {
+    await this.allowances.clear();
+    await this.payments.clear();
+    await this.blocklist.clear();
+    await this.permissions.clear();
+
+    await browser.storage.local.set({
```
rolznz, line 95: There is also browser.storage.local.clear() - would this be better to ensure future entries are also cleared?

reneaaron (changes requested):
Thanks for the PR! 

As @rolznz noted I think the `browser.storage.clear()` APIs would be a better fit here as they delete all data and do not need to be adjusted in case there are new tables added in future.

src/extension/background-script/db.ts
mquasar: From what I've figured from the code, there's a sort of replication/mirroring of browser.storage.local and IndexedDB. Also, 'to keep consistency' with the code.

mquasar:
> Thanks for the PR! 
> 
> As @rolznz noted I think the `browser.storage.clear()` APIs would be a better fit here as they delete all data and do not need to be adjusted in case there are new tables added in future. 

We tried that but the data still 'lives on' in the IndexedDB. I might be wrong here, obviously 😊

taboca:
> Thanks for the PR!
> 
> As @rolznz noted I think the `browser.storage.clear()` APIs would be a better fit here as they delete all data and do not need to be adjusted in case there are new tables added in future.

Although there is a situation that loading data ( such as payments ) comes from browser.storage API - which would indeed suggest to that deleting browser.storage is enough - yet based on local and limited tests, we have stumbled in a situation where IndexedDB data was persistent. 

A part of the code that gives us a hint the function loadFromStorage: 

https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#L94

In addition, notice how the code logic gives up the sync fro the browser.storage API in the situation where IndexedDB data is already in there: 

https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#LL116C6-L129C12

src/extension/background-script/db.ts
taboca: I also felt that the notion of "attribute by attribute" clearing would be more verbose - at this point - in contrast with the idea of a whole wipe using browser.storage.local.clear() and the whole wipe in the DB too.

rolznz:
> > Thanks for the PR!
> > As @rolznz noted I think the `browser.storage.clear()` APIs would be a better fit here as they delete all data and do not need to be adjusted in case there are new tables added in future.
> 
> Thanks for supporting us @reneaaron and @rolznz
> 
> Although there is a situation that loading data ( such as payments ) comes from browser.storage API - which would indeed suggest that just deleting browser.storage is enough - yet based on local and limited tests, we have stumbled in a situation where IndexedDB data was persistent.
> 
> The part of the code that gives us that idea is the function loadFromStorage:
> 
> https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#L94
> 
> In addition, notice the following section example where the logic gives up the load sync from browser.storage API in the situation where IndexedDB data exists: https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#LL116C6-L129C12
> 
> With the above said it is correct that there is a situation where IndexedDB is not present, which indeed require us to also clear the data from browser.storage
> 
> https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#L106
> 
> Some other observations [#2435 (comment)](https://github.com/getAlby/lightning-browser-extension/issues/2435#issuecomment-1559747137)
> 
> With the above said, if the suggestion has to do with easy maintenance, and if we do the "browser.storage.clear()" then we would have to keep some consistency with wiping the IndexedDB database entirely too? Another idea is to try to do the maintenance from the loadFromStorage function in db.ts which is the code that gets kicked in app initialization time.

@mquasar @taboca thanks for testing and looking deeper into the issue. I think the goal is to delete absolutely everything when resetting the extension. So any and all storage mechanisms (indexedDb, browser.storage, etc) we save data to should be completely cleared.

taboca:
> > > Thanks for the PR!
> > > As @rolznz noted I think the `browser.storage.clear()` APIs would be a better fit here as they delete all data and do not need to be adjusted in case there are new tables added in future.
> > 
> > 
> > Thanks for supporting us @reneaaron and @rolznz
> > Although there is a situation that loading data ( such as payments ) comes from browser.storage API - which would indeed suggest that just deleting browser.storage is enough - yet based on local and limited tests, we have stumbled in a situation where IndexedDB data was persistent.
> > The part of the code that gives us that idea is the function loadFromStorage:
> > https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#L94
> > 
> > In addition, notice the following section example where the logic gives up the load sync from browser.storage API in the situation where IndexedDB data exists: https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#LL116C6-L129C12
> > With the above said it is correct that there is a situation where IndexedDB is not present, which indeed require us to also clear the data from browser.storage
> > https://github.com/getAlby/lightning-browser-extension/blob/5bddad27cd0f58dcb83ffbceb5f3cf1317bb517c/src/extension/background-script/db.ts#L106
> > 
> > Some other observations [#2435 (comment)](https://github.com/getAlby/lightning-browser-extension/issues/2435#issuecomment-1559747137)
> > With the above said, if the suggestion has to do with easy maintenance, and if we do the "browser.storage.clear()" then we would have to keep some consistency with wiping the IndexedDB database entirely too? Another idea is to try to do the maintenance from the loadFromStorage function in db.ts which is the code that gets kicked in app initialization time.
> 
> @mquasar @taboca thanks for testing and looking deeper into the issue. I think the goal is to delete absolutely everything when resetting the extension. So any and all storage mechanisms (indexedDb, browser.storage, etc) we save data to should be completely cleared.

Ok, I will check then the other aspects of DB that seem to be there, such as "LBE-AVAILABILITY-CHECK" and the ways to clear all the DBs.

fczuardi:
> Thanks for the PR!
> 
> As @rolznz noted I think the `browser.storage.clear()` APIs would be a better fit here as they delete all data and do not need to be adjusted in case there are new tables added in future.

We have just updated the patch with the suggestions. It's much simpler now :) Thanks @rolznz !!

rolznz:
@fczuardi I made some more simplifications. Could you please review the changes?

src/extension/background-script/actions/setup/reset.ts
```diff
@@ -1,16 +1,11 @@
...
 const reset = async (message: MessageReset) => {
-  await state.getState().password(null);
-  state.setState({
-    accounts: {},
-    account: null,
-    connector: null,
-    currentAccountId: null,
-  });
-  await state.getState().saveToStorage();
+  await state.getState().reset();
```
fczuardi, line 7: I see that `reset()` will change back `accounts` and `currentAccountId` back to the default values, but I couldn't see if `account` and `connector` are being reset as well. Should they?

fczuardi:
> @fczuardi I made some more simplifications. Could you please review the changes?

@rolznz from your 3 commits, the first 2 LGTM, the third is probably fine as well, I just left a comment asking if your version is equivalent because I couldn't see 2 attributes nullified.

src/extension/background-script/actions/setup/reset.ts
rolznz: @fczuardi thanks for the review! I think I have fixed the issue now.

rolznz:
@reneaaron I think this PR is ready for review now. 💪

bumi:
looks good to me. @reneaaron have a look and merge.

reneaaron (approved):
Thanks for the PR @fczuardi! :muscle: 

tACK

PR #2447, opened 2023-05-23
feat: add send screen amount field max balance error

### LINKED ISSUE #2415: [BUG] Should show a message when inserting a value that is higher than the available balance

### Describe the bug

Hi,

Should show a message when inserting a value that is higher than the available balance. And disable confirm button.

https://github.com/getAlby/lightning-browser-extension/assets/1178856/bfb08505-8e9c-453f-b364-9f140a033c92

### Screenshots [optional]

_No response_

### Steps To Reproduce

1- Paste a lightning address 
2- Insert a value higher than your available balance

### Expected behavior

Show a message telling that the value is higher than the available balance and disable confirm button

### Alby information

Alby version: 2.01

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

No

lujakob:

Add error color to field border and label to inform user about selected amount exceeds balance. 

Fixes #2415

### Type of change

Bildschirmfoto 2023-05-23 um 16 42 42

### How has this been tested?

Manually

### Checklist


lujakob:
@reneaaron I only made changes in the LNURLPay page. Does it make sense to add it to the Keysend page as well? Any more I am missing?

lujakob:
@reneaaron I noticed the max value of the range mentioned and shown in the screenshot above is fixed. I'll adapt it to be either the current fixed or `auth.account.balance`, because it doesn't it is contrarian to display a range higher than the max, agree?

` const amountMax = Math.min(Math.floor(+details.maxSendable / 1000), auth?.account?.balance || 0);`

reneaaron:
Very nice, thanks for the PR! :muscle: 

> @reneaaron I noticed the max value of the range mentioned and shown in the screenshot above is fixed. I'll adapt it to be either the current fixed or auth.account.balance, because it doesn't it is contrarian to display a range higher than the max, agree?

I think the intention of this label is to show the range the receiver is able to accept and would probably try to keep that separate from the balance display (which is also marked red in case of an error). 

I currently can't think of anything else than the Keysend screen, no. (other screens like the receive screen are not limited by your balance)

lujakob:
Thanks for checking.

> Very nice, thanks for the PR! 💪
> 
> > @reneaaron I noticed the max value of the range mentioned and shown in the screenshot above is fixed. I'll adapt it to be either the current fixed or auth.account.balance, because it doesn't it is contrarian to display a range higher than the max, agree?
> 
> I think the intention of this label is to show the range the receiver is able to accept and would probably try to keep that separate from the balance display (which is also marked red in case of an error).
> 

I reverted this change.

> I currently can't think of anything else than the Keysend screen, no. (other screens like the receive screen are not limited by your balance)

I added the new changes to keysend. What I noticed, there is not `hint` about the amount underneath left side. So it's not really obvious, why the input field suddenly has a red border. I suppose it's just missing and added it. See my two screens before and after. I key keysend page should not have a balance hint, I can remove it. But then I think it needs to be improved somehow else...

Bildschirmfoto 2023-05-24 um 13 20 02

Bildschirmfoto 2023-05-24 um 13 22 01

reneaaron:
@lujakob Including the balance here absolutely makes sense, thanks for adding it! I'll do some final testing later today :+1:

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -9,6 +10,8 @@ export type Props = {
   fiatValue: string;
   label: string;
   hint?: string;
+  amountExceeded?: boolean;
+  maxExceeded?: boolean;
```
reneaaron, line 14: Since there is also a minimum value for the range it would probably make sense to change this to `rangeExceeded` and also handle amounts < min for consistency?

src/app/screens/LNURLPay/index.tsx
```diff
@@ -518,7 +524,7 @@ function LNURLPay() {
                         isFocused={false}
                         label={tCommon("actions.confirm")}
                         loading={loadingConfirm}
-                        disabled={loadingConfirm || !valueSat}
+                        disabled={loadingConfirm || !valueSat || amountExceeded}
```
reneaaron, line 527: Should the `rangeExceeded` be included in here as well?

reneaaron (changes requested):

src/app/screens/LNURLPay/index.tsx
lujakob: Updated.

lujakob:
@reneaaron I updated both change requests and adapted the changes for both Lnurlpay and Keysend.

reneaaron (commented):
Isn't it a bit strange that every consumer of this component has to implement those checks again? 

Would it be more consistent to add an optional field for the current balance to the `DualCurrencyField` component and check for those values in there?

lujakob:
> Isn't it a bit strange that every consumer of this component has to implement those checks again?
> 
> Would it be more consistent to add an optional field for the current balance to the `DualCurrencyField` component and check for those values in there?

I'm not really sure what you mean. Passing the `auth?.account?.balance` to the `DualCurrencyField` instead of calculating `const amountExceeded = +amountSat > (auth?.account?.balance || 0);` in the parent and passing `amountExceeded` into `DualCurrencyField`?

I think a problem is we need `amountExceeded` in the parent as well to disable the button.

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -85,8 +91,13 @@ export default function DualCurrencyField({
           {label}
         </label>
         {(min || max) && (
-          <span className="text-xs text-gray-700 dark:text-neutral-400">
-            <RangeLabel min={min} max={max} /> sats
+          <span
+            className={classNames(
+              "text-xs text-gray-700 dark:text-neutral-400",
+              (!!min || !!max) && !!rangeExceeded && "text-red-500"
```
rolznz, line 97: Why is `(!!min || !!max)` check needed here? it is already checked in the enclosing scope (line 93)

rolznz: You shouldn't need to do `!!` on a boolean? values

rolznz (approved):

reneaaron (approved):
tACK

rolznz:
@reneaaron sorry, I mean it doesn't accept `undefined` as false. Some other classnames libraries like https://www.npmjs.com/package/clsx do

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -125,7 +138,12 @@ export default function DualCurrencyField({
         )}
       </div>
       {hint && (
-        <p className="my-1 text-xs text-gray-700 dark:text-neutral-400">
+        <p
+          className={classNames(
+            "my-1 text-xs text-gray-700 dark:text-neutral-400",
+            !!amountExceeded && "text-red-500 dark:text-red-500"
```
rolznz, line 144: why is the dark: version the same? can't it be omitted? (multiple places in this file)

src/app/components/form/DualCurrencyField/index.tsx
lujakob: The reason why I had `(!!min || !!max)` here is, in Keysend component I have `const rangeExceeded = +amountSat < amountMin;` - `amountMin` is `1` and if for some reason `amountSat` would be `undefined` I would end up with `+undefined < 1` which is `false`. This cased a small amount of uncertainty and I wanted to make sure this rangeExceeded is only being applied when there actually is a min or max value. But probably it is not needed.

src/app/components/form/DualCurrencyField/index.tsx
reneaaron: no, because the dark:text-neutral-400 class supersedes the dark:text-red-500 if it's not set.

reneaaron (approved):
tACK

PR #2448, opened 2023-05-23
feat: get payments from connector; add interface; add lndhub implemen…

### LINKED ISSUE #2434: Load outgoing transactions from connectors
### Feature description

Currently all outgoing transactions are loaded from the local database, while incoming transactions are loaded from the connector. This leads to some confusion around transaction lists.

### Describe the solution

Load outgoing transactions from the connectors and display them in the extension. Enhance transactions loaded from the connector with data stored in the local DB (e.g. the website this payment has been made on, if available)

Existing connectors need to be enhanced with an `getTransactions` method that loads all outgoing transactions from the respective source. Merging this info with data from the local database (using the preimage or a unique transaction id) should be done outside of the connectors. 

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

lujakob:
…tation

Fetch transactions data from connectors

Fixes #2434

### Type of change

### How has this been tested?

### Checklist

### DIFF
--- a/src/extension/background-script/actions/payments/listByAccount.ts
+++ b/src/extension/background-script/actions/payments/listByAccount.ts
@@ -1,24 +1,64 @@
-import type { MessagePaymentListByAccount } from "~/types";
+import { ConnectorTransaction } from "~/extension/background-script/connectors/connector.interface";
+import state from "~/extension/background-script/state";
+import type {
+  DbPayment,
+  MessagePaymentListByAccount,
+  Transaction,
+} from "~/types";
 
 import db from "../../db";
 
+// @Todo: renmae, due to fact that "byAccount" will be default
+
 const listByAccount = async (message: MessagePaymentListByAccount) => {
-  const { accountId } = message.args;
+  const accountId = state.getState().currentAccountId;
+
   // TODO: Add pagination instead of limiting to 2121
   const limit = message?.args?.limit || 2121;
 
-  const payments = await db.payments
+  const connector = await state.getState().getConnector();
+
+  // get IndexDB payments
+  const dbPayments = await db.payments
     .toCollection()
     .and((p) => p.accountId === accountId)
     .limit(limit)
     .reverse()
     .sortBy("createdAt");
 
-  return {
-    data: {
-      payments,
-    },
-  };
+  try {
+    const result = await connector.getTransactions();
+
+    // map IndexDB payment data onto ConnectorTransaction
+    const transactions: Transaction[] = result.data.transactions.map(
+      (connectorPayment: ConnectorTransaction) => {
+        const dbPayment = dbPayments.find(
+          (p: DbPayment) => p.preimage === connectorPayment.payment_preimage
+        );
+        // @Todo: map transactions correctly
+
+        return {
+          ...dbPayment,
+          id: "",
+          amount: connectorPayment.value + "",
+          date: "",
+          title: "",
+        } as Transaction;
+      }
+    );
+
+    return {
+      data: {
+        // @Todo: adapt property name
+        payments: transactions,
+      },
+    };
+  } catch (e) {
+    console.error(e);
+    if (e instanceof Error) {
+      return { error: e.message };
+    }
+  }
 };
 
 export default listByAccount;
--- a/src/extension/background-script/connectors/citadel.ts
+++ b/src/extension/background-script/connectors/citadel.ts
@@ -7,6 +7,7 @@ import Connector, {
   GetBalanceResponse,
   GetInfoResponse,
   GetInvoicesResponse,
+  GetTransactionsResponse,
   KeysendArgs,
   MakeInvoiceArgs,
   MakeInvoiceResponse,
@@ -77,6 +78,14 @@ class CitadelConnector implements Connector {
     throw new Error(
       `${this.constructor.name}: "getInvoices" is not yet supported. Contact us if you need it.`
     );
+  }
+
+  // @Todo: implement function call
+  async getTransactions(): Promise<GetTransactionsResponse> {
+    console.error(
+      `${this.constructor.name} has not implementation for getTransactions yet`
+    );
+    throw new Error("Not yet implemented with the currently used connector.");
   }
 
   // not yet implemented
--- a/src/extension/background-script/connectors/connector.interface.ts
+++ b/src/extension/background-script/connectors/connector.interface.ts
@@ -26,6 +26,14 @@ export interface ConnectorInvoice {
   type: "received";
 }
 
+// @Todo: extend missing properties accordingly
+export interface ConnectorTransaction {
+  r_hash: string;
+  memo: string;
+  value: number;
+  payment_preimage: string;
+}
+
 export interface MakeInvoiceArgs {
   amount: string | number;
   memo: string;
@@ -52,6 +60,12 @@ export type GetBalanceResponse = {
 export type GetInvoicesResponse = {
   data: {
     invoices: ConnectorInvoice[];
+  };
+};
+
+export type GetTransactionsResponse = {
+  data: {
+    transactions: ConnectorTransaction[];
   };
 };
 
@@ -115,6 +129,7 @@ export default interface Connector {
   getInfo(): Promise<GetInfoResponse>;
   getBalance(): Promise<GetBalanceResponse>;
   getInvoices(): Promise<GetInvoicesResponse>;
+  getTransactions(): Promise<GetTransactionsResponse>;
   makeInvoice(args: MakeInvoiceArgs): Promise<MakeInvoiceResponse>;
   sendPayment(args: SendPaymentArgs): Promise<SendPaymentResponse>;
   keysend(args: KeysendArgs): Promise<SendPaymentResponse>;
--- a/src/extension/background-script/connectors/lndhub.ts
+++ b/src/extension/background-script/connectors/lndhub.ts
@@ -15,10 +15,12 @@ import Connector, {
   CheckPaymentArgs,
   CheckPaymentResponse,
   ConnectorInvoice,
+  ConnectorTransaction,
   ConnectPeerResponse,
   GetBalanceResponse,
   GetInfoResponse,
   GetInvoicesResponse,
+  GetTransactionsResponse,
   KeysendArgs,
   MakeInvoiceArgs,
   MakeInvoiceResponse,
@@ -119,6 +121,39 @@ export default class LndHub implements Connector {
     return {
       data: {
         invoices,
+      },
+    };
+  }
+
+  async getTransactions(): Promise<GetTransactionsResponse> {
+    // this connector endpoint has a fixed result set of 100 entries
+    const data = await this.request<
+      {
+        r_hash: string;
+        payment_hash: string;
+        payment_preimage: string;
+        value: number;
+        type: string;
+        fee: number;
+        timestamp: number;
+        memo: string;
+        keysend: boolean;
+        custom_records: Record<number, number>;
+      }[]
+    >("GET", "/gettxs", undefined);
+
+    const transactions: ConnectorTransaction[] = data.map(
+      (p): ConnectorTransaction => ({
+        r_hash: p.r_hash,
+        payment_preimage: p.payment_preimage,
+        memo: p.memo,
+        value: p.value,
+      })
+    );
+
+    return {
+      data: {
+        transactions,
       },
     };
   }
--- a/src/extension/background-script/router.ts
+++ b/src/extension/background-script/router.ts
@@ -32,7 +32,7 @@ const routes = {
   signMessage: ln.signMessage,
   makeInvoice: ln.makeInvoice,
   connectPeer: ln.connectPeer,
-  getPayments: payments.all,
+  getPayments: payments.listByAccount,
   getPaymentsByAccount: payments.listByAccount,
   accountInfo: accounts.info,
   accountDecryptedDetails: accounts.decryptedDetails,
--- a/src/extension/background-script/connectors/commando.ts
+++ b/src/extension/background-script/connectors/commando.ts
@@ -13,6 +13,7 @@ import Connector, {
   GetBalanceResponse,
   GetInfoResponse,
   GetInvoicesResponse,
+  GetTransactionsResponse,
   KeysendArgs,
   MakeInvoiceArgs,
   MakeInvoiceResponse,
@@ -210,6 +211,14 @@ export default class Commando implements Connector {
           },
         };
       });
+  }
+
+  // @Todo: implement function call
+  async getTransactions(): Promise<GetTransactionsResponse> {
+    console.error(
+      `${this.constructor.name} has not implementation for getTransactions yet`
+    );
+    throw new Error("Not yet implemented with the currently used connector.");
   }
 
   async getInfo(): Promise<GetInfoResponse> {
--- a/src/extension/background-script/connectors/connector.example
+++ b/src/extension/background-script/connectors/connector.example
@@ -57,6 +57,14 @@ class ConnectorExample implements Connector {
   makeInvoice(args: MakeInvoiceArgs): Promise<MakeInvoiceResponse> {
     // Add your own implementation.
   }
+
+  getInvoices(): Promise<GetInvoicesResponse> {
+    // Add your own implementation.
+  }
+
+  getTransactions(): Promise<GetTransactionsResponse> {
+    // Add your own implementation.
+  }
 }
 
 export default ConnectorExample;
--- a/src/extension/background-script/connectors/eclair.ts
+++ b/src/extension/background-script/connectors/eclair.ts
@@ -10,6 +10,7 @@ import Connector, {
   GetBalanceResponse,
   GetInfoResponse,
   GetInvoicesResponse,
+  GetTransactionsResponse,
   KeysendArgs,
   MakeInvoiceArgs,
   MakeInvoiceResponse,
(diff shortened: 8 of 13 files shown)


lujakob:
This is a WIP draft. Connector inferface change, types and placeholder methods for all connectors. Alby lndhub has a simple POC implementation.

bumi:
this will be so greatly needed and will solve a lot of confusion!

rolznz:
@lujakob do you need any help with this one?

lujakob:
> @lujakob do you need any help with this one?

I'll check this on Monday and get back to you, thanks.

lujakob:
@rolznz @reneaaron @bumi @im-adithya 

I refined my working draft and it is at a state where the outgoing transactions are fetched from the connector and IndexDB payments data are being mapped onto them. This currently works only for the Alby Lndhub connector.

Most of the magic is happening in 

https://github.com/lujakob/lightning-browser-extension/blob/44af2167d65ff2f3c59643613789105ad2ef155d/src/extension/background-script/actions/payments/listByAccount.ts

I have some uncertainties and questions, mostly regarding properties and mapping them.

* I created a ConnectorTransaction interface. Right now the properties are taken from the lndhub.go `gettxs` response. I could imagine some of the connectors response have differences here. Which properties should be mandatory which optinal? Any clues? https://github.com/lujakob/lightning-browser-extension/blob/4fb06da930b721095a86120643ced30c423aab76/src/extension/background-script/connectors/connector.interface.ts#L30-L47
* Which properties from `DbPayment` should be mapped onto it? https://github.com/lujakob/lightning-browser-extension/blob/4fb06da930b721095a86120643ced30c423aab76/src/types.ts#L600-L615
* Are we going to deprecate and remove some of them in a database migration?
* How are we going to apply/map "title" and "description" and "memo"?
* See `@Todo` comments

So most of my uncertainties revolve around those data mappings. I guess this also was the reason for my procrastination of this issue ;-)

Would be great if you could have a POC test and give me some ideas/suggestions in general and also about the property mappings in general. Thx.

src/extension/background-script/router.ts
```diff
@@ -32,7 +32,7 @@ const routes = {
   signMessage: ln.signMessage,
   makeInvoice: ln.makeInvoice,
   connectPeer: ln.connectPeer,
-  getPayments: payments.all,
+  getPayments: payments.listByAccount,
```
rolznz, line 35: This now does the same as `getPaymentsByAccount`. Should one of these be removed?

src/extension/background-script/connectors/citadel.ts
```diff
@@ -79,6 +80,14 @@ class CitadelConnector implements Connector {
     );
   }
 
+  // @Todo: implement function call
+  async getTransactions(): Promise<GetTransactionsResponse> {
+    console.error(
```
rolznz, line 85: I think we should not implement all these functions in a single PR.

Maybe we could just do a console warning and return an empty list for the other connectors (their functionality can stay unchanged )?

src/extension/background-script/actions/payments/listByAccount.ts
```diff
@@ -1,24 +1,71 @@
...
-  return {
-    data: {
-      payments,
-    },
-  };
+  try {
+    const result = await connector.getTransactions();
+
+    // map IndexDB payment data onto ConnectorTransaction
+    const transactions: Transaction[] = result.data.transactions.map(
```
rolznz, line 33: Because there are currently limits on the number of invoices (and TBC: maybe we do not do the getTransactions() implementation for all connectors) maybe we should create a list from both the DB and the result of getTransactions, only merging the transactions that exist in both. What do you think?

src/extension/background-script/router.ts
lujakob: Exactly, it will be removed.

src/extension/background-script/connectors/lndhub.ts
```diff
@@ -123,6 +125,51 @@ export default class LndHub implements Connector {
...
+        payment_preimage: string;
+        r_hash: {
+          data: number[];
+          type: string;
+        };
+        timestamp: number;
+        type: string;
+        value: number;
+      }[]
+    >("GET", "/gettxs", undefined);
```
rolznz, line 149: I think we should use the v2 endpoint (invoices/outgoing): https://ln.getalby.com/swagger/index.html#/Invoice/get_v2_invoices_outgoing

rolznz: EDIT: sorry, this would only work for Alby

src/extension/background-script/connectors/citadel.ts
lujakob: ... and already merge into master? How would it work until finishing it, one connector uses the new logic of fetching transactions from the connector all of the rest fetch from IndexDB as currently is?

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -26,6 +26,26 @@ export interface ConnectorInvoice {
   type: "received";
 }
 
+// @Todo: extend missing properties accordingly
+export interface ConnectorTransaction {
+  custom_records: Record<number, number>;
+  fee: number;
+  keysend: boolean;
+  memo: string;
+  payment_hash: {
```
rolznz, line 35: I think payment_hash and r_hash should just be a string. Can they be converted in the lndhub connector?

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -26,6 +26,26 @@ export interface ConnectorInvoice {
   type: "received";
 }
 
+// @Todo: extend missing properties accordingly
+export interface ConnectorTransaction {
```
rolznz, line 30: Maybe as many properties as possible should be shared with ConnectorInvoice (they could be extracted to a separate interface that both inherit from).

That way we don't have inconsistencies with the naming (e.g. preimage, amount, etc).

src/extension/background-script/connectors/connector.interface.ts
rolznz: maybe only the payment_hash is needed (I think they are both the same thing)

src/extension/background-script/actions/payments/listByAccount.ts
lujakob: @reneaaron mentioned this in the last community call. I think this gets to confusing for the user.  E.g.: the first 100 (which is the result set for many unpaginated connector result sets) comes from the connector, the rest from IndexDB. Users might wonder why older transactions are missing? Also explaining this in the UI is not really cool.

Talking about the pagination topic René mentioned he is quite surprised the missing pagination for incoming transactions was never really an issue for users so far. It seems only latest are relevant.

I would tend to implement as is, implement pagination for connectors that support it (our lndhub.go doesn't but it's in our control), LND does support it, and implement the rest as it will be supported in the future. 

@reneaaron @rolznz what do you think?

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -26,6 +26,26 @@ export interface ConnectorInvoice {
...
+    data: number[];
+    type: string;
+  };
+  payment_preimage: string;
+  r_hash: {
+    data: number[];
+    type: string;
+  };
+  timestamp: number;
+  type: string;
```
rolznz, line 45: what is the type field for?

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -26,6 +26,26 @@ export interface ConnectorInvoice {
   type: "received";
 }
 
+// @Todo: extend missing properties accordingly
+export interface ConnectorTransaction {
+  custom_records: Record<number, number>;
```
rolznz, line 31: I think this should be optional, probably not all connectors support it

src/extension/background-script/connectors/connector.interface.ts
lujakob: I just implemented this one (as all other fields) because it is part of the response.

src/extension/background-script/connectors/connector.interface.ts
lujakob: Hmm, not sure if this makes any sense. We don't have a choice in the namings, they are dictated by the connector API results.

src/extension/background-script/actions/payments/listByAccount.ts
rolznz: Currently we see no transactions at all that happened outside of the extension, so I think seeing the 100 latest transactions is a huge improvement.

I am just worried if we try implement all the connectors it will delay this getting merged. I think we should focus on the most important connectors and improve in small steps, especially when we have to rely on other dependencies to also be updated.

src/extension/background-script/connectors/connector.interface.ts
rolznz: This is the common connector interface, right? the individual connectors should map their results to this interface.

src/extension/background-script/connectors/citadel.ts
rolznz: Yes. Or we change the merging strategy (my other comment)

src/extension/background-script/connectors/connector.interface.ts
lujakob: So far they are not mapped, because they are not used in the mapped Transaction interface.

https://github.com/lujakob/lightning-browser-extension/blob/4fb06da930b721095a86120643ced30c423aab76/src/extension/background-script/actions/payments/listByAccount.ts#L44-L53

src/extension/background-script/actions/payments/listByAccount.ts
```diff
@@ -1,24 +1,71 @@
...
+        );
+
+        // @Todo: map transactions correctly
+        // @Todo: how to we use/map "title" and "description" and "memo"
+
+        // @Todo: this comes from convertPaymentToTransaction function - remove there if this is right
+        const title = dbPayment ? dbPayment.name || dbPayment.description : "";
+        return {
+          ...(dbPayment ? dbPayment : {}),
+          id: "",
```
rolznz, line 46: I guess the id should come from the transaction if it exists, otherwise use the dbPayment id

src/extension/background-script/router.ts
lujakob: I removed it now.

src/extension/background-script/actions/payments/listByAccount.ts
lujakob: Transactions don't have ids I think. I explicitly use dbPayment.id now. I think the original issue was `DbPayment` has id optional and `Transaction` not. Maybe this can be simplified if it turns out the interfaces can be optimized.

src/extension/background-script/actions/payments/list.ts
```diff
@@ -0,0 +1,65 @@
...
+      (transaction: ConnectorTransaction) => {
+        const dbPayment = dbPayments.find(
+          (p: DbPayment) => p.preimage === transaction.payment_preimage
+        );
+
+        // @Todo: map transactions correctly
+        // @Todo: how to we use/map "title" and "description" and "memo"
+
+        // @Todo: this comes from convertPaymentToTransaction function - remove there if this is right
+        const title = dbPayment ? dbPayment.name || dbPayment.description : "";
```
rolznz, line 37: It seems weird to me that the title is set to the description. I know this is existing code. Maybe it can be improved in a later PR.

src/extension/background-script/actions/payments/list.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    // map IndexDB payment data onto ConnectorTransaction
+    const transactions: Transaction[] = result.data.transactions.map(
+      (transaction: ConnectorTransaction) => {
+        const dbPayment = dbPayments.find(
+          (p: DbPayment) => p.preimage === transaction.payment_preimage
+        );
+
+        // @Todo: map transactions correctly
+        // @Todo: how to we use/map "title" and "description" and "memo"
```
rolznz, line 34: I think in most cases there is only one field (either memo or description or comment). I think we should try to have a single field if possible? it is called `description` in the Transaction type right now, right?

lujakob: Exactly...

https://github.com/lujakob/lightning-browser-extension/blob/aab05df5550c57b9c771dbe52ba0abab1dd41297/src/types.ts#L577

src/extension/background-script/connectors/connector.interface.ts
reneaaron: I also understood it the way Roland does: All of the `getTransactions` implementations would return an array of `ConnectorTransaction` that maps any fields that differ in the connectors to this one common interface. 

Both `ConnectorTransaction` and `ConnectorInvoice` could implement a `ITransaction` interface to enforce the same naming of fields that are available on both types.

lujakob: > Hmm, not sure if this makes any sense. We don't have a choice in the namings, they are dictated by the connector API results.

I forgot, the mapping happens here yes...

https://github.com/lujakob/lightning-browser-extension/blob/aab05df5550c57b9c771dbe52ba0abab1dd41297/src/extension/background-script/connectors/lndhub.ts#L152-L163

rolznz: If possible it would be good to add more properties there - amount, date, id, custom_records. An (incoming) invoice and transaction should have many similar properties. I guess the fee only applies to transactions, and the type should be a different constant value ("received" for invoices and "sent" for transactions), everything else could be moved to the `ITransaction` interface?

lujakob:
```
export interface ConnectorInvoice extends ITransaction {
  custom_records?: {
    "696969": string;
    "7629169": string;
    "5482373484": string;
  };
  id: string;
  settled: boolean;
  settleDate: number;
  totalAmount: string;
  type: "received";
}

// @Todo: extend missing properties accordingly
export interface ConnectorTransaction extends ITransaction {
  // @Todo: can this be the same custom type as in ConnectorInvoice above?
  custom_records?: Record<number, number>;
  fee: number;
  keysend: boolean;
  timestamp: number;
  type: string;
  value: number;
}

* for `custom_records` which type of the two should I use?
* `ConnectorTransaction` has no `id`
* can I make `ConnectorInvoice settleDate` and `ConnectorTransaction timestamp` into a shared `date`?

That's all the similarities I see..
 

```

src/extension/background-script/connectors/connector.interface.ts
lujakob: Outgoing and Incoming are not shown in a single list, they have separate lists with tabs. Is this intended to be changed? How should I proceed, just remove it?

src/extension/background-script/actions/payments/listByAccount.ts
lujakob: @reneaaron @rolznz what do you propose? Have two concurrent logics, getting transactions from the connector for lndhub (as a first implementation) and all others with the current logic of getting data from IndexDB? I think this restricts us in refactoring the code as well. @im-adithya offered to help out implementing the connectors. What do you guys think?

lujakob: I see your point, but I don't feel good about having separate logics for connectors and as well would like to have the freedom in refactoring the interfaces/data in the transactions list to make it clean.

src/extension/background-script/connectors/connector.interface.ts
rolznz: maybe it could be explicitly sent to "sent" instead of a string for now.

src/extension/background-script/connectors/connector.interface.ts
rolznz: Maybe ```
custom_records?: {
    "696969": string;
    "7629169": string;
    "5482373484": string;
  } & Record<string, unknown>;
``` could be used. It should cover all cases.

I am not sure regarding the timestamp (whether it is the settled date or the created date?). It might be good to have both dates. For invoices, only settled invoices will show in the extension. But for transactions, we might want to show unsettled/pending transactions. So I think `settled` and `settledDate` could be moved to ITransaction.

File: src/extension/background-script/connectors/citadel.ts (hunk 2 above)
Comment (lujakob) [reply src/extension/background-script/connectors/citadel.ts:85]: In the community call we decided to try to implement all in one PR for now, so I'll resolve this conversation.

**Comment by lujakob** [comment]
@reneaaron @rolznz I added the `getTransactions` for the Alby connector, but I realized I am really confused with the definitions of the interfaces. I have no clarity about the outgoing transactions interfaces. I need to have a clear idea of required properties from a UI side that align with the IndexDB payments we already have. So I will merge the required properties of IndexDB payments and connector outgoing transactions into a new interface `ConnectorTransaction` that inherits from `ITransaction` (we also have `ConnectorInvoice`)

If this is clear, I can just map the available properties that I get from the `getTransactions` of the different connectors.

```
export interface ITransaction {
  totalAmount: string;
  memo: string;
  preimage: string;
  settled: boolean;
  settleDate: number;
  custom_records?: TransactionCustomRecords;
}

export interface ConnectorInvoice extends ITransaction {
  id: string;
  type: "received";
}

export interface ConnectorTransaction extends ITransaction {
  totalFee: number;
  keysend: boolean;
  timestamp: number;
  type: "sent";
}
```

* Which properties are mandatory from the connectors data? 
* which we take from IndexDB payments?
* can we remove some deprecated IndexDB payments?
* Do we have "description" and "memo" or is this the same? In case, which name do we prefer?
* In IndexDB payments we have totalAmount and totalFees, should we keep those names?
* alby sdk response gives me "amount" and "value", which one should I use?

I'm not sure what makes sense, if some of you can define the interfaces or should we team up and have a call to sort this out?

itstomekk:
@rolznz should we move it to backlog?

rolznz:
@itstomekk no, this is an important task. I will try to prioritize it in the next few weeks.

reneaaron:
Closing as this PR is stale and has been implemented in #2888

PR #2453, opened 2023-05-28
Update artifacts-comment.yml

bumi:
remove "powered by" - it was not used and could lead to potential link injections.


reneaaron:
Okay, sad to see this go :wink: Don't forget to delete the fly app, too.

PR #2454, opened 2023-05-29
fix: upgrade playwright

reneaaron:

Fix error in test runs caused by playwright.


im-adithya:
I think you forgot to commit the `package.json` file

PR #2467, opened 2023-06-09
feat: add liquid

### LINKED ISSUE #2074: [Feature] Signing functionality for Liquid

im-adithya:

Extends the mnemonic for use in liquid btc.

Fixes #2074

### Type of change

### How has this been tested?

Manually

### Checklist


src/extension/background-script/liquid/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  getPublicKey() {
+    const publicKey = secp256k1.schnorr.getPublicKey(
+      secp256k1.utils.hexToBytes(this.privateKey)
+    );
+    const publicKeyHex = secp256k1.utils.bytesToHex(publicKey);
+    return publicKeyHex;
+  }
+
+  async signSchnorr(sigHash: string): Promise<string> {
```
rolznz, line 18: It would be great to add signSchnorr for webbtc. I can add it once this PR is merged

src/extension/content-script/onstart.ts
```diff
@@ -18,10 +19,14 @@ async function onstart() {
   injectScript(browser.runtime.getURL("js/inpageScriptWebBTC.bundle.js"));
 
   // window.nostr
-  const nostrEnabled = (await api.getAccount()).nostrEnabled;
-  if (nostrEnabled) {
+  if (accountDetails.nostrEnabled) {
     injectScript(browser.runtime.getURL("js/inpageScriptNostr.bundle.js"));
   }
+
+  // window.secp256k1
```
rolznz, line 26: should this be window.liquid?

src/i18n/locales/en/translation.json
```diff
@@ -453,6 +454,47 @@
             "title": "Your Secret Key"
           }
         },
+        "liquid": {
+          "title": "Liquid",
+          "hint": "have to fill this.",
```
rolznz, line 459: fix hint

src/i18n/locales/en/translation.json
```diff
@@ -453,6 +454,47 @@
...
+          "advanced_settings": {
+            "label": "Advanced Settings",
```
rolznz, line 461: most or all this is the same as Nostr right? maybe we should consider using a variable {coinType} or something instead, since you made a shared component?

src/i18n/locales/en/translation.json
```diff
@@ -825,6 +867,18 @@
       "allow": "Allow this website to execute:",
       "always_allow": "Remember my choice and don't ask again"
     },
+    "liquid": {
+      "title": "Nostr",
```
rolznz, line 871: should be Liquid

src/i18n/locales/en/translation.json
```diff
@@ -1053,6 +1107,12 @@
       "nip04decrypt": "Decrypt data.",
       "signmessage": "Sign message with your key."
     },
+    "liquid": {
+      "getpublickey": "Read your public key.",
+      "nip04encrypt": "Encrypt data.",
```
rolznz, line 1112: nip keys should not be needed for liquid?

src/app/screens/Accounts/AdvancedSettings/index.tsx
```diff
@@ -12,35 +12,35 @@ import Button from "~/app/components/Button";
 import InputCopyButton from "~/app/components/InputCopyButton";
 import TextField from "~/app/components/form/TextField";
 import { useAccount } from "~/app/context/AccountContext";
+import { KeyOrigin, getKeyOrigin } from "~/app/utils/getKeyOrigin";
```
rolznz, line 15: maybe this file/component should be called `AdvancedKeySettings` or something. It's for Nostr and Liquid key management, right?

src/extension/content-script/onstart.ts
im-adithya: Ah, this is decided by Bumi and the person from Ionio (Tiero?) // @bumi

src/i18n/locales/en/translation.json
im-adithya: yeah I was thinking to make one but missed a few things which were to be shared as there were lot of files haha, thanks for reminding 🙌

src/app/screens/Accounts/AdvancedSettings/index.tsx
im-adithya: Yup was thinking the same, will change 👍

rolznz (approved):
~~utACK~~

src/extension/content-script/onstart.ts
rolznz: But it will always use the liquid key? won't webbtc also have signSchnorr?

src/common/lib/mnemonic.ts
```diff
@@ -2,10 +2,15 @@ import * as secp256k1 from "@noble/secp256k1";
 import { HDKey } from "@scure/bip32";
 import * as bip39 from "@scure/bip39";
 
+export const LIQUID_DERIVATION_PATH = "m/44'/1776'/0'/0/0";
```
rolznz, line 5: We should consider adding a _REGTEST version too which will be controlled by the network setting.

im-adithya:
Converting this into WIP PR

src/extension/background-script/mnemonic/index.ts
```diff
@@ -3,7 +3,11 @@ import { HDKey } from "@scure/bip32";
 import * as bip39 from "@scure/bip39";
 import Hex from "crypto-js/enc-hex";
 import sha256 from "crypto-js/sha256";
+import { SLIP77Factory } from "slip77";
+import * as tinysecp from "tiny-secp256k1";
```
louisinger, line 7:
```suggestion
import * as tinysecp from "../liquid/secp256k1";
```

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -0,0 +1,232 @@
...
+import {
+  Address,
+  PsetPreview,
+  fetchAssetRegistry,
+  getPsetPreview,
+} from "~/common/lib/pset";
+import type { OriginData } from "~/types";
+
+function toLiquidNetworkName(net: "bitcoin" | "testnet" | "regtest") {
+  return net === "bitcoin" ? "liquid" : "testnet"; // TODO: how do we handle regtest ?
```
rolznz, line 27: this can be updated now 🎉

src/types.ts
```diff
@@ -546,6 +560,22 @@ export interface MessageGetAddress extends MessageDefault {
...
+  args: {
+    pset: string;
+  };
+  action: "signPsetWithPrompt";
+}
+
+export interface MessageSignPset extends MessageDefault {
+  args: {
+    pset: string;
+    network: "liquid" | "testnet" | "regtest";
```
rolznz, line 574: can we use `LiquidNetworkType` here?

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -0,0 +1,232 @@
...
+import msg from "~/common/lib/msg";
+import {
+  Address,
+  PsetPreview,
+  fetchAssetRegistry,
+  getPsetPreview,
+} from "~/common/lib/pset";
+import type { OriginData } from "~/types";
+
+function toLiquidNetworkName(net: "bitcoin" | "testnet" | "regtest") {
```
rolznz, line 26: can we use the BitcoinNetworkType here?

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: Edit: if "regtest" is unsupported maybe it would be better to throw an exception here?

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -0,0 +1,232 @@
...
+  const pset = navState.args?.pset;
+  const origin = navState.origin as OriginData;
+  const [loading, setLoading] = useState(false);
+  const [successMessage, setSuccessMessage] = useState("");
+  const [preview, setPreview] = useState<PsetPreview | undefined>(undefined);
+  const [assetRegistry, setAssetRegistry] = useState<
+    Record<string, EsploraAssetInfos>
+  >({});
+  const [liquidNetwork, setLiquidNetwork] = useState<
+    "liquid" | "testnet" | "regtest"
```
rolznz, line 47: can we use `LiquidNetworkType` here?

src/common/lib/liquid.ts
```diff
@@ -0,0 +1,14 @@
+import { schnorr } from "@noble/curves/secp256k1";
+import * as secp256k1 from "@noble/secp256k1";
+
+const liquid = {
+  generatePublicKey(privateKey: string) {
```
rolznz, line 5: This is no longer used right? can we remove this file?

src/common/lib/pset.ts
```diff
@@ -0,0 +1,155 @@
...
+  address: string;
+  asset: string;
+};
+
+export type PsetPreview = {
+  inputs: Address[];
+  outputs: Address[];
+};
+
+export function getPsetPreview(
```
rolznz, line 22: @im-adithya  After our review of the mnemonic feature we moved a lot of code to the background script and tried to avoid introducing new lib files where possible. Is it possible to move `getPsetPreview` to `background-script/liquid` and add an action for it?

src/extension/background-script/mnemonic/index.ts
```diff
@@ -3,7 +3,12 @@ import { HDKey } from "@scure/bip32";
 import * as bip39 from "@scure/bip39";
 import Hex from "crypto-js/enc-hex";
 import sha256 from "crypto-js/sha256";
+import { SLIP77Factory } from "slip77";
 
+import * as tinysecp from "../liquid/secp256k1";
+
+export const LIQUID_DERIVATION_PATH = "m/84'/1776'/0'/0/0";
```
rolznz, line 10: Can these constants be moved to `src/extension/background-script/liquid/index.ts`?

src/extension/background-script/mnemonic/index.ts
```diff
@@ -16,6 +21,23 @@ class Mnemonic {
     this._hdkey = HDKey.fromMasterSeed(seed);
   }
 
+  deriveLiquidMasterBlindingKey(): string {
```
rolznz, line 24: Can these liquid methods be moved to `src/extension/background-script/liquid/index.ts`?

src/extension/background-script/state.ts
```diff
@@ -133,6 +137,21 @@ const state = createState<State>((set, get) => ({
...
+    if (currentLiquid) {
+      return currentLiquid;
+    }
+    const mnemonic = await get().getMnemonic();
+    const currentAccountId = get().currentAccountId as string;
+    const account = get().accounts[currentAccountId];
+
+    const isMainnet = account.bitcoinNetwork === "bitcoin";
+
+    const liquid = new Liquid(mnemonic, isMainnet ? "liquid" : "testnet");
```
rolznz, line 151: we have this check in a few places, it would be nice to move to one utility function if possible (in case we support regtest later)

src/extension/providers/liquid/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+    return await this.execute("enable");
+  }
+
+  async getAddress() {
+    await this.enable();
+    return await this.execute("getAddressOrPrompt");
+  }
+
+  async signPset(pset: string) {
```
rolznz, line 28: @im-adithya I cannot see signSchnorr here. Do we want to expose it? otherwise should we remove the confirmation dialogs?

src/i18n/locales/en/translation.json
```diff
@@ -386,11 +386,13 @@
           "export_uri": "LNDHub Credentials URI",
           "scan_qr": "Import this wallet into Zeus or BlueWallet by scanning the QRCode."
         },
+        "no_mnemonic_hint": "💡 You don't have a secret key yet. <0>Click here</0></div> to create your secret key first to unlock these settings.",
         "mnemonic": {
           "title": "Secret Key",
           "description": "Your Account Secret Key allows you to use Alby to interact with protocols such as Nostr.",
           "description2": "Your Secret Key is a set of 12 words that will allow you to access your keys to protocols such as Nostr on a new device or in case you lose access to your account.",
           "saved": "Secret Key encrypted & saved successfully.",
+          "existing_liquid_key_notice": "⚠️ This account already has a liquid private key set and will not be derived from this secret key. You can manage your liquid key from your account settings.",
```
rolznz, line 395: This is no longer true, right?

src/i18n/locales/en/translation.json
```diff
@@ -434,6 +436,15 @@
             }
           }
         },
+        "liquid": {
+          "enable": {
+            "title": "Liquid Network",
+            "description": "Enable interactions with Liquid Network apps and protocols."
+          },
+          "errors": {
+            "failed_to_load": "Failed to load the Liquid key. Is it a valid Liquid key?"
```
rolznz, line 445: This is no longer possible, right?

src/i18n/locales/en/translation.json
```diff
@@ -771,6 +782,30 @@
       "title": "Sign",
       "content": "This website asks you to sign:"
     },
+    "confirm_sign_psbt": {
```
rolznz, line 785: Sign PSBT is not supported yet, I think this section should be removed

src/types.ts
```diff
@@ -676,6 +706,11 @@ export interface Payment extends Omit<DbPayment, "id"> {
   id: number;
 }
 
+export enum PermissionMethodLiquid {
+  LIQUID_SIGNSCHNORR = "liquid/signSchnorr",
```
rolznz, line 710: what about sign PSET?

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: @louisinger could you please review this one?

louisinger: @rolznz do u mean to support regtest ? If yes, it's fine to return "regtest" here: everything else should be unchanged (it's just a matter of address encoding)

src/types.ts
rolznz: @louisinger could you link me to the line of code in the source? how do I import it?

src/types.ts
rolznz: ah, we always prompt for signing

src/types.ts
louisinger: sorry, my bad, we removed that type some time ago

src/extension/providers/liquid/index.ts
louisinger: +1 to do not expose signSchnorr not needed thanks to signPset and could be dangerous

src/types.ts
```diff
@@ -676,6 +706,11 @@ export interface Payment extends Omit<DbPayment, "id"> {
...
+  LIQUID_GETPUBLICKEY = "liquid/getPublicKey",
```
louisinger, line 711: maybe we should rename to `LIQUID_GETADDRESS` ? (getPublicKey -> getAddress)

src/extension/providers/liquid/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  interface Window {
+    liquid: LiquidProvider;
+  }
+}
+
+export default class LiquidProvider {
+  enabled: boolean;
+
+  constructor() {
+    this.enabled = false;
```
louisinger, line 13: any reason to initialize to `false`? even if the user has enabled the app

src/extension/background-script/actions/liquid/getAddressOrPrompt.ts
```diff
@@ -0,0 +1,70 @@
+import utils from "~/common/lib/utils";
+import type { MessageGetLiquidAddress } from "~/types";
+import { PermissionMethodLiquid } from "~/types";
+
+import state from "../../state";
+import { addPermissionFor, hasPermissionFor } from "../nostr/helpers";
```
louisinger, line 6: nit: depending on `nostr` can be misleading, perhaps we could move these helpers into a common module?

louisinger (approved):

src/extension/providers/liquid/index.ts
rolznz: Thanks @louisinger for confirmation. I will remove it along with the confirmation dialogs in the other PR.

src/types.ts
rolznz: fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/i18n/locales/en/translation.json
rolznz: Fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/i18n/locales/en/translation.json
rolznz: Removed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/i18n/locales/en/translation.json
rolznz: Removed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/extension/background-script/state.ts
rolznz: Fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/extension/background-script/mnemonic/index.ts
rolznz: Fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/extension/background-script/mnemonic/index.ts
rolznz: Fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/common/lib/pset.ts
rolznz: I moved the pset preview code to the background script in https://github.com/getAlby/lightning-browser-extension/pull/2583

I am not sure about esplora

src/common/lib/liquid.ts
rolznz: Removed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: Fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: Fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/extension/providers/liquid/index.ts
louisinger: >  If the user adds a permission then it can enable without a popup. Is this what you mean?

yeah in such case, this.enabled should be `true` no ?

src/types.ts
rolznz: I guess it doesn't matter if we have another type with the same values. Updated in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: Fixed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/extension/providers/liquid/index.ts
rolznz: No, enable() has to be called either way. It's just what happens inside enable() will change if there is a permission set.

src/extension/providers/liquid/index.ts
rolznz: @louisinger @im-adithya removed in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/extension/background-script/actions/liquid/getAddressOrPrompt.ts
rolznz: @louisinger thanks, moved in https://github.com/getAlby/lightning-browser-extension/pull/2583

src/extension/providers/liquid/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  async enable() {
+    if (this.enabled) {
+      return { enabled: true };
+    }
+    return await this.execute("enable");
```
pavanjoshi914, line 20: we do need to update the enabled variable after enabling  the provider
refer: https://github.com/getAlby/lightning-browser-extension/pull/2594

src/extension/providers/liquid/index.ts
```diff
@@ -17,7 +17,11 @@ export default class LiquidProvider {
     if (this.enabled) {
       return { enabled: true };
     }
-    return await this.execute("enable");
+    const result = await this.execute("enable");
+    if (typeof result.enabled === "boolean") {
+      this.enabled = result.enabled;
```
louisinger, line 22: So I have to call `enable` to init the value? it means I can't show the enable state to the user without any interaction (calling `enable` may show a popup if it's the first time). 

is it possible to recover the value of "enabled" directly in the constructor? or have an async getter  `isEnabled()` just checking if the user has already called `enable` function in the past ?

rolznz: @louisinger that's a good point. I will see if there is a way that can be done.

src/i18n/locales/en/translation.json
```diff
@@ -386,6 +386,7 @@
           "export_uri": "LNDHub Credentials URI",
           "scan_qr": "Import this wallet into Zeus or BlueWallet by scanning the QRCode."
         },
+        "no_mnemonic_hint": "💡 You don't have a secret key yet. <0>Click here</0></div> to create your secret key first to unlock these settings.",
```
rolznz, line 389: master key

src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -455,7 +456,27 @@ function AccountDetail() {
               </div>
             </div>
             <MenuDivider />
-            <div className="flex justify-between items-center">
+            {!hasMnemonic && (
```
rolznz, line 459: Bitcoin network should also be moved into this section

rolznz: also rename to "network"?

src/common/lib/esplora.ts
```diff
@@ -0,0 +1,127 @@
+import fetchAdapter from "@vespaiach/axios-fetch-adapter";
+import axios, { AxiosInstance } from "axios";
+import { networks } from "liquidjs-lib";
+import { PsetPreview } from "~/types";
+
+// TODO: can this file be moved to the background script?
```
rolznz, line 6: Move this file

yarn.lock
```diff
@@ -2481,6 +2500,13 @@ available-typed-arrays@^1.0.5:
   resolved "https://registry.yarnpkg.com/available-typed-arrays/-/available-typed-arrays-1.0.5.tgz#92f95616501069d07d10edb2fc37d3e1c65123b7"
   integrity sha512-DMD0KiN46eipeziST1LPP/STfDU0sufISXmjSgvVsoU2tqxctQeASejWcfNtxYKqETM1UxQ8sp2OrSBWpHY6sw==
 
+axios@^0.21.1:
```
reneaaron, line 2503: Can we get rid of that one?

package.json
```diff
@@ -56,6 +56,7 @@
     "html5-qrcode": "^2.3.8",
     "i18next": "^22.5.1",
     "i18next-browser-languagedetector": "^7.1.0",
+    "liquidjs-lib": "^6.0.2-liquid.28",
```
rolznz, line 59: review how much bundle size has changed

src/extension/providers/liquid/index.ts
rolznz: @louisinger I chatted with @reneaaron and we think enable() should always be called after a user interaction (e.g. button click), it could be up to an application itself to store whether or not the user has ever logged in, to call enable immediately on page load.

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -0,0 +1,230 @@
...
+                <Hyperlink onClick={toggleShowAddresses}>
+                  {showAddresses ? t("hide_addresses") : t("view_addresses")}
+                </Hyperlink>
+                <span>{"•"}</span>
+                <Hyperlink onClick={toggleShowHex}>
+                  {showHex ? t("hide_hex") : t("view_hex")}
+                </Hyperlink>
+              </div>
+
+              {showAddresses && (
```
rolznz, line 141: @louisinger should we show the fee somewhere on the confirmation dialog? it can be calculated just subtracting the input from the output?

package.json
rolznz: not much, only a few hundred kb.

src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -481,6 +503,38 @@ function AccountDetail() {
...
+                <p className="text-gray-900 dark:text-white font-medium">
+                  {t("liquid.enable.title")}
+                </p>
+                <p className="text-gray-500 text-sm dark:text-neutral-500">
+                  {t("liquid.enable.description")}
+                </p>
+              </div>
+
+              <div className="w-1/5 flex justify-end align-middle">
+                <Toggle
```
rolznz, line 522: @louisinger what is the reason you added the toggle? we are thinking maybe it is more user-friendly not to have it. Then anyone who sets up a master key can use liquid immediately. CC @reneaaron @stackingsaunter

src/app/screens/Liquid/ConfirmSignPset.tsx
louisinger: Because of confidential feature, the subtract trick won't work. That's why Liquid txs have a "fee output" which is an output with a value (the fees amount) and an empty script. It should be easy to get that value in `getPsetPreview`. 

Beside that, not sure if we really need that feature. FYI Marina does not show the fee amount (because most of the app/explorers show them)

src/app/screens/Accounts/Detail/index.tsx
louisinger: :+1: +1 to remove the toggle step ! I just did it to respect what u did with the `window.liquid` provider (to do not inject by default into webpages)

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: @louisinger I wonder if it would be good to have it in the PsetPreview since Alby should be the source of truth (another app could get the fee wrong and users might not know?). But this could be a future improvement.

louisinger: @rolznz as you prefer, let me know, I can open a PR to update the PsetPreview :grin: 

> Alby should be the source of truth

To be honest, because Alby does not handle unblinding, this can't be true. the values/assets in the preview are computed from data in clear set up by the app requesting signing (in case of confidential outputs). In other words, _Alby can't unblind confidential inputs/outputs to check if the app isn't lying_. Maybe user should be aware of this (via a warning or something in the popup) ?

rolznz: @louisinger The signing screen is to make sure the user knows exactly what they are signing before confirming signing the transaction. Alby is the sender and generates the blinding key, shouldn't it know exactly what it is signing? sorry, I do not know enough about Liquid here, but it seems wrong to me if the user cannot verify what they are signing.

@bumi @reneaaron do you have any thoughts here?

src/app/screens/Accounts/Detail/index.tsx
rolznz: We reviewed and it is only a small script being injected. We don't think it is an issue

src/extension/background-script/liquid/index.ts
```diff
@@ -0,0 +1,141 @@
...
+
+class Liquid {
+  private slip77: Slip77Interface;
+  private network: networks.Network;
+  private mnemonic: Mnemonic;
+  public privateKey: string;
+
+  constructor(mnemonic: Mnemonic, networkType: LiquidNetworkType) {
+    // fix usages of window (unavailable in service worker)
+    globalThis.window ??= globalThis.window || {};
```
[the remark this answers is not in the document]
rolznz: Not needed, removed. Thanks!

src/app/screens/Liquid/ConfirmSignPset.tsx
louisinger: > shouldn't it know exactly what it is signing?

to do this, I would need https://github.com/vulpemventures/secp256k1-zkp module (in order to verify the ZKP proofs). So it means another dependencie (as far as I know `noble/curves` does not handle ZKP, right?)

yarn.lock
rolznz: @louisinger axios is listed as both a dependency and a dev dependency in liquidjs-lib. It seems it is not used except in the tests, could you convert it to just a dev dependency and let us know when it's updated?

src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -481,6 +503,38 @@ function AccountDetail() {
                 />
               </div>
             </div>
+            <MenuDivider />
+            <div
+              className={`flex justify-between items-end ${
+                !hasMnemonic && "opacity-30"
+              }`}
+            >
+              <div className="w-7/12">
```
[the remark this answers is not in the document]
rolznz: Removed the liquid toggle instead

src/app/screens/Accounts/Detail/index.tsx
rolznz: I've removed the toggle

yarn.lock
louisinger: Oh right ! nice catch thanks, here is the [PR](https://github.com/vulpemventures/liquidjs-lib/pull/105), should be released soon

louisinger: @rolznz liquidjs-lib `6.0.2-liquid.29` has been released without axios

rolznz (approved):
tACK

yarn.lock
rolznz: @louisinger thanks!

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: @louisinger we chatted with @bumi and @reneaaron, we think there is no need to do this now and users should have some level of trust with the site they are using. We can re-review if there is more usage. Thanks!

src/i18n/locales/en/translation.json
```diff
@@ -796,10 +795,26 @@
       "title": "Sign",
       "content": "This website asks you to sign:"
     },
-    "confirm_get_address": {
-      "title": "Get Address",
-      "heading": "This website asks you to read:",
-      "content": "Your Bitcoin receive address"
+    "confirm_sign_pset": {
+      "warning": "⚠️ Signing a PSET may result in loosing your funds or Ordinals. Sign only if you understand what you're doing and trust this transaction",
```
rolznz, line 799: Remove this warning

src/i18n/locales/en/translation.json
```diff
@@ -796,10 +795,26 @@
...
+      "title": "Sign PSET",
+      "allow_sign": "This website asks you to sign a Partially Signed Elements (Liquid) Transaction:",
```
rolznz, line 801: Simplify wording, maybe link to a help page (learn more button)

rolznz: @bumi @reneaaron changed to 
image

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -0,0 +1,219 @@
...
+                  </div>
+                </Hyperlink>
+              </div>
+              <div className="flex gap-2">
+                <Hyperlink onClick={toggleShowAddresses}>
+                  {showAddresses ? t("hide_addresses") : t("view_addresses")}
+                </Hyperlink>
+                <span>{"•"}</span>
+                <Hyperlink onClick={toggleShowHex}>
+                  {showHex ? t("hide_hex") : t("view_hex")}
```
bumi, line 126: We say HEX here and Base64 in the UI?

and what do I typically do with this string? can we hide that link behind the "show addresses" (if we call it "show details")?

I think less disctraction here is better.

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -0,0 +1,219 @@
...
+              image={origin.icon}
+              url={origin.host}
+            />
+            <div className="p-4 shadow bg-white dark:bg-surface-02dp rounded-lg overflow-hidden flex flex-col gap-4">
+              <div className="flex justify-between items-start">
+                <h2 className="font-medium dark:text-white">
+                  {t("allow_sign", { host: origin.host })}{" "}
+                </h2>
+                <Hyperlink
+                  href="https://medium.com/blockstream/pset-constructing-confidential-transactions-437109bb4ac5"
```
bumi, line 112: we should link to something we control. So let's make an Alby guide page for this - and from there link to such resources.

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -0,0 +1,219 @@
...
+                  target="_blank"
+                >
+                  <div className="bg-blue-500 rounded-full">
+                    <InfoIcon className="h-4 w-4 text-white" />
+                  </div>
+                </Hyperlink>
+              </div>
+              <div className="flex gap-2">
+                <Hyperlink onClick={toggleShowAddresses}>
+                  {showAddresses ? t("hide_addresses") : t("view_addresses")}
```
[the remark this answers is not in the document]
rolznz: image

src/app/screens/Liquid/ConfirmSignPset.tsx
rolznz: I made a page here. Needs some more work: https://guides.getalby.com/overall-guide/alby-browser-extension/features/liquid

PR #2476, opened 2023-06-13
Update signup

bumi:


src/app/router/Options/Options.tsx
```diff
@@ -17,20 +17,30 @@ import Send from "@screens/Send";
...
 import ChooseConnector from "~/app/screens/connectors/ChooseConnector";
 import ChooseConnectorPath from "~/app/screens/connectors/ChooseConnectorPath";
+import { getAlbyWalletOptions } from "~/app/utils";
 import i18n from "~/i18n/i18nConfig";
 
 function Options() {
   const connectorRoutes = getConnectorRoutes();
+  const [options, setOptions] = useState({ signup_disabled: false });
+
+  useEffect(() => {
```
reneaaron, line 39: What is the reason you include this on the Options / Welcome page and not in the individual components? 

Now the wallet options would be loaded every time I open the options / welcome page while that would only be necessary on the login / signup?

webpack.config.js
```diff
@@ -13,8 +13,8 @@ const BundleAnalyzerPlugin =
   require("webpack-bundle-analyzer").BundleAnalyzerPlugin;
 
 // default value is set in the code where it is used
-if (!process.env.WALLET_CREATE_URL) {
-  process.env.WALLET_CREATE_URL = ""; // env variables are passed as string. empty strings are still falsy
+if (!process.env.WALLET_ROOT_URL) {
```
reneaaron, line 16: Update build.yml and all occurances of WALLET_CREATE_URL (docs + playwright.config)

PR #2481, opened 2023-06-14
fix: return to home when switching accounts

reneaaron:

Takes you to `/` when switching between accounts. 

Pro:
 - Clean start after switching accounts (no broken behavior e.g. when switching accounts on the receive page with the QR code and polling for invoice updates)

Contra: 
 - If you switch accounts on all other pages you first have to navigate there again (e.g. when you notice that you have selected the wrong account on the send page)


stackingsaunter:
> * If you switch accounts on all other pages you first have to navigate there again (e.g. when you notice that you have selected the wrong account on the send page)

This is a bit unfortunate, I personally used switch while login in or sending a few times. I think I never switch receiving request.

reneaaron:
Yes, switching in prompts is especially useful (you noticed that you are on the wrong account while you authenticate, pay, etc). I think this has to be limited to only non-prompt usages.

PR #2490, opened 2023-06-16
feat: improve secret key security

rolznz:
- only reveal one word at a time and hide word length
- add warning on copy secret key button

**Note: needs translation. Are you happy with the flow? if so I will add the translation keys.**

Maybe we can also add some extra copy regarding the recommended approach of writing down the keys on a piece of paper and storing somewhere safe.

image

image

image

image


bumi:
maybe can we even remove the copy option? 
this means less code and it's not good to copy them anyway.

rolznz:
Note: copy button (and confirmation dialog) has now been removed

PR #2494, opened 2023-06-16
feat: add receive via bitcoin address button

pavanjoshi914:
add receive via bitcoin address button
styling button
adding translations
add ReceiveOnChain component
resolve missing key prop for translations
show button only when its an alby account

### Type of change

image
image

### Checklist

- [X] checked if I need to make corresponding changes to the documentation (and made those changes if needed)


src/app/components/OnChainReceive/index.tsx
```diff
@@ -0,0 +1,68 @@
+import { CaretLeftIcon } from "@bitcoin-design/bitcoin-icons-react/filled";
+import Header from "@components/Header";
+import { Trans, useTranslation } from "react-i18next";
+import { useNavigate } from "react-router-dom";
+import Button from "~/app/components/Button";
+import Container from "~/app/components/Container";
+import IconButton from "~/app/components/IconButton";
+
+export default function OnChainReceive() {
```
rolznz, line 9: I think this file should be in "screens" instead of "components"

src/app/components/OnChainReceive/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+      />
+      <div className="pt-4">
+        <Container justifyBetween maxWidth="sm">
+          <div className="text-center dark:text-neutral-200 h-full flex flex-col justify-center items-center">
+            <div className="mb-8">
+              <p>
+                <Trans
+                  i18nKey={"instructions1"}
+                  t={t}
+                  components={[<strong key="instruction1-strong"></strong>]}
```
rolznz, line 36: I think `-strong` should be removed from the key name. It is just decoration, not related to the translation.

src/app/components/OnChainReceive/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </div>
+          <div className="mb-4">
+            <a href="https://getalby.com/node/receive">
+              <Button
+                type="submit"
+                label={t("go")}
+                fullWidth
+                primary
+                onClick={() =>
+                  window.open(`https://getalby.com/user`, "_blank")
```
rolznz, line 59: Why is this button wrapped in a link and also has an `onClick` function? The onClick one has the wrong url too.

src/app/screens/Receive/index.tsx
```diff
@@ -237,44 +239,72 @@ function Receive() {
...
+                    fullWidth
+                    primary
+                    loading={loading}
+                    disabled={loading}
                   />
                 </div>
+              </Container>
+            </fieldset>
+          </form>
+          {isAlbyUser ? (
```
rolznz, line 281: You can do `isAlbyUser &&` since there is no "else" component

src/app/screens/Receive/index.tsx
```diff
@@ -237,44 +239,72 @@ function Receive() {
...
-                  <TextField
-                    id="description"
-                    label={t("description.label")}
-                    placeholder={t("description.placeholder")}
-                    onChange={handleChange}
+                  <Button
+                    type="button"
+                    label={t("receive_via_bitcoin_address")}
+                    fullWidth
+                    loading={loading}
```
rolznz, line 297: when I click create invoice, this button also shows a loading spinner. This is strange IMO

rolznz:
@stackingsaunter do you have a figma file @pavanjoshi914 can review to ensure he gets the styles right?

image

For example, I think the action button normally should be at the bottom of the popup (to be confirmed)

stackingsaunter:
Let's put the button at the bottom and maybe center the text on the screen (so it's in the middle not on the top). Not super elegant but I think this is temporary solution anyway, as soon we can have swaps within the extension, so let's just merge it!

src/app/components/OnChainReceive/index.tsx
pavanjoshi914: this is from on chain Tabs pull request. i thought it is already reviewed and considered valid.
what shall be the behaviour here?

pavanjoshi914:
> Let's put the button at the bottom and maybe center the text on the screen (so it's in the middle not on the top). Not super elegant but I think this is temporary solution anyway, as soon we can have swaps within the extension, so let's just merge it!

do this looks good?
image

src/app/components/OnChainReceive/index.tsx
pavanjoshi914: changed to single url. /node/receive

src/app/screens/Receive/index.tsx
pavanjoshi914: resolved

rolznz:
@pavanjoshi914 I found one minor spacing issue. I think the OR buttons should be aligned at the bottom. @stackingsaunter could you confirm?

image


Otherwise looks good to me!

rolznz (approved):

reneaaron (approved):
Let's do another feedback loop and review the copy of this PR, other than that:

tACK

pavanjoshi914:
thanks for the changes. shall we replacd the "or" in lnd screen as "or" is placed in common set of strings now

image

rolznz:
@pavanjoshi914 I think it is fine to change to "or". It might be nice to have a Divider component, and then we could also use it in the LND connect screen. What do you guys think? (Maybe this could be done as a follow up PR?)

pavanjoshi914:
> @pavanjoshi914 I think it is fine to change to "or". It might be nice to have a Divider component, and then we could also use it in the LND connect screen. What do you guys think? (Maybe this could be done as a follow up PR?)

let's do this in follow up pr in think

pavanjoshi914:
@reneaaron can we merge this?

PR #2497, opened 2023-06-20
fix: firefox suggesting unlock passcode on nostr private keys

### LINKED ISSUE #2489: [BUG] Unlock password gets used as nostr private key

### Describe the bug

My password manager generated a "safe password" when starting the extension for the first time and I found that password again as the Nostr -> private key:

Screenshot from 2023-06-15 20-31-50

Now I "generated a new key" but the nsec got then replaced by the password again. I suppose the npub relates to the valid nsec still.

### Screenshots [optional]

_No response_

### Steps To Reproduce

1. install Alby
2. try to use it

### Expected behavior

the nsec field should not trigger the password manager to think this was the same as the unlock password

### Alby information

Alby 2.1.1

### Device information

Debian
FF 96.0.3 (64-bit)

### Additional context

_No response_

### Are you working on this?

None

fczuardi:

Hints Firefox password manager that the Nostr private keys are not password for the same (no user) that is the unlock password.

Fixes #2489

### Type of change

### How has this been tested?

On Firefox (Linux) I have set up 3 accounts with 3 different nostr private keys, later I opened `about:logins` and searched for "moz". Firefox didn't replace my unlock passcode, the password for the `(no username)` user.

### Checklist

### DIFF
--- a/src/app/screens/Accounts/Detail/index.tsx
+++ b/src/app/screens/Accounts/Detail/index.tsx
@@ -473,6 +473,7 @@ function AccountDetail() {
                 {/* We use a hidden field to hint Firefox's password manager
                     that this password field is for an account different
                     than the unlock passcode (no username)
+                    See: https://github.com/getAlby/lightning-browser-extension/issues/2489
                 */}
                 <input value={accountName} style={{ display: "none" }} />
                 <TextField
--- a/src/app/screens/Accounts/Detail/index.tsx
+++ b/src/app/screens/Accounts/Detail/index.tsx
@@ -475,7 +475,7 @@ function AccountDetail() {
                     than the unlock passcode (no username)
                     See: https://github.com/getAlby/lightning-browser-extension/issues/2489
                 */}
-                <input value={accountName} style={{ display: "none" }} />
+                <input value={accountName} className="hidden" />
                 <TextField
                   id="nostrPrivateKey"
                   label={t("nostr.private_key.label")}
--- a/src/app/screens/Accounts/Detail/index.tsx
+++ b/src/app/screens/Accounts/Detail/index.tsx
@@ -470,6 +470,11 @@ function AccountDetail() {
               className="mb-4 flex justify-between items-end"
             >
               <div className="w-7/12">
+                {/* We use a hidden field to hint Firefox's password manager
+                    that this password field is for an account different
+                    than the unlock passcode (no username)
+                */}
+                <input value={accountName} style={{ display: "none" }} />
                 <TextField
                   id="nostrPrivateKey"
                   label={t("nostr.private_key.label")}


reneaaron:
But you still want to be able to save your nostr key in the password manager, is that right?

fczuardi:
> But you still want to be able to save your nostr key in the password manager, is that right?

Yes, the password manager will still be able to save the nostr private key of an account as a password, however now it will use the account name as the username, instead of the `(no username)` user.

In other words: the Firefox password manager will still save your nostr private key as a password if you choose to, and it will not overwrite the existing unlock passcode as it did before.

bumi:
interesting! great find @fczuardi wow.

it sounds like a bit of a strange FF behavior.
@reneaaron any objections to merge this? 
@fczuardi maybe we can add a link to the issue in the comment.

fczuardi:
> interesting! great find @fczuardi wow.
> 
> it sounds like a bit of a strange FF behavior. @reneaaron any objections to merge this? @fczuardi maybe we can add a link to the issue in the comment.

reference to the issue added

src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -470,6 +470,12 @@ function AccountDetail() {
               className="mb-4 flex justify-between items-end"
             >
               <div className="w-7/12">
+                {/* We use a hidden field to hint Firefox's password manager
+                    that this password field is for an account different
+                    than the unlock passcode (no username)
+                    See: https://github.com/getAlby/lightning-browser-extension/issues/2489
+                */}
+                <input value={accountName} style={{ display: "none" }} />
```
reneaaron, line 478:
```suggestion
                <input value={accountName} className="hidden" />
```

pavanjoshi914:
hey @fczuardi  thanks for making up the PR. we had a lot of changes in Alby screens the past few days. so I decided to make a new pr. we corrected the autocomplete problem on all the password fields that firefox is causing. closing this in favour of #2595.

fczuardi:
Hi @pavanjoshi914,

I remember we tried the autocomplete new-password path and the issue persisted, that's why our approach was the hidden field path. But I can try to reproduce it on my Firefox and your branch tomorrow and let you know if we should reopen the issue or not.

-------- Original Message --------

pavanjoshi914:
thanks for helping out there! please do test and let me know! on my side. issue is resolved for firefox password manager as well. let me know if issue persist on your side

pavanjoshi914:
where did you tried ```new-password``` field? on form level or on text field?

fczuardi:
Hi @pavanjoshi914 I just confirmed on the master branch (that have your patch) and Firefox+Linux the issue is still reproducible. Entering a Nostr private key on the Nostr Settings and accepting the Firefox prompt to save the password, it will overwrite the saved unlocking passcode (the password with "no username" in the password manager), see screenshots below:

01-Firefox-prompt-Screenshot_20230731_214948
02-Alby-with-unlocking-password-filled-Screenshot_20230731_215430
03-Alby-with-unlocking-password-visible-Screenshot_20230731_215509

pavanjoshi914:
> 

we use new-password on all the fields except unlock field. even if you save nostr private key in pasword manager. it will not autocomplete the field. in the nostr text field. as it will always default to new-password. the current discussion was to just disable autocomplete for all other fields or don't allow any saving feature on other fields. and only accept auto-complete for unlock fields. 

if we use the hidden field on nostr private key. it will also disable autocomplete feature. then why to save private key in password manager?

regarding unlock fields, we don't put any restrictions there. probably we can disable autocomplete there as well. cc @rolznz @reneaaron @bumi

fczuardi:
The password manager wont autocomplete the nostr private key field, thats correct. But if someone accepts the firefox prompt asking to save the password (by accident or by habit) on the setting page it will overwrite the unlocking password with nostr private key.

Issue https://github.com/getAlby/lightning-browser-extension/issues/2489 was indeed about the other way around (unlocking password getting autocompleted on nostr field) and that is not happening anymore. However, after the fix we now have a new problem: nostr private key getting autocompleted on the unlocking passcode field.

Should we open a new issue then? instead of reopening #2489 ?

@pavanjoshi914 do you see my point? What my screenshots show is something that is not the unlocking passcode being completed on the unlocking passcode field.

The statement below is not true:

>if we use the hidden field on nostr private key. it will also disable autocomplete feature.

It wont.

pavanjoshi914:
if we use the hidden field on nostr private key. it will also disable autocomplete feature.

pavanjoshi914:
if we use the hidden field on nostr private key. it will also disable autocomplete feature.

pavanjoshi914:
>The password manager wont autocomplete the nostr private key field, that's correct. But if someone accepts the firefox prompt asking to save the password (by accident or by habit) on the settings page it will overwrite the unlocking password with nostr private key.

regarding unlock fields, we don't put any restrictions there. probably we can disable autocomplete there as well

PR #2504, opened 2023-06-22
feat: send account switched messages to website

rolznz:

Send a message to the website when user switches account.

To listen to the event the website would add an event listener like:

```
window.addEventListener(
  "message",
  (event) => {
      if (event.data === "accountSwitched") {
         //... do something here
      }
  },
);
```

Fixes _#2503_

### Type of change

### How has this been tested?

Manually

### Checklist

- [ ] Review security implications (we do not give any info about the account)
- [ ] Review the action name. Should it have some sort of prefix?
- [ ] How to document / promote this feature?


bumi:
best first also write a spec for window.nostr and window.webln

src/extension/content-script/onend.js
```diff
@@ -32,11 +32,15 @@ async function init() {
...
 
-  // extract LN data from websites
   browser.runtime.onMessage.addListener((request, sender, sendResponse) => {
+    // extract LN data from websites
     if (request.action === "extractLightningData") {
       extractLightningData();
     }
+    // forward account switched messages to the website
+    else if (request.action === "accountSwitched") {
+      window.postMessage("accountSwitched", "*");
```
bumi, line 42: how should the user use this? 

as a developer we want to do something like?: 
```
winow.addEventListener("webln:accountChanged", (old, new) => { ... }); 
```
is this correct?

reneaaron: Just for reference: 

https://docs.metamask.io/wallet/reference/provider-api/#accountschanged

Seems they offer an `on()` method directly on the provider.

rolznz: @bumi  I added an example above but it requires listening to generic window messages. I like the solution @reneaaron linked because we can add the method to the existing `webln` and `nostr` objects, messages need to be explicitly subscribed to by the web app (and we can control whether or not the user needs to approve the request) and also have better control over the response object.

For the specific "accountSwitched" case I think we should be able to execute the method without asking for the user's permission because this is yet another modal users have to accept (We could consider having a different message type if we want to provide sensitive details in the callback?).

Should we go with a similar method as metamask?

`window.webln.on("accountSwitched", callback) `
`window.nostr.on("accountSwitched", callback) `

I think Alby should have a working implementation before we submit a PR for the specs.

Thank you both for bring these points up! much better than my original implementation, but I am glad the ball is rolling now :-)

MoritzKa:
Thanks for initiating this. With the possibility of having different Accounts in the extension, users have from time to time the wrong account selected without knowing it. 
If we follow the NIP standardization process, we would need a working implementation first. But we should propose it maybe here and let people know? https://github.com/nostr-protocol/nips/blob/master/07.md
After that I can reach out to coracle, iris. I know the developers of Primal and snort but did not have direct contact yet.

rolznz:
@MoritzKa I think the only way to successfully propose it is to create a PR. If we already have a working implementation then the PR may be reviewed more closely so maybe it is best to wait for that?

src/extension/content-script/onend.js
rolznz: @reneaaron @bumi any thoughts on the event name? I called it "switched" here as "changed" could be misinterpreted as "updated". What do you think?

rolznz:
I have assigned @pavanjoshi914 to help updating this PR to use the new .on() method for window.webln and window.nostr

src/extension/content-script/onend.js
reneaaron: To me `accountChanged` feels more natural as I think there is no reason for a website to be notified when something IN the account changes, right? It's maybe also more in line with other HTML events that offer onChange event handlers?

src/extension/background-script/actions/accounts/select.ts
```diff
@@ -39,3 +42,25 @@ const select = async (message: MessageAccountSelect) => {
...
+// which will then be posted to the window so websites can sync with the switched account
+async function notifyAccountSwitched() {
+  try {
+    const tabs = await browser.tabs.query({
+      active: true,
+      currentWindow: true,
+    });
+
+    const currentUrl = tabs.length && tabs[0].url;
+    if (currentUrl && currentUrl.startsWith("http")) {
```
pavanjoshi914, line 56: why http here?

pavanjoshi914:
to handle events in best way we can use ```EventEmitters``` node environments provide such functionality. in our case we can use events package to handle events such as ```accountSelected```.

here is the useCase
1. ```webln.on("accountSelected", callback)``` : let users subscribe to the event and add to the list of events
image

package provides complete set of methods required to handle events under EventEmitter class such as removing listener, emiting events etc

other supported events

https://nodejs.org/dist/v11.13.0/docs/api/events.html#events_class_eventemitter

pavanjoshi914:
https://github.com/getAlby/lightning-browser-extension/assets/55848322/e3796c55-050f-4daa-a9f9-6b1567f10e08


website registering event
```
webln.on("accountSelected", callback);
```

website remove the event, insure to pass same function to remove listener successfully

```
webln.removeListener("accountSelected", callback)
```

src/extension/content-script/onend.js
rolznz: CC @pavanjoshi914 - I think you preferred `accountChanged` too. Should we update it?

src/extension/background-script/actions/accounts/select.ts
rolznz: for sites running on localhost

src/extension/background-script/actions/accounts/select.ts
```diff
@@ -27,6 +28,8 @@ const select = async (message: MessageAccountSelect) => {
     // save the current account id once the connector is loaded
     await state.getState().saveToStorage();
 
+    await notifyAccountSwitched();
```
rolznz, line 31: Rene has voted for `accountChanged` so maybe we should update all the usages - `notifyAccountChanged`, etc.

src/extension/content-script/onendwebln.js
```diff
@@ -32,11 +32,15 @@ async function init() {
 
   injectScript(browser.runtime.getURL("js/inpageScript.bundle.js")); // registers the DOM event listeners and checks webln again (which is also loaded onstart
 
-  // extract LN data from websites
   browser.runtime.onMessage.addListener((request, sender, sendResponse) => {
+    // extract LN data from websites
     if (request.action === "extractLightningData") {
       extractLightningData();
     }
+    // forward account selected messages to the website
```
rolznz, line 40: this comment is now wrong, it should be the inpage-script right? I am not sure if we can use a different value than "*", but I don't think it matters really, at least for this event.

src/extension/inpage-script/index.js
```diff
@@ -121,6 +121,13 @@ if (document) {
     },
     { capture: true }
   );
+  // Listen for webln events from the extension
+  // emit more events as received
+  window.addEventListener("message", (event) => {
+    if (event.source === window && event.data === "accountSelected") {
+      window.webln.emit("accountSelected");
```
rolznz, line 128: - should we check if webln exists first?
- also we should fire this event for window.nostr, if it exists.

rolznz:
@pavanjoshi914 after this is merged and deployed, it would be great to do these follow up tasks:
- [ ] Create PR for updating NIP-07
- [ ] Update https://github.com/WebBTC/webln-types
- [ ] Update webln.guide with new .on and .off methods and the supported event types (only `accountChanged` for now)
- [ ] Possibly we can submit PRs to snort.social and other open source nostr clients to add this feature

src/extension/content-script/onend.js
pavanjoshi914: sure!

src/extension/background-script/actions/accounts/select.ts
pavanjoshi914: resolved

src/extension/content-script/onendwebln.js
pavanjoshi914: resolved

src/extension/inpage-script/index.js
pavanjoshi914: resolved

src/extension/inpage-script/index.js
```diff
@@ -121,6 +121,27 @@ if (document) {
     },
     { capture: true }
   );
+  // Listen for webln events from the extension
+  // emit events to the websites
+  window.addEventListener("message", (event) => {
+    if (event.source === window && event.data.action === "accountChanged") {
+      if (
+        (window.webln && window.webln.enabled) ||
```
rolznz, line 129: I think these checks should be moved into the eventEmitter, it would be safer to check the scope and check if the window object (nostr or webln) exists in the same place

src/extension/inpage-script/index.js
```diff
@@ -121,6 +121,27 @@ if (document) {
...
+      )
+        eventEmitter(event.data.action, event.data.scope);
+    }
+  });
 } else {
   console.warn("Failed to inject WebLN provider");
 }
+
+function eventEmitter(action, scope) {
+  if (scope == "nostr") {
```
rolznz, line 140: could it be simpler like this? then we don't need to add more checks for e.g. window.webbtc or window.alby
```
if (window[scope]) {
  window[scope].emit(action);
}
```

src/extension/providers/webln/index.ts
```diff
@@ -109,6 +113,27 @@ export default class WebLNProvider {
     });
   }
 
+  //webln events
+  on(...args: Parameters<EventEmitter["on"]>) {
```
rolznz, line 117: All this new code is duplicated in the two providers. I wonder if we can extend a common provider class or something. We could also move the common stuff like enabling and executing there.

I would do this as a follow up PR I guess.

src/extension/providers/nostr/index.ts
```diff
@@ -10,16 +12,23 @@ declare global {
...
 
   constructor() {
     this.enabled = false;
+    this._eventEmitter = new EventEmitter();
   }
 
   async enable() {
     if (this.enabled) {
-      return { enabled: true };
+      return Promise.resolve({ enabled: true });
```
bumi, line 24: as it's an async function I don't need a Promise.resolve here, do I?

src/extension/providers/nostr/index.ts
```diff
@@ -10,16 +12,23 @@ declare global {
...
     }
-    return await this.execute("enable");
+    return this.execute("enable").then((result) => {
```
bumi, line 26: should this be awaited becuse of the async function?

src/extension/providers/nostr/index.ts
reneaaron: > The word “async” before a function means one simple thing: a function always returns a promise. Other values are wrapped in a resolved promise automatically. So, async ensures that the function returns a promise, and wraps non-promises in it.

I think we should just return the plain objects in such cases. No need to fiddle around with promises.

src/extension/providers/nostr/index.ts
reneaaron: +1. I think mixing promises & non-promises (async/await) should be avoided because it makes it harder to read. 

Isn't that a lot easier to read? 

```
var result = await this.execute("enable");
if (typeof result.enabled === "boolean") {
  this.enabled = result.enabled;
}
return result;
```

src/extension/inpage-script/index.js
pavanjoshi914: resolved

src/extension/inpage-script/index.js
pavanjoshi914: cool! i tried this earlier but didn't worked in console so avoided it. but works well enough in code. thanks for this suggestion

src/extension/providers/nostr/index.ts
pavanjoshi914: resolved

src/extension/providers/nostr/index.ts
pavanjoshi914: all the instances have been updated to this. resolved

src/extension/inpage-script/index.js
```diff
@@ -121,6 +121,21 @@ if (document) {
...
+    }
+  });
 } else {
   console.warn("Failed to inject WebLN provider");
 }
+
+function eventEmitter(action, scope) {
+  if (window[scope] && window[scope]._eventEmitter) {
+    window[scope].emit(action);
+  } else {
```
rolznz, line 138: `else return` is redundant?

Why check for a private variable `_eventEmitter` but then call `.emit`? maybe you can check  for `.emit`?

Can action and scope variables be typed?

pavanjoshi914: why typing is needed here scope will always be passed as string as it's not on user end and on extension side

rolznz:
Looks good to me. I cannot approve it as I originally made the PR.

CC @bumi @reneaaron

PR #2505, opened 2023-06-23
fix: improved validation with latest nostr-tools validateEvent

### LINKED ISSUE #2446: [BUG] window.nostr.signEvent does not validate the input and returns undefined for invalid input

### Describe the bug

Currently if you call signEvent with one of the following, it returns undefined rather than throwing an exception. This leads to confusion.

`const signedEvent = await window.nostr.signEvent(undefined);`
`const signedEvent = await window.nostr.signEvent({a: 1});`

### Screenshots [optional]

_No response_

### Steps To Reproduce

1. call `window.nostr.enable()`
2. Execute one of the following:

`const signedEvent = await window.nostr.signEvent(undefined);`
`const signedEvent = await window.nostr.signEvent({a: 1});`

### Expected behavior

Alby throws an exception saying the input is not a valid nostr event.

### Alby information

Alby 2.0.1

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

None

mquasar:

Updated validateEvent with latest nostr-tool version and changes do signEventOrPrompt to return error in case of invalid events

Fixes  #2446

### Type of change

### How has this been tested?

1. call window.nostr.enable()
2. Execute one of the following:

```
const signedEvent = await window.nostr.signEvent(undefined);
const signedEvent = await window.nostr.signEvent({a: 1});

```
### Checklist

### DIFF
--- a/src/extension/background-script/actions/nostr/helpers.ts
+++ b/src/extension/background-script/actions/nostr/helpers.ts
@@ -58,15 +58,14 @@ export async function addPermissionFor(method: string, host: string) {
   return !!permissionIsAdded && (await db.saveToStorage());
 }
 
-// from: https://github.com/nbd-wtf/nostr-tools/blob/160987472fd4922dd80c75648ca8939dd2d96cc0/event.ts#L61
+// from: https://github.com/nbd-wtf/nostr-tools/blob/b9a7f814aaa08a4b1cec705517b664390abd3f69/event.ts#L95
 // to avoid the additional dependency
 export function validateEvent(event: Event): boolean {
   if (!(event instanceof Object)) return false;
   if (typeof event.kind !== "number") return false;
   if (typeof event.content !== "string") return false;
   if (typeof event.created_at !== "number") return false;
-  //if (typeof event.pubkey !== "string") return false;
-  //if (!event.pubkey.match(/^[a-f0-9]{64}$/)) return false;
+  // ignore pubkey checks because we always override it with the pubkey from the extension's account
 
   if (!Array.isArray(event.tags)) return false;
   for (let i = 0; i < event.tags.length; i++) {
--- a/src/extension/background-script/actions/nostr/helpers.ts
+++ b/src/extension/background-script/actions/nostr/helpers.ts
@@ -61,11 +61,12 @@ export async function addPermissionFor(method: string, host: string) {
 // from: https://github.com/nbd-wtf/nostr-tools/blob/160987472fd4922dd80c75648ca8939dd2d96cc0/event.ts#L61
 // to avoid the additional dependency
 export function validateEvent(event: Event): boolean {
+  if (!(event instanceof Object)) return false;
+  if (typeof event.kind !== "number") return false;
   if (typeof event.content !== "string") return false;
   if (typeof event.created_at !== "number") return false;
-  // ignore these checks because if the pubkey is not set we add it to the event. same for the ID.
-  // if (typeof event.pubkey !== "string") return false;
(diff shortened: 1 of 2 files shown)


src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -47,6 +44,8 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
       }
     }
 
+    if (!event.id) event.id = nostr.getEventHash(event);
```
rolznz, line 47: I think we should always set the pubkey and id. The pubkey must be the pubkey of the event creator, and the ID can only be generated once all fields including the pubkey are set on the event.

**important** the pubkey must be set in order to calculate the event ID. When you moved them you swapped the order around which will break `getEventHash`.

fczuardi: nice catch! after we move this to after the validation we swapped pubkey and event.id order, you are right, the event id should be after everything else is part of the event object

fczuardi: as for the "always set the pubkey and id" the idea then is to always override whatever event we receive to sign with the pubkey from alby, makes sense

rolznz: I think so. Otherwise we need to validate them, and the only way to validate them is to compare against the generated values, so we might as well overwrite them?

src/extension/background-script/actions/nostr/helpers.ts
```diff
@@ -58,14 +58,14 @@ export async function addPermissionFor(method: string, host: string) {
   return !!permissionIsAdded && (await db.saveToStorage());
 }
 
-// from: https://github.com/nbd-wtf/nostr-tools/blob/160987472fd4922dd80c75648ca8939dd2d96cc0/event.ts#L61
+// from: https://github.com/nbd-wtf/nostr-tools/blob/b9a7f814aaa08a4b1cec705517b664390abd3f69/event.ts#L95
 // to avoid the additional dependency
 export function validateEvent(event: Event): boolean {
+  if (!(event instanceof Object)) return false;
```
rolznz, line 64: We now have multiple changes from the above link, so maybe we should update the comment to say "based upon"?

rolznz (approved):
tACK

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -47,6 +44,8 @@ const signEventOrPrompt = async (message: MessageSignEvent) => {
       }
     }
 
+    event.pubkey = nostr.getPublicKey();
```
reneaaron, line 47: Not sure about this change. Aren't the clients supposed to fill this event object? 

I think we should only override those properties if they are not set: 

https://github.com/fiatjaf/nos2x/blob/master/extension/background.js#L170-L171

reneaaron: Signing clients are supposed to fill those fields according to the spec here: 

https://github.com/nostr-protocol/nips/blob/master/07.md

> // takes an event object, adds `id`, `pubkey` and `sig` and returns it

I'd say we just fill them if they are empty :+1:

mquasar: Ok, just commited a new version with the checks back.

rolznz (approved):

reneaaron (approved):
tACK

PR #2509, opened 2023-06-23
fix: update keysend screen

reneaaron:

Cleans up the keysend screen
 - remove duplicate node id
 - remove box around amount input (similar to the LNURL screen)
 - spacing

### DIFF
--- a/src/app/screens/Keysend/index.tsx
+++ b/src/app/screens/Keysend/index.tsx
@@ -4,7 +4,6 @@ import ConfirmOrCancel from "@components/ConfirmOrCancel";
 import ContentMessage from "@components/ContentMessage";
 import Header from "@components/Header";
 import IconButton from "@components/IconButton";
-import PublisherCard from "@components/PublisherCard";
 import ResultCard from "@components/ResultCard";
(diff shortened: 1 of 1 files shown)


src/app/screens/Keysend/index.tsx
```diff
@@ -110,29 +109,28 @@ function Keysend() {
...
-                </div>
+                <DualCurrencyField
+                  id="amount"
+                  label={t("amount.label")}
+                  min={1}
+                  onChange={(e) => setAmountSat(e.target.value)}
+                  value={amountSat}
+                  fiatValue={fiatAmount}
+                />
+                <SatButtons onClick={setAmountSat} />
```
rolznz, line 124: Is SatsButtons needed? doesn't the amount need to be fixed for keysend?

But it doesn't seem to show anyway

image

rolznz (approved):

src/app/screens/Keysend/index.tsx
reneaaron: I think you are on the wrong screen. :thinking: 

You would have to paste a nodes public key on the send screen to get to the this screen.

PR #2512, opened 2023-06-26
feat: redeem-lnurl screen

pavanjoshi914:
redeem lnurl screen to redeem via lnurl

image
image
image

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


src/app/screens/LNURLRedeem/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+    const invoice = data.match(reqExp);
+
+    if (invoice) {
+      return invoice[1];
+    } else {
+      return data.replace(/^lightning:/i, "");
+    }
+  }
+
+  if (qrIsOpen) {
```
rolznz, line 79: I think a lot of this code is copied from src/app/screens/Send/index.tsx ?

I think it would be good to extract it into a single place if possible.

src/i18n/locales/en/translation.json
```diff
@@ -634,6 +634,19 @@
       },
       "success": "Withdraw request of {{amount}} sent successfully to {{sender}}"
     },
+    "lnurlredeem": {
```
rolznz, line 637: I think the key should follow our conventions, should it be an _ between words? and maybe lnurl_withdraw rather than redeem. "Redeem" is the English translation shown to users.

src/i18n/locales/en/translation.json
```diff
@@ -634,6 +634,19 @@
...
+      "title": "Redeem bitcoin",
+      "input": {
+        "label": "LNURL-withdraw code",
+        "placeholder": "LNURL..."
+      },
+      "actions": {
+        "withdraw": "Withdraw"
+      },
+      "qrcode": {
```
rolznz, line 646: QR code translations should not be duplicated. Could you look if these can be moved to a common place?

src/app/screens/LNURLRedeem/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+    } catch (e) {
+      if (e instanceof Error) {
+        toast.error(e.message);
+      }
+    } finally {
+      setLoading(false);
+    }
+  }
+
+  function extractInvoiceFrom(data: string) {
```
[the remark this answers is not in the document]
pavanjoshi914: this will be used if we scan qr codes which contains  url where lnurl-w code is given in the tag named ```lightning```

eg. https://legend.lnbits.com/?lightning=LNURL1DP68GURN8GHJ7MR9VAJKUEPWD3HXY6T5WVHXXMMD9AMKJARGV3EXZAE0V9CXJTMKXYHKCMN4WFKZ7KZDD4E455ECGAJ8Q5698P55W6M8DGENSATK9AT52K26GAU5CJJ32ET5K6RWTP38WATSDDK5CL2QW0E

src/app/screens/LNURLRedeem/index.tsx
pavanjoshi914: tried to extract the scanner in at single place but we do set a state variable here. we need to return back a lnurl-withdraw code back. not sure if we shall go this way

src/i18n/locales/en/translation.json
pavanjoshi914: we already have a translations for lnurl_withdraw for lnurlWithdraw screen. is it fine to have redeem here?
image
as other translations for lnurl don't contain underscore for the sake of consistency

src/i18n/locales/en/translation.json
pavanjoshi914: resolved

src/app/utils/index.ts
```diff
@@ -78,3 +78,15 @@ export async function getAlbyWalletOptions() {
     console.error(e);
   }
 }
+
+export function extractLightningTagData(url: string) {
+  const reqExp = /lightning=([^&|\b]+)/i;
```
reneaaron, line 83: We should add a comment here why we search for that (basically for LNBits links iirc)

src/i18n/locales/en/translation.json
rolznz: sounds ok to me, thanks

src/app/screens/LNURLRedeem/index.tsx
rolznz: Can we have a screen component for the QR code which includes this?
image

rolznz: This could be moved into a ScanQRAdornment component with a callback function for qrCodeSuccessCallback and qrCodeErrorCallback. Would that work?

image

pavanjoshi914: we need to consider here passing value back and getting back on the lnurl-redeem screen as well. if scanner extracted into separate screen. how callbacks will work here. in terms of passing value back and navigating?

currently i proposed a solution using navigation. instead of callbacks kindly have a look

src/app/components/QrcodeAdornment/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default function QrcodeAdornment({ route }: Props) {
+  const navigate = useNavigate();
+  return (
+    <button
+      aria-label="Scan QR"
+      type="button"
+      className="flex justify-center items-center w-10 h-8"
+      onClick={() => {
+        navigate("/scanQRCode", { state: { route: route }, replace: true });
```
rolznz, line 16: This should not be `replace: true`, right? When you exit instead of going back to /send or /lnurlRedeem it will go back two pages

src/app/screens/LNURLRedeem/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { toast } from "react-toastify";
+import QrcodeAdornment from "~/app/components/QrcodeAdornment";
+import lnurlLib from "~/common/lib/lnurl";
+import { isLNURLDetailsError } from "~/common/utils/typeHelpers";
+
+function LNURLRedeem() {
+  const { t } = useTranslation("translation", { keyPrefix: "lnurlredeem" });
+  const location = useLocation();
+
+  const [lnurlwithdraw, setlnurlwithdraw] = useState(
```
rolznz, line 19: lnurlwithdraw seems like the wrong variable name here. Should this be `lnurl` or something?

rolznz: setter should have a capital letter after `set`

rolznz: Can you add a comment explaining why location.state is used? (comes from QRScanner screen). I wonder if `decodedQR` is better than `decodedText`?

src/app/screens/ScanQRCode/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const location = useLocation();
+  const route = location.state?.route;
+
+  return (
+    <div>
+      <Header
+        title={t("title")}
+        headerRight={
+          <IconButton
+            onClick={() => navigate(-1)}
```
rolznz, line 21: can you navigate to the route here too, rather than -1?

src/app/screens/LNURLRedeem/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+            onClick={() => navigate(-1)}
+            icon={<CaretLeftIcon className="w-4 h-4" />}
+          />
+        }
+      />
+      <form onSubmit={handleSubmit} className="h-full">
+        <Container justifyBetween maxWidth="sm">
+          <div className="pt-4">
+            <TextField
+              id="invoice"
```
rolznz, line 79: this is not an invoice, maybe the id should be renamed or removed if it is not necessary

src/app/screens/Send/index.tsx
```diff
@@ -1,29 +1,26 @@
...
 import lnurlLib from "~/common/lib/lnurl";
 import { isLNURLDetailsError } from "~/common/utils/typeHelpers";
 
 function Send() {
   const { t } = useTranslation("translation", { keyPrefix: "send" });
   const { t: tCommon } = useTranslation("common");
+  const location = useLocation();
 
-  const [invoice, setInvoice] = useState("");
+  const [invoice, setInvoice] = useState(location.state?.decodedText || "");
```
rolznz, line 22: Can you add a comment explaining why location.state is used? (comes from QRScanner screen). I wonder if decodedQR is better than decodedText?

src/app/utils/index.ts
```diff
@@ -78,3 +78,16 @@ export async function getAlbyWalletOptions() {
     console.error(e);
   }
 }
+
+// to extract lightning data associated with the lightning tag within the URL. eg. LNBits QR codes
+export function extractLightningTagData(url: string) {
```
rolznz, line 83: Is this only used in one place? does it make sense to be moved back to the Send screen?

src/app/components/QrcodeAdornment/index.tsx
pavanjoshi914: this is needed, cause if we go back we will have two routes in navigation stack for same page. instead, i changed code for exist button to navigate with replace: true. and things work fine for both back button and exit button

src/app/screens/LNURLRedeem/index.tsx
pavanjoshi914: but the expected text from the form is lnurlWithrdraw string right? we are just validating after that if its lnurl and has tag of withdraw request or not. WDYT?

src/app/screens/ScanQRCode/index.tsx
pavanjoshi914: resolved

src/app/screens/LNURLRedeem/index.tsx
pavanjoshi914: resolved

src/app/screens/Send/index.tsx
pavanjoshi914: changed this to decodedQR. added comment for the use case of the location.state

src/app/utils/index.ts
pavanjoshi914: shall we use this in redeemLNURL screen as well!. if user pastes the url directly in the form field this function will filter and get use withdraw code there as well?

pavanjoshi914: currently adding it

src/app/screens/LNURLRedeem/index.tsx
rolznz: I think "LNURL-withdraw" is the name of the protocol, that's why I suggested call it just "lnurl", maybe "lnurlWithdrawLink" could also be good?

For `decodedText`, this is not explanatory at all where it comes from. That's why I suggested `decodedQR`

pavanjoshi914: changed name from lnurlwithdraw to lnurlWithdrawLink

src/app/components/QrcodeAdornment/index.tsx
pavanjoshi914: Reason to do this:

* we go to lnulr redeem screen - first entry in stack
* we go to scan qr code screen via qr adornment component - second entry in stack - it replaces previous lnurl redeem screen present in stack
* we scan qr and go back to lnurl redeem screen. we remove scanQrcode screen entry
* in the end we will have single lnurl-redeem screen entry
* now if we press back button it will go directly to receive

rolznz: sounds good, thank you

rolznz (approved):
tACK

pavanjoshi914:
cc @reneaaron

reneaaron (approved):
Added some more improvements and minor style fixes. 

tACK

PR #2514, opened 2023-06-27
feat: onchain swaps

reneaaron:

Sending to onchain addresses.


rolznz:
image

Minor UI issue - recipient address does not wrap

rolznz:
Inconsistency: service fee is not shown but network fee is

image

rolznz:
Minor UI issue: all review content does not fit on one page:

image

rolznz:
image

Should we change this message?

"100000 sats are on their way to <address>". Your sats will appear after X block confirmations.

src/app/screens/Send/index.tsx
```diff
@@ -91,6 +97,13 @@ function Send() {
             },
           },
         });
+      } else if (
+        isAlbyOAuthAccount(auth.account?.connectorType) &&
```
rolznz, line 101: Why are other connectors not supported?

rolznz:
image

Should we update this translation to include "onchain address"?

src/i18n/locales/en/translation.json
```diff
@@ -818,6 +818,27 @@
...
+        "label": "Recipient"
+      },
+      "amount": {
+        "label": "Amount (sats)"
+      },
+      "network_fee": {
+        "label": "Network fee"
+      },
+      "service_fee": {
+        "label": "Service fee"
```
rolznz, line 833: Should we explicitly say "Alby service fee" here? I think it's easier to differentiate for users

src/app/screens/Send/index.tsx
rolznz: From Rene: this endpoint is only available for logged-in users (it is part of the OAuth API)

src/app/screens/SendToBitcoinAddress/index.tsx
```diff
@@ -0,0 +1,379 @@
...
+              message={tCommon("success_message", {
+                amount: getFormattedSats(amountSat),
+                fiatAmount: showFiat ? ` (${fiatAmount})` : ``,
+                destination: bitcoinAddress,
+              })}
+            />
+            <div className="text-center my-4">
+              <Hyperlink
+                href={`https://mempool.space/address/${bitcoinAddress}`}
+                rel="noopener nofollow"
```
rolznz, line 359: I think this should be `target="_blank"`

src/app/utils/index.ts
```diff
@@ -63,6 +63,12 @@ export function getAlbyAccountName(info: GetAccountInformationResponse) {
   return info.email || info.lightning_address || "getalby.com";
 }
 
+export function isBitcoinAddress(address: string) {
+  return /^(?:[13]{1}[a-km-zA-HJ-NP-Z1-9]{25,34}|(bc1|tb1)[a-z0-9]{39,59})$/i.test(
```
rolznz, line 67: where does this come from? should we add a comment here?

src/extension/background-script/actions/swaps/create.ts
```diff
@@ -0,0 +1,23 @@
+import Alby from "~/extension/background-script/connectors/alby";
+import state from "~/extension/background-script/state";
+import { MessageCreateSwap } from "~/types";
+
+const createSwap = async (message: MessageCreateSwap) => {
+  try {
+    const connector = (await state
```
rolznz, line 7: maybe we should add createSwap as an optional function to the connector interface, then we can do it without casting/fetching the Alby one specifically

src/app/utils/index.ts
reneaaron: :+1: https://stackoverflow.com/questions/21683680/regex-to-match-bitcoin-addresses + slightly modified to support testnet addresses

src/extension/background-script/actions/swaps/create.ts
reneaaron: Good idea.

src/app/screens/Send/index.tsx
bumi: can this actually be a function on the `auth` itself?

src/extension/background-script/actions/swaps/create.ts
```diff
@@ -0,0 +1,24 @@
+import state from "~/extension/background-script/state";
+import { MessageCreateSwap } from "~/types";
+
+const createSwap = async (message: MessageCreateSwap) => {
+  try {
+    const connector = await state.getState().getConnector();
```
bumi, line 6: we need it like this now,
but I am wondering if this should be a function on the connector. We move in a direction to add more specialized functions to the connector interface that are actually not universal. 

I think we should try to move this out of the connector and onto the action level here at some point. 
this would make it also easier to have configurable swap services. Those then also just use the connector do pay/create invoices, etc.

src/extension/background-script/actions/swaps/create.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+    const data = await connector.createSwap(message.args);
+
+    return {
+      data,
+    };
+  } catch (e) {
+    console.error(e);
+    return {
+      error: "Creating swap failed: " + e,
```
bumi, line 19: `e` is an object here? should this be `e.message` ?

src/extension/background-script/actions/swaps/info.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    const data = await connector.getSwapInfo();
+
+    return {
+      data,
+    };
+  } catch (e) {
+    console.error(e);
+    return {
+      error: "getSwapInfo failed: " + e,
```
bumi, line 20: also here should this be `e.message`

src/extension/content-script/onendnostr.js
```diff
@@ -1,7 +1,6 @@
-import browser from "webextension-polyfill";
-
 import getOriginData from "./originData";
 import shouldInject from "./shouldInject";
+import browser from "webextension-polyfill";
```
bumi, line 3: why did this get all changed? 

didn't we have the config somewhere that first there are extenal package imports and then the internal ones?

src/i18n/locales/en/translation.json
bumi: I actually think splitting this is too complicated 
the user has no choice in setting either fee, so why not just have one amount the user has to pay?

src/extension/background-script/actions/swaps/create.ts
rolznz: It was like this originally and we did a check if the connector type was Alby. Is there a better way?

bumi: I think for now not. but maybe at some point we support others there and it's open.

src/i18n/locales/en/translation.json
reneaaron: I'd leave it in for now as (especially during high fee environments) users know what they pay for.

src/i18n/locales/en/translation.json
```diff
@@ -924,7 +951,7 @@
     "pasted": "Pasted!",
     "description": "Description",
     "description_full": "Full Description",
-    "success_message": "{{amount}}{{fiatAmount}} sent to {{destination}}",
+    "success_message": "{{amount}}{{fiatAmount}} are on their way to {{destination}}",
```
reneaaron, line 954: @rolznz Did you change the default message that is being used everywhere on purpose? Should we add a separate success message for that?

bumi (approved):

PR #2523, opened 2023-06-29
fix: double slash in url

dylancom:

I got an error trying to connect my Umbrel. The error toast currently shows me a URL that contains a double slash: "https://umbrel.local:8080//v1/getinfo". Should be -> https://umbrel.local:8080/v1/getinfo

### DIFF
--- a/src/app/screens/connectors/ConnectUmbrel/index.tsx
+++ b/src/app/screens/connectors/ConnectUmbrel/index.tsx
@@ -31,7 +31,8 @@ export default function ConnectUmbrel() {
       let lndconnect = new URL(lndconnectUrl);
       lndconnect.protocol = "http:";
       lndconnect = new URL(lndconnect.toString());
-      const url = `https://${lndconnect.host}${lndconnect.pathname}`;
+      const pathname = lndconnect.pathname.replace(/\/$/, "");
+      const url = `https://${lndconnect.host}${pathname}`;
       let macaroon = lndconnect.searchParams.get("macaroon") || "";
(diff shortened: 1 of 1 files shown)


reneaaron (commented):
The pathname always contains a trailing slash, however in L90 another slash is appended:

https://github.com/getAlby/lightning-browser-extension/pull/2523/files#diff-569ed28eef6144884f31508cd6d7118e4d05a49070a40b241a3d2dc37b0cb4d8R91

Wouldn't it be more consistent to remove the duplicate slash there as `formData.url` is supposed to have a trailing slash anyway?

reneaaron:
Other connectors affected by this: 
 - ConnectMyNode
 - ConnectStart9

src/app/screens/connectors/ConnectUmbrel/index.tsx
```diff
@@ -31,7 +31,8 @@ export default function ConnectUmbrel() {
       let lndconnect = new URL(lndconnectUrl);
       lndconnect.protocol = "http:";
       lndconnect = new URL(lndconnect.toString());
-      const url = `https://${lndconnect.host}${lndconnect.pathname}`;
+      const pathname = lndconnect.pathname.replace(/\/$/, "");
```
bumi, line 34: is this `$` correct - that's the end of the line? 

but where does the double slash actually come from? `.host` should not have a slash but `.pathname` starts with a slash. 

so I am wondeing where the double slash comes from. 🤔

reneaaron: @bumi See my review comment.

PR #2524, opened 2023-06-29
feat!: Remove Citadel connector

AaronDewes:

The last Citadel update made some breaking API changes, so this was not working anymore. The next Citadel release will instead utilize the new API to add an account on the dashboard and will use LND/Commando directly, which will also expose more features and prevent issues with 2FA.

### Type of change

### Checklist

- [ ] Manual testing
- [ ] Update Docs & Guides
- For UI-related changes
- [ ] Darkmode
- [ ] Responsive layout


bumi:
What should we do with current users? How will they upgrade Citadel?

reneaaron:
@AaronDewes Also the `alby.addAccount()` APIs that you initiated recently got merged which could make connecting nodes to Alby a lot easier for users, feel free to check it out:  

https://guides.getalby.com/alby-extension-apis/

AaronDewes:
> What should we do with current users? How will they upgrade Citadel?

They will need to disconnect and re-add their wallet. But only using this API from the Citadel dashboard won't work probably because the initial setup wouldn't work...

I'll experiment a bit.

reneaaron:
@AaronDewes What is the current status here? Is this still relevant? Do the existing APIs even work with the current release of Citadel? What connectors will be available for Citadel in future?

PR #2526, opened 2023-06-30
feat: add default keyboard focus style

dylancom:

Currently some elements such as buttons have an explicit keyboard focus style with an orange ring.
Other things like links fallback to the default (blue) styling that the browser gives.
I think we should set a default here that matches with the explicit defined styles and overall brand (using the primary color).

Before:
Scherm­afbeelding 2023-06-30 om 10 34 54

After:
Scherm­afbeelding 2023-06-30 om 10 34 07

### DIFF
--- a/src/app/styles/index.css
+++ b/src/app/styles/index.css
@@ -37,6 +37,10 @@ body {
   font-weight: normal;
 }
 
+:focus-visible {
+  @apply outline outline-2 outline-primary;
+}
+
 .ReactModal__Overlay {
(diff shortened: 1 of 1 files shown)


reneaaron:
Thanks for the PR! 

I now see two different outlines: (orange + white)

![image](https://github.com/getAlby/lightning-browser-extension/assets/100827540/366b8f67-d4c8-4edd-90a7-eda5d247e881

Not sure if that is intentional?

PR #2527, opened 2023-06-30
fix: remove fixed height when no image available

dylancom:

Currently the Publisher card is only showing a tipping button (and the rest is hidden) when there is no image to display.
This happens because of a fixed height being set when there is no image.
I think this is some legacy code?

While visiting: https://github.com/getAlby/lightning-browser-extension

Before:
Scherm­afbeelding 2023-06-30 om 11 10 40

After:
Scherm­afbeelding 2023-06-30 om 11 10 06


reneaaron (approved):
Good catch, thanks for the PR! :raised_hands: 

tACK

PR #2530, opened 2023-07-03
feat: add connection-error component and account error state

### LINKED ISSUE #2377: Improve handling of errors in connectors
### Feature description

Currently errors during account initialization are in some cases not handled, which can lead to some erroneous states in the extension. 

While there is a toast message informing the user that something went wrong, some parts of the UI do not reflect this change. 

**Example**
1. Unlock the extension
2. Choose account A
3. Switch to account B that throws an error during initialization (e.g. node is down, wrong credentials, etc)
4. The extension still displays the balance of the  account A

### Describe the solution

Errors should be handled and a message should be displayed to the user. @stackingsaunter Created some UI to reflect the failed loading in the extension. Feel free to add the designs here. :raised_hands: 

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

No

lujakob:

Add AccountMenu account error state. Add ConnectionError component.

Fixes #2377

### Type of change

### How has this been tested?

### Checklist

- [ ] Manual testing
- [ ] Update Docs & Guides
- For UI-related changes
- [ ] Darkmode
- [ ] Responsive layout

### DIFF
--- a/src/app/router/Popup/Popup.tsx
+++ b/src/app/router/Popup/Popup.tsx
@@ -1,3 +1,4 @@
+import ConnectionError from "@components/ConnectionError";
 import Navbar from "@components/Navbar";
 import ConfirmPayment from "@screens/ConfirmPayment";
 import Home from "@screens/Home";
@@ -9,8 +10,10 @@ import LNURLWithdraw from "@screens/LNURLWithdraw";
 import Receive from "@screens/Receive";
 import Send from "@screens/Send";
 import Unlock from "@screens/Unlock";
+import { useEffect, useState } from "react";
 import { HashRouter, Outlet, Route, Routes } from "react-router-dom";
 import { ToastContainer } from "react-toastify";
+import { useAccount } from "~/app/context/AccountContext";
 import Providers from "~/app/context/Providers";
 import OnChainReceive from "~/app/screens/OnChainReceive";
 
@@ -56,12 +59,23 @@ function Popup() {
 }
 
 const Layout = () => {
+  const { account } = useAccount();
+  // keep last tries result, to prevent component flick during loading state
+  const [lastAccountFailed, setLastAccountFailed] = useState<boolean>(false);
+
+  useEffect(() => {
+    // skip the state during loading phase where only the id is present
+    if (Object.prototype.hasOwnProperty.call(account, "name")) {
+      setLastAccountFailed(!!account?.error);
+    }
+  }, [setLastAccountFailed, account]);
+
   return (
     <div className="flex flex-col h-full">
       <Navbar />
 
       <main className="flex flex-col grow min-h-0">
-        <Outlet />
+        {lastAccountFailed || account?.error ? <ConnectionError /> : <Outlet />}
         <ToastContainer autoClose={10000} hideProgressBar={true} />
       </main>
     </div>
--- a/src/common/lib/api.ts
+++ b/src/common/lib/api.ts
@@ -33,6 +33,7 @@ export interface AccountInfoRes {
   currentAccountId: string;
   info: { alias: string; pubkey?: string };
   name: string;
+  error?: string;
 }
 
 export interface GetAccountRes
@@ -70,7 +71,8 @@ export const swrGetAccountInfo = async (
   const accountsCache = await getAccountsCache();
 
   return new Promise((resolve, reject) => {
-    if (accountsCache[id]) {
+    // don't apply cache for connectors that failed to connect
+    if (accountsCache[id] && !accountsCache[id].error) {
       if (callback) callback(accountsCache[id]);
       resolve(accountsCache[id]);
     }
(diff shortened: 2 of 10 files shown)


lujakob:
@reneaaron this is a working draft. Refinement (especially in styles) is needed, but you could check out my changes/architecture if you want.

src/common/lib/api.ts
```diff
@@ -70,26 +71,37 @@ export const swrGetAccountInfo = async (
   const accountsCache = await getAccountsCache();
 
   return new Promise((resolve, reject) => {
-    if (accountsCache[id]) {
+    // don't apply cache for connectors that failed to connect
+    if (accountsCache[id] && !accountsCache[id].error) {
```
reneaaron, line 75: Can we keep accounts with errors out of the cache? Why would we cache them if we exclude them from the cache here? :thinking:

src/common/lib/api.ts
```diff
@@ -70,26 +71,37 @@ export const swrGetAccountInfo = async (
...
-        const account = {
-          id,
-          name,
-          alias,
-          balance,
-          currency: currency || "BTC", // set default currency for every account
-        };
+        let account: AccountInfo;
+        if (response.error) {
+          account = {
```
reneaaron, line 85: Shouldn't this case be handled in the catch() block here? 

msg.request should throw an Error():

https://github.com/lujakob/lightning-browser-extension/blob/feat/not-reachable-error/src/common/lib/msg.ts#L19-L21

src/app/router/Popup/Popup.tsx
```diff
@@ -56,12 +59,23 @@ function Popup() {
...
+  // keep last tries result, to prevent component flick during loading state
+  const [lastAccountFailed, setLastAccountFailed] = useState<boolean>(false);
+
+  useEffect(() => {
+    // skip the state during loading phase where only the id is present
+    if (Object.prototype.hasOwnProperty.call(account, "name")) {
+      setLastAccountFailed(!!account?.error);
+    }
+  }, [setLastAccountFailed, account]);
+
```
reneaaron, line 72: That feels a bit complex on the first sight. What is it that you want to prevent here? 

Isn't it natural to go from error -> loading -> success / error?

reneaaron (commented):
Thanks for the PR @lujakob! Left some first feedback, generally I think we are on the right track here. :+1:

src/common/lib/api.ts
lujakob: Your suggestion was my first approach. The problem is, the argument of `throw new Error()` is of type string. But I need to pass an object, to be able to get the `name` of the connector and `error`. To change this, I need to refactor dozens of places where `msg.request` is called in a try/catch block. So on the one hand this is really a big refactor, on the other hand I think what we do now is not completely wrong: we return a connector in error state, that has `name` as its only property. Considering these two points I opted for the current approach. (Maybe it doesn't feel so wrong for me, because I used this approach in other applications, where the error returns data and therefore is not dealt with as a real error) What do you think?

src/app/router/Popup/Popup.tsx
lujakob: You can disable it and see it. If you replace 
`{lastAccountFailed || account?.error ? <ConnectionError /> : <Outlet />}` with 
`{account?.error ? <ConnectionError /> : <Outlet />}` you see during the loading state after clicking on "Try again" you see the `Outlet` component for a short moment. A glitch...

src/app/components/Button/index.tsx
```diff
@@ -67,12 +67,15 @@ const Button = forwardRef(
...
         disabled={disabled}
       >
-        {loading && (
-          <div className={direction === "row" ? "mr-2" : ""}>
-            <Loading color={primary ? "white" : "black"} />
+        {(loading || icon) && (
+          <div
+            className={classNames(
+              direction === "row" ? "mr-2 flex justify-center min-w-[24px]" : ""
+            )}
```
lujakob, line 74: @reneaaron this change in styles is necessary, because having an icon and loading state results in a very small horizontal jump of the text, due to the rotating loading icon and the icon itself not having the same width. I'm kind of unsatisfied with this change, but I found no better way in preventing this rather annoying UX problem. Please have a look and let me know. You can revert by "show diff" branch master and revert the changes to see what I mean.

src/common/lib/api.ts
lujakob: I changed it.

lujakob:
@reneaaron I updated with master, added the connection error to the options page and abstracted the "connectionFailed" state to a hook function. I added extensive comments to the hook function for explanation. I think this is 95% done now, so let me know what you think. The LND getInfo request url is erroneous for testing. If you throttle/delay the network requests, you can see the loading spinner on the button and test the screen not flicking back to the normal content during loading.

rolznz:
We currently have an issue where if you're unable to connect to a connector, then it's impossible to use LNURL auth there. I think it would it make sense to be able to still switch connector, but it be in a failed state (like this). And you can still use the LNURL auth  and master key features. What do you think? @lujakob @reneaaron @bumi

reneaaron:
Yeah, but that should be the case already, right? The failed state would only be reflected on the wallet page (but not in the prompts), right?

lujakob:
> > Yeah, but that should be the case already, right? The failed state would only be reflected on the wallet page (but not in the prompts), right?
> 
> I don't think this has been implemented yet. I can't see any change to how the connector init works so that it will still return the connector if it fails

@rolznz do you think it makes sense for me to look into this? I would need some more information as I don't have the full picture here. 

My understanding of lnurl auth is it's a single sign on to log into websites. If I am on www.getalby.com and "login with lightning" the popup opens. Is this the whole thing or is there more to know? Why is there a path for the lnurl auth page in the Options.tsx?

By the way, how do you debug/test lnurl auth in dev mode? If I can successfully login to my www.getalby.com account with the production extension, but not with the dev mode extension. 

Maybe a short chat would be helpful?

PR #2536, opened 2023-07-04
fix: move primary buttons above fold for smaller screen sizes

reneaaron:

Optimize height of welcome screens so they fit for smaller screen sizes without moving the main button below the fold:

image

For larger screen sizes the elements on the page are centered vertically.


bumi:
with errors: 
image
without errors:
image

PR #2538, opened 2023-07-05
feat: add account detail pages header

### LINKED ISSUE #2521: [Feature] Display profile header for account related pages
### Feature description

Currently you can have one account selected and edit another one, but in account subpages it is not obvious which account you are editing.

### Describe the solution

- Display the profile header on subpages of the account. (currently I think only the secret key related things are "below" the account).

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

lujakob:

Add header with account name to account detail pages.

Fixes #2521

### Type of change

Bildschirmfoto 2023-07-05 um 07 33 43

### How has this been tested?

Manually

### Checklist

- [x] Manual testing
- [x] Update Docs & Guides
- For UI-related changes
- [x] Darkmode
- [x] Responsive layout


lujakob:
@reneaaron please have a look. Not sure if you addressed all desired pages.

* I created an AccountDetailHeader component
* I added a cancel button to the generate page, as the import page had one as well
* I changed the loading state for the backup page and added a `backup.no_mnemonic_found`, because when I visited this page via typing the url there was a continuous loading state, due to the mnemonic data being null.

reneaaron:
@lujakob Thanks for the PR, I refactored it to use subroutes / layouts instead of implementing the same things over and over on different pages. 

> I added a cancel button to the generate page, as the import page had one as well

Cool, thanks.

> I changed the loading state for the backup page and added a backup.no_mnemonic_found, because when I visited this page via typing the url there was a continuous loading state, due to the mnemonic data being null.

I don't think we need to handle that and removed it. 

Please have a look :blush:

lujakob:
> @lujakob Thanks for the PR, I refactored it to use subroutes / layouts instead of implementing the same things over and over on different pages.
> 
> > I added a cancel button to the generate page, as the import page had one as well
> 
> Cool, thanks.
> 
> > I changed the loading state for the backup page and added a backup.no_mnemonic_found, because when I visited this page via typing the url there was a continuous loading state, due to the mnemonic data being null.
> 
> I don't think we need to handle that and removed it.
> 
> Please have a look 😊

Looks good, thanks for helping out.

src/app/components/AccountDetailLayout/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+    setExportModalIsOpen(false);
+  }
+
+  return (
+    <>
+      <Header
+        title={t("title1")}
+        headerLeft={
+          <IconButton
+            onClick={() => navigate("/accounts")}
```
bumi, line 67: Can we make this a back button? when I am on a "subpage" and want to go back I end up on the accounts index page again.

reneaaron: Good catch. Will update :ok_hand:

src/app/components/AccountDetailLayout/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  useEffect(() => {
+    (async () => {
+      const account = await api.getAccount(id);
+      setAccount(account);
+    })();
+  }, [id, navigate]);
+
+  async function exportAccount({ id, name }: AccountAction) {
```
rolznz, line 46: Isn't this only possible from the account detail page? maybe it should be in there instead.

Also, I cannot see the export functions when I add an Lndhub account, strange

image

rolznz: Should the lndhub-specific code be moved to a separate component?

src/app/screens/Accounts/ImportSecretKey/index.tsx
```diff
@@ -13,31 +13,35 @@ import MnemonicInputs from "~/app/components/mnemonic/MnemonicInputs";
...
           // do not allow user to import a mnemonic if they already have one for the current account
           // go to account settings
           navigate(`/accounts/${id}`);
         }
-        setHasFetchedData(true);
       } catch (e) {
         console.error(e);
         if (e instanceof Error) toast.error(`Error: ${e.message}`);
+      } finally {
+        setLoading(false);
```
rolznz, line 44: what happens if the catch block occurs? it might be better to not set loading to false in that case

src/app/components/AccountDetailLayout/index.tsx
reneaaron: I think this stuff will be moved to the Account Detail page. I'd rather wait for @stackingsaunter's proposal before we put more work into creating components right now.

reneaaron (commented):
Addressed review comments and improved "back" navigation.

src/app/components/AccountDetailLayout/index.tsx
lujakob: @reneaaron didn't we talk about this in Telegram DM? It used to be only in the Detail Page, which was the issue with having the layout parent page. Because the layout page needs to know (e.g. by path segment) which page we are in and show the export button only for the detail page. You mentioned why should it only be in the detail page. If I remember correctly.

The problem is, this csv export button is part of the header markup. But it should only be displayed in the detail page (if this is the case). That's why I thought it might be better to have the header as a component, with a prop to configure displaying the export button. This way it can be controlled within the subpages.

stackingsaunter:
I propose changes to the accounts header to make it smaller and take less space and move "export" ino settings as an item (now known as "Connect to mobile wallet"

image

Zeus links to: https://zeusln.app/
Blue Wallet links to: https://bluewallet.io/

[Figma designs and handoff](https://www.figma.com/file/NHFCZOjPedXkAutglKmhzA/Accounts-revamp?type=design&node-id=0%3A1&mode=design&t=7T8V9tQcuTmW71hZ-1) 

This would also require change the export modal name to "Connect to mobile wallet".

rolznz:
@reneaaron why did this get marked as a draft? is it still planned to be worked on?

reneaaron:
@rolznz I noticed some problems during the final tests: 

When you change your account name it's not updated in the account header (but it is updated in the account dropdown).

reneaaron:
L

lujakob:
> @rolznz I noticed some problems during the final tests:
> 
> When you change your account name it's not updated in the account header (but it is updated in the account dropdown). Let me know if you have any ideas how to fix that. Need to take another look at that...

I see the name displayed in the `AccountLayoutPage` is fetched on component mount via the parameter id.

https://github.com/lujakob/lightning-browser-extension/blob/ad5ef0d234c7e00f82a301c29d43769a90812362/src/app/components/AccountDetailLayout/index.tsx#L22C1-L23

It is updated in the account header only when the account name being changed is the currently selected account. Which makes sense.

The parent `AccountLayoutPage` component where the name is displayed has no way of knowing when the child `AccountDetail` updates the name. So the only way I can think of right now is, reloading the page after updating the name, which is kinda dirty :(

lujakob:
> > @rolznz I noticed some problems during the final tests:
> > When you change your account name it's not updated in the account header (but it is updated in the account dropdown). Let me know if you have any ideas how to fix that. Need to take another look at that...
> 
> I see the name displayed in the `AccountLayoutPage` is fetched on component mount via the parameter id.
> 
> https://github.com/lujakob/lightning-browser-extension/blob/ad5ef0d234c7e00f82a301c29d43769a90812362/src/app/components/AccountDetailLayout/index.tsx#L22C1-L23
> 
> It is updated in the account header only when the account name being changed is the currently selected account. Which makes sense.
> 
> The parent `AccountLayoutPage` component where the name is displayed has no way of knowing when the child `AccountDetail` updates the name. So the only way I can think of right now is, reloading the page after updating the name, which is kinda dirty :(

I saw you found a good solution, Rene, perfect.

src/app/screens/Accounts/BackupSecretKey/index.tsx
```diff
@@ -10,28 +10,33 @@ import SecretKeyDescription from "~/app/components/mnemonic/SecretKeyDescription
...
   const fetchData = useCallback(async () => {
     try {
+      setLoading(true);
       const accountMnemonic = await api.getMnemonic(id as string);
       setMnemonic(accountMnemonic);
     } catch (e) {
       console.error(e);
       if (e instanceof Error) toast.error(`Error: ${e.message}`);
+    } finally {
+      setLoading(false);
```
rolznz, line 31: I think this should only be set to false if the fetch actually works

PR #2546, opened 2023-07-07
fix: update keyboard shortcut to mv3

reneaaron:

Action commands seem to have been moved with MV3: 

https://developer.chrome.com/docs/extensions/reference/commands/#action-commands


reneaaron:
The keyboard shortcut doesn't work for me in chrome (at least not with the dev build) :thinking:

bumi:
does it work with this?

reneaaron:
No, only after setting it manually in the chrome settings: 

chrome://extensions/shortcuts

Don't know if that is because this shortcut might already be used.

pavanjoshi914:
there is a bug in chrome, https://bugs.chromium.org/p/chromium/issues/detail?id=1478453#c_ts1695708024 it works for firefox currently maybe we merge it for now. since it will take time to get this fix from their side

PR #2555, opened 2023-07-10
fix: missing lang key

reneaaron:


bumi:
do you know how this happens?

reneaaron:
I guess if none of the required keys is translated on weblate:

https://github.com/getAlby/lightning-browser-extension/blob/fix/missing-language-key/src/i18n/i18nConfig.ts#L60-L61

PR #2556, opened 2023-07-10
fix: add default for locale

reneaaron:

i18next changed `resolvedLanguage` to be optional in a recent upgrade which causes a type error.

### DIFF
--- a/src/common/settings.ts
+++ b/src/common/settings.ts
@@ -10,7 +10,7 @@ export const DEFAULT_SETTINGS: SettingsStorage = {
   isUsingLegacyLnurlAuthKey: false,
   userName: "",
   userEmail: "",
-  locale: i18n.resolvedLanguage,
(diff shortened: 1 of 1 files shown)


bumi:
sorry, I had merged this. the changelog looked fine. 

I am wondering if we should use that one there even? does it add too much complexity to the defaults use in the state?

reneaaron:
> I am wondering if we should use that one there even? does it add too much complexity to the defaults use in the state?

We've used it all along, so I wouldn't change anything here now. 

To me it would still make sense to use it, as otherwise always "en" will be persisted into your settings. If you now have a browser that has a default language set to some other language I would assume that this language is also persisted into your settings. (didn't test it)

bumi:
I just think that the `state` should be small and we should be careful there of having much logic in there. (as here hidden in detecting the browser language)

PR #2579, opened 2023-07-20
fix: number formatting for range label

reneaaron:

Formats numbers in range label.


rolznz (approved):
utACK

PR #2581, opened 2023-07-20
fix: add chrome prod pubkey to manifest json

rolznz:
Ensures production builds always have the same extension ID

This wasn't the case when loading an unpacked extension (e.g. nightly build)

Manually tested locally by importing an unpacked production extension build (built locally).

Note: this public key is freely available (included in extension folder e.g. 
```
C:\Users\MyUsername\AppData\Local\Google\Chrome\User Data\Default\Extensions\iokeahhehimjnekafflcihljlcjccdbe\2.1.1_0\manifest.json
```


reneaaron (approved):
Looks good to me.

utACK

PR #2587, opened 2023-07-24
fix: credential loading for different builds

reneaaron:

 - Remove build variables from GH actions (not needed anymore as they are maintained in our webpack config)
 - Added support for loading oAuth credentials from environment variables (for chrome & firefox simultaneously through running yarn package)


rolznz (approved):
utACK

PR #2588, opened 2023-07-24
fix: remove getInfo side effect in alby connector

rolznz:
We want to completely remove this side effect in getInfo() - for now avatar will still get updated but I added a TODO for that one


bumi:
why can't we remove it now? 
such side effects are not good and saving is problematic there. 

instead of rewriting it let's fix the root cause.

reneaaron (approved):
utACK

PR #2589, opened 2023-07-25
feat: nostr setup wizard

reneaaron:

Better onboarding for users that want to use Alby with nostr.


src/app/components/RadioButton/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  onClick: () => void;
+};
+
+interface IconTypeProps {
+  className: string;
+}
+
+type IconType = (props: IconTypeProps) => JSX.Element;
+
+export default function RadioButton({
```
rolznz, line 16: This doesn't feel like a RadioButton to me, it feels more like a card (I know we already have a card component, which should probably be renamed/moved because it is not for general use)

src/app/screens/Accounts/NostrSetup/NostrSetup.tsx
```diff
@@ -0,0 +1,97 @@
...
+import Hyperlink from "~/app/components/Hyperlink";
+import RadioButton from "~/app/components/RadioButton";
+import RadioButtonGroup from "~/app/components/RadioButton/Group";
+
+function NostrSetup() {
+  const { t } = useTranslation("translation", {
+    keyPrefix: "accounts.account_view.nostr.setup",
+  });
+  const navigate = useNavigate();
+  const [step, setStep] = useState("start");
```
rolznz, line 21: Can you add typing: `useState<"start" | "import">`

src/app/components/RadioButton/index.tsx
reneaaron: You are right, especially without the "radio" behavior that makes no sense anymore. I chose to rename to "CardButton". The Card should be renamed yes, seems like a very specific use case.

rolznz (approved):

PR #2594, opened 2023-07-26
fix: resolve enable parameter bug

pavanjoshi914:

set enable after promise is returned so that enabled parameter is upto date

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/extension/providers/nostr/index.ts
+++ b/src/extension/providers/nostr/index.ts
@@ -19,13 +19,13 @@ export default class NostrProvider {
     if (this.enabled) {
       return { enabled: true };
     }
-    return await this.execute("enable").then((result) => {
-      if (typeof result.enabled === "boolean") {
-        this.enabled = result.enabled;
-      }
-      return result;
-    });
+    const result = await this.execute("enable");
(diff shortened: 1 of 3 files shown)


src/extension/providers/nostr/index.ts
```diff
@@ -19,9 +19,13 @@ export default class NostrProvider {
     if (this.enabled) {
       return { enabled: true };
     }
-    return await this.execute("enable");
+    return await this.execute("enable").then((result) => {
+      if (typeof result.enabled === "boolean") {
+        this.enabled = result.enabled;
+      }
+      return result;
+    });
```
bumi, line 27: what's the best practice? is mixing await and then ok? 

or is something like this better?

```
const result = await this.execute("enable");
if (typeof result.enabled === "boolean") {
  this.enabled = result.enabled;
}
return result;
```

pavanjoshi914: thanks for updating it

PR #2595, opened 2023-07-27
feat: disable autocomplete on other form fields

pavanjoshi914:
don't allow password manager to fill password field other then fields on unlock screen add types to text fields

Fixes  #2480 new pr for #2497  

### Type of change

### How has this been tested?
manually checking auto complete on firefox

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/app/screens/LNURLPay/index.tsx
+++ b/src/app/screens/LNURLPay/index.tsx
@@ -469,7 +469,6 @@ function LNURLPay() {
                       <div className="mt-4">
                         <TextField
                           id="comment"
-                          type="text"
                           label={t("comment.label")}
                           placeholder={tCommon("optional")}
                           onChange={(e) => {
@@ -498,7 +497,6 @@ function LNURLPay() {
                           <div className="mt-4">
                             <TextField
                               id="name"
-                              type="text"
                               label={t("name.label")}
                               placeholder={tCommon("optional")}
                               value={userName}
@@ -512,7 +510,6 @@ function LNURLPay() {
                           <div className="mt-4">
                             <TextField
                               id="email"
-                              type="text"
                               label={t("email.label")}
                               placeholder={tCommon("optional")}
                               value={userEmail}
--- a/src/app/screens/LNURLPay/index.tsx
+++ b/src/app/screens/LNURLPay/index.tsx
@@ -469,6 +469,7 @@ function LNURLPay() {
                       <div className="mt-4">
                         <TextField
                           id="comment"
+                          type="text"
                           label={t("comment.label")}
                           placeholder={tCommon("optional")}
                           onChange={(e) => {
@@ -497,6 +498,7 @@ function LNURLPay() {
                           <div className="mt-4">
                             <TextField
                               id="name"
+                              type="text"
                               label={t("name.label")}
                               placeholder={tCommon("optional")}
                               value={userName}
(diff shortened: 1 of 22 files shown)


src/app/screens/LNURLPay/index.tsx
```diff
@@ -497,6 +498,7 @@ function LNURLPay() {
                           <div className="mt-4">
                             <TextField
                               id="name"
+                              type="text"
```
bumi, line 501: I think type text is default

src/app/screens/connectors/ConnectLnd/index.tsx
```diff
@@ -175,6 +176,7 @@ export default function ConnectLnd() {
         <div>
           <TextField
             id="macaroon"
+            type="text"
```
bumi, line 179: let's make this also a password field

src/app/screens/connectors/ConnectMyNode/index.tsx
```diff
@@ -133,6 +133,7 @@ export default function ConnectMyNode() {
       <div className="mt-6">
         <TextField
           id="lndconnect"
+          type="text"
```
bumi, line 136: let's make this also a password field

src/app/screens/Send/index.tsx
```diff
@@ -126,6 +126,7 @@ function Send() {
           <div className="pt-4">
             <TextField
               id="invoice"
+              type="text"
```
reneaaron, line 129: Why did you add the type everywhere? It's "text" by default?

src/app/screens/connectors/ConnectRaspiBlitz/index.tsx
```diff
@@ -164,6 +165,7 @@ export default function ConnectRaspiBlitz() {
         <div>
           <TextField
             id="macaroon"
+            type="text"
```
bumi, line 168: let's make this also a password field

reneaaron (changes requested):

src/app/screens/connectors/ConnectStart9/index.tsx
```diff
@@ -136,6 +136,7 @@ export default function ConnectStart9() {
       <div className="mt-6">
         <TextField
           id="lndconnect"
+          type="text"
```
bumi, line 139: let's make this also a password field. 

lndconnect holds secrets and thus we can handle it as a password field I guess.

src/app/screens/connectors/ConnectUmbrel/index.tsx
```diff
@@ -133,6 +133,7 @@ export default function ConnectUmbrel() {
       <div className="mt-6">
         <TextField
           id="lndconnect"
+          type="text"
```
bumi, line 136: let's make this also a password field

bumi (changes requested):

src/app/screens/Send/index.tsx
pavanjoshi914: we have type added as text at few of the text field and few of them doesn't contain them. so shall we remove type field at all the fields containing type="text"

it's better to be consistent!

pavanjoshi914: having type explicitly in the code would be better. isn't it?

src/app/screens/LNURLPay/index.tsx
pavanjoshi914: we have type="text" at few of the text field and few of them doesn't contain them.  it's better to be consistent!

shall we remove it from everywhere or add it everywhere? isn't this better and readable to have type explicitly defined

PR #2597, opened 2023-07-28
refactor: refactor password view into separate component

pavanjoshi914:

add password toggle view to all the password fields
refactor password view into separate component
add toggle field to new password fields

### How has this been tested?

tested all the instances

### Checklist

- [X] Manual testing
- [x] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


reneaaron:
Do an outer component (a consumer of such a component) ever need to control the visibility of this field? 

I couldn't come up with a practical use case where this is needed and was wondering if we could move the handling of switching between text / password into this component so outside APIs could just consume this component and wouldn't need to 
 - track the state (passwordVisible)
 - register a function to change the state (https://github.com/getAlby/lightning-browser-extension/pull/2597/files#diff-dd145cd63524105648c5b0361bd65880f955aab182d77391c261fba2de4d8c36R175)
themselves.

pavanjoshi914:
> 



> Do an outer component (a consumer of such a component) ever need to control the visibility of this field?
> 
> I couldn't come up with a practical use case where this is needed and was wondering if we could move the handling of switching between text / password into this component so outside APIs could just consume this component and wouldn't need to
> 
> * track the state (passwordVisible)
> * register a function to change the state ([#2597 (files)](https://github.com/getAlby/lightning-browser-extension/pull/2597/files#diff-dd145cd63524105648c5b0361bd65880f955aab182d77391c261fba2de4d8c36R175))
>   themselves.


I agree! but if we move whole text field into separate component, it would be too complex to pass other fields suchs as placeholder, id, and labels of textfield into separate component. leading to more complexity

current approach probably will be less complex and also allows us to achieve refactoring. now we can simply put a textfield in a seamless way and add a toggle view adornment without considering much details to pass

src/app/components/PasswordForm/index.tsx
```diff
@@ -135,25 +123,16 @@ export default function PasswordForm<
...
-                  setPasswordConfirmationView(!passwordConfirmationView)
-                }
-              >
-                {passwordConfirmationView ? (
-                  <HiddenIcon className="h-6 w-6" />
-                ) : (
-                  <VisibleIcon className="h-6 w-6" />
-                )}
-              </button>
+              <PasswordViewAdornment
```
rolznz, line 131: Just on the naming here normally we start with a Verb "ViewPasswordAdornment", but I wonder if "TogglePasswordVisibilityAdornment" or something would be more accurate (I know it is a bit long)

rolznz:
@reneaaron @pavanjoshi914 one example of having to control the visibility from the outside is the MnemonicInputs component. Actually it would be nice to use this new component there too - if you pass a value to the component, then it should use that. Otherwise it will control the value itself. What do you think?

src/app/components/PasswordForm/index.tsx
pavanjoshi914: didn't we follow a similar convention for qrcode scanner component? name was ```QrcodeAdornment``` what we want to specify action in the name there?

viewpassword - action.
passwordView -  adornment for the password view component. makes sense?

rolznz:
image

@pavanjoshi914 I spotted a couple bugs:
- icon shows visible when the field is hidden
- there should be 3 columns.

Please check the styling on all the usages of the new component and make sure nothing has changed

pavanjoshi914:
> 

which browser you are using? i can't reproduce this
image

rolznz:
@pavanjoshi914 I merged master into your branch. Please pull the latest changes

pavanjoshi914:
> image
> 
> @pavanjoshi914 I spotted a couple bugs:
> 
> * icon shows visible when the field is hidden
> * there should be 3 columns.
> 
> Please check the styling on all the usages of the new component and make sure nothing has changed

added feature to allow components to handle viewing the password field itself
2 column bug happened after introducing latest changes. worked on latest master and resolved , thanks @rolznz

rolznz (approved):
tACK

PR #2599, opened 2023-07-28
fix: change webln enable to void

### LINKED ISSUE #2596: why does webln.enable() return an object?
According to [this spec](https://www.webln.guide/building-lightning-apps/webln-reference/webln.enable) the enable function has a void return type

However the webln object injected by Alby has an enable function that does not adhere to this spec so it seems some app devs have assumed that enable may return an object with `enabled: true` or `isEnabled: true` 

Seems this creates confusion for WebLN providers since `enable()` could be implemented in inconsistent ways

Cross-posting this in more than one place (https://github.com/BoltzExchange/boltz-web-app/issues/229) since I'm not sure what the best approach to fix this is... implementations are either building on an outdated spec or they are not compliant with a spec that cannot be enforced 🤔 

any suggestions on what to do here?

reneaaron:

Fixes #2596


otech47 (commented):
looks nice and spec-compliant :)

tracking potentially affected apps here: https://github.com/getAlby/lightning-browser-extension/issues/2596#issuecomment-1656836070

PR #2600, opened 2023-07-28
refactor: providers

pavanjoshi914:
1. add common provider class
2. extend existing providers with common provider
3. extract event methods in common provider class
4. add events in all the providers
5. refactor: pass scope to the base class via constructor
6. remove explicit usage of scope form this.execute
7. rename basec class and refactor it
8. add _checkEnabled function to check if provider is enabled and throw error
9. remove all the instances of explict checks from all the provider methods
10. shift enable function to base class and remove all the instances of enable from providers 
11. handle nostr, on and off error throwing case explicitly

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


rolznz:
@pavanjoshi914 awesome! can we move the enabled / enable() functionality there too?

src/extension/providers/events/index.ts
```diff
@@ -0,0 +1,20 @@
+import EventEmitter from "events";
+
+export default class EventsProvider {
+  private _eventEmitter: EventEmitter;
+
+  constructor() {
+    this._eventEmitter = new EventEmitter();
+  }
+
+  async on(...args: Parameters<EventEmitter["on"]>) {
```
bumi, line 10: does not need to be async and it should not return anything

src/extension/providers/commonProvider/index.ts
```diff
@@ -0,0 +1,43 @@
+import { EventEmitter } from "events";
+import { postMessage } from "~/extension/providers/postMessage";
+import { PromiseQueue } from "~/extension/providers/promiseQueue";
+
+export default class CommonProvider {
+  enabled: boolean;
+  private _eventEmitter: EventEmitter;
+  private _queue: PromiseQueue;
+
+  constructor() {
```
reneaaron, line 10: Can we pass the scope to the base class? That would allow us to get rid of the first parameter for every `execute()` call from deriving classes.

src/extension/providers/commonProvider/index.ts
```diff
@@ -0,0 +1,43 @@
+import { EventEmitter } from "events";
+import { postMessage } from "~/extension/providers/postMessage";
+import { PromiseQueue } from "~/extension/providers/promiseQueue";
+
+export default class CommonProvider {
```
reneaaron, line 5: I think `ProviderBase` would be a better name. Also maybe move it to the `src/extensions/providers/` directory, because it's not really a provider in that sense but just a base class.

reneaaron (commented):
Great refactoring, love to see that quite some code got deleted! :rocket:

src/extension/providers/liquid/index.ts
```diff
@@ -1,26 +1,17 @@
...
-    this._queue = new PromiseQueue();
-  }
-
+export default class LiquidProvider extends CommonProvider {
   async enable(): Promise<void> {
     if (this.enabled) {
       return;
     }
-    const result = await this.execute("enable");
+    const result = await this.execute("liquid", "enable");
```
rolznz, line 14: can the `enable` method also be moved to the base class?

src/extension/providers/commonProvider/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  private _queue: PromiseQueue;
+
+  constructor() {
+    this.enabled = false;
+    this._eventEmitter = new EventEmitter();
+    this._queue = new PromiseQueue();
+  }
+
+  on(...args: Parameters<EventEmitter["on"]>) {
+    if (!this.enabled) {
```
rolznz, line 17: Could this enabled check be moved to the base class, and just pass the method of the name being called? something like `this._checkEnabled("on");`

src/extension/providers/liquid/index.ts
pavanjoshi914: resolved! i didn't refactored it because of the way i was using introducing scope in refactored code! replacing the logic, now allow us to refactor this as well. thanks

rolznz (approved):
tACK

PR #2620, opened 2023-08-02
fix: e2e tests

rolznz:


bumi:
still some only error?

tests/e2e/helpers/loadExtension.ts
```diff
@@ -149,23 +155,83 @@ export async function loginToExistingAlbyAccount(page: Page) {
...
+  const oauthDocument4 = await getDocument(oauthPage);
+
+  const oauthConfirmAuthButton = await getByText(
+    oauthDocument4,
+    "Connect with Alby Extension"
+  );
+  oauthConfirmAuthButton.click();
+
+  let retries = 0;
+  const MAX_RETRIES = 20;
```
reneaaron, line 219: Why would we want to retry here? Is there some reason it wouldn't work at first, but then it would after retrying?

rolznz: The variable maybe could be renamed, hmm. What it's doing is waiting for the OAuth modal to close, and the redirection to occur. I couldn't find a better way to do it.

PR #2628, opened 2023-08-03
fix: make wallet page the startpage for the options view

reneaaron:

On the options.html page `/` currently redirects you to the websites. 
Since we redirect users to `/` when switching accounts this takes you to the websites page. 

I think it would be better to take users to the wallet page as it's probably the most commonly used one.

### DIFF
--- a/src/app/router/Options/Options.tsx
+++ b/src/app/router/Options/Options.tsx
@@ -54,7 +54,7 @@ function Options() {
               </RequireAuth>
             }
           >
-            <Route index element={<Navigate to="/publishers" replace />} />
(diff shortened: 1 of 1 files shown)


bumi:
hmm. I am not sure about that. 
We still show only the outgoing transactions from within Alby for example

reneaaron:
I think the wallet just makes more sense here: from there users can access the most commonly used actions (send, receive, etc). 

I can't imagine a situation where it would be more helpful to be taken to the websites than the wallet page when I switch accounts. (it's also the same behavior as in the popup)

PR #2630, opened 2023-08-03
fix: events improvements

pavanjoshi914:

remove enable from emit function which causes unnecessary dialogue open post message only for webln enabled pages, so that malicious pages doesn't listen to such events this also results in emitting the event only when we are on webln enabled page

### Type of change

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/extension/content-script/onendwebln.js
+++ b/src/extension/content-script/onendwebln.js
@@ -41,7 +41,10 @@ async function init() {
     }
     // forward account changed messaged to inpage script
     else if (request.action === "accountChanged" && isEnabled) {
-      window.postMessage({ action: "accountChanged", scope: "webln" }, "*");
+      window.postMessage(
+        { action: "accountChanged", scope: "webln" },
+        window.location.origin
+      );
     }
   });
 
--- a/src/extension/content-script/onendwebln.js
+++ b/src/extension/content-script/onendwebln.js
@@ -26,7 +26,6 @@ const disabledCalls = ["webln/enable"];
 let isEnabled = false; // store if webln is enabled for this content page
 let isRejected = false; // store if the webln enable call failed. if so we do not prompt again
 let callActive = false; // store if a webln call is currently active. Used to prevent multiple calls in parallel
-
 async function init() {
   const inject = await shouldInject();
   if (!inject) {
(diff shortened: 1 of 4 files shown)


bumi:
I think the event should be published for all the enabled tabs. 
and not only the currently active one as selected [here](https://github.com/getAlby/lightning-browser-extension/pull/2504/files#diff-1dfe211fc8f084873d2cbcb1d2b9203b35f6212f12ce78a696962bde6b9e0744R50-R53)

//cc @rolznz what do you think?

src/extension/content-script/onendwebln.js
```diff
@@ -41,7 +40,7 @@ async function init() {
       extractLightningData();
     }
     // forward account changed messaged to inpage script
-    else if (request.action === "accountChanged") {
+    else if (request.action === "accountChanged" && isEnabled) {
```
rolznz, line 43: The line below, can we replace "*"? we only want the inpage script to receive the message so it can emit the event for any listeners. Currently all frames on the website are receiving it.

rolznz:
> I think the event should be published for all the enabled tabs. and not only the currently active one as selected [here](https://github.com/getAlby/lightning-browser-extension/pull/2504/files#diff-1dfe211fc8f084873d2cbcb1d2b9203b35f6212f12ce78a696962bde6b9e0744R50-R53)
> 
> //cc @rolznz what do you think?

@pavanjoshi914 please could you make this change too?

src/extension/content-script/onendwebln.js
pavanjoshi914: we can use window.location.origin here. which returns domain without any routes/ query parameters

pavanjoshi914:
> > I think the event should be published for all the enabled tabs. and not only the currently active one as selected [here](https://github.com/getAlby/lightning-browser-extension/pull/2504/files#diff-1dfe211fc8f084873d2cbcb1d2b9203b35f6212f12ce78a696962bde6b9e0744R50-R53)
> > //cc @rolznz what do you think?
> 
> @pavanjoshi914 please could you make this change too?



made those changes.

PR #2631, opened 2023-08-04
fix: remove access-control-allow-origin

reneaaron:

This header is a response header. Sending it from a client makes no sense. 

> The Access-Control-Allow-Origin header is included in the response from one website to a request originating from another website, and identifies the permitted origin of the request. A web browser compares the Access-Control-Allow-Origin with the requesting website's origin and permits access to the response if they match.


bumi:
this might be some legacy from some bluewallet calls (and that got c&p )

PR #2644, opened 2023-08-09
fix: remove "only connect with sites you trust"

reneaaron:

Removes warning from different prompts as you already connected with this site. The only place where this message would make sense is in the enable() screen. 

I found those messages do more harm than good and leave the user puzzled. What does it mean to "trust a website"?


bumi:
hmmm, on the one hand we sometimes we put huge scary messages on the screens and the other hand we remove the general messages which should tell the user to be aware of using it on the right website.

PR #2645, opened 2023-08-09
fix: add the account name to prompt

reneaaron:

Seems it's not clear what users need to enter here, so I added the current account name to those prompts.


reneaaron:
@bumi A good example where we would definitely need to deprecate current translations :thinking:

bumi:
so how do we do it? can we change the key?

reneaaron:
@bumi I don't think we should change the keys. Only option that comes to mind is to manually reset the translations in code to an empty string. That should work, right?

scripts/remove-outdated-translations.js
```diff
@@ -0,0 +1,67 @@
+const fs = require("fs");
+const glob = require("glob");
+const path = require("path");
+
+const keyToUpdate = process.argv[2]; // Get the key to update from command line argument
```
rolznz, line 5: Can you add something to the documentation to guide users on how to execute this and what parameters to pass?

reneaaron: Definitely, just wanted to get your feedback about whether you think this is a feasible approach going forward.

scripts/remove-outdated-translations.js
```diff
@@ -0,0 +1,67 @@
...
+function updateNestedKey(obj, nestedKey) {
+  const keys = nestedKey.split(".");
+  let currentObj = obj;
+  let foundAndUpdated = false;
+
+  for (let i = 0; i < keys.length; i++) {
+    const key = keys[i];
+
+    if (i === keys.length - 1 && currentObj.hasOwnProperty(key)) {
+      currentObj[key] = "";
```
reneaaron, line 21: Check if this can be removed completely?

scripts/remove-outdated-translations.js
reneaaron: I've now integrated a Github action that directly outputs the command you might consider to run to reset the translations on your branch: 

image

scripts/remove-outdated-translations.js
reneaaron: I now changed it to remove the key instead of resetting it to an empty string. Let's see if that makes a difference for weblate, I don't think so tbh.

PR #2647, opened 2023-08-10
fix: e2e tests

reneaaron:


rolznz (approved):

tests/e2e/helpers/loadExtension.ts
```diff
@@ -170,48 +170,77 @@ export async function loginToExistingAlbyAccount(page: Page) {
...
   //get the new page object:
   const oauthPage = await newTarget.page();
   if (!oauthPage) {
     throw new Error("OAuth page not found");
   }
   const oauthDocument = await getDocument(oauthPage);
 
-  const oauthLoginButton = await getByText(oauthDocument, "Log in to connect");
-  oauthLoginButton.click();
+  const connectButtonText = process.env.CI
```
reneaaron, line 181: Can we maybe just make the search less specific?

tests/e2e/helpers/loadExtension.ts
```diff
@@ -228,6 +257,8 @@ export async function loginToExistingAlbyAccount(page: Page) {
     retries++;
   }
   if (retries >= MAX_RETRIES) {
+    //console.info("oauthPage", await oauthPage.content());
```
reneaaron, line 260: Remove

tests/e2e/helpers/loadExtension.ts
rolznz: I tried but wasn't successful, seems it needs a specific selector if you search for a substring. Regex search also didn't work for me. This shouldn't change now, I would just keep it as is if that's ok

PR #2650, opened 2023-08-10
feat: use react-hot-toast 🍞

reneaaron:

Replace react-toastify with react-hot-toast: 

https://react-hot-toast.com/

**TODO**
 - dark mode


escapedcat:
What's the main reason for the switch? The raspiblitz UI is also using toastify but it might be good to switch as well? :)

reneaaron:
@escapedcat They just feel much nicer and fit the style of Alby imo.

PR #2657, opened 2023-08-12
feat: custom webpack plugin to inline provider

pavanjoshi914:


reneaaron:
Closing in favor of #2691

PR #2671, opened 2023-08-15
fix: use official alby sdk

bumi:


rolznz:
@bumi fix is also in https://github.com/getAlby/lightning-browser-extension/pull/2514/files in case we can merge that one

Issue #2613, opened 2023-08-01
[BUG] Sometimes window.nostr/liquid/webln takes too long to load and is unavailable when website checks for it


rolznz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The inpage script is sometimes loaded too slowly because it is loaded asynchronously, causing sites to think the objects aren't available.

### Steps To Reproduce

1. Visit https://louisinger.github.io/alby-liquid-wallet/#
2. Sometimes it will throw an exception because window.liquid does not exist

### Expected behavior

window.liquid/nostr/webln should be available instantly.

### Alby information

Latest

### Additional context

https://github.com/getAlby/lightning-browser-extension/issues/2462

### Are you working on this?

None

PR #2672, opened 2023-08-16
feat: fast webln via injection in main world

pavanjoshi914:

currently webln is slow and lags, sometimes it's unavailable for the user. did some quite research to inject provider as soon as possible and here is the PR for fast WebLN

chromium browsers  - we are MV3 here

https://github.com/getAlby/lightning-browser-extension/assets/55848322/1f6de3f1-3c5e-4f74-a6f7-ebffc1dc084a

firefox::

https://github.com/getAlby/lightning-browser-extension/assets/55848322/26181125-dfd3-4d4a-98ed-05fda54b6914

Fixex #2613 

### Type of change

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/build-utils/InjectWindowProvider.js
+++ b/build-utils/InjectWindowProvider.js
@@ -3,7 +3,8 @@ const { sources } = require("webpack");
 const PLUGIN_NAME = "InjectWindowProvider";
 const WINDOW_PROVIDER_FILENAME = "js/inpageScript.bundle.js";
 const PROVIDER_BRIDGE_FILENAME = "js/contentScriptOnEndWebLN.bundle.js";
-
+// plugin to intercept webln bundle during built times and inline/attach provider to make it available immediately
+// works only on MV2 and designed specifically for firefox as main world execution is not supported for firefox
 class InjectWindowProvider {
   apply(compiler) {
     compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
--- a/build-utils/InjectWindowProvider.js
+++ b/build-utils/InjectWindowProvider.js
@@ -1,0 +1,38 @@
+const { sources } = require("webpack");
+
+const PLUGIN_NAME = "InjectWindowProvider";
+const WINDOW_PROVIDER_FILENAME = "js/inpageScript.bundle.js";
+const PROVIDER_BRIDGE_FILENAME = "js/contentScriptOnEndWebLN.bundle.js";
+
+class InjectWindowProvider {
+  apply(compiler) {
+    compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tap(
+        {
+          name: PLUGIN_NAME,
+          stage: compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_INLINE,
+        },
+        (assets) => {
+          let windowProviderSource =
+            assets[WINDOW_PROVIDER_FILENAME].source().toString();
+
+          // need to encode so it can be used as a string
+          // in non optimized builds the source is a multi-line string > `` needs to be used
+          // but ${ needs to be escaped separately otherwise it breaks the ``
+          windowProviderSource = JSON.stringify(windowProviderSource);
+          let providerBridgeSource =
+            assets[PROVIDER_BRIDGE_FILENAME].source().toString();
+
+          assets[PROVIDER_BRIDGE_FILENAME] = new sources.RawSource(
+            providerBridgeSource.replace(
+              '"@@@WINDOW_PROVIDER@@@"',
+              windowProviderSource
+            )
+          );
+        }
+      );
+    });
+  }
+}
+
+module.exports = InjectWindowProvider;
--- a/src/extension/background-script/index.ts
+++ b/src/extension/background-script/index.ts
@@ -141,6 +141,24 @@ browser.runtime.onInstalled.addListener(handleInstalled);
 async function init() {
   console.info("Loading background script");
 
+  const registerInPageContentScript = async () => {
+    try {
+      await chrome.scripting.registerContentScripts([
+        {
+          id: "inpage",
+          matches: ["file://*/*", "http://*/*", "https://*/*"],
+          js: ["js/inpageScript.bundle.js"],
+          runAt: "document_start",
+          world: "MAIN",
+        },
+      ]);
+    } catch (err) {
+      console.warn(`Dropped attempt to register inpage content script. ${err}`);
+    }
+  };
+
+  registerInPageContentScript();
+
   await state.getState().init();
   console.info("State loaded");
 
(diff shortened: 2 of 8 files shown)


src/manifest.json
```diff
@@ -166,6 +164,13 @@
       "matches": ["*://*/*"],
       "run_at": "document_start",
       "js": ["js/contentScriptOnStart.bundle.js"]
+    },
+    {
+      "allFrames": true,
+      "matches": ["file://*/*", "http://*/*", "https://*/*"],
+      "runAt": "document_start",
+      "js": ["js/inpageScript.bundle.js"],
+      "world": "MAIN"
```
reneaaron, line 173: I couldn't find any docs about this "world" property. Is it needed?

src/extension/content-script/onstart.ts
```diff
@@ -12,9 +12,6 @@ async function onstart() {
   }
 
   const account = await api.getAccount();
-  // window.webln
-  injectScript(browser.runtime.getURL("js/inpageScriptWebLN.bundle.js"));
```
rolznz, line 16: @pavanjoshi914 can you inject all the providers below in this way? I want to test it with `window.liquid` for example.

src/manifest.json
pavanjoshi914: this is the key for the solution. this will execute the script in browser envrionment rather than in isolate environment. new changes introduce world property which hold 2 values
1. isolated
2. MAIN

by default content scripts run in isolated environment

rolznz:
@pavanjoshi914 I tested this using https://github.com/getAlby/lightning-wallet-connect and removing the delay in [this file](src/connectors/ExtensionConnector.ts).

But it still looks like the window.webln is not immediately available.

```
store.ts:61 Error: window.webln does not exist
    at ExtensionConnector.init (Connector.ts:13:13)
    at Object.connect (store.ts:51:23)
    at loadConfig (store.ts:103:22)
    at store.ts:107:1
```

src/manifest.json
```diff
@@ -166,6 +164,13 @@
       "matches": ["*://*/*"],
       "run_at": "document_start",
       "js": ["js/contentScriptOnStart.bundle.js"]
+    },
+    {
+      "allFrames": true,
```
rolznz, line 169: there is an inconsistency here with the entry above which has "all_frames": true,

rolznz:
@pavanjoshi914 how did you test this and on which browsers? are you sure this has made an improvement?

src/manifest.json
reneaaron: i just fixed that :+1:

src/manifest.json
```diff
@@ -38,6 +36,7 @@
     "activeTab",
     "nativeMessaging",
     "notifications",
+    "scripting",
```
reneaaron, line 39: Should this be in here? There is a separate definition for __chrome below and FF shouldn't need that with the current implementation, right?

src/extension/content-script/onstart.ts
```diff
@@ -12,9 +12,6 @@ async function onstart() {
   }
 
   const account = await api.getAccount();
-  // window.webln
```
reneaaron, line 15: Don't we need to inject this script for FF still?

src/extension/background-script/index.ts
```diff
@@ -141,6 +141,24 @@ browser.runtime.onInstalled.addListener(handleInstalled);
 async function init() {
   console.info("Loading background script");
 
+  const registerInPageContentScript = async () => {
```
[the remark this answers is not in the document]
pavanjoshi914: this is what leads to the speed. registering via service worker is the key to get provider injected immediately

src/extension/content-script/onstart.ts
pavanjoshi914: yup! need to decide fallback for firefox

src/extension/background-script/index.ts
```diff
@@ -141,6 +141,24 @@ browser.runtime.onInstalled.addListener(handleInstalled);
...
+        {
+          id: "inpage",
+          matches: ["file://*/*", "http://*/*", "https://*/*"],
+          js: ["js/inpageScript.bundle.js"],
+          runAt: "document_start",
+          world: "MAIN",
+        },
+      ]);
+    } catch (err) {
+      console.warn(`Dropped attempt to register inpage content script. ${err}`);
```
rolznz, line 156: Does the app still work if this fails?

pavanjoshi914: for chromium based browser? i don't think this fails. it's just that we are registering script programatically instead of using manifest. but core thing is same.

for other browsers. if function fails rest of the flow will still continue. on firefox
```
![image](https://github.com/getAlby/lightning-browser-extension/assets/55848322/c4a562d7-a94e-4a47-af62-84d4ae27ae18)

```

src/manifest.json
pavanjoshi914: resolved

src/extension/background-script/index.ts
pavanjoshi914: resolved

src/extension/content-script/onstart.ts
rolznz: @pavanjoshi914 this needs to be addressed still, right?

src/extension/content-script/onendwebln.js
```diff
@@ -33,7 +33,11 @@ async function init() {
     return;
   }
 
-  injectScript(browser.runtime.getURL("js/inpageScript.bundle.js")); // registers the DOM event listeners and checks webln again (which is also loaded onstart
+  const providerScript = "@@@WINDOW_PROVIDER@@@";
```
rolznz, line 36: this is only used if (!isMv3) right? can you move it into the if block?

pavanjoshi914: resolved

build-utils/InjectWindowProvider.js
```diff
@@ -0,0 +1,38 @@
+const { sources } = require("webpack");
+
+const PLUGIN_NAME = "InjectWindowProvider";
+const WINDOW_PROVIDER_FILENAME = "js/inpageScript.bundle.js";
+const PROVIDER_BRIDGE_FILENAME = "js/contentScriptOnEndWebLN.bundle.js";
+
+class InjectWindowProvider {
```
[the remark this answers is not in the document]
pavanjoshi914: resolved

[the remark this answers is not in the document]
pavanjoshi914: this is only for mv2 browsers. also if we migrated firefox to mv3 main world execution will not work there as it is not supported by firefox on MV3. we have to keep firefox on mv2 till they add support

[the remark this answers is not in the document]
rolznz: @pavanjoshi914 thanks for clarifying and adding the comment. This is very useful info!

reneaaron:
Closing in favor of #2691. Great work @pavanjoshi914!

PR #2677, opened 2023-08-16
fix: webln enable usage

bumi:


bumi:
merging this because it is broken in master. please still have a look.

PR #2679, opened 2023-08-17
fix: add promise queues for providers

### LINKED ISSUE #2613: [BUG] Sometimes window.nostr/liquid/webln takes too long to load and is unavailable when website checks for it

reneaaron:

Fixes #2613

### TODO
 - Implement queue for other providers


src/extension/providers/nostr/index.ts
```diff
@@ -1,21 +1,33 @@
...
 import { Event } from "./types";
-import EventEmitter from "events";
 
 declare global {
   interface Window {
     nostr: NostrProvider;
   }
 }
 
+class PromiseQueue {
```
rolznz, line 11: Can you move this class to a new file?

src/extension/providers/nostr/index.ts
```diff
@@ -1,21 +1,33 @@
...
+      this.queue = this.queue.then(operation).then(resolve).catch(reject);
+    });
+  }
+}
+
 export default class NostrProvider {
   nip04 = new Nip04(this);
   enabled: boolean;
   private _eventEmitter: EventEmitter;
+  queue: PromiseQueue;
```
rolznz, line 25: this should be private

src/extension/providers/nostr/index.ts
```diff
@@ -30,23 +42,31 @@ export default class NostrProvider {
...
+    });
   }
 
   async getRelays() {
-    await this.enable();
-    return this.execute("getRelays");
+    return this.queue.add(async () => {
+      await this.enable();
+      return this.execute("getRelays");
+    });
```
rolznz, line 69: `enable` in the on and off methods should also be wrapped in the queue, right?

rolznz: For this reason I am wondering if the execute method itself should handle the queue, hmm..

bumi: I was wondering the same. we always call execute, could execute implement what is currently in the `PromiseQueue #add`

reneaaron: Yes, but maybe we could even move it to execute<T>(), I'll check.

bumi (approved):

PR #2682, opened 2023-08-17
fix: add explicit enable call for window.liquid

reneaaron:

Require clients to explicitly call enable() before any other method calls. (this is also how Marina wallet seems to do it right now)

### DIFF
--- a/src/extension/providers/liquid/index.ts
+++ b/src/extension/providers/liquid/index.ts
@@ -13,22 +13,27 @@ export default class LiquidProvider {
     this.enabled = false;
   }
 
-  async enable() {
+  async enable(): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/extension/providers/liquid/index.ts
```diff
@@ -25,12 +25,10 @@ export default class LiquidProvider {
   }
 
   async getAddress() {
-    await this.enable();
```
rolznz, line 28: shouldn't these have a check if it is enabled or not, and if not throw an exception (see the webln provider)

reneaaron: yeah :see_no_evil:

rolznz (approved):
utACK

PR #2691, opened 2023-08-21
feat: fast provider injection 🔥

reneaaron:

Improvements for #2672


build-utils/ProviderInjectionPlugin.js
```diff
@@ -0,0 +1,61 @@
...
+const PLUGIN_NAME = "ProviderInjectionPlugin";
+const PROVIDER_BRIDGE_FILENAME = "js/contentScriptOnStart.bundle.js";
+
+// plugin to intercept webln bundle during built times and inline/attach provider to make it available immediately
+// works only on MV2 and designed specifically for firefox as main world execution is not supported for firefox
+class ProviderInjectionPlugin {
+  apply(compiler) {
+    // declare any new provider here
+    const inpageScripts = [
+      {
```
reneaaron, line 12: Use an array instead of object.

build-utils/ProviderInjectionPlugin.js
```diff
@@ -0,0 +1,61 @@
...
+        {
+          name: PLUGIN_NAME,
+          stage: compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_INLINE,
+        },
+        (assets) => {
+          const windowProviderSource = inpageScripts
+            .map((provider) => assets[provider.fileName].source().toString())
+            .join("\n");
+
+          const encodedWindowProviderSource =
```
reneaaron, line 43: Add a comment why this is needed.

src/extension/background-script/index.ts
```diff
@@ -141,6 +142,31 @@ browser.runtime.onInstalled.addListener(handleInstalled);
 async function init() {
   console.info("Loading background script");
 
+  const registerInPageContentScript = async () => {
+    try {
+      await chrome.scripting.registerContentScripts([
+        {
+          id: "inpage",
```
reneaaron, line 149: Remove optional id

build-utils/ProviderInjectionPlugin.js
pavanjoshi914: resolved

src/extension/background-script/index.ts
pavanjoshi914: can't remove its a compulsion

pavanjoshi914:
1. new changes include, doing browser side checks in the inpagescript itself so that those checks work for mv3 and mv2 both

2. handle browser side check + extension side check such as blocklist in the providers content script itself. so that it works both on mv2 and mv3

like this we achieve common behaviour on both mv2 and mv3 

1. script is injected without any check
2. inpage script itself checks documenttype, suffix and script element if everything is fine script is injected. if not script will not injected
3. if inpage is injected, if user performs some call. shouldInject from contentscript is responsible to handle blocklist via shouldInject present in there

src/extension/inpage-script/index.js
```diff
@@ -4,87 +4,109 @@ import LiquidProvider from "~/extension/providers/liquid";
 import NostrProvider from "~/extension/providers/nostr";
 import WebBTCProvider from "~/extension/providers/webbtc";
 import WebLNProvider from "~/extension/providers/webln";
+import shouldInjectInpage from "./shouldInject";
 
-if (document) {
-  // Intercept any `lightning:` requests
+function init() {
+  if (document) {
```
reneaaron, line 10: This document check is in shouldInjectInpage already, right? Remove duplicate check.

src/extension/inpage-script/index.js
```diff
@@ -4,87 +4,109 @@ import LiquidProvider from "~/extension/providers/liquid";
...
 
-  window.liquid = new LiquidProvider();
-  window.alby = new AlbyProvider();
-  window.nostr = new NostrProvider();
-  window.webbtc = new WebBTCProvider();
-  window.webln = new WebLNProvider();
+    const readyEvent = new Event("webln:ready");
+    window.dispatchEvent(readyEvent);
+    // Intercept any `lightning:` requests
+    window.addEventListener(
```
reneaaron, line 22: Can we move this to a function like "registerLightningLinkClickHandler"?

src/extension/inpage-script/index.js
```diff
@@ -1,86 +1,112 @@
...
+            ) {
+              alert(`Error: ${e.message}`);
+            }
+          }
+        }
 
-      if (lnurl) {
         try {
-          const response = await window.webln.lnurl(lnurl);
+          const response = await window.webln.sendPayment(paymentRequest);
```
bumi, line 105: these things here just changed because it was moved around, corret? 
we did not change anything on how the links are handled?

src/extension/inpage-script/index.js
pavanjoshi914:
```
export function doctypeCheck() {
  const doctype = window.document.doctype;
  if (doctype) {
    return doctype.name === "html";
  } else {
    return true;
  }
}
```

does window.document will always exist? i don't know much in this context

src/extension/inpage-script/index.js
pavanjoshi914: or may be just, lightningClickHandler?

src/extension/inpage-script/index.js
pavanjoshi914: resolved

PR #2700, opened 2023-08-27
chore: remove lnurl-pay description hash check

bumi:
as discussed in: https://github.com/lnurl/luds/pull/234


reneaaron (approved):
tACK

Tested with calle@legend.8333.space (didn't work before, but now does :rocket: )

PR #2701, opened 2023-08-28
feat: provider onboard

### LINKED ISSUE #2713: [Feature] Alby should prompt user to setup a master key if using LNURL auth and none is set
### Feature description

Similar to https://github.com/getAlby/lightning-browser-extension/issues/2605 but for LNURL auth

### Describe the solution

Use the same method as the nostr signing if possible, however this also needs to support lightning: lnurl login links as well as webln.lnurl.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:

onboard users for different providers such as liquid, nostr and webbtc

Fixes #2713 

TODO:
1. response handling
2. UI/UX
3. extend screen for other providers
4. make text of the prompt customizable depending from which provider it was called

all the todo's are done

### Type of change

### How has this been tested?

### Checklist

- [ ] Manual testing
- [ ] Update Docs & Guides
- For UI-related changes
- [ ] Darkmode
- [ ] Responsive layout


src/extension/content-script/onendnostr.js
```diff
@@ -42,7 +43,7 @@ async function init() {
   // message listener to listen to inpage nostr calls
   // those calls get passed on to the background script
   // (the inpage script can not do that directly, but only the inpage script can make nostr available to the page)
-  window.addEventListener("message", (ev) => {
+  window.addEventListener("message", async (ev) => {
```
bumi, line 46: why async now?

bumi:
@pavanjoshi914 can you merge master?

src/app/screens/ProviderOnboard/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  action: string;
+};
+
+export default function ProviderOnboard(props: Props) {
+  function openOptions(path: string) {
+    // if we are in the popup
+    if (window.location.pathname !== "/options.html") {
+      utils.openPage(`options.html#/${path}`);
+      // close the popup
+      window.close();
```
bumi, line 21: we somehow need to reply otherwise the calling promise will be stuck forever.

src/extension/background-script/actions/onboard/providerOnboard.ts
```diff
@@ -0,0 +1,32 @@
...
+  const host = getHostFromSender(sender);
+  if (!host) return;
+
+  try {
+    const response = await utils.openPrompt<{
+      enabled: boolean;
+      remember: boolean;
+    }>(message);
+
+    return {
```
bumi, line 18: I think we should return an error here, as we will not get this data

src/app/router/Prompt/Prompt.tsx
```diff
@@ -125,6 +126,14 @@ function Prompt() {
             <Route path="confirmKeysend" element={<ConfirmKeysend />} />
             <Route path="confirmSignMessage" element={<ConfirmSignMessage />} />
             <Route path="confirmAddAccount" element={<ConfirmAddAccount />} />
+            <Route
+              path="public/nostr/providerOnboard"
+              element={<ProviderOnboard />}
+            />
+            <Route
+              path="public/liquid/providerOnboard"
```
rolznz, line 134: we don't have it public yet, but could you add webbtc here too?

src/extension/content-script/liquid.js
```diff
@@ -66,9 +70,14 @@ async function init() {
...
       };
 
+      if (!account.hasMnemonic) {
+        messageWithOrigin.action = `public/liquid/providerOnboard`;
+      }
+
       const replyFunction = (response) => {
         if (ev.data.action === `${SCOPE}/enable`) {
           isEnabled = response.data?.enabled;
+          isOnboardingStarted = response.data?.onBoardStarted;
```
rolznz, line 80: `onBoardStarted` should be `onboardStarted`

rolznz: Can you use consistent naming for both?

pavanjoshi914:
@rolznz we had a huddle to discuss potential flow of the process. so i quickly pushed commit to disucss things. code is not cleaned yet and contains many unnecessary things. making proper changes

src/extension/content-script/liquid.js
pavanjoshi914: i was resolving promise with success reply and use this to handle next execution of function and avoid double popup of onboarding screen. 

for eg. nostr we call enable and call actual function. content script runs two times 

now we resolve promise with the error this is no more the case

rolznz:
@pavanjoshi914 can you link this PR to the original issue?

src/extension/background-script/actions/onboard/providerOnboard.ts
```diff
@@ -0,0 +1,22 @@
+import utils from "~/common/lib/utils";
+import { getHostFromSender } from "~/common/utils/helpers";
+import type { MessageAllowanceEnable, Sender } from "~/types";
+
+const providerOnboard = async (
+  message: MessageAllowanceEnable,
```
rolznz, line 6: is the message type correct here? `MessageAllowanceEnable`?

src/extension/background-script/router.ts
```diff
@@ -83,6 +84,7 @@ const routes = {
   // Public calls that are accessible from the inpage script (through the content script)
   public: {
     webbtc: {
+      providerOnboard: onboard.providerOnboard,
```
rolznz, line 87: the naming feels wrong here. It should be a verb. "setupProvider"? does it do more than "setupMasterKey"?

src/extension/content-script/liquid.js
```diff
@@ -66,6 +69,10 @@ async function init() {
         origin: getOriginData(),
       };
 
+      if (!account.hasMnemonic) {
+        messageWithOrigin.action = `public/liquid/providerOnboard`;
```
rolznz, line 73: can you add a comment here? so it overrides the action so the user can go through onboarding to setup their master key. Then they need to refresh this page to do what they were going to do before, right?

src/extension/content-script/nostr.js
```diff
@@ -82,12 +83,19 @@ async function init() {
         origin: getOriginData(),
       };
 
+      const account = await api.getAccount();
```
bumi, line 86: why do we load the account here in the message handler, but load it for liquid on the init() function?

src/extension/background-script/actions/onboard/providerOnboard.ts
pavanjoshi914: we are intercepting only enable calls here. this particular interface addresses right message type here i think

```
export interface MessageAllowanceEnable extends MessageDefault {
  origin: OriginData;
  args: {
    host: Allowance["host"];
  };
  action:
    | "public/webln/enable"
    | "public/nostr/enable"
    | "public/liquid/enable"
    | "public/alby/enable";
}
```

src/extension/background-script/router.ts
pavanjoshi914: let's use openOnboardingPrompt as it can be used for other purposes or other upcoming providers as well, not just for setupMasterKey

src/extension/content-script/nostr.js
pavanjoshi914: resolved! extended same code on other providers. keeping account check in listener function helps in executing commands without need of reload

rolznz: Good catch -It should be done in the message handler for each right, so that the user does not need to refresh after the key is set (and also so it checks every time, after a user switches account for example)

src/extension/background-script/router.ts
rolznz: Sounds good, thanks!

src/app/screens/Onboard/Onboard/index.tsx
```diff
@@ -0,0 +1,89 @@
+import {
```
rolznz, line 1: Why is this file inside Onboard/Onboard (twice?)

pavanjoshi914: image
cause that is somehow the file structure is

rolznz: This feels confusing, should it at least be called Onboard/OnboardingPrompt? CC @reneaaron

reneaaron: I kind of agree. I'll change the structure a bit during my final review :ok_hand:

reneaaron (approved):
tACK

PR #2710, opened 2023-08-30
fix: support for non-square avatars

jankoegel:
# Scenario
+ I have a legacy, non-square avatar that was uploaded before https://github.com/getAlby/getalby.com/commit/079710187ed4b09c405f23e6b35ec6a82a97759b:
  https://uploads.getalby-assets.com/uploads/lightning_address/avatar/2893/thumb_jan-koegel_yours-truly.jpg

## Notes
+ semi-tested, I only modified the CSS via my web inspector but did not rebuild the extension on my machine.
+ `.object-cover` is documented here: https://tailwindcss.com/docs/object-fit

## Before
image

## After
image

### DIFF
--- a/src/app/components/Avatar/index.tsx
+++ b/src/app/components/Avatar/index.tsx
@@ -15,6 +15,10 @@ const Avatar = (props: Props) => {
   }
 };
 
+// NOTE: avatar images are all square since this commit from 26 July 2023:
+//   https://github.com/getAlby/getalby.com/commit/079710187ed4b09c405f23e6b35ec6a82a97759b
+//   Legacy avatars created before this commit can have non-square dimensions and have not
+//   been converted, that's why we need `object-cover` and not `object-fill`.
 const AvatarImage = (props: Props) => {
   return (
     <div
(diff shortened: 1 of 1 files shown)


reneaaron (approved):
tACK

src/app/components/Avatar/index.tsx
```diff
@@ -15,10 +15,8 @@ const Avatar = (props: Props) => {
   }
 };
 
-// NOTE: avatar images are all square since this commit from 26 July 2023:
-//   https://github.com/getAlby/getalby.com/commit/079710187ed4b09c405f23e6b35ec6a82a97759b
-//   Legacy avatars created before this commit can have non-square dimensions and have not
-//   been converted, that's why we need `object-cover` and not `object-fill`.
+// Use object-cover to support non-square avatars that might be loaded by
+// different connectors
```
jankoegel, line 19: I don't understand why you changed this.

reneaaron: Because getalby-com is a private repository and imo shouldn't be referenced in here.

bumi: yes, only public references in public code.

there are also no "legacy avatars" in the extension.

PR #2714, opened 2023-09-01
fix: remove obsolete activetab permission

reneaaron:

This permission is not needed anymore, since we use host_permissions that allow us to inject / read from any website (not only the active tab). 

`activeTab` would allow temporary access to do those kind of things AFTER the user opened the extension popup on this particular site.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -14,7 +14,6 @@
   "description": "The Bitcoin Lightning wallet for direct payments across the globe, Bitcoin Lightning applications and passwordless logins.",
   "homepage_url": "https://getAlby.com/",
   "permissions": [
-    "activeTab",
     "nativeMessaging",
     "notifications",
     "storage",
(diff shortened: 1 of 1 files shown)


pavanjoshi914:
we are using host permissions only for chrome right? activeTab is still needed for firefox?

reneaaron:
I don't think so, see the activeTab permission on FF docs and our manifest: 

https://github.com/getAlby/lightning-browser-extension/blob/nostr-key-prompt/src/manifest.json#L23

pavanjoshi914:
> I don't think so, see the activeTab permission on FF docs and our manifest:
> 
> [`nostr-key-prompt`/src/manifest.json#L23](https://github.com/getAlby/lightning-browser-extension/blob/nostr-key-prompt/src/manifest.json?rgh-link-date=2023-09-04T07%3A39%3A10Z#L23)

yup! looks good. ff mv2 doesn't require explicit ```host_permissions:```

pavanjoshi914 (approved):

PR #2717, opened 2023-09-02
fix: inject content script into all frames

reneaaron:

Inject into all frames as well so `window` APIs are also available on sites like codepen.io, etc.


bumi:
goooood catch! very important

PR #2721, opened 2023-09-05
feat: unify-queue

### LINKED ISSUE #2501: [BUG] Unlock dialog launched multiple times

### Describe the bug

A second dialog can launch while the first is open, when the extension is locked.

### Screenshots [optional]

https://github.com/getAlby/lightning-browser-extension/assets/33993199/4cc15d6e-9088-48ab-86de-e1e57edf783c

### Steps To Reproduce

1. Launch a new browser with extension locked
2. Open snort.social (Note: Maybe you need snort premium for it to launch 2 modals like this)

### Expected behavior

Only one modal is opened

### Alby information

2.1.1

### Device information

Chrome Windows

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:

1. unify the queue for all providers to solve multiple dialogue problems. we use a global queue for all the providers.
2. separate queue logic from providers and shift it to postMessage function

the logic behind making it global is that, allowance is set in the first enable call. and when we make the next enable call (even from another window provider API)  enable call will not open a prompt as allowance is set for the website in the first enable call. and we avoid multiple prompts like this

Fixes  #2501 

### Type of change

### How has this been tested?
use webln.enable(), webbtc.enable() calls in the console. you will see only single dialogue

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/extension/inpage-script/index.js
+++ b/src/extension/inpage-script/index.js
@@ -4,16 +4,18 @@ import LiquidProvider from "~/extension/providers/liquid";
 import NostrProvider from "~/extension/providers/nostr";
 import WebBTCProvider from "~/extension/providers/webbtc";
 import WebLNProvider from "~/extension/providers/webln";
+import { PromiseQueue } from "../providers/promiseQueue";
 import shouldInjectInpage from "./shouldInject";
 
 function init() {
   const inject = shouldInjectInpage();
   if (!inject) return;
-  window.liquid = new LiquidProvider();
-  window.alby = new AlbyProvider();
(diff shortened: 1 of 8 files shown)


reneaaron (approved):
tACK

I like this change, because it simplifies the API surface of connectors and hides complexity in the background. 

Also this change merges all provider queues into one, so currently there can't be any more than 1 open popup by the extension, no matter which APIs websites use.

PR #2722, opened 2023-09-05
feat: new receive ui

### LINKED ISSUE #2366: [Feature] Improve QR code colors in dark mode
### Feature description

QR colors should be inverted in dark mode. They need to be passed to the QRCode component. Since it isn't css, we can't apply the styles using Tailwind. We need a way to get the theme which I think will be available when this commit is merged: https://github.com/getAlby/lightning-browser-extension/pull/2347

image

### Describe the solution

Invert the QR code colors in dark mode so it is easier on the eyes

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

### LINKED ISSUE #2659: Revamp Receive screen
Proposition to revamp "Receive" screen.

[Figma link with a user flow and handoff](https://www.figma.com/file/MCzf4t1e3pi7Dre6huwq1y/Receive-%2F-Send-revamp?type=design&node-id=251-23749&mode=design&t=H0l2ZM0Vq4IPEY20-0)

## Changes:
- displaying user's lightning address
- displaying QR code with user's lightning address + user avatar inside
- "list item" buttons with icons and better explained ways of receiving bitcoin

# Alby Accounts:
image

# Non-Alby Accounts:
image
Alternatively, this way could be implemented: 
image

# New icons used:
Bitcoin circle

Note:

In "dev mode" you can see the color tokens that are not yet used in the extension
image

Here's the equivalent of those from tailscale:

image
image
image

reneaaron:

Fixes #2659 

Fixes #2366


package.json
```diff
@@ -135,6 +135,7 @@
     "puppeteer": "^21.0.3",
     "stream-browserify": "^3.0.0",
     "swc-loader": "^0.2.3",
+    "tailwindcss-3d": "^1.0.0",
```
reneaaron, line 138: Needed for translate-z class generation.

rolznz:
image

Should we update this to "Lightning Address Copied"?

I think the toast style will be improved in a [different PR](https://github.com/getAlby/lightning-browser-extension/pull/2650) - would be nice to have dark mode styling there.

src/i18n/locales/en/translation.json
```diff
@@ -717,7 +717,18 @@
...
       "actions": {
         "create_invoice": "Create Invoice",
-        "copy_lightning_address": "Copy Lightning Address"
+        "invoice": {
+          "title": "Lightning invoice",
+          "description": "Request instant and specific amount payments"
+        },
+        "bitcoin_address": {
+          "title": "Bitcoin address",
+          "description": "Request instant and specific amount payments"
```
rolznz, line 726: this is wrong right? onchain is not instant

rolznz: What about "Receive on-chain payments to your Alby account"?

src/app/screens/Receive/index.tsx
```diff
@@ -1,374 +1,187 @@
...
-    event.preventDefault();
-    createInvoice();
-  }
-
   async function getLightningAddress() {
-    setLoadingLightningAddress(true);
-    const response = await api.getAccountInfo();
-    const lightningAddress = response.info.lightning_address;
+    // Wait for a minimum delay to prevent flickering
+    const [accountInfo] = await Promise.all([api.getAccountInfo(), delay(250)]);
```
rolznz, line 42: this wouldn't be needed if we use SWR I think. Currently it reloads the lightning address every time you visit this page

src/app/screens/Receive/index.tsx
```diff
@@ -1,374 +1,187 @@
...
-      )}
+        </Container>
+      </div>
     </div>
   );
 }
 
 export default Receive;
+
+type IconLinkCardProps = {
```
rolznz, line 157: This looks like a useful component, should it be moved to a new file?

src/app/components/QRCode/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { classNames, useTheme } from "~/app/utils";
+
+export type Props = {
+  value: string;
+  size?: number;
+  className?: string;
+};
+
+export default function QRCode({ value, size, className }: Props) {
+  const theme = useTheme();
```
rolznz, line 11: Nice 🚀

Will fix https://github.com/getAlby/lightning-browser-extension/issues/2366

reneaaron: Yes :+1:

src/i18n/locales/en/translation.json
reneaaron: Good catch, that should be a different message here! :sweat_smile:

reneaaron:
> Should we update this to "Lightning Address Copied"?
I don't think it's worth maintaining separate translations for that. I think it's clear from the context what is put into your clipboard. 

> I think the toast style will be improved in a https://github.com/getAlby/lightning-browser-extension/pull/2650 - would be nice to have dark mode styling there.
Yes, I'd love to get that merged ASAP, still facing some problems I need to sort out. 

> Also feels like the edges of the qr are a bit cut-off, but it still scans ok!
Yeah, I did that on purpose. :+1:  Let's see if all QR-code scanners work with the inverted colors as well.

src/common/utils/helpers.ts
```diff
@@ -61,3 +61,7 @@ export async function poll<T>({
 
   return new Promise(executePoll);
 }
+
+export function delay(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
reneaaron, line 67:
```suggestion
```

rolznz (approved):
tACK

PR #2729, opened 2023-09-06
feat: new toasts 🔥🍞

reneaaron:

[Screencast from 2023-09-06 15:10:23.webm](https://github.com/getAlby/lightning-browser-extension/assets/100827540/9ae00f64-1c4b-41af-8411-513ef0171d11)


bumi:
Will the bottom always be well visible? on a bigger screen when I enter something at the top?

rolznz:
How does the user close the toast right now? do we want/need that?

rolznz (approved):

reneaaron:
> Will the bottom always be well visible? on a bigger screen when I enter something at the top?

It's a fixed container, so yes.

> How does the user close the toast right now? do we want/need that?

They can't currently, the automatically go away after 4s (success) / 8s (error) unless there is a different duration defined. We could easily add that though...

reneaaron:
I've added a close icon for toasts that show longer than 10s. :+1:

PR #2733, opened 2023-09-07
feat: handle case when user goes to onboard but doesn't setup master key

pavanjoshi914:

user makes a call. starts the onboarding process but doesn't setup the click. now the user makes the call again. in this case route will not be overridden because of account check. also i feel we don't need that

### Type of change

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


bumi:
what was the reason to cache this?

pavanjoshi914:
> what was the reason to cache this?

i don't know it was added later

rolznz (approved):
utACK

reneaaron (approved):
tACK

PR #2735, opened 2023-09-07
fix: Use relative path for the extension path for testing

volfiros:

The extension path used for testing is relative to the root project directory and not relative path. Windows is unable to detect it, to tackle this problem I have changed to relative path for the development which works on every system.

### Type of change

Windows unable to detect the extension path from the dist folder so unable to add the extension path to the place where the chrome executable is present.

image

### How has this been tested?

Ran all the tests and now everything is automated and works fine.

### Checklist

- [x] Manual testing


Rithvik-padma:
@im-adithya Can u review this PR.
Thank you

im-adithya:
Yeah, this works but just want others to test in different OSes just to be sure

- [x] Mac (done by me)
- [x] Windows (done by @Rithvik-padma)
- [ ] Linux (@roland can you please help here?)

rolznz:
@im-adithya I do not think this is needed. Tests are run through a yarn command and the dist folder is directly in the root of the project. Why do we need to use `"../../../"` which in my opinion indicates there is something wrong with the project on @Rithvik-padma 's machine

Rithvik-padma:
> @im-adithya I do not think this is needed. Tests are run through a yarn command and the dist folder is directly in the root of the project. Why do we need to use `"../../../"` which in my opinion indicates there is something wrong with the project on @Rithvik-padma 's machine

This may be the case but I see that even other windows users are also facing similar kind of issue in loading the extension by puppeteer using the absolute path from the project root.
Can you please look into this [issue](https://github.com/puppeteer/puppeteer/issues/2733) on puppeteer repo. 
It is old but I think it is still relevant.

pavanjoshi914:
additionally we also suggested to use wsl for windows users as linux systems are best to test debug and feature development for extension, closing

PR #2736, opened 2023-09-07
fix: additionally check message origin

bumi:
this checks that the message is from the same origin and can only come from the content script with the same origin


src/extension/providers/postMessage.ts
```diff
@@ -30,6 +30,7 @@ export function postMessage<T>(
           // check if it is a relevant message
           // there are some other events happening
           if (
+            messageEvent.origin != window.location.origin ||
```
reneaaron, line 33:
```suggestion
            messageEvent.origin !== window.location.origin ||
```

reneaaron (approved):
tACK 

Tested with both sites & iframes on FF.

PR #2737, opened 2023-09-07
feat: move onboarding prompt to enable screen

reneaaron:

 - Have a separate enable screen for different providers (nostr, webbtc, liquid) 
   - This will enable us to do further provider specific improvements (like permission presets, etc)
 - Get rid of all the content-script changes (no more api calls in there) 
 - Enable screens would check for respective keys (nostr -> nostr, webbtc -> mnemonic, etc)


rolznz:
@reneaaron I think this is a good idea. Would @pavanjoshi914 work on it?

reneaaron:
@pavanjoshi914 After discussing this with @bumi we think it would be best to create separate enable screens (I've updated the issue description with

src/app/screens/Enable/NostrEnable.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  useEffect(() => {
+    async function fetchAccountAndSetComponent() {
+      try {
+        const fetchedAccount = await api.getAccount();
+
+        if (fetchedAccount.nostrEnabled) {
+          setAccountComponent(<Enable origin={props.origin} />);
+        } else {
+          setAccountComponent(<Onboard />);
```
bumi, line 26: is it good to set this as a state or should we do this in the render block depending on the state variables that is set?

src/extension/background-script/router.ts
```diff
@@ -85,7 +85,7 @@ const routes = {
   public: {
     webbtc: {
       onboard: onboard.prompt,
-      enable: allowances.enable,
+      enable: allowances.mnemonicEnable,
```
bumi, line 88: move those under the webbtc/nostr/etc. namespaces then we also indicated there that those are called from the window providers

bumi (commented):
let's use different components for each enable screen that we can also adjust the content.
and on nostr talk about nostr
for liquid talk about liquid, etc.

src/extension/background-script/router.ts
pavanjoshi914: moved enable prompts to their respective namespaces

pavanjoshi914:
> let's use different components for each enable screen that we can also adjust the content. and on nostr talk about nostr for liquid talk about liquid, etc.

added separate component for each provider screen

src/app/screens/Enable/NostrEnable.tsx
pavanjoshi914: now we return component directly based on the state variable

src/app/components/Enable/LiquidEnable.tsx
```diff
@@ -0,0 +1,122 @@
...
+      setLoading(false);
+    }
+  }, [tCommon]);
+
+  function reject(event: React.MouseEvent<HTMLAnchorElement>) {
+    event.preventDefault();
+    msg.error(USER_REJECTED_ERROR);
+  }
+
+  async function block(event: React.MouseEvent<HTMLAnchorElement>) {
```
rolznz, line 44: I have seen time and time again when we have 5+ duplicated files like this, changes are incorrectly propagated across all of them and this leads to bugs. @bumi I know you said just copy these screens, but all the code for handling allowances/blocklists is also duplicated.

src/app/router/Prompt/Prompt.tsx
```diff
@@ -74,23 +78,33 @@ function Prompt() {
...
-              element={<Enable origin={navigationState.origin as OriginData} />} // prompt will always have an `origin` set, just the type is optional to support usage via PopUp
+              element={
+                <NostrEnable origin={navigationState.origin as OriginData} />
+              } // prompt will always have an `origin` set, just the type is optional to support usage via PopUp
             />
             <Route
               path="public/webbtc/enable"
-              element={<Enable origin={navigationState.origin as OriginData} />} // prompt will always have an `origin` set, just the type is optional to support usage via PopUp
+              element={
+                <WebbtcEnable origin={navigationState.origin as OriginData} />
```
rolznz, line 106: Further down you have these routes, they should be removed right?

```
<Route path="public/nostr/onboard" element={<Onboard />} />
            <Route path="public/liquid/onboard" element={<Onboard />} />
            <Route path="public/webbtc/onboard" element={<Onboard />} />
            <Route path="public/webln/onboard" element={<Onboard />} />
```

src/app/screens/Enable/LiquidEnable.tsx
```diff
@@ -0,0 +1,43 @@
...
+      }
+    }
+
+    fetchAccountAndSetComponent();
+  }, [props.origin, account]);
+
+  return (
+    <div>
+      {hasMnemonic ? (
+        <LiquidEnableComponent origin={props.origin} />
```
rolznz, line 37: Isn't this back to front? now we have duplicated enable components, but the Onboard component (which should be different) is the same across all enable screens

src/app/router/Prompt/Prompt.tsx
pavanjoshi914: thanks i missed to remove them

src/app/components/Enable/LiquidEnable.tsx
bumi: it's just a function call, there is no logic. 
I think you add more logic and complexity to the code when you try to unify it.

src/app/components/Enable/LiquidEnable.tsx
```diff
@@ -0,0 +1,122 @@
...
+      try {
+        const allowance = await msg.request("getAllowance", {
+          domain: props.origin.domain,
+          host: props.origin.host,
+        });
+        if (allowance && allowance.enabled) {
+          enable();
+        }
+      } catch (e) {
+        if (e instanceof Error) console.error(e.message);
```
bumi, line 65: let's remove this. 
I think we already check this in the action and the prompt should not even open if there is an allowance. 

if it still opens, then we can show it I think. 

so we don't need this code.

src/app/screens/Enable/LiquidEnable.tsx
rolznz: Decided to leave this for now

src/app/screens/Enable/LiquidEnable.tsx
```diff
@@ -0,0 +1,43 @@
...
+type Props = {
+  origin: OriginData;
+};
+
+export default function LiquidEnable(props: Props) {
+  const { account } = useAccount();
+  const [hasMnemonic, setHasMnemonic] = useState(false);
+
+  useEffect(() => {
+    async function fetchAccountAndSetComponent() {
```
rolznz, line 17: can this method be renamed? that's not what it does

What about `fetchAccountInfo` or something? (issue is for all providers)

src/app/screens/Enable/NostrEnable.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  useEffect(() => {
+    async function fetchAccountAndSetComponent() {
+      try {
+        const fetchedAccount = await api.getAccount();
+
+        if (fetchedAccount.nostrEnabled) {
+          setHasNostrkeys(true);
+        } else {
+          setHasNostrkeys(false);
```
rolznz, line 24: minor: can you add a capital K for "keys"

src/extension/background-script/actions/allowances/__tests__/enable.test.ts
```diff
@@ -4,7 +4,7 @@ import state from "~/extension/background-script/state";
 import { allowanceFixture } from "~/fixtures/allowances";
 import type { DbAllowance, MessageAllowanceEnable, Sender } from "~/types";
 
-import enableAllowance from "../enable";
+import enableAllowance from "../../webln/enable";
```
rolznz, line 7: this seems wrong related to the filename. Now do we need tests for each provider?

src/extension/background-script/actions/liquid/enable.ts
```diff
@@ -0,0 +1,78 @@
+import utils from "~/common/lib/utils";
```
rolznz, line 1: there is a lot of duplication in all the enable methods now

pavanjoshi914: ok! but that was already decided to do. because of the additional checks we need in the prompt. they look same but still they have different actions now because of onboarding

src/extension/background-script/actions/allowances/__tests__/enable.test.ts
pavanjoshi914: what shall be do here?

pavanjoshi914: this shall be not in allowance anymore. separate file in each provider?

src/i18n/locales/en/translation.json
```diff
@@ -957,6 +971,7 @@
     }
   },
   "common": {
+    "allow": "Allow this website to:",
```
rolznz, line 974: should the enable-related translations be put in their own section?: 
`"common": { "enable": {"allow": "..."} }`

pavanjoshi914: resolved

src/extension/background-script/actions/nostr/enable.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    .where("host")
+    .equalsIgnoreCase(host)
+    .first();
+
+  if (
+    isUnlocked &&
+    allowance &&
+    allowance.enabled &&
+    account?.nostrPrivateKey
```
rolznz, line 25: why is the `nostrPrivateKey` check needed?

rolznz: That's to force onboarding in the enable prompt, right?

pavanjoshi914: this is needed because if allowance is set we directly return from the prompt itself. we don't go to screen

pavanjoshi914: we need to open enable screen and render screen conditionally

rolznz (approved):
tACK

PR #2744, opened 2023-09-08
feat: transaction list revamp

reneaaron:

transaction list revamp

### Design
https://www.figma.com/file/IvlUHnC3vkIWnQ14FppZqT/%F0%9F%8E%A8-Extension-UI-revamp?node-id=101%3A62629&mode=dev


rolznz:
@pavanjoshi914 I get some NaN sats transactions
image

rolznz:
@pavanjoshi914 can you please update the PR description and link to a ticket?

rolznz:
Arrow icons do not match the figma (sizing and thickness), can they be updated?

src/app/hooks/useTransactions.ts
```diff
@@ -17,16 +18,38 @@ export const useTransactions = () => {
           accountId,
           limit,
         });
+        const fetchInvoice = await api.getInvoices({ isSettled: true, limit });
```
rolznz, line 21: the naming is wrong for this variable. Maybe `getInvoicesResponse`?

src/app/hooks/useTransactions.ts
```diff
@@ -17,16 +18,38 @@ export const useTransactions = () => {
...
+
+        const invoices: Transaction[] = fetchInvoice.invoices.map(
+          (invoice) => ({
+            ...invoice,
+            title: invoice.memo,
```
rolznz, line 26: are both title and description supposed to be the same?

src/app/hooks/useTransactions.ts
rolznz, line 21: it would be nice to use SWR, there's a delay each time you go back to the home page right now. Maybe that could be done in a follow up PR?

rolznz, line 21: Talking with @bumi , he suggested we add a new connector method to get all transactions. Then the alby connector can be simplified because the logic can be implemented in the backend.

Should we do this as a follow up PR?

src/app/components/TransactionsTable/index.tsx
```diff
@@ -30,145 +54,215 @@ export default function TransactionsTable({
...
+                    </div>
+                  )}
+                </div>
+                <div className="overflow-hidden mr-3">
+                  <div
+                    className="
+      text-sm font-medium text-black truncate dark:text-white"
+                  >
+                    <p className="truncate">
+                      {tx.publisherLink && tx.title ? (
```
rolznz, line 85: it's strange and not intuitive how the title link works. What about we always expand to show more details on click, and then have an explicit link inside there instead?

src/app/components/TransactionsTable/index.tsx
```diff
@@ -30,145 +54,215 @@ export default function TransactionsTable({
...
+            <div
+              key={tx.id}
+              className="px-3 py-2 hover:bg-gray-100 dark:hover:bg-surface-02dp cursor-pointer rounded-md"
+              onClick={() => openDetails(tx)}
+            >
+              <div className="flex gap-3">
+                <div className="flex items-center">
+                  {type == "outgoing" ? (
+                    <div className="flex justify-center items-center bg-orange-100 rounded-full w-8 h-8">
+                      <ArrowUpIcon className="w-6 h-6 text-orange-400" />
```
rolznz, line 71: icon does not match figma

src/app/components/TransactionsTable/index.tsx
```diff
@@ -30,145 +54,215 @@ export default function TransactionsTable({
...
-                            textColor={badge.textColor}
-                          />
-                        ))}
+                    {transaction.totalFees !== undefined && (
+                      <div className="grid grid-cols-3 gap-2 px-0 p-1">
+                        <dt className="text-md text-right font-medium leading-6 text-gray-400 dark:text-neutral-600">
+                          Fee
+                        </dt>
+                        <dd className=" text-md leading-6  text-gray-900 dark:text-white col-span-2 mt-0 break-all">
+                          {transaction.totalFees} Sats
```
rolznz, line 195: can an existing translation be used?

rolznz: `getFormattedSats` can be used, right?

src/app/components/TransactionsTable/index.tsx
```diff
@@ -30,145 +54,215 @@ export default function TransactionsTable({
...
                       </p>
+                    )}
+                  </div>
+                </div>
+              </div>
+            </div>
+          );
+        })}
+      </div>
+      <Modal
```
rolznz, line 130: could this be moved to a separate file?

pavanjoshi914:
> @pavanjoshi914 I get some NaN sats transactions image

unable to reproduce this on my side. kindly check in production

pavanjoshi914:
> Arrow icons do not match the figma (sizing and thickness), can they be updated?

bitcoin designs has only those icon, we need custom fat arrow icons if we want the icon similar to design. i didn't put much priority on icons currently

src/app/hooks/useTransactions.ts
pavanjoshi914: resolved

src/app/hooks/useTransactions.ts
pavanjoshi914: we did that before as well. we only have the memo field in the invoice which is meaninfull so we set title and description both as memo

src/app/hooks/useTransactions.ts
pavanjoshi914: let's decide this along with a new method that returns both incoming and outgoing transactions. we also need to decide it on other connectors as well. i will probably prefer to do it as follow up pr

pavanjoshi914: we also need to discuss feasibility for this for other connectors as well. something that could be done as follow up pr. yup!

src/app/components/TransactionsTable/index.tsx
pavanjoshi914: resolved

src/app/components/TransactionsTable/index.tsx
pavanjoshi914: bitcoin designs has only those icon, we need custom fat arrow icons if we want the icon similar to design. i didn't put much priority on icons currently

src/app/components/TransactionsTable/index.tsx
pavanjoshi914: resolved

pavanjoshi914:
@rolznz icons thickness and size is corrected as well matching the original design

reneaaron:
|          | Alby | Citadel | Commando (core-lightning) | Eclair | Galoy | Kollider | LNBits | LND | LNDHub |
|----------|------|---------|---------------------------|--------|-------|----------|--------|-----|--------|
| Incoming |      |         |                           |        |       |          |        |     |        |
| - Empty  |      |         |                           |        |       |          |        |     |        |
| - Memo   |      |         |                           |        |       |          |        |     |        |
| - LNURLP Comment |      |         |                           |        |       |          |        |     |        |
| - Boostagram |      |         |                           |        |       |          |        |     |        |
| Outgoing |      |         |                           |        |       |          |        |     |        |
| - Empty  |      |         |                           |        |       |          |        |     |        |
| - Memo   |      |         |                           |        |       |          |        |     |        |

src/extension/background-script/actions/ln/transactions.ts
```diff
@@ -0,0 +1,45 @@
+import utils from "~/common/lib/utils";
+import state from "~/extension/background-script/state";
+import type { Invoice, MessageTransactions } from "~/types";
+
+const transactions = async (message: MessageTransactions) => {
+  const isSettled = message.args.isSettled;
```
reneaaron, line 6: Are there any use cases for loading invoices that are not yet settled? I think we don't display them anywhere, do we?

src/extension/background-script/actions/ln/transactions.ts
```diff
@@ -0,0 +1,45 @@
...
+  const limit = message.args.limit;
```
reneaaron, line 7: FYI: This is quite messed up implementation here. 

I think the limit / offset would need to be passed from the UI all the way down to the  connector.getTransactions() call since the connector is the one loading the actual transactions. 

Just to keep in mind when we want to add pagination for transactions one day.

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -17,19 +17,20 @@ interface Route {
   total_fees: number;
 }
 
-export interface ConnectorInvoice {
+export interface ConnectorTransaction {
```
reneaaron, line 20: Renamed because we use this type for both invoices and payments.

src/i18n/locales/en/translation.json
```diff
@@ -1105,25 +1100,22 @@
         "allow_camera_access": "Please allow camera access in the settings screen."
       }
     },
-    "transactionsTable": {
+    "transactions_table": {
```
reneaaron, line 1103: Renamed for consistency, also renamed all translation files.

src/types.ts
```diff
@@ -884,14 +890,15 @@ export type SupportedExchanges = "alby" | "coindesk" | "yadio";
 
 export interface Invoice {
```
reneaaron, line 891: I think this type should be renamed for consistency or removed if possible. Couldn't the ConnectorTransaction already decode the custom_records? 

Transaction: Invoice + Payment
ConnectorTransaction: Invoice + Payment on connector level

reneaaron (commented):
Looks quite good so far, some feedback to address :arrow_down:

src/app/components/TransactionsTable/index.tsx
```diff
@@ -18,143 +24,104 @@ export default function TransactionsTable({
...
-              {({ open }) => (
-                <>
-                  <div className="flex">
+  function openDetails(transaction: Transaction) {
+    setTransaction(transaction);
+    setModalOpen(true);
+  }
+
+  function getTransactionType(tx: Transaction): "incoming" | "outgoing" {
+    return [tx.type && "sent", "sending"].includes(tx.type)
```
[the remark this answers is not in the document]
pavanjoshi914: resolved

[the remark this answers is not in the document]
pavanjoshi914: resolved

src/app/components/TransactionsTable/index.tsx
```diff
@@ -18,143 +24,104 @@ export default function TransactionsTable({
...
-                              rel="noopener noreferrer"
-                            >
-                              {tx.title}
-                            </a>
-                          ) : (
-                            tx.title || tx.boostagram?.message || "\u00A0"
-                          )}
+                          {tx.title ||
+                            tx.boostagram?.message ||
+                            (type == "incoming" ? t("received") : t("sent"))}
```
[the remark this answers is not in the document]
pavanjoshi914: why is this needed. in the end it will be either sent or received?

src/app/hooks/useTransactions.ts
```diff
@@ -11,22 +11,30 @@ export const useTransactions = () => {
...
           limit,
         });
-        const _transactions: Transaction[] =
-          await convertPaymentsToTransactions(payments);
 
-        for (const transaction of _transactions) {
+        const transactions = getTransactionsResponse.transactions.map(
+          (transaction) => ({
+            ...transaction,
+            title: transaction.memo,
```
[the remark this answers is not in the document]
pavanjoshi914: we get only memos from the endpoints. so on the connector level, it's the same. removing it

src/extension/background-script/actions/ln/transactions.ts
pavanjoshi914: we don't load unsettled invoices that is why we do check. we can remove that. but we wait for multiple testings of endpoint so that we are sure, no unsettled invoices are returned by getTransactions

src/extension/background-script/connectors/connector.interface.ts
pavanjoshi914: thanks!

src/extension/background-script/actions/ln/transactions.ts
pavanjoshi914: we use the same hook, for showing all transactions and for showing limited transactions, so we do need to pass  according to context.

pavanjoshi914: we are not passing it to connector, connector will return all transactions by default !

we are doing slicing in the action itself
```
    if (limit) {
      transactions = transactions.slice(0, limit);
    }
```

src/types.ts
pavanjoshi914: we still need Invoice type for getInvoice method as it returns invoice itself and not connectorTRansaction

src/extension/background-script/actions/ln/transactions.ts
```diff
@@ -0,0 +1,46 @@
+import utils from "~/common/lib/utils";
+import { ConnectorTransaction } from "~/extension/background-script/connectors/connector.interface";
+import state from "~/extension/background-script/state";
+import type { MessageTransactions } from "~/types";
+
+const transactions = async (message: MessageTransactions) => {
```
rolznz, line 6: should this be `getTransactions`?

src/app/hooks/useTransactions.ts
```diff
@@ -11,22 +11,29 @@ export const useTransactions = () => {
...
-          accountId,
+        const getTransactionsResponse = await api.getTransactions({
+          isSettled: true,
           limit,
         });
-        const _transactions: Transaction[] =
-          await convertPaymentsToTransactions(payments);
 
-        for (const transaction of _transactions) {
+        const transactions = getTransactionsResponse.transactions.map(
```
rolznz, line 21: Do we still need to do the mapping here, or should this be handled somewhere else?

src/extension/background-script/actions/ln/transactions.ts
rolznz: you can still filter them out, right? maybe we just do not need isSettled in message.args?

src/extension/background-script/actions/ln/transactions.ts
```diff
@@ -0,0 +1,46 @@
...
+        );
+        return {
+          ...transaction,
+          boostagram,
+          paymentHash: transaction.payment_hash,
+        };
+      });
+
+    if (limit) {
+      transactions = transactions.slice(0, limit);
```
rolznz, line 30: should the limit be passed to connector.getTransactions()? it'd be better to only request `limit` items rather than filtering them out in memory afterwards

src/extension/background-script/connectors/alby.ts
```diff
@@ -92,15 +94,16 @@ export default class Alby implements Connector {
       client.incomingInvoices({})
     )) as Invoice[];
 
-    const invoices: ConnectorInvoice[] = incomingInvoices.map(
-      (invoice, index): ConnectorInvoice => ({
+    const invoices: ConnectorTransaction[] = incomingInvoices.map(
```
rolznz, line 97: are both `getInvoices() and getTransactions()` still needed? shouldn't we have just one method?

src/extension/background-script/connectors/commando.ts
```diff
@@ -222,6 +247,54 @@ export default class Commando implements Connector {
       });
   }
 
+  async getTransactions(): Promise<GetTransactionsResponse> {
+    const incomingInvoicesResponse = await this.getInvoices();
+    const outgoingInvoicesResponse = await this.getPayments();
+
+    const transactions: ConnectorTransaction[] = [
```
rolznz, line 254: we do this merging in multiple connectors, should it be extracted to a utility function? e.g. `mergeTransactions(invoices,payments)`

src/extension/background-script/actions/ln/transactions.ts
reneaaron: Also rename the message type: MessageGetTransactions

reneaaron:
**TODO**
- [ ] Add decode boostagrams to outgoing transactions
- [ ] Invoices API: Remove `isSettled` from args (not used at all)
- [ ] Create a new issue for passing `offset`, `limit`. 
- [ ] LNC: Filter unsettled invoices in memory
- [ ] #2839: Show unsettled payments in transaction lists (e.g. HODL invoices) 
- [ ] Follow-up: Try to get rid of the Invoice type 
  - [ ] ([memo vs. title](https://github.com/getAlby/lightning-browser-extension/blob/feat/transaction-list/src/app/hooks/useTransactions.ts#L24)
  - [ ] tackle overlaps of types, e.g. `payment_hash` and `paymentHash`, maybe map explicitly instead of ...) - not a problem in terms of TS types, but the resulting JS object contains all fields (e.g. when you `console.log` it)

src/extension/background-script/actions/ln/transactions.ts
pavanjoshi914: resolved

src/extension/background-script/actions/ln/transactions.ts
pavanjoshi914: resolved

src/extension/background-script/connectors/commando.ts
pavanjoshi914: resolved

reneaaron:
tACK

PR #2748, opened 2023-09-11
feat: send message to all the valid tabs which has receiving end

pavanjoshi914:

we query active tabs and iterate over them. if we have the tab that doesn't have the content script for such tab when we do postMessage. the error, receiving end does not exist will be thrown and we will get out of the function. this can lead to not notifying to active tabs

for eg. [tab not having content script, tab having content script]

tab having a content script is in the second place of the array. when we iterate and post message. we will get error for tab present in first position. and hence no notification logic will be executed  for tab which is on second position of the array.

solution would be. handle errors inside for loop instead of top level function

image

Fixes _#ISSUE-NUMBER_

### Type of change

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/extension/background-script/actions/accounts/select.ts
+++ b/src/extension/background-script/actions/accounts/select.ts
@@ -47,20 +47,20 @@ export default select;
 // Send a notification message to the content script
 // which will then be posted to the window so websites can sync with the switched account
 async function notifyAccountChanged() {
-  try {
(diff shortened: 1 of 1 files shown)


rolznz (approved):
tACK

PR #2753, opened 2023-09-15
feat: add isEnabled provider method, which returns provider was ever enabled

### LINKED ISSUE #2692: [Feature] webln should remember enabled status
### Feature description

Currently you need to invoke window.webln.enable() or track it in your app. It would be nice for the extension (the provider) to remember whether webln has been enabled before or not.

### Describe the solution

`window.webln.isEnabled()` should return true if webln has ever been enabled (and not disabled again)

Requires an update to the webln spec (add `isEnabled(): boolean`)

**NOTE: Should also work with other providers that require explicit .enable()**

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:

1. add isEnabled method which indicate provider was ever enabled or not

behaviour:

1. when extension is locked, isenabled return false
2. when provider was not enabled before, isEnabled returns false
3. when provider was enabled before and allowance was set, return true and override the enabled properties of content script and provider objects
4. when provider was enabled before and then allowance was removed, return false on reload 

Fixes  #2692 

### Type of change

https://github.com/getAlby/lightning-browser-extension/assets/55848322/e546410b-4479-4fe7-8e8a-d425abe0b6ac

### How has this been tested?

by  using providers on the websites

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


bumi:
I am not sure if local storage is enough, because it will not catch the case when the user removes the allowance. 
then it will be still in localstorage enabled.

rolznz:
> I am not sure if local storage is enough, because it will not catch the case when the user removes the allowance. then it will be still in localstorage enabled.

Who would do this in real usage? can't we also delete the localstorage value if the user deletes the allowance?

If it's not in localstorage it'll probably need a request to the background script, which would be async and then the provider wouldn't instantly be setup which may affect sites that check window.webln.

rolznz:
Can we confirm if there is a way the two could get out of sync? if so, this could be a problem. Sites that use a `window.(webln|webbtc|liquid).isEnabled()` check might not try to call enable() and then continuously fail.

src/extension/providers/providerBase.ts
```diff
@@ -8,11 +8,20 @@ export default class ProviderBase {
   private _scope: string;
 
   constructor(scope: string) {
-    this.enabled = false;
+    this.enabled = this.loadEnabledState();
     this._eventEmitter = new EventEmitter();
     this._scope = scope;
   }
 
+  private loadEnabledState(): boolean {
```
rolznz, line 16: @pavanjoshi914 we were planning to add an `isEnabled()` function right? can this be added to the same PR?

src/extension/providers/providerBase.ts
```diff
@@ -8,11 +8,20 @@ export default class ProviderBase {
...
+    const storedEnabled = localStorage.getItem("enabled");
+    return storedEnabled === "true";
+  }
+
+  private saveEnabledState(enabled: boolean): void {
+    localStorage.setItem("enabled", enabled.toString());
```
rolznz, line 22: I think "enabled" as a key is quite unsafe as any site can write this value. If we decide to use localStorage, we should prefix Alby-specific keys.

Also, we need a different key per provider, right?

For example `LBE-${this.scope}-enabled`?

src/extension/content-script/webln.js
```diff
@@ -24,7 +24,12 @@ const weblnCalls = [
 // calls that can be executed when webln is not enabled for the current content page
 const disabledCalls = ["webln/enable"];
 
-let isEnabled = false; // store if webln is enabled for this content page
+function loadEnabledState() {
```
rolznz, line 27: why do we do this in the webln provider? doesn't it make sense to store enabled state for all providers?

pavanjoshi914: haven't extended this for all the providers. let's first decide the flow

src/extension/providers/providerBase.ts
bumi: first try to identify and think about which local storage is used here. 

+ the website/inpage script context?
+ the conent script context?
+ the background script context?

we must understand where this is executed and what scope/access it has.

bumi:
> Who would do this in real usage? can't we also delete the localstorage value if the user deletes the allowance?


no, I don't think so. 
removing an allowance does not know anything about some localstorage somewhere else. Thus tight-coupling must be avoided. 
And the action also will not have access to the local storage.

bumi:
I am not sure if we should do some local storage here or if not a `await webln.isEnabled()` that checks for the allowance (and caches the result if called once) is enough? 

why do you avoide the async call?

rolznz:
> I am not sure if we should do some local storage here or if not a `await webln.isEnabled()` that checks for the allowance (and caches the result if called once) is enough?
> 
> why do you avoide the async call?

Websites might check webln.isEnabled() on init. Now that window.webln is available immediately, shouldn't it also be setup correctly? otherwise it might cause some unexpected behaviour.

src/extension/providers/providerBase.ts
pavanjoshi914: image


it's stored in website-specific local storage. which is accessible by in page script and content script as well. and hence from content script it can be accessed by background script also

pavanjoshi914:
I am still unclear about the behavior that we want to achieve here. afaik, we don't want to use local storage as well. so i have written a base code that doesn't include the localstorage usage but uses db.

1. does isProviderEnabled() shall return true after allowance is deleted? the current implementation doesn't do so, it will return false if allowance is deleted. (i am using recently added enabledFor field for the same which tracks providers for which allowance is present)
2. does the enabled property of the provider need to be overridden?  as we fetch things from db, and providers are in inpage script it's complicated + it will slow down the calls

@bumi @rolznz

bumi:
so I would say: 

`webln.isEnabled()` does a normal call to the background script - similar to `enable()` but it only checks if there is a matching allowance available or not. 
if not it resolves to false, 
if it is it resolves to true but also sets the `enabled` properties (as you do in the content script, but also in the provider object itself)

if the allowance is deleted `isEnabled` will return false after the next page reload

pavanjoshi914:
1. i enable the webln, enabled sets to true
2. i reload the site, enabled will be reset to false
3. i now call isProviderEnabled(), then it returns true cause webln was enabled before and i override the value this.enabled with the response of the function

is this case valid?

bumi:
yes, that `enabled` property is only internal.

extanally one always has to go through the function.

pavanjoshi914:
cool! also shall we rename the function to webln.hasEnabled()? we already have isEnabled in webbtc provider which conflicts with the function name, unless i tweak webbtc provider a little more

bumi:
I'd say we use `isEnabled()` for the pubic facing API. 
internally we can use something else.

src/extension/providers/providerBase.ts
pavanjoshi914: resolved

src/extension/background-script/actions/alby/isEnabled.ts
```diff
@@ -0,0 +1,30 @@
+import { getHostFromSender } from "~/common/utils/helpers";
+import db from "~/extension/background-script/db";
+import type { MessageAllowanceEnable, Sender } from "~/types";
+
+import state from "../../state";
+
+const isEnabled = async (message: MessageAllowanceEnable, sender: Sender) => {
+  const host = getHostFromSender(sender);
```
rolznz, line 8: @bumi do you think it would be ok to have a function that just takes the provider name and then all actions can use it? 

The only difference between these files is "alby" vs "liquid" etc.

src/extension/content-script/liquid.js
```diff
@@ -76,6 +77,10 @@ async function init() {
           }
         }
 
+        if (ev.data.action === `${SCOPE}/isEnabled`) {
```
rolznz, line 80: Now we use `${SCOPE}` it would be nice to reduce duplication in these content script provider files as a followup PR. What do you think?

src/extension/providers/webbtc/index.ts
```diff
@@ -7,12 +7,12 @@ declare global {
 }
 
 export default class WebBTCProvider extends ProviderBase {
-  isEnabled: boolean;
+  isWebBTCEnabled: boolean;
```
rolznz, line 10: this should be undone, right?

src/extension/content-script/liquid.js
pavanjoshi914: yup in some files there is scope variable declared and in some there is not. i will clean it up a bit in follow up pr

src/extension/providers/webbtc/index.ts
pavanjoshi914: this conflicts with the function name isEnabled. so I changed boolean name

src/extension/providers/providerBase.ts
```diff
@@ -8,6 +8,7 @@ export default class ProviderBase {
   private _scope: string;
 
   constructor(scope: string) {
+    this._scope = scope;
```
rolznz, line 11: Could you rename `enabled` to `_isEnabled` and make it private?

src/extension/providers/webbtc/index.ts
rolznz: Edit: both this and executing should be removed

pavanjoshi914: resolved

src/extension/content-script/liquid.js
bumi: these are very critical, a mistake must not be made. thus I kinda actually also like hardcoded stuff here.

bumi (approved):
utACK. looks good to me. simple, helpful!

rolznz (approved):
tACK

PR #2755, opened 2023-09-16
feat: Adding necessary labels to increase accessibility of screens

### LINKED ISSUE #2573: Changes to be done on alby extension to increase accessibility
### Feature description

1. The balance show is not labeled so to check balance we have to go to alby accounts.
image 

2. Recent Transactions are also not labeled. Outgoing and incoming are not changing with the tab button. We have to select incoming by right arrow key, so if any person who is blind uses it he/she may be unaware about this field and how to select it.

  image

3. Incoming transactions are not labeled. We can make them dynamic also so we can also see who send them.
image

4. With keyboard we are unable to select ‘See all Transactions’
image

5. On the receiving screen we have only these fields labeled so we can also label qr code to scan and other remaining also.
image

6. Here on the send screen there is no label on amount which can help for the amount confirmation and how much amount they are sending.
image

7. On the success screen there is no level or any announcement for the success of the transaction.
image

8. When we click on these dropdowns then we are not able to select the dropdown option with  tab , we have to use arrow keys which make it difficult for any blink user to judge what he/she has to do.
image

### Describe the solution

1. Every information on the screen or any CTA should be labeled.
2.  To switch between outgoing and incoming we can use tab and to get more detail about individual we can get them by click enter.
3. 'See all transactions' should be labeled and clickable.
4.  For buttons there should be different label like 'Close Button'.
5. For sending there should be conformation of amount.
6.  Also there should be label for the success screen also.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

Yes

Rithvik-padma:

Added labels at various places in the following screens - 
1. Default screen
2. Receive screen
3. Send screen
4. Result screen

Fixes #2573

### Type of change

 `feat`: New feature (non-breaking change which adds functionality)

### How has this been tested?

Tested the labels using a screen reader but few screens have not been tested as I am using a testnet (send and result screens).

### Checklist

- [x] Manual testing
- [x] Darkmode
- [x] Responsive layout


Rithvik-padma:
@stackingsaunter @rolznz Can you review this PR.
Thanks!

src/app/components/AccountMenu/index.tsx
```diff
@@ -60,7 +60,10 @@ function AccountMenu({ showOptions = true }: Props) {
   return (
     <div className="relative pl-2 flex justify-end w-72">
       <Menu as="div">
-        <Menu.Button className="h-full px-2 rounded-md hover:bg-gray-100 dark:hover:bg-white/10 transition-colors duration-200">
+        <Menu.Button
+          className="h-full px-2 rounded-md hover:bg-gray-100 dark:hover:bg-white/10 transition-colors duration-200"
+          aria-label={`Account menu`}
```
rolznz, line 65: you can use standard quotes`"Account menu"` unless you need to use a template string

src/app/components/BalanceBox/index.tsx
```diff
@@ -16,6 +16,10 @@ function BalanceBox({ className }: Props) {
         "w-full flex flex-col items-center justify-center dark:text-white p-4",
         className ?? ""
       )}
+      tabIndex={0}
+      aria-label={`Balance: ${balancesDecorated.accountBalance}${
```
rolznz, line 20: why do we use aria-label here and not where the balance is rendered?

could a simple label like "balance" be added to the div that actually renders the value?

src/app/components/TransactionsTable/index.tsx
```diff
@@ -27,7 +27,13 @@ export default function TransactionsTable({
       <Loading />
     </div>
   ) : !transactions?.length && noResultMsg ? (
-    <p className="text-gray-500 dark:text-neutral-400">{noResultMsg}</p>
+    <p
+      className="text-gray-500 dark:text-neutral-400"
+      tabIndex={0}
+      aria-label={noResultMsg}
```
rolznz, line 33: why do we add a label when the value is already visible and it's exactly the same? Is there any benefit?

rolznz: There are quite a few similar issues here. Could you re-review the changes and make sure the labels provided are actually helpful and necessary? then I think @stackingsaunter can hopefully test it.

src/app/components/AccountMenu/index.tsx
Rithvik-padma: Okay

src/app/components/BalanceBox/index.tsx
Rithvik-padma: We can add the aria-label to the BalanceBox component directly at the place where it is rendering I have added it in the div inside the component so that at every place the BalanceBox component is being used the screen reader can detect it but ig in this case BalanceBox is used only at one place so I will add it there itself.

I did not get your second point here, can you elaborate a bit?

Rithvik-padma: Yes, but in this case by using the `aria-label` attribute we are erasing the default label (which is the child elements' text) and creating a new label which replaces the default label rather than adding/extending to it. So that is hy there is a need of adding the data again to the `aria-label`.

Rithvik-padma:
@rolznz I have made the changes as per your review.
@stackingsaunter Can u test the changes.
Thanks!

src/app/components/BalanceBox/index.tsx
rolznz: @Rithvik-padma I do not believe that is true. aria-label does not replace anything, it just enhances and provides more information.

A simple example would be:
`<div aria-label="Wallet Balance">{walletBalance}</div>`

A visually impaired user does not know the number inside the div is the balance. But the aria-label enhances that div and reads both the value, and the label.

rolznz:
> @rolznz I have made the changes as per your review. @stackingsaunter Can u test the changes. Thanks!

Before the changes can be tested we need to make sure we do this right. I think there needs some more changes first.

Rithvik-padma:
> > @rolznz I have made the changes as per your review. @stackingsaunter Can u test the changes. Thanks!
> 
> Before the changes can be tested we need to make sure we do this right. I think there needs some more changes first.

Sure got it!

src/app/components/BalanceBox/index.tsx
Rithvik-padma: @rolznz 
No, I felt it does the same initially but I have tested this format using a screen reader but this doesn't work, the screen reader only reads only the aria-label value but not the value inside it.
These are the resources I have referred to confirm - [Stackoverflow](https://stackoverflow.com/questions/72883953/screenreader-should-read-both-aria-label-and-original-content), [Medium post](https://bootcamp.uxdesign.cc/when-to-use-aria-label-or-screen-reader-only-text-cd778627b43b) and also MDN docs.

Anyway there is an alternative which is using a label at the necessary places and making it visually hidden using the `sr-only` class provided by tailwindcss. I have tested this method using screen reader, it works fine and looks like a better solution for labels.
Should I use this method instead of using `aria-label` attribute?

Rithvik-padma: > @rolznz No, I felt it does the same initially but I have tested this format using a screen reader but this doesn't work, the screen reader only reads only the aria-label value but not the value inside it. These are the resources I have referred to confirm - [Stackoverflow](https://stackoverflow.com/questions/72883953/screenreader-should-read-both-aria-label-and-original-content), [Medium post](https://bootcamp.uxdesign.cc/when-to-use-aria-label-or-screen-reader-only-text-cd778627b43b) and also MDN docs.
> 
> Anyway there is an alternative which is using a label at the necessary places and making it visually hidden using the `sr-only` class provided by tailwindcss. I have tested this method using screen reader, it works fine and looks like a better solution for labels. Should I use this method instead of using `aria-label` attribute?

@rolznz any thoughts on this?

Rithvik-padma:
@rolznz @stackingsaunter any update on this?

src/app/components/BalanceBox/index.tsx
rolznz: @Rithvik-padma I like the suggestion in the stack overflow answer you linked:

`<p><span class="visually-hidden">Participant</span>
  Martin Clark</p>`
  
There is support in tailwind for this: https://tailwindcss.com/docs/screen-readers

That way we can remove some of the duplication. What do you think?

Rithvik-padma: Yes I will make changes for it

reneaaron:
There seem to be quite some merge conflicts and also all of the accessibility strings seem to miss translations :thinking: Can you update this PR and add them (try to use existing translations wherever possible)

stackingsaunter:
@Rithvik-padma sorry to keep you waiting that long for the review! I can't test it right now, but will gladly do with fixes that @rolznz and @reneaaron mentioned

PR #2764, opened 2023-09-19
Fix: UI feedback

rolznz:

Address feedback from @stackingsaunter 

- fix: only use white background on qr avatar
- fix: do not allow user to setup nostr again if they already have a key
- chore: minor receive translation changes to prevent orphans


bumi:
How do people get to the nostr setup again? they should not see that link I guess

PR #2769, opened 2023-09-20
fix: validate the nostr private key only while saving

### LINKED ISSUE #2763: [BUG] Not nice experience manually entering nostr private key

### Describe the bug

private key is validated on each keypress rather than on form submission.

### Screenshots [optional]

image

### Steps To Reproduce

1. Go to nostr private key settings
2. Clear the form and start typing

### Expected behavior

Key is only validated on save

### Alby information

3.1.1

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

None

Rithvik-padma:

In the Nostr settings when we try to change the nostr private key the validation takes place on every change of character in the input. Now I have changed it so that validation take only when the user clicks on the save button.

Fixes #2763 

### Type of change

### How has this been tested?

Tested by manually populating the nostr private key.

### Checklist

- [x] Manual testing


Rithvik-padma:
@rolznz Can you review this PR.
Thanks!

src/app/screens/Accounts/NostrSettings/index.tsx
```diff
@@ -88,38 +67,56 @@ function NostrSettings() {
...
-    }
-
     try {
-      if (nostrPrivateKey) {
-        await api.nostr.setPrivateKey(id, nostrPrivateKey);
-      } else {
-        await api.nostr.removePrivateKey(id);
+      // TODO: is there a way this can be moved to the background script and use the Nostr object?
+      // NOTE: it is done this way to show the user the new public key before saving
+      setNostrPublicKey(
```
rolznz, line 73: this doesn't work - the nostrPublicKey is now blank until you click save.

What about we undo the changes and just not show an error toast if we fail to set the pubkey?

Rithvik-padma: Okay then we will just set the public key to an empty string if the input private key is not valid and then when after the user clicks on the save button then we can validate, if the public key does not exist then we can error toast it.
Does this method sound good?

Rithvik-padma:
@rolnz I have made the changes as mentioned, can you please review it.

src/app/screens/Accounts/NostrSettings/index.tsx
```diff
@@ -47,27 +47,6 @@ function NostrSettings() {
     fetchData();
   }, [fetchData]);
 
-  useEffect(() => {
```
rolznz, line 50: Could we keep this code and just not do the toast in the catch block? we could log the error to the console instead

src/app/screens/Accounts/NostrSettings/index.tsx
rolznz: @Rithvik-padma sorry, I think this actually can be fixed by a minor change of changing a toast to logging the error to the console. We already have error handling when we save the key.

rolznz (approved):
tACK

Thanks @Rithvik-padma and sorry for not considering fully what needed to be changed there.

Rithvik-padma:
@rolnz I have one suggestion to make here would it be better not to navigate to the account settings screen whenever there is an error?

rolznz:
> @rolnz I have one suggestion to make here would it be better not to navigate to the account settings screen whenever there is an error?

Why would that be better? and in what case would there be an error like this?

Here: if you paste an invalid nostr key, it'll show you an error toast. You can check your key again, re-paste and save.

PR #2776, opened 2023-09-25
feat: set allowance per provider

### LINKED ISSUE #2758: [Feature] enable popups should only enable allowance for the current provider
### Feature description

Currently if you enable nostr (e.g. by calling `window.nostr.getPublicKey()`) on a website without an allowance, then an allowance will be set. Then if you call window.webln.enable() it'll enable instantly because an allowance already exists.

This is problematic because each provider requests different functionality, and the enable screens only show what the current provider enables, which is quite confusing, and a user can give more access to a site than they might wish to.

### Describe the solution

allowances needs to be changed to allow enabling for a specific provider rather than all providers in one go.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:

set allowance per provider

Fixes #2758 

### Type of change

https://github.com/getAlby/lightning-browser-extension/assets/55848322/0bf778fa-6162-49be-9740-433c88d0af7c

### How has this been tested?

by enabling allowance per provider

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


src/types.ts
```diff
@@ -806,6 +806,7 @@ export interface Blocklist extends DbBlocklist {}
 
 export interface DbAllowance {
   createdAt: string;
+  enabledFor?: string[];
```
rolznz, line 809: what about `providers`? is there any downside of this name? I feel like it's more specific

src/extension/background-script/actions/webln/enable.ts
```diff
@@ -38,17 +43,24 @@ const enable = async (message: MessageAllowanceEnable, sender: Sender) => {
           if (!allowance.id) {
             return { data: { error: "id is missing" } };
           }
-          await db.allowances.update(allowance.id, {
-            enabled: true,
-            name: message.origin.name,
-            imageURL: message.origin.icon,
-          });
+          if (allowance.enabledFor) {
+            const updatedEnabledFor = allowance.enabledFor.includes("webln")
```
rolznz, line 47: why do we do a check here when further up there is `allowance.enabledFor?.includes("webln")` further up?

rolznz: Or is the check above wrong?

rolznz: We need to make sure we can update an allowance to include multiple providers, so I think the above should be removed?

Also, maybe we could convert the array to a set and back to avoid the tertiary operator

pavanjoshi914: we don't want to duplicate the entry next time. that is why we need to check provider name is already there or not

to reproduce: unlock the extension and do the provider call. if we don't do that duplicate entries will be there in the list

pavanjoshi914:
@rolznz  do you have idea about that one failing test. i am still unable to get why is that failing?

src/extension/background-script/actions/webln/enable.ts
rolznz: @pavanjoshi914 how does it work when you have the allowance.enabledFor = ["nostr"] and then you try to enable it to include "webln". It doesn't seem possible with the current checks?

rolznz: @pavanjoshi914 oh sorry, I read it wrong. Thanks!

This comment would be good to fix though I think: `maybe we could convert the array to a set and back to avoid the tertiary operator`

rolznz:
> @rolznz do you have idea about that one failing test. i am still unable to get why is that failing?

@pavanjoshi914 it seems you defined the array wrong in the dexie schema: https://dexie.org/docs/MultiEntry-Index

src/types.ts
pavanjoshi914: allowance are set not just by provider. but also during the payments. for eg. while doing a lnurl-pay to the website, in future we might add some other entries there as well, isn't it good to keep it generalized. so that it can hold any kind of entries

src/extension/background-script/actions/webln/enable.ts
pavanjoshi914: resolved

src/extension/background-script/actions/liquid/enable.ts
```diff
@@ -39,8 +47,13 @@ const enable = async (message: MessageAllowanceEnable, sender: Sender) => {
           if (!allowance.id) {
             return { data: { error: "id is missing" } };
           }
+
+          if (!enabledFor.has("liquid")) {
+            enabledFor.add("liquid");
+          }
           await db.allowances.update(allowance.id, {
             enabled: true,
+            enabledFor: [...enabledFor],
```
rolznz, line 56: why `[...enabledFor]`?

rolznz:
I've tested it and it seems to work well!

The allowance screen hasn't been changed, I think it would it be good to show which providers have been enabled there. Would it be good to do this as a follow up PR?

src/extension/background-script/actions/liquid/enable.ts
pavanjoshi914: that's a set variable name

rolznz: I don't understand. Can't you just do `enabledFor,` instead of `enabledFor: [...enabledFor],`?

pavanjoshi914: resolved. i just pushed values back in the form of an array notation. but this works as well

rolznz (approved):
tACK, I think we need another PR to display the enabled providers in the allowance screen and on the connected sites page.

Best to get a review from @reneaaron or @bumi though before merging

src/extension/background-script/actions/alby/enable.ts
```diff
@@ -38,8 +40,14 @@ const enable = async (message: MessageAllowanceEnable, sender: Sender) => {
           if (!allowance.id) {
             return { data: { error: "id is missing" } };
           }
+
+          if (!enabledFor.has("alby")) {
```
rolznz, line 44: @pavanjoshi914 this check is not needed any more since we are using a set now

pavanjoshi914: why is it different if we use set there? it's done to avoid duplicate entries in enabledFor. 

for example, enable the allowance. we will have entry. and then lock the extension and then enable the allowance again using same provider. see duplicate entries. next time we update the allowance we add duplicate entries. i think this is still needed. kindly cross check again

rolznz: @pavanjoshi914 it uses the javascript `Set` class. Which is a collection without duplicates.

pavanjoshi914: ohh okay! i haven't used set before thanks for the change

PR #2780, opened 2023-09-27
fix: update msw depedency

bumi:
this fixes a version conflict issue with npm install

msw 1.3.0 or greater has support for TS 5.2
https://github.com/mswjs/msw/releases/tag/v1.3.0

otherwise npm install fails


rolznz (approved):
utACK

PR #2803, opened 2023-10-08
Linear min & max sats buttons

matjaz:

_Sats buttons 4 linear values from min to max, with nice rounding & formating (1M, 1k, 100). min & max values are unchanged._

### Type of change

### Checklist

- [x] Manual testing
- [ ] Update Docs & Guides
- For UI-related changes
- [x] Darkmode
- [x] Responsive layout


pavanjoshi914:
can you please try and fix code quality checks in the code?

reneaaron:
@matjaz I don't think a linear distribution of the amounts between min and max is what I would want as a user. When I tip someone I use those buttons as kind of shortcuts for smaller amounts. 

I don't think people would use a 1M sat button a lot, do you? :thinking:

matjaz:
Using LNURL, `LN SERVICE` specify min & max. So it's up to service what values they want.
Example I was considering is service balance withdrawal where max is balance.

When you tip someone you don't use LNURL (no min&max).
When min & max are not specified it fallbacks to current values(1k, 5k 10k, 25k).

There is no point of having disabled buttons.

What do you suggest, @reneaaron ?

@pavanjoshi914 fixed formatting.

matjaz:
https://lntictactoe.com & https://www.lnflip.com enables balance withdrawal.
Let's say user has a balance of 2500 sats, most of the buttons are disabled. For 150000 sats button values are too low.

So for tag
- payRequest, Alby could have a max value of wallet balance or `LNURL` `maxSendable`
- withdrawRequest, `LN SERVICE` defines `maxWithdrawable`

They also have built-in fees which covers withdrawal fees. Probably they also have throttling for withdrawals.

matjaz:
Any feedback on this?
I don't see the point in buttons with higher values than available. For both, payRequest and withdrawRequest.

PR #2804, opened 2023-10-09
fix: improve QR code readability for different wallets

reneaaron:

 - Remove white background for dark mode
 - Selective error correction for lightning addresses (but not invoice QRs)
 - Remove `lightning:` prefix for lightning addresses QR code (consistent with what is displayed for invoices) 
 - Improved sizing of QR codes for bigger screens
 - Deleted some custom code for the success message on the `ReceiveInvoice` screen and replaced it with our `ResultCard`
 
Test results with different readers: 

| Wallet     | Lightning Address | Invoice |
|------------|:-----------------:|:-------:|
| Bitkit     |         ❌         |    ✅    |
| Blixt*      |         ✅         |    ✅    |
| Blink      |         ✅         |    ✅    |
| Breez*      |         ✅         |    ✅    |
| BlueWallet* |        ✅         |    ✅    |
| Fedi       |         ✅         |    ✅    |
| Muun       |         ❌         |    ✅    |
| Phoenix    |         ✅         |    ✅    |
| WoS        |         ✅         |    ✅    |

**Legend**
 - ❌ don't support Lightning Addresses 
 - `*` don't support prefixed QR codes (`lightning:`)

Since some of the wallets don't seem to support prefixed QR codes at the moment (Blixt, Breez, BlueWallet for lightning addresses) I think it's better to remove the prefix for now at the cost of not being able to scan the QR codes with QR code readers or your camera app and being able to directly click a link (but having to copy & paste the contents of the QR code, or just directly scan it with your wallet).


src/app/components/QRCode/index.tsx
```diff
@@ -5,9 +5,12 @@ export type Props = {
   value: string;
   size?: number;
   className?: string;
+
+  // set the level to Q if there are overlays
```
bumi, line 9: why? 

and why can't it always be "Q"?

rolznz: Always using Q will make the invoice QR codes unnecessarily more detailed, but it still works, right?

reneaaron: Q will improve error correction (so we can add overlays covering up to 25% of the QR) at the price of decreased information density (meaning the QR codes "pixels" have to be smaller to encode the same information). 

While that isn't that much of a problem for lightning addresses (because they are usually quite short), for invoices that contain larger amount of data those QR codes can get "harder" to read. (meaning you have to aim your phone very precisely and have to wait longer for the reader to recognize the QR code)

The QR codes still work, but since we don't have overlays there (yet) I would prefer not to use the error correction and keep the QR codes pixels as big as possible for improved readability.

bumi: ok, thanks for the explanation. makes sense.
can you add this to that comment there. the comment explains the what but not the why which we will not know in the future

rolznz: Done

rolznz (approved):
tACK

Issue #1568, opened 2022-10-05
back and forth navigation on the publisher screen 


MoritzKa:
### Feature description

The latest release allows navigating to the home screen in the pop up. But there is no way to navigate back to the website/publisher screen

### Describe the solution

Add the possibility to navigate back and forth
just for illustration purposes
![image](https://user-images.githubusercontent.com/13707997/194049771-967cf955-a48a-40dd-97db-ab20bbf96da3.png)
![image](https://user-images.githubusercontent.com/13707997/194050019-f42c5008-f6f3-4ac9-a8fa-9ff7d074a674.png)

### Additional context [optional]

_Add any other additional context or screenshots about the feature request here_

### Are you working on this?

_(Yes/No)_

MoritzKa:
not urgent.

bumi:
I think this has been there for a while already and not part of the latest release. 
is this really a case? why did you want to do that? 

(closing and opening the popup does that)

ranjithjupaka:
I would like to work on the issue can you assign this to me.

escapedcat:
> I would like to work on the issue can you assign this to me.

Before any action maybe describe how you would tackle this first. Would like to avoid you wasting time.  
I think we still need at least a mockup and/or a design as well.

DineshGupta-droid:
Hello @bumi @escapedcat, I am Dinesh I wanted to contribute to this task can you guide me through this as I'm new to it and I am A SOB participant.

escapedcat:
@im-adithya is already working on this, maybe he has an idea how you could support.

escapedcat:
Not sure what the priority is here. Maybe @stackingsaunter has some input/feedback.

DineshGupta-droid:
Hello @stackingsaunter  I wanted to contribute to the task can you guide me through this as I'm new to it and I am A SOB participant.

stackingsaunter:
@MoritzKa  What would you say if instead of showing website and it's history, we could show main balance with some "connected" card that could allow you to view history?

This also solves the problem that on website view we dont have send/receive buttons

![CleanShot 2024-01-29 at 16 19 11@2x](https://github.com/getAlby/lightning-browser-extension/assets/68239231/c52726f9-6d49-4a9c-9e97-821c24d125b9)

MoritzKa:
Sounds good. Makes sense for me.

MoritzKa:
from my side yes. @stackingsaunter what do you think?

[...]

PR #2821, opened 2023-10-14
feat: add allowance screen send link

lujakob:

Add "Send" button/link to popup Allowance View to spare one user click to send a payment. Optionally preset invoice field value from clipboard.

Fixes 2812

### Type of change

### How has this been tested?

### Checklist

- [ ] Manual testing
- [ ] Update Docs & Guides
- For UI-related changes
- [ ] Darkmode
- [ ] Responsive layout

### DIFF
--- a/src/app/screens/Home/AllowanceView/index.tsx
+++ b/src/app/screens/Home/AllowanceView/index.tsx
@@ -15,6 +15,8 @@ import { useSettings } from "~/app/context/SettingsContext";
 import { PublisherLnData } from "~/app/screens/Home/PublisherLnData";
 import { convertPaymentsToTransactions } from "~/app/utils/payments";
 import type { Allowance, Battery, Transaction } from "~/types";
+import Hyperlink from "@components/Hyperlink";
+import { useNavigate } from "react-router-dom";
 
 dayjs.extend(relativeTime);
 
@@ -33,6 +35,8 @@ const AllowanceView: FC<Props> = (props) => {
     getFormattedFiat,
     getFormattedNumber,
   } = useSettings();
+
+  const navigate = useNavigate();
 
   const [transactions, setTransactions] = useState<Transaction[] | null>(null);
   const [isLoadingTransactions, setIsLoadingTransactions] = useState(true);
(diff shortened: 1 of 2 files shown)


lujakob:
@stackingsaunter I would welcome your opinion about the link/button. @bitcoinuser suggested a button at the bottom of the screen, but I think the bottom part is reserved for the "Recent transactions" list. I would suggest to place it underneath the website image box (or even the Hyperlink "Send a budget for one-click payments") - a button could be more prominent, but takes a lot of space, so maybe a Hyperlink as well? (clicking the link navigates to the send screen, same as arrow back at the top left)


Bildschirmfoto 2023-10-14 um 11 07 31

lujakob:
Anyone knows why a `navigator.clipboard.readText()` call throws an exception?

`DOMException: Document is not focused.
code: 0
message: "Document is not focused."
name: "NotAllowedError"`

clipboard-read permissions are set as far as I can see. I test calling this function delayed and see the field is focused, so it should work.

Maybe @rolznz @reneaaron got an idea?

lujakob:
TBH I'm not sure if we want to automatically paste an address anyways, because it is less explicit. Dealing with payments and addresses is very delicate. Am I really sure this is the right address I want to pay and not another address I copied just seconds before. I'm not sure if this might confuse or lead to unexpected behaviour. What do you guys think? Safe to apply automatic clipboard paste?

rolznz:
I think actually the allowance view should not require the user to go "back" to the home page in order to send a payment. Could it just be a small section at the top of the popup but still on the home page, so the user can scroll down to see the normal actions? then we do not need to add send/receive/etc buttons here too.

For automatic paste, I think @bumi had some hesitations about this. Even if we try to make sure we only paste values that are relevant by using a regex, access to the clipboard without a user action may be a bit unsafe?

rolznz:
@reneaaron @stackingsaunter  I feel like being able to see the current site's transaction history is less important, what do you think?

stackingsaunter:
We talked about this today with @reneaaron but there's one problem, that any Send button placed on this screen leaves an impression like you're about to send a payment to this website (little bit like in battery payment to Bumi above). So I'd actually not show Send button on this screen at all.

Another approach would be something like this: 


image

By default we show main popup screen with a box indicating where we are connected. Then by clicking it we enter website screen. Thoughts? @reneaaron @rolznz

stackingsaunter:
I think this could be closed in favour of https://github.com/getAlby/lightning-browser-extension/issues/1568#issuecomment-2000028368 ? cc @reneaaron

PR #2830, opened 2023-10-24
feat: use token events to subscribe to the token changes

pavanjoshi914:

listen for token changes via alby-js-sdk events

### Type of change

### How has this been tested?

trying oauth in extension

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


src/extension/background-script/connectors/alby.ts
```diff
@@ -311,10 +317,9 @@ export default class Alby implements Connector {
 
       throw error;
     } finally {
-      const newToken = this._authUser.token;
-      if (newToken && newToken !== oldToken) {
-        await this._updateOAuthToken(newToken);
-      }
+      this._authUser.on("tokenRefreshed", async (token: Token) => {
```
bumi, line 320: shouldn't we define this event listener once globally whenever we set the `this._authUser`?  (as I think you do above)
and then remove this finally.

pavanjoshi914: my bad! I missed the link, this will also duplicate the event listener.

so the flow would be.
1. user creates the account, event is registered as it's registered in authorize function.
2. if token is expired when making a request, token will be updated, and callback is called registered via event listener.

src/extension/background-script/connectors/alby.ts
```diff
@@ -254,11 +254,18 @@ export default class Alby implements Connector {
         token: this.config.oAuthToken, // initialize with existing token
       });
 
+      authClient.on("tokenRefreshed", async (token: Token) => {
+        await this._updateOAuthToken(token);
```
rolznz, line 258: I do not think async and await are needed here, there's only one method call. Also, js-sdk just uses .emit() and I don't believe it can wait for the result.

src/extension/background-script/connectors/alby.ts
```diff
@@ -254,11 +254,18 @@ export default class Alby implements Connector {
...
+      });
+
+      authClient.on("tokenRefreshFailed", async (error: Error) => {
+        console.error("Failed to Refresh token", error);
```
rolznz, line 262: It would be nice to comment here what happens when the refresh fails in the different scenarios. Can we make sure they are always handled correctly (e.g. the auth modal is re-launched)

src/extension/background-script/connectors/alby.ts
```diff
@@ -254,11 +254,18 @@ export default class Alby implements Connector {
...
+      });
+
       if (this.config.oAuthToken) {
         try {
           if (authClient.isAccessTokenExpired()) {
-            const token = await authClient.refreshAccessToken();
-            await this._updateOAuthToken(token.token);
+            await authClient.refreshAccessToken();
```
rolznz, line 268: Further down in this file (Lines 290-292)
```
 const token = await authClient.requestAccessToken(code);
      await this._updateOAuthToken(token.token);
      return authClient;
```
We should not need to manually call this._updateOAuthToken. It should be fired from the `tokenRefreshed` callback (needs change in alby-js-sdk).

src/extension/background-script/connectors/alby.ts
rolznz: Maybe add a comment:

```
// Currently the JS SDK guarantees request of a new refresh token is done synchronously.
// The only way a refresh should fail is if the refresh token has expired, which is handled when the connector is initialized.
// If a token refresh fails after init then the connector will be unusable, but we will still log errors here so that this can be debugged if it does ever happen.
```

src/extension/background-script/connectors/alby.ts
pavanjoshi914: resolved

src/extension/background-script/connectors/alby.ts
pavanjoshi914: resolved

src/extension/background-script/connectors/alby.ts
pavanjoshi914: made a pr for that https://github.com/getAlby/js-sdk/pull/121

rolznz: We decided this is not a good way to go, so let's leave this as-is

src/extension/background-script/connectors/alby.ts
```diff
@@ -254,11 +254,22 @@ export default class Alby implements Connector {
         token: this.config.oAuthToken, // initialize with existing token
       });
 
+      //// Currently the JS SDK guarantees request of a new refresh token is done synchronously.
```
rolznz, line 257: I think this comment is in the wrong place

Can you put it in the `authClient.on("tokenRefreshFailed")` function?

rolznz (approved):
tACK

PR #2833, opened 2023-10-29
feat: Disable 'Create Invoice' button until an amount is entered

### LINKED ISSUE #2802: [BUG] Create invoice button should be disabled until user add a value in the amount field

### Describe the bug

Hi,

Create invoice button should be disabled until user add a value in the amount field

Captura de tela de 2023-10-06 23-41-48

### Screenshots [optional]

_No response_

### Steps To Reproduce

1- Click in receive, then Lightning Invoice

### Expected behavior

Create invoice button should be disabled until user add a value in the amount field

### Alby information

Alby: 3.2.1

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

No

AdarshRawat1:
The button is now disabled until the user enters a value in the 'Amount' field. This ensures that the user cannot create an invoice without specifying the amount first.

This change helps to prevent invalid or incomplete invoice creation.

Fixes #2802

### Type of change

Current  - ( Button is active by default ) 

https://github.com/getAlby/lightning-browser-extension/assets/100958893/68f2fb21-60bb-445d-b43e-9a55d30e1873

After PR -( Button is disabled by default and enabled when user enters a value ) 

https://github.com/getAlby/lightning-browser-extension/assets/100958893/45cf92a0-1a5a-47e7-803d-c68c724a5bef

### Checklist

- [x] Manual testing
- [ ] Update Docs & Guides
- For UI-related changes
- [x] Darkmode
- [x] Responsive layout

### DIFF
--- a/src/app/screens/ReceiveInvoice/index.tsx
+++ b/src/app/screens/ReceiveInvoice/index.tsx
@@ -264,7 +264,11 @@ function ReceiveInvoice() {
                   fullWidth
                   primary
                   loading={loadingInvoice}
-                  disabled={loadingInvoice}
+                  disabled={
+                    loadingInvoice ||
+                    !formData.amount ||
+                    formData.amount === "0"
+                  }
                 />
               </Container>
             </fieldset>


src/app/screens/ReceiveInvoice/index.tsx
```diff
@@ -264,7 +264,11 @@ function ReceiveInvoice() {
                   fullWidth
                   primary
                   loading={loadingInvoice}
-                  disabled={loadingInvoice}
+                  disabled={
+                    loadingInvoice ||
+                    !formData.amount ||
+                    formData.amount === "0"
```
pavanjoshi914, line 270: @rolznz how many of the connectors support amount less invoice, I think lndhub supports that. even if it not supports we throw the error. so probably we remove the third condtion. rest looks good

pavanjoshi914 (changes requested):

src/app/screens/ReceiveInvoice/index.tsx
rolznz: I don't think a user should have to manually specify "0" for an amountless invoice. I see Lndhub works fine right now - I paid an amountless invoice created in Alby using a different wallet. (I know the Alby extension cannot pay amountless invoices yet).

So maybe we should actually not merge this PR and not change anything.

@bitcoinuser do you remember why you made this issue?

bitcoinuser: > @bitcoinuser do you remember why you made this issue?

To be consistent with others windows of Alby, like the send one, that the button are disabled until we insert an information in the field.

bitcoinuser: > I don't think a user should have to manually specify "0" for an amountless invoice.

Why not show the qrcode on receive window for amountless invoice like what we have for custodial alby wallet? In the case of Alby account show lightning address qrcode, but for non custodial wallets, could show a amountless qrcode, no?

rolznz: I don't think we can always rely on a QR code (unless there is a copy button as well). But I think it's unrelated to this issue. Like Pavan said I would just remove this check and not allow 0-amount invoices and then this PR looks ok to me :rocket:

AdarshRawat1: @bitcoinuser @rolznz @pavanjoshi914 Thank you for your valuable feedback. I have removed the third check.

reneaaron:
So we make 0-amount invoices unusable for all connectors that support that currently for the sake of disabling a button?

I like the approach suggested here with the 0-amount invoice being shown by default on the receive screen (if there is no lightning address and the connector supports it) but until we have that I would prefer to keep this feature in there.

PR #2836, opened 2023-11-01
Fix the small padding issue

### LINKED ISSUE #2788: [BUG] Small padding issue for the confirm/cancel buttons in the LNURL pay send screen

### Describe the bug

The buttons are breaking a little (padding issue) when more options are open, this hinders the uniformity of padding.

### Screenshots [optional]

mzBlJky1Mv

### Steps To Reproduce

1. Visit the Alby extension
2. Click on the `Send` button
3. Populate the `Recipient` field and click on `Continue` button.
4. Click on the `More` button at the bottom and you will observe that the padding is not uniform and there is a little bit of breaking issue.

### Expected behavior

The padding issue to be fixed to make the alignment of buttons uniform

### Alby information

Alby Version - 3.2.1

### Device information

- OS - Windows
- Browser - Chrome

### Additional context

_No response_

### Are you working on this?

Yes

Rithvik-padma:

fix the small padding issue for confirm/cancel buttons in the LNURL pay screen when the "more options" button is clicked

Fixes #2788 

### Type of change

VvcZLQbvCc

### How has this been tested?

Tested manually

### Checklist

- [x] Manual testing
- [x] Responsive layout


Rithvik-padma:
@rolznz can u please review this PR.
Thanks!

pavanjoshi914:
this will always break in the future for other screens where we use show more button. better to make a permanent fix here



an alternate to fix this would be to assign bottom paddings to the last buttons div itself and not in the container.tsx (https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/Container.tsx#L32 notice pb-4 here) but this will lead to change in other files where container component is used. 

wdyt @rolznz

Rithvik-padma:
> this will always break in the future for other screens where we use show more button. better to make a permanent fix here!
> 
> 
> 
> the problem here is when we expand fields, the bottom padding present in container field is not reflecting.
> 
> 
> 
> 
> 
> 
> 
> an alternate to fix this would be to assign bottom paddings to the last button div itself and not in the container.tsx (https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/components/Container.tsx#L32 notice pb-4 here) but this will lead to changes in other files where the container component is used. 
> 
> 
> 
> wdyt @rolznz 
> 
> 
> 
> 

Yes I agree, some change has to be made in `Container` component for a permanent fix or change the button padding everywhere `Container` component is used

rolznz:
@pavanjoshi914 @Rithvik-padma I noticed in the Container component there is a `h-full` class which is problematic if the height of the component becomes larger than the height of the popup because the bottom will be cut off (here you see the padding, but it's actually worse than that). It seems to fix it if this is changed to `min-h-full` without any need to change padding elsewhere and I can't see any side effects. Does this look good to you? CC @reneaaron

Rithvik-padma:
> @pavanjoshi914 @Rithvik-padma I noticed in the Container component there is a `h-full` class which is problematic if the height of the component becomes larger than the height of the popup because the bottom will be cut off (here you see the padding, but it's actually worse than that). It seems to fix it if this is changed to `min-h-full` without any need to change padding elsewhere and I can't see any side effects. Does this look good to you? CC @reneaaron

@rolznz I have already tried doing that but the alignment of the child components is getting disturbed in the this screen and its possible to happen in other screens where `Container` component is being used as the alignment is done considering the `h-full` class in the `Container` component. The `LNURL` pay screen looks like this if `min-h-full` instead of the `h-full`.

image

rolznz:
@Rithvik-padma can you highlight the issue in the screenshot? is it the extra spacing before the description? if you inspect it, now the css rules (like justify-between) on the container are correctly being applied and you can see the first element is just the publisher card.

I think the fix I suggested is now revealing an existing issue. If we don't want spacing there, we should not be using justifyBetween to add space between the publisher card and the form

Rithvik-padma:
> @Rithvik-padma can you highlight the issue in the screenshot? is it the extra spacing before the description? if you inspect it, now the css rules (like justify-between) on the container are correctly being applied and you can see the first element is just the publisher card.
> 
> I think the fix I suggested is now revealing an existing issue. If we don't want spacing there, we should not be using justifyBetween to add space between the publisher card and the form

@rolznz So I just found out that the possible reason for this issue is the `fieldset` element, it is not aligning as expected for `flex`, I have tried using the `div` tag instead of it and it works fine by considering `min-h-full` class in `Container` component.

So ig this means that there is no breaking if we use the `min-h-full` class in the `Container` component but the problem is with the `fieldset` tag.

I will make some changes for this issue by replacing the `fieldset` tag, can you please take a look into it.

rolznz:
@Rithvik-padma the send page looks good to me. The close button on the success page is a bit broken - could you please check that one?
image

Rithvik-padma:
> @Rithvik-padma the send page looks good to me. The close button on the success page is a bit broken - could you please check that one?
> 
> image
> 
> 

Sure!

Rithvik-padma:
@rolnz how can I test the success page while using a test account

Rithvik-padma:
@rolznz I have fixed the padding issue in the result page, apparently `min-h-full` class was causing an issue there.

rolznz (approved):
tACK

rolznz:
Thanks @Rithvik-padma !

PR #2841, opened 2023-11-03
fix: remove kollider

### LINKED ISSUE #2840: Remove Kollider connector
### Feature description

Kollider is shutting down: 

https://twitter.com/kollider_trade/status/1720428682072772708

### Describe the solution

Remove connector and all other things related to that (logos, translations, etc)

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

reneaaron:

Fixes #2840


bumi:
We can not do this, people currently might have accounts using it and for them the class would now be missing. 

We can remove it from the UI but the connector must stay there.

reneaaron:
Good point, we could also think about removing the connector config from the list of accounts, right? The check could potentially be done once after a version upgrade? (similar to DB migrations?)

bumi:
I would just keep the connector around but remove it from the UI where people can connect it.

reneaaron:
Good idea, thanks. Updated the PR and re-added the connector. :ok_hand:

PR #2847, opened 2023-11-06
fix: LNC typings

### LINKED ISSUE #2844: Add LNC typings
### Feature description

LNC is currently used as an `FixMe` type, which prevents our tooling from catching errors.

https://github.com/getAlby/lightning-browser-extension/blob/feat/transaction-list/src/extension/background-script/connectors/lnc.ts#L159

### Describe the solution

Using types provided by the LNC library.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

reneaaron:

Fixes #2844


pavanjoshi914 (approved):
do we have lnc node to test this? code wise lgtm

reneaaron:
> do we have lnc node to test this? code wise lgtm

I don't think so, I've tested it with my own one.

PR #2864, opened 2023-11-14
fix: caching getInfo calls

reneaaron:

Cache `getInfo()` calls on connector level for the lifetime of the connector. (changing accounts will reload the info as every time a new instance is created)


bumi (approved):

reneaaron:
tACK 

cc @rolznz

PR #2869, opened 2023-11-16
feat: handle auth flow natively

pavanjoshi914:

one way we can eliminate identity api is to use completely native browser api so that we can support all the browsers including firefox for android

doesn't require any hacks or content script. just using tabs api and event listeners which works on all the browsers including android

### DIFF
--- a/src/extension/background-script/connectors/alby.ts
+++ b/src/extension/background-script/connectors/alby.ts
@@ -339,7 +339,7 @@ export default class Alby implements Connector {
       return null;
     }
     const urlSearchParams = new URLSearchParams(url.split("?")[1]);
-    return urlSearchParams.get("code") || null;
+    return urlSearchParams.get("code");
   }
 
   private async _request<T>(func: (client: Client) => T) {
--- a/src/extension/background-script/connectors/alby.ts
+++ b/src/extension/background-script/connectors/alby.ts
@@ -282,8 +282,6 @@ export default class Alby implements Connector {
         authorizeUrl: process.env.ALBY_OAUTH_AUTHORIZE_URL,
       });
 
-      console.log(authUrl);
-
       authUrl += "&webln=false"; // stop getalby.com login modal launching lnurl auth
 
       // Use the tabs API to create a new tab with the authorization URL
@@ -296,17 +294,8 @@ export default class Alby implements Connector {
           changeInfo: browser.Tabs.OnUpdatedChangeInfoType,
           tab: browser.Tabs.Tab
         ) => {
-          console.log(changeInfo.status);
           if (changeInfo.status === "complete" && tabId === createTab.id) {
-            // The tab has finished loading, you can check the tab URL for the authorization code
-            // Extract the code from the URL and proceed with token exchange
-
-            // Note: You might need to handle errors and corner cases depending on the authorization flow.
             const authorizationCode = this.extractCodeFromTabUrl(tab.url);
-
-            console.log(changeInfo);
-
-            console.log(authorizationCode);
 
             if (authorizationCode) {
               authClient
@@ -320,6 +309,7 @@ export default class Alby implements Connector {
                   reject(error);
                 })
                 .finally(() => {
+                  browser.tabs.remove(tabId);
                   // Remove the event listener once the code is received
                   browser.tabs.onUpdated.removeListener(handleUpdated);
                 });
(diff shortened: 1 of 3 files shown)


pavanjoshi914:
everything else is done only this is to create client. which requires a redirect url. not sure what to use as redirect url

src/extension/background-script/connectors/alby.ts
```diff
@@ -259,9 +261,7 @@ export default class Alby implements Connector {
       authClient.on("tokenRefreshed", (token: Token) => {
         this._updateOAuthToken(token);
       });
-      // Currently the JS SDK guarantees request of a new refresh token is done synchronously.
```
rolznz, line 262: can you re-add the comments? any reason why you removed them?

src/extension/background-script/connectors/alby.ts
```diff
@@ -285,28 +283,60 @@ export default class Alby implements Connector {
...
+      return new Promise<auth.OAuth2User>((resolve, reject) => {
+        const handleUpdated = (
+          tabId: number,
+          changeInfo: browser.Tabs.OnUpdatedChangeInfoType,
+          tab: browser.Tabs.Tab
+        ) => {
+          if (changeInfo.status === "complete" && tabId === oAuthTab.id) {
+            const authorizationCode = this.extractCodeFromTabUrl(tab.url);
+
+            if (authorizationCode) {
```
rolznz, line 298: can you handle the `else` case or maybe better, reverse this check

`if (!authorizationCode) { throw new Error("...") }`

src/extension/background-script/connectors/alby.ts
```diff
@@ -285,28 +283,60 @@ export default class Alby implements Connector {
...
+                  reject(error);
+                })
+                .finally(() => {
+                  browser.tabs.remove(tabId);
+                  browser.tabs.onUpdated.removeListener(handleUpdated);
+                });
+            }
+          }
+        };
+        const handleRemoved = (tabId: number) => {
```
rolznz, line 316: can you rename to `handleTabUpdated` and `handleTabRemoved`

src/extension/background-script/connectors/alby.ts
pavanjoshi914: aah! accidently deleted while cleaning up other comments. resolved

src/extension/background-script/connectors/alby.ts
pavanjoshi914: resolved

src/extension/background-script/connectors/alby.ts
pavanjoshi914: we can do that. but the first ever event listener will trigger for original authURL as it's valid http url but we don't get any code there and for the second time after completiing auth flow we will have redirecturl tab where we get code.

so it will throw error for the first case unnecessarily. is it ok for you?

rolznz: Can we throw an error if there is no authorization code and the URL is the callback URL, but if it's the original auth URL we do not need that check. What do you think?

pavanjoshi914: on the other way, it seems fine to me as well. instead of introducing one more additional check just throw error if listener is triggered and no code is returned

rolznz: But then you mentioned we would throw an error when we should not, right? I don't think we should do that.

I just think we need to add error handling when we expect a code but one is not returned.

bumi:
Is there a way we can append the amount of accounts the user has to the URL?

pavanjoshi914:
> Is there a way we can append the amount of accounts the user has to the URL?

not sure about this. why we need that?

bumi:
> > Is there a way we can append the amount of accounts the user has to the URL?

> not sure about this. why we need that?

There are those reports that people struggling with adding another account when they are logged in because we don't prompt them. 
So I thought maybe we auto authorize the first call but not when the user has multiple accounts already.

pavanjoshi914:
we build auth url and add search-params in the following way in js-sdk

```
  url.search = buildQueryString({
      ...options,
      client_id,
      scope: scopes.join(" "),
      response_type: "code",
      redirect_uri: callback,
      code_challenge_method,
      code_challenge,
    });
```

I don't think we have any option to pass additional metadata such as a number of accounts. maybe we introduce such fields in options? or any separate field?

reneaaron (approved):
tACK

Created a follow-up-issue for improving the account connection in multi-account scenarios: 

#2909

Issue #2872, opened 2023-11-16
 Swap "accounts" for "wallets" copy


stackingsaunter:
Part of #2871 epic, which full desired effect [can be viewed in this Figma](https://www.figma.com/file/w3ZLyJ2ScwPfGKNuzcaNhx/Extension-accounts-%26-settings-problem?type=design&node-id=54-10065&mode=design&t=ML7bvnXqIyn1Ntit-0)

Goal is to change copy related to "extension accounts" that from now on will be called "wallets". 

List of changes:

### Dropdowns
- "Settings" --> "Extension Settings"
- "Account settings" --> "Wallet Settings"
- "Add an account" --> "Connect a Wallet"
- "Manage accounts" --> "Manage Wallets
- "Switch account" --> "Switch Wallet"

![image](https://github.com/getAlby/lightning-browser-extension/assets/68239231/ada2f5ae-4d90-464b-9f5c-6e0c1bb6e56a)

### "Settings" page
- Page title changes to "Extensions Settings" 

### "Account settings" page
- Page title changes to "Wallet Settings" 

### "Accounts" (Manage accounts)
- Title changes to "Wallets"
- Button copy changes to "Connect a Wallet"
- 
![CleanShot 2023-11-16 at 21 18 42@2x](https://github.com/getAlby/lightning-browser-extension/assets/68239231/ef91b945-34a3-439c-bfc3-e27ec09150c4)

ayushdhiman-py:
@stackingsaunter Could you please provide details of the files that need to be changed? It would be really helpful.

rolznz:
@ayushdhiman-py this is part of the task to figure that out and make the changes.

Harshit-Prasad:
@stackingsaunter can you assign me this #2872 issue.
I went through the figma files and went through all the required changes I will get back to you once i finish setting up the project.

Harshit-Prasad:
@stackingsaunter sorry but i won't be continuing the issue, thank you as I got to learn about react-i18n. Also this was my first time working on an extension.

rolznz:
@Harshit-Prasad thanks for you work. I don't think the translations are an issue here and we can remove the outdated keys of the other languages by running a script.

stackingsaunter:
@AdarshRawat1 forgot to mention this in initial post

On Wallet Settings page in "Danger" zone copy also must be updated to wallets:

[...]

PR #2876, opened 2023-11-19
fix: swapped accounts with wallet

Harshit-Prasad:
This PR swaps the word account with wallet. Mainly in user-menu, accounts-menu and in connect-a-wallet page.

fixes: #2872

### Type of Change

account-menu
user-menu
connect-a-wallet

### How has this been tested?

Changes are reflected in the extension without breaking anything.


Rithvik-padma:
@Harshit-Prasad u need to change the translations for other languages too not only for english

rolznz:
> @Harshit-Prasad u need to change the translations for other languages too not only for english

@Rithvik-padma I don't think this is right - we cannot know what the correct translation is for all languages. There is a script created by @reneaaron in `.github/workflows/translations.yml` which should check keys from other languages to see if any need to be removed, and a script you can run here to remove them: `scripts/remove-outdated-translations.js`

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
reneaaron, line 1: Do we need this file? Otherwise please remove it...

yarn.lock
reneaaron, line 1: Please undo those changes.

reneaaron (commented):
@Harshit-Prasad You just need to update the English translation keys and run the script @rolznz mentioned, which will remove the keys from the translated files. 

This will remove existing translations and allows translators to add the new translations (since they are very different from what we have now, e.g. "Add account" vs "Add wallet").

AdarshRawat1:
> @Harshit-Prasad You just need to update the English translation keys and run the script @rolznz mentioned, which will remove the keys from the translated files.

@reneaaron 
I tried running the scripts and I can't seem to get things right.
image

p.s.-  I even tried modifying the script, trying different arguments nothing seems to work

reneaaron:
That's strange. I just executed the commands in your branch locally and they seemed to work. :thinking: 

Maybe you could try to debug the script by adding console.log statements here:
https://github.com/getAlby/lightning-browser-extension/blob/fix/swapped-account-for-wallet/scripts/remove-outdated-translations.js#L13

Otherwise just skip it and I'll have another look before we merge it.

PR #2882, opened 2023-11-22
feat: add support for tabs api as an alternate in prompt opening logi…

pavanjoshi914:
…c of extension

if not able to open popups open tabs

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


src/common/lib/utils.ts
```diff
@@ -85,89 +86,101 @@ const utils = {
...
-          }
+      const createTab = (options: browser.Tabs.CreateCreatePropertiesType) => {
+        return browser.tabs.create(options);
+      };
 
-          // Re-focus the popup after 2 seconds to mitigate the problem of lost popups
-          // (e.g. when a user clicks the website)
-          setTimeout(() => {
-            if (!window.id) return;
+      const optionsPopup: browser.Windows.CreateCreateDataType = {
```
rolznz, line 97: why declare these outside of createTab/createPopup? if you put them inside then you don't need to pass arguments. Also, maybe it would be nice to suffix with "Options" or "Properties" since they aren't actually a popup/tab, only the config passed to create the popup or tab

src/common/lib/utils.ts
```diff
@@ -85,89 +86,101 @@ const utils = {
...
-                  return resolve(responseMessage);
-                }
-              });
-            }
-          };
+      const createPromise = browser.windows
+        ? createPopup(optionsPopup)
+        : createTab(optionsTab);
+
+      return createPromise.then(async (windowOrTab) => {
```
rolznz, line 114: Maybe we can use await instead of callbacks to help simplify this?

src/common/lib/utils.ts
pavanjoshi914: resolved

src/common/lib/utils.ts
pavanjoshi914: we had callbacks before as well! i think we should do minimal possible changes here so that its easy to review the logic behind the code

reneaaron (approved):
tACK

reneaaron (approved):
tACK

PR #2887, opened 2023-11-24
fix: Blink connector to use the API key from the Blink Dashboard

### LINKED ISSUE #2881: [Feature] Switch to the new API key scheme for Blink wallet
### Feature description

Blink has a new way to manage API keys: https://dashboard.blink.sv/

image

The header has changed to:
```
--header 'X-API-KEY' 'blink_...'
```
from the previously used:
```
--header "Authorization: Bearer ory_..."
```

The old way is only used in the mobile wallet and should not be used for Alby any more as that key can only be revoked by deleting the whole account while the API keys can be created and revoked on demand.

### Describe the solution

Will update the connection instructions and correct the header used.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

Yes

openoms:
Fix the problem that the current method to connect the Blink wallet does not work.
Change the header to use the new method for the API key
Add instructions to the connector page about generating the API key.
Remove translations referencing the deprecated method of log in through the discontinued web wallet.
Use the staging endpoint of a `galoy_staging_` API key is used.
Fix the checkPayment function

Fixes #2881

### Screenshots:
image

Screenshot from 2023-11-24 17-57-32

image

image

### How has this been tested?
Tested manually through building the extension, connected to the Galoy staging and Blink mainnet.
Generating an invoice, checking if received and sending to LN invoice functions are working.

### Known remaining issues (to be addressed separately):
* getTransactions() is not yet supported with the currently used account: Galoy
* default Stablesats wallet is not supported for generating invoices
* cannot pay a signet LN invoice (`lntbs..`): `Unknown coin bech32 prefix`
### Checklist
- [x] Manual testing
- [x] Update Docs & Guides (built in to the connector page)


rolznz:
Hi @openoms 

I tried this out and managed to successfully pay and receive lightning invoices. But I seem to have no history:

image


Also, what happens if the user's token is invalid (e.g. an old jwt token or a new API token that has expired, especially since the default is 90 days now). Do they just receive an error message and have to delete their extension account and add a new one? (This becomes more problematic as users have to migrate e.g. their master key / nostr key. For example, the Alby connector will allow you to re-login if your auth token has expired)

openoms:
Thank you for the feedback!
Rebased to integrate the latest changes in `master`. 

Implemented and tested the transaction history now (returns the last 100 transactions max):
image

In Blink there are two balances: BTC and Stablesats (synthetic USD)
Changed that there is only the BTC wallet is used as Alby is expecting the balances in sats and Stablesats is denominated in USDcents. Using a Stablesats wallet would need bigger changes involving other parts of Alby as well. Would prefer to discuss this separately.

Agree that it is important to be able to relogin to the same account so I am testing the workflow of the token expiration next.

openoms:
After discussing the various possibilities of dealing with the expired API key came to the conclusion of simply creating a longer (or non expiring) API key would be the best trade-off. 

If the user would have other services (like Nostr) connected to the account they should can still migrate with the Master Key.

The Blink Dashboard will have the option to select  a non-expiring key as well so added the relevant notes to the connector screen:

image

Please let me know if there are more changes expected.

openoms:
@rolznz was you old account logged in with and API key (`blink_..`)? It does not work any more with `ory_...`. In case of the API key you will be albe to check if it expired in the dashboard.

You should be able to log in to the dashboard with the number again if you click on `Sign in with phone`. If you are logged in on a phone or in the dashboard you can add email which then you can use to log in as well.

I have run a couple new rounds of tests manually and still working.

rolznz:
> @rolznz was your old account logged in with and API key (`blink_..`)? It does not work any more with `ory_...`. In case of the API key you will be able to check if it is expired in the dashboard.
> 
> You should be able to log in to the dashboard with the number again if you click on `Sign in with phone`. If you are logged in on a phone or in the dashboard you can add an email address which then you can use to log in as well.
> 
> I have run a couple new rounds of tests manually and still working.

@openoms I only created the API key when testing this PR, so it should still be valid.

But I cannot login with my phone number (starts with +666) as the input at https://dashboard.blink.sv/ only seems to allows logging in with email. My phone number doesn't pass the regex validation. Without entering anything in the field, it asks me to fill it out before continuing.

How do I login with phone number?

reneaaron:
Thanks for your PR @openoms! Just tested it and worked like a charm! 

I noticed one minor detail: 
Incoming transactions don't seem to yield the memo (that's why every entry in the transaction list just shows "received"). Do you know why that could be the case? :thinking: 

Existing users that use the Blink with the current implementation will see errors after the upgrade: 
image

I don't see an easy path to migrate them. I guess it shouldn't be too many users though...

reneaaron:
> Changed that there is only the BTC wallet is used as Alby is expecting the balances in sats and Stablesats is denominated in USDcents. Using a Stablesats wallet would need bigger changes involving other parts of Alby as well. Would prefer to discuss this separately.

+1 for tackling that in a separate PR, we already have the foundation to support fiat-denominated accounts (was introduced quite some time back for the Kollider connector). However, in Alby they would have 3 separate accounts for that (BTC, USD, EUR) as we don't have any hierarchy for wallets yet. (users select which currency they want to use at the time of connecting their account) 

If you want to get a first idea of how that could work take a look at the Kollider connector: 
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/kollider.ts#L65

Happy to collaborate and answer any questions you might have. :raised_hands:

openoms:
Thank you for the review @reneaaron!

I also found an outstanding edge case where the `checkPayment` doesn't work if the user sets Stablesats as their default wallet. Will fix this in the next commit and look at the option to preserve the `ory_...` tokens.

The memo issue looks like a bug in the backend: https://github.com/GaloyMoney/galoy/issues/3659 which when fixed should correctly show up in Alby.

Thank you for the clue about following the Kollider connector for Stablesats, will do that and will ask the related question separately.

openoms:
As  a last cosmetic change removed the now inappropriate alpha warning and displaying a note about using only the BTC wallet for now (as the scope of the PR):
image

openoms:
Rebased on the latest master, awaiting the checks.

reneaaron:
@openoms Just did a final round of testing and the legacy accounts now work as well! Thanks for the PR! :100:

rolznz:
@openoms I didn't notice in this PR (these lines were not changed) but @bumi mentioned sendPayment in this connector does not return the preimage. Can this be added?

Also I noticed there it also will return without a preimage if the payment is in "PENDING" state. But lightning payments should technically be instant in most cases, right? or is this for HODL invoices?

image

CC @reneaaron

rolznz:
Update: I saw Rene made an issue here: https://github.com/getAlby/lightning-browser-extension/issues/2946

PR #2890, opened 2023-11-27
clearer account description

bumi:
more recommendations for running an own node


bumi:
@reneaaron let's see what happens with the translations :)

PR #2894, opened 2023-11-28
fix: update copy

reneaaron:

Update copy on nostr enable screen. (apps can't sign without your permission)


bumi:
shall we run the translation script?

reneaaron:
Yeah, probably a good idea. Otherwise those things won't get translated.

PR #2897, opened 2023-11-30
feat: refactor nip-04 screens

### LINKED ISSUE #2889: Improve NIP07 encrypt prompt
### Feature description

The current encrypt prompt only asks you to allow the client to encrypt things, but it doesn't show any details:

image

### Describe the solution

There are 2 things that could be added here: 
 - The message that is being encrypted
 - Which public key this will be encrypted for

pavanjoshi914:

imrprove nip04 screens

Fixes #2889 

### Type of change

image
image
image

### How has this been tested?

trying encrypt and decrypt on nostr chat

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


src/app/screens/Nostr/Confirm.tsx
```diff
@@ -62,21 +72,29 @@ function NostrConfirm() {
...
+              />
+            )}
+            <div className="flex justify-center mb-4 gap-4">
+              <Hyperlink onClick={toggleShowDetails}>
+                {showDetails ? t("hide_details") : t("view_details")}
+              </Hyperlink>
             </div>
+            {showDetails && (
+              <div className="whitespace-pre-wrap break-words p-2 mb-4 shadow bg-white rounded-lg dark:bg-surface-02dp text-gray-500 dark:text-gray-400">
+                {tPermissions("nostr.peer")}: {peer}
```
bumi, line 95: is there always a peer in this confirm screen?

pavanjoshi914: yup! cause confirm screen is only used for encrypt and decrypt actions. for other nostr actions we have different screens

bumi: should we then name the screen EncryptDecryptConfirm? or something?

pavanjoshi914: resolved

src/extension/background-script/actions/nostr/decryptOrPrompt.ts
```diff
@@ -33,9 +33,13 @@ const decryptOrPrompt = async (message: MessageDecryptGet, sender: Sender) => {
...
       }>({
         ...message,
-        action: "public/nostr/confirm",
+        action: "public/nostr/confirmEncryptOrDecrypt",
         args: {
-          description: i18n.t("permissions:nostr.nip04decrypt"),
+          description: {
+            action: i18n.t("permissions:nostr.nip04decrypt"),
+            peer: message.args.peer,
+            message: message.args.ciphertext,
```
bumi, line 41: @reneaaron what do you think. 
This ciphertext will not be useful for the user. so I'd say we should leave this out here. What do you think?

reneaaron: I agree, but then we'd need to have an alternative layout here. I'd suggest creating a follow-up task to split encrypt / decrypt into two different flows. I'd rather have them separate...

reneaaron: Created an issue cc @pavanjoshi914 #2942

PR #2898, opened 2023-11-30
feat: NWC connector 🤙

### LINKED ISSUE #2891: [Feature] Nostr Wallet Connect
### Feature description

Nostr Wallet Connect is becoming a viable way to connect to wallets.
We should create a new connector for NWC which will allow the extension to be used with more new wallets.

### Describe the solution

Implement a new connector implementing the interface: https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/connector.interface.ts

The SDK should be used for this:
https://github.com/getAlby/js-sdk

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

reneaaron:

Fixes #2891 

**TODO** 
 - [x] Implement missing connector methods
 - [x] Store secret of NWC in account config
 - [x] Instructions @ Connector setup
 - [x] Draft specs for missing methods (lookupInvoice, listTransactions)
 - [ ] Implement new specs in the extension

Still quite some stuff work on, but I figured I'd share my excitement for this early on. :fire:


src/extension/background-script/connectors/connector.example
reneaaron, line 1: Template is outdated, we have an interface for this purpose.

src/app/router/connectorRoutes.tsx
```diff
@@ -149,6 +151,12 @@ const connectorMap: { [key: string]: ConnectorRoute } = {
     title: i18n.t("translation:choose_connector.btcpay.title"),
     logo: btcpay,
   },
+  nwc: {
```
rolznz, line 154: We could also now add NWC to the Umbrel page, right?

src/extension/background-script/connectors/nwc.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  constructor(account: Account, config: Config) {
+    this.config = config;
+    this.nwc = new webln.NostrWebLNProvider({
+      nostrWalletConnectUrl: this.config.connectionString,
+    });
+  }
+
+  async init() {
+    await this.nwc.enable();
```
rolznz, line 48: minor: can just return without await if it's only one line

src/extension/background-script/connectors/nwc.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  async unload() {}
```
rolznz, line 51: can disconnect from the relay here, something like `this.nwc.disconnect()`?

src/app/screens/connectors/ConnectNWC/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <Trans i18nKey={"title"} t={t} />
+        </h1>
+      }
+      description={t("page.instructions")}
+      logo={logo}
+      submitLoading={loading}
+      submitDisabled={formData.nostrWalletConnectUrl === ""}
+      onSubmit={handleSubmit}
+    >
+      <div className="mb-6">
```
rolznz, line 85: I wonder if we can do other options than just pasting a NWC URL. Can we even embed Bitcoin Connect here? :o

src/app/router/connectorRoutes.tsx
reneaaron: Yeah, good point! :+1:

src/extension/background-script/connectors/nwc.ts
```diff
@@ -0,0 +1,191 @@
...
+          total_amt: args.amount,
+          // TODO: How to get fees from WebLN?
+          total_fees: 0,
+        },
+      },
+    };
+  }
+
+  async checkPayment(args: CheckPaymentArgs): Promise<CheckPaymentResponse> {
+    // TODO: update WebLN types package
```
rolznz, line 169:
```suggestion
```

src/extension/background-script/connectors/nwc.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  async getInfo(): Promise<GetInfoResponse> {
+    const info = await this.nwc.getInfo();
+    return {
+      data: info.node,
+    };
+  }
+
+  async getBalance(): Promise<GetBalanceResponse> {
+    const balance = await this.nwc.getBalance();
```
bumi, line 68: this methods might raise an error if the call fails (like for example if the method is not implemented).
so I think we need to catch them here and return a proper internal error

bumi: we should also be able to know if the method is supported here, isn't it?

rolznz: Not here unless we call `getInfo` first somewhere in the connector, like in the `init` function. Is that fine?

rolznz: @bumi I tested it with mutiny but mutiny doesn't even support get_info yet so the TestConnection doesn't even work. I will let them know.

rolznz: EDIT: if getInfo fails it should just fallback to ["pay_invoice"] as the supported methods. Something must be wrong here, I'll check again.

rolznz: fix here: https://github.com/getAlby/js-sdk/pull/171

But it does mean that currently we have to wait for the timeout (60 seconds) before the fallback is returned. This is a bit unusable. I think maybe we should be able to provide a shorter timeout to Alby JS SDK for some NWC methods, what do you think?

PR #2930, opened 2023-12-13
feat: secure allowance creationby including protocol information

### LINKED ISSUE #2437: Check for protocol and port in allowances
### Feature description

Currently allowances only consider the domain they are being created on regardless of the protocol or port being used. 

### Describe the solution

 - Include the fully qualified origin for allowances (maybe `origin.domain` could be used here instead of `origin.host`)
 - Migrate existing allowances (prefix with https)
 - Potentially display a warning on the enable screen when connecting to non-ssl domains (MITM)

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:

Include the fully qualified origin for allowances
Migrate existing allowances (prefix with https)
Potentially display a warning on the enable screen when connecting to non-ssl domains (MITM)

Fixes #2437 

### Type of change

image

### How has this been tested?

create allowance on http site

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/app/components/Enable/WeblnEnable.tsx
+++ b/src/app/components/Enable/WeblnEnable.tsx
@@ -4,6 +4,7 @@ import Container from "@components/Container";
 import PublisherCard from "@components/PublisherCard";
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
+import Alert from "~/app/components/Alert";
 import ScreenHeader from "~/app/components/ScreenHeader";
 import toast from "~/app/components/Toast";
 import { USER_REJECTED_ERROR } from "~/common/constants";
@@ -15,6 +16,7 @@ type Props = {
 };
 function WeblnEnableComponent(props: Props) {
   const [loading, setLoading] = useState(false);
+  const hasHttp = props.origin.domain.startsWith("http://");
   const { t } = useTranslation("translation", {
     keyPrefix: "webln_enable",
   });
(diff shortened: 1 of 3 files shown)


src/app/components/Enable/WeblnEnable.tsx
```diff
@@ -62,6 +64,14 @@ function WeblnEnableComponent(props: Props) {
             isSmall={false}
           />
 
+          <div className="pt-3">
+            {hasHttp && (
+              <Alert type="warn">
+                ⚠️ you are connecting to a unsecure domain
```
bumi, line 70: we should put this in i18n

pavanjoshi914: resolved

bumi (commented):
when we save new allowances, don't we there also need a change to save it with the protocol?

pavanjoshi914:
> when we save new allowances, don't we there also need a change to save it with the protocol?

is there any other way we can create allowances? if we use providers to create one. we already store allowance host along with https prefix

bumi:
what do we use in here? https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/allowances/add.ts#L5

pavanjoshi914:
> what do we use in here? [`master`/src/extension/background-script/actions/allowances/add.ts#L5](https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/allowances/add.ts?rgh-link-date=2023-12-18T12%3A21%3A38Z#L5)

aah! yes you are right, we add allowances while confirmPayment + doing keysends as well. fixed

bumi:
This currently breaks the links on the connected sites pages. 

I think we should only add the warning and not change the allowance and the host usage for now. We can do this in a follow-up

bumi (changes requested):
let's only add the warning. the change in the allowance has some issues like for example the links to the site a broken. 

let's do this in a follow-up PR.

pavanjoshi914:
> e change in the allowance has some issues like for example the links to the site a broken.

resolved! + for existing connected sites when we have migrations running. they will not cause the problem

PR #2947, opened 2023-12-19
fix: styles for paymentsummary

### LINKED ISSUE #2944: Fix copy items with labels on transaction screens
Currently we have 2 different stylings of 2 very similar components that appear across different actions in the extension:

image

I propose following improvements:

- [x] Label: text-sm medium, gray800 dark: neutral200
- [x] Copy: text-lg regular, gray600 dark: neutral400
- [x] Delete brackets ( ) from fiat element
- [x] ideally it could be just the same defined component, but that may be a seperate PR

CleanShot 2023-12-19 at 20 07 43@2x

CleanShot 2023-12-19 at 20 08 32@2x

reneaaron:

Fixes #2944

### DIFF
--- a/src/app/components/PaymentSummary/index.tsx
+++ b/src/app/components/PaymentSummary/index.tsx
@@ -4,50 +4,45 @@ import { useSettings } from "~/app/context/SettingsContext";
 
 export type Props = {
   amount: string | number;
-  amountAlt?: string;
   description?: string | React.ReactNode;
   fiatAmount: string;
 };
 
-const PaymentSummary: FC<Props> = ({
-  amount,
-  amountAlt,
-  description,
-  fiatAmount,
-}) => {
+const PaymentSummary: FC<Props> = ({ amount, description, fiatAmount }) => {
   const { t: tCommon } = useTranslation("common");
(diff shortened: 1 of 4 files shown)


src/app/components/PaymentSummary/index.tsx
```diff
@@ -4,50 +4,45 @@ import { useSettings } from "~/app/context/SettingsContext";
...
-            className="text-gray-500 dark:text-neutral-500"
-            data-testid="fiat_amount"
-          >
-            {" "}
-            (~{fiatAmount})
-          </span>
-        )}
-      </dd>
-      {amountAlt && <dd className="text-gray-400">{amountAlt}</dd>}
+    <dl className="space-y-4">
```
pavanjoshi914, line 16: this have no effect. maybe add this style on div element instead of dl. but spacing is quite large maybe just remove this

reneaaron: Without `space-y-4`:
image

With `space-y-4`:
image

reneaaron: Implemented spacing according to the designs, space is needed desperately.

PR #2950, opened 2023-12-20
chore: remove topup tip

bumi:

Topup is currently unsupported

### Type of change


reneaaron (approved):
tACK

rolznz:
Why not just add a flag? or do we never plan to have topup tips in the future?

reneaaron:
The onboarding cards will probably restructured quite a bit anyway, if we'd ever need it again we can recover the code from git :+1:

PR #2952, opened 2023-12-20
feat: new onboarding

reneaaron:

Some new designs for the onboarding flow that help users explore available options:

image


src/app/screens/connectors/ChooseConnectorPath/index.tsx
```diff
@@ -1,66 +1,135 @@
...
             }
+            actions={<ConnectAlby />}
           />
         </div>
       </div>
     </div>
   );
 }
+
+const FeatureItem = ({
```
bumi, line 114: is this missing some light mode font config? 
there the text is white.

src/app/screens/connectors/ChooseConnectorPath/index.tsx
```diff
@@ -1,66 +1,135 @@
...
+                <FeatureItem type="success">{t("voltage.point3")}</FeatureItem>
+                <FeatureItem type="success">{t("voltage.point4")}</FeatureItem>
+              </>
+            }
+            actions={
               <Link
-                to="choose-connector"
-                className="flex flex-wrap gap-6 w-72 lg:w-80 mx-auto my-5 items-center"
-                tabIndex={-1}
+                to="https://info.getalby.com/voltage"
```
bumi, line 75: is this a URL that we easily can redirect? otherwise I'd recommend getalby.com/voltage

bumi: What do you think about copying the LND connector screen and using that with custom Voltage code? 

we can add a link to the guide there and this could be the image:
CleanShot 2023-12-22 at 00 16 10@2x

src/app/screens/connectors/ChooseConnectorPath/index.tsx
reneaaron: Great catch! :sweat:

pavanjoshi914:
we need to adopt this for smaller screens as well

PR #2960, opened 2023-12-23
feat: add screen reader labels for better accessibility

### LINKED ISSUE #2573: Changes to be done on alby extension to increase accessibility

Rithvik-padma:

Adding screen reader labels to elements wherever necessary for better accessibility.

For both the menus (user and account menus) as the `Menu` component is being used tab keys cannot be used to traverse menu items so instead I have added a screen reader message at the beginning of menu so that the user is notified that he can only use the arrow keys to access menu items.

Fixes #2573 

### Type of change

### How has this been tested?

Tested manually using a screen reader.

### Checklist

- [x] Manual testing
- For UI-related changes
- [x] Darkmode
- [x] Responsive layout


Rithvik-padma:
@rolnz Please review this PR.
Thanks!

Rithvik-padma:
> @rolnz Please review this PR. Thanks!

I have added some translation keys for some component screen reader labelling. I guess there is no need to run any script for newly added keys. If it is not the case please give some feedback for further translations changes.

PR #2973, opened 2023-12-29
feat: responsive testconnection

pavanjoshi914:
make test connection screen responsive and refactor headings

image

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


reneaaron (changes requested):
I think the names of those fields are not necessary at this point. A user doesn't know what an "alias" is, anyway.

pavanjoshi914:
will this work?
image

pavanjoshi914:
the only motive here is to split alias and account name as it's too long for smaller screens as it was inline before

reneaaron:
I don't understand. The account name & alias have always been separated by a dash, right? 

image

pavanjoshi914:
yup but on smaller screens it looks something like this. and the account name are mostly continuous strings. so instead of breaking words or hiding overflows i wished to separate both on new line
image

PR #2978, opened 2024-01-02
feat: use separate instances to avoid provider being invaded by other wallets

pavanjoshi914:

### Checklist

- [X] Manual testing
- [ ] Update Docs & Guides
- For UI-related changes
- [ ] Darkmode
- [ ] Responsive layout

### DIFF
--- a/src/extension/inpage-script/index.js
+++ b/src/extension/inpage-script/index.js
@@ -13,12 +13,7 @@ function init() {
   window.nostr = new NostrProvider();
   window.webbtc = new WebBTCProvider();
   window.webln = new WebLNProvider();
-  window.alby = new AlbyProvider(
-    window.liquid,
-    window.nostr,
-    window.webbtc,
-    window.webln
-  );
+  window.alby = new AlbyProvider();
   const readyEvent = new Event("webln:ready");
   window.dispatchEvent(readyEvent);
 
--- a/src/extension/providers/alby/index.ts
+++ b/src/extension/providers/alby/index.ts
@@ -5,22 +5,13 @@ import WebBTCProvider from "~/extension/providers/webbtc";
 import WebLNProvider from "~/extension/providers/webln";
 
 export default class AlbyProvider extends ProviderBase {
-  liquid: LiquidProvider;
-  nostr: NostrProvider;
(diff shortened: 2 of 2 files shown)


src/extension/providers/alby/index.ts
```diff
@@ -5,22 +5,13 @@ import WebBTCProvider from "~/extension/providers/webbtc";
 import WebLNProvider from "~/extension/providers/webln";
 
 export default class AlbyProvider extends ProviderBase {
-  liquid: LiquidProvider;
-  nostr: NostrProvider;
-  webbtc: WebBTCProvider;
-  webln: WebLNProvider;
+  webln = new WebLNProvider();
```
reneaaron, line 8: Can we, instead of creating completely new ones, reference our own objects here so they only exist once? Otherwise `enable()` states are not shared between objects. 

```js
const liquid = new LiquidProvider();
window.liquid = liquid;
...
window.alby = new AlbyProvider(liquid, nostr, webbtc, webln);
```

reneaaron (changes requested):

src/extension/providers/alby/index.ts
pavanjoshi914: resolved as per discussion in slack

src/extension/inpage-script/index.js
```diff
@@ -13,12 +13,7 @@ function init() {
   window.nostr = new NostrProvider();
   window.webbtc = new WebBTCProvider();
   window.webln = new WebLNProvider();
-  window.alby = new AlbyProvider(
```
bumi, line 16: does this work? 

```js
const liquid = new LiquidProvider();
const nostr = new NostrProvider();
const webbtc = new WebBTCProvider();
const webln = new WebLNProvider();

window.liquid = liquid;
window.nostr = nostr;
window.webbtc = webbtc
window.webln = webln;

const alby = new AlbyProvider(liquid, nostr, webbtc, webln);
window.alby = alby;
```

pavanjoshi914: image

pavanjoshi914: unfortunately no

PR #3046, opened 2024-02-15
fix lnurl channel

pavanjoshi914:
show proper error messages

fix lnurl callback
proper error messages while doing requests

### Type of change

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/app/screens/LNURLChannel/index.tsx
+++ b/src/app/screens/LNURLChannel/index.tsx
@@ -44,9 +44,6 @@ function LNURLChannel() {
       const nodeId = infoResponse.node.pubkey;
 
       if (!nodeId) {
-        toast.error(
-          `No nodeId available. Your account might not support channel requests`
-        );
         throw new Error(
           `No nodeId available. Your account might not support channel requests`
         );
--- a/src/app/screens/LNURLChannel/index.tsx
+++ b/src/app/screens/LNURLChannel/index.tsx
@@ -31,7 +31,7 @@ function LNURLChannel() {
 
   const [loadingConfirm, setLoadingConfirm] = useState(false);
   const [successMessage, setSuccessMessage] = useState("");
-  const [isPrivateChannel, setIsPrivateChannel] = useState(false);
+  const [privateChannel, setPrivateChannel] = useState(false);
 
   async function confirm() {
     try {
@@ -56,7 +56,7 @@ function LNURLChannel() {
         params: {
           k1: details.k1,
           remoteid: nodeId,
-          private: isPrivateChannel ? 1 : 0,
+          private: privateChannel ? 1 : 0,
         },
         adapter: fetchAdapter,
       });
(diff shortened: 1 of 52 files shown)


src/app/screens/LNURLChannel/index.tsx
```diff
@@ -73,8 +76,12 @@ function LNURLChannel() {
         msg.reply(callbackResponse?.data);
       }
     } catch (e) {
-      console.error(e);
-      if (e instanceof Error) toast.error(`Error: ${e.message}`);
+      if (axios.isAxiosError(e)) {
```
reneaaron, line 79: Why only if it's an axios error? if e.g. api.getInfo() fails there is no error shown?

src/app/screens/LNURLChannel/index.tsx
```diff
@@ -115,6 +122,22 @@ function LNURLChannel() {
...
+                checked={isPrivateChannel}
+                onChange={(event) => {
+                  setIsPrivateChannel(event.target.checked);
+                }}
+              />
+              <label
+                htmlFor="is_private_channel"
+                className="cursor-pointer ml-2 block text-sm text-gray-900 font-medium dark:text-white"
+              >
+                Open a private channel?
```
reneaaron, line 138: Add to translations

src/app/screens/LNURLChannel/index.tsx
```diff
@@ -115,6 +122,22 @@ function LNURLChannel() {
               heading={`${t("content_message.heading")}:`}
               content={uri}
             />
+            <div className="flex">
+              <Checkbox
+                id="is_private_channel"
+                name="private channel checkbox"
```
reneaaron, line 128: Can we use a proper name here? (it's allowed and works, but to me this feels strange)

reneaaron:
```suggestion
                name="is_private_channel_checkbox"
```

reneaaron (changes requested):
Nice! :rocket:

pavanjoshi914:
sry i just quickly coded it. need to improve the pr a bit. thanks for the review addressing it

src/app/screens/LNURLChannel/index.tsx
```diff
@@ -74,7 +77,13 @@ function LNURLChannel() {
       }
     } catch (e) {
       console.error(e);
-      if (e instanceof Error) toast.error(`Error: ${e.message}`);
+      if (axios.isAxiosError(e)) {
+        const error =
+          (e.response?.data as { reason?: string })?.reason || e.message;
+        toast.error(`Error: ${error}`);
+      } else if (e instanceof Error) {
+        toast.error(`Error: ${e.message}`);
```
reneaaron, line 85: If we add this here we should be able to remove https://github.com/getAlby/lightning-browser-extension/pull/3046/files#diff-a082864a4adc3604b70d8b843b52fc353f4debd06482a72388a209bb6701e1f2R47-R49, right?

reneaaron (approved):
One little thing to change. 

Otherwise utACK

src/app/screens/LNURLChannel/index.tsx
pavanjoshi914: yup true! that's not needed it will eventually be catched and shown in toast. thanks!

pavanjoshi914:
will merge this. i think functionality is complete now. if there are further improvements we need to do. i can do a follow up pr

PR #3065, opened 2024-03-05
fix: styling and other issues on Apple Devices

im-adithya:
#### Common changes among iOS, iPad and Mac (Safari only)
Fixes blank screen issue on start page
Prevents Start Page(s) from being removed resulting in closing of Safari while locking the extension

#### iOS
Replaces w-96 with w-full

#### iPad
Removes max-w-full as that is disturbing the popup size

### Type of change

<p>

</p>

### How has this been tested?

Manually

### Checklist

- [x] Manual testing
- [x] Update Docs & Guides
- For UI-related changes
- [x] Darkmode
- [x] Responsive layout


src/app/screens/Home/index.tsx
```diff
@@ -62,12 +62,14 @@ const Home: FC = () => {
...
-        browser.tabs.sendMessage(tabs[0].id as number, {
-          action: "extractLightningData",
-        });
+        if (currentUrl.startsWith("http")) {
+          browser.tabs.sendMessage(tabs[0].id as number, {
+            action: "extractLightningData",
+          });
+        }
+      } else {
+        setLoadingAllowance(false);
```
reneaaron, line 72: Why do we need to set that explicitly? 

`loadAllowance` should only be triggered as the currentUrl is set? 

https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/Home/index.tsx#L46-L48

im-adithya: But in cases where there's no `currentUrl` to be set i.e. Start Page in Safari, we go into a state where the `useEffect` is not invoked and the `loadAllowance` is never called, resulting in `loadingAllowance` not changing at all (initially set to true, which means it's loading but actually it isn't as the currentUrl is already set (to `null`)) and showing blank screen.

src/app/router/Popup/index.tsx
```diff
@@ -9,6 +9,26 @@ import Popup from "./Popup";
...
+    navigator.userAgent.match(/Macintosh/i) &&
+    navigator.userAgent.match(/Safari/i) &&
+    navigator.maxTouchPoints &&
+    navigator.maxTouchPoints > 1;
+  if (isSafariOniOS) {
+    document.body.classList.remove("w-96");
+    document.body.classList.add("w-full");
+  }
+  if (isSafariOniPad) {
+    document.body.classList.remove("max-w-full");
```
reneaaron, line 27: Why remove the max-width? The width is still set to w-96 then, right?

src/app/screens/Home/index.tsx
reneaaron: I think that's somewhat strange as it's not at all obvious why this is being done. 

I'd probably try to get rid of the useEffect() and instead just call loadAllowance() in here. It's the only call, anyway and it makes things more obvious imho. (instead of having a hidden dependency make it explicit)

src/extension/background-script/state.ts
```diff
@@ -214,12 +214,14 @@ const state = create<State>((set, get) => ({
 
     // https://stackoverflow.com/a/54317362/1667461
     const allTabIds = Array.from(allTabs, (tab) => tab.id).filter(
-      (i): i is number => {
-        return typeof i === "number";
+      (id, index): id is number => {
+        // Safari: allTabs consist of Start Pages too
+        return typeof id === "number" && allTabs[index].title !== "";
```
reneaaron, line 219: Line 213 queries all tabs with the Title "Alby", however here we filter out tabs with empty titles? They shouldn't be included in this list, right?

im-adithya: Yeah that's something done for Safari, the start page tabs also get queried (which is weird as they have `title:""`) so here we are removing those again so we don't close the window on closing Alby in the start page (given there are no other tabs opened in Safari with proper url)

Safari is weird.

src/app/router/Popup/index.tsx
reneaaron: wtf :shrug:

src/app/screens/Home/index.tsx
```diff
@@ -62,13 +55,15 @@ const Home: FC = () => {
...
-          action: "extractLightningData",
-        });
+        if (currentUrl.startsWith("http")) {
+          browser.tabs.sendMessage(tabs[0].id as number, {
+            action: "extractLightningData",
+          });
+        }
       }
+
+      loadAllowance();
```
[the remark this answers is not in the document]
reneaaron: getCurrentURLAndLoadAllowance()? (not very creative, but hey :shrug: )

reneaaron (approved):
utACK

PR #3066, opened 2024-03-06
chore: bump Alby JS SDK to 3.4.0

rolznz:
This fixes error messages not being displayed from errors in the NWC connector due to Alby JS SDK previously not throwing proper error objects (therefore the `message` field being undefined and the error falling back to "Something went wrong")

image


pavanjoshi914 (approved):
maybe we shall avoid longer error messages. it shall be upto the client to add string like "something went wrong". extension scheme is simple "Error: error.message" where error.message shall indicate actual reason of error directly

PR #2653, opened 2023-08-11
feat: add NIP 44 encryption support

staab:
I've added added for NIP-44 encryption and decryption so that users can access nostr features built on the new primitives.

Fixes #2652

- `feat`: New feature (non-breaking change which adds functionality)

### How has this been tested?

No testing yet

### Checklist

- [x] Self-review of changed code
- [ ] Manual testing
- [x] Added automated tests where applicable
- [ ] Update Docs & Guides

PR #3075, opened 2024-03-11
NIP44 Encryption Support

pavanjoshi914:
1. NIP-44 encryption and decryption so that users can access nostr features built on the new primitives (derived from https://github.com/getAlby/lightning-browser-extension/pull/2653)
2. extract NIP44 into the extension itself until we update js-sdk to nostr 2.x


src/i18n/locales/en/translation.json
```diff
@@ -1176,18 +1176,21 @@
...
-    },
     "bitcoin": {
       "getaddress": "Read your Bitcoin receive address"
     },
     "liquid": {
       "getaddress": "Read your Liquid receive address",
       "signschnorr": "Sign message with your key"
     },
+    "nostr": {
+      "getpublickey": "Read your public key.",
```
reneaaron, line 1187: Remove those dots pls

reneaaron (approved):
Please fix the translations, otherwise: 

utACK

src/i18n/locales/en/translation.json
pavanjoshi914: resolved

src/extension/background-script/nostr/index.ts
```diff
@@ -69,6 +82,14 @@ class Nostr {
     return Utf8.stringify(decrypted);
   }
 
+  nip44Encrypt(peer: string, plaintext: string) {
```
rolznz, line 85: Can the existing encrypt and functions be prefixed with `nip04`?

rolznz (approved):
utACK

src/extension/background-script/nostr/index.ts
pavanjoshi914: resolved

alexgleason:
Hey y'all can you do a release please? Wanting to nip44 encrypt events but it only works if people can decrypt them. Thank you! :orange_heart:

pavanjoshi914:
Yeah its already present in the production!

Issue #2511, opened 2023-06-24
Add unit switching within the "Amount" input


stackingsaunter:
Currently there are two important missing features while creating an invoice that are often used and requested:

- creating invoiced priced in fiat currency
- being able to see/use units in BTC not sats

Let's add both being able to set an input by user in fiat and also change sats to BTC.
You can play how desired outcome should work in [this prototype](https://www.figma.com/proto/X45Mrk7OPlF5zu7NreyCxe/Input-units-swapping?page-id=4%3A2674&type=design&node-id=4-2697&viewport=2835%2C521%2C1&scaling=min-zoom&starting-point-node-id=4%3A2697&mode=design)

You can see the behaviour flow and hanoff [HERE](https://www.figma.com/file/X45Mrk7OPlF5zu7NreyCxe/Input-units-swapping?type=design&node-id=146%3A3781&mode=design&t=DepFvtEUY0dIMzkw-1). 

## Changes description:

1. **Add trailing users input "sats" or "BTC" unit:**
![image](https://github.com/getAlby/lightning-browser-extension/assets/68239231/eb404acc-0113-4f2c-bce0-a10994336028)

2. **Those units should be clickable and change to BTC (and vice versa)**
![image](https://github.com/getAlby/lightning-browser-extension/assets/68239231/0d2290a7-ff9b-4cb9-bbd8-79e744649134)

3. **Clicking on fiat unit should switch it's place with bitcoin units, thus making it possible for the user to create an invoice priced in fiat:**
![image](https://github.com/getAlby/lightning-browser-extension/assets/68239231/cf215b4d-976a-4294-abf8-2a9306c9fb4a)

4. **If fiat unit has a symbol, we show it in front of the number. If it uses 3 letters (USD, ARS, TRY, PLN etc) we should show it after the fiat amount:**
![image](https://github.com/getAlby/lightning-browser-extension/assets/68239231/275d3dde-0c86-4438-97d4-57cd733c93e2)

5. **All units have hover states** I also found one colouring inconsistency. Amount input in dark mode uses "gray" instead of "neutral" like all the rest units, so we could upgrade it to "neutral" with PR related to this issue^ 
[...]

PR #3089, opened 2024-03-16
Support fiat input in dual currency field

riccardobl:
This PR is an initial effort in dealing with https://github.com/getAlby/lightning-browser-extension/issues/2511 , in its current state it doesn't tweak the rendering of the input field but takes care only of the underlying changes needed to support the currency switch.

To make the implementation effective and reduce code duplication i've moved the switching and conversion logic under the DualCurrencyFieldChangeEvent component that now has an optional boolean  `showFiat` field instead of  `fiatValue` .

The `onChange` event is captured and transformed so that `event.target.value` is always a string with the value in satoshis (no matter which currency is used as input), `event.target` is also extended with the following custom fields
```ts
valueInFiat: number;
formattedValueInFiat: string;
valueInSats: number;
formattedValueInSats: string;
```
to provide all the transformed values needed throughout the various ui components.

DualCurrencyFieldChangeEvent is set to default to use fiat input whenever the account is using a fiat currency ( `account.currency !== "BTC"`), the manual switching is temporarily  binded to the onClick event of the "alt currency"  field 
Screenshot from 2024-03-16 15-08-02

Let me know what you think about this draft.


reneaaron:
Wow, thanks for tackling that right away! :100: 

> Let me know what you think about this draft, the proposed ux design is obviously not there yet, but i think it is better to get the logic sorted out first.

Generally speaking this is great! There might be some nitpicks around naming & event properties but I am well aware this is just a draft so I'll skip that for now. We'll be able to sort those things out as the PR progresses. 

So looking forward to finally be able to send 10 USD to a friend without having to lookup rates! :muscle:

riccardobl:
I propose to use the bitcoin symbol ₿ as prefix for bitcoin inputs (like for $). The input placeholder will still say "Amount in Satoshis..." so it will be clear that it is asking a bitcoin input denominated in sats.
Both the prefixed symbol and the "fiat amounts" will toggle the main currency  when clicked.

The reason to have a prefix symbol for sats inputs instead of the "sats" suffix proposed in https://github.com/getAlby/lightning-browser-extension/issues/2511 is to keep implementation simple and consistent, since as noted in the issue, implementing a cross-browser growing input field adds quite a lot of complexity.

I think this is overall a pretty good alternative, since it is clear, simple and makes use of the ₿ symbol.

I've changed the dual currency input to automatically set the parameterized "Amount in {{currency}}..." placeholder when no placeholder is passed to the component and i've adapted the code to not pass a placeholder.

Some screenshots:


Screenshot from 2024-03-23 14-51-58

Screenshot from 2024-03-23 14-52-12


Screenshot from 2024-03-23 14-51-44

Screenshot from 2024-03-23 14-51-52


Looking forwards to your feedback. Also pinging @stackingsaunter  for feedbacks on the design since ux is not my forte.

The latest commit contains all the aforementioned changes.

pavanjoshi914:
hey @riccardobl i did some testing. the receive screen is broken that is why tests are failing as well. the right sat amount is not getting reflected to the form change functions in receive screen. it always defaults to 0 can you please try and fix that?

riccardobl:
I think i fixed the issues, it seems to be working as expected now. However it think i'll need some help to get the unit tests updated or figure out why they fail

pavanjoshi914:
image
why i can't remove the zero from the set budget form? i have to type currently for eg. 0100 to set budget for 100 sats. if I press back space it throws error

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -38,10 +51,149 @@ export default function DualCurrencyField({
...
   const inputEl = useRef<HTMLInputElement>(null);
   const outerStyles =
     "rounded-md border border-gray-300 dark:border-gray-800 bg-white dark:bg-black transition duration-300";
 
+  const initialized = useRef(false);
+  const [useFiatAsMain, _setUseFiatAsMain] = useState(false);
+  const [altFormattedValue, setAltFormattedValue] = useState("");
+  const [minValue, setMinValue] = useState(min);
+  const [maxValue, setMaxValue] = useState(max);
+  const [inputValue, setInputValue] = useState(value || 0);
```
pavanjoshi914, line 71: this shall just be value?

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -38,10 +51,149 @@ export default function DualCurrencyField({
...
+        formattedSats,
+        valueInFiat,
+        formattedFiat,
+      };
+    },
+    [getCurrencyRate, getFormattedInCurrency, settings]
+  );
+
+  const setUseFiatAsMain = useCallback(
+    async (v: boolean) => {
```
pavanjoshi914, line 104: can you please change to meaningful variable name

pavanjoshi914:
@riccardobl have you tested things up? is it ready for another round of review?

riccardobl:
> @riccardobl have you tested things up? is it ready for another round of review?
Yes, it should be ready for review

stackingsaunter (changes requested):
Hey @riccardobl thanks for those changes! I have a few requests:

1. Behaviour is a bit inconsistent, because the second unit appears only when you click on fiat unit. I think both units should be visible all the time in the input.

2. If it's possible it would be great If units itself would have hover unit (and change the cursor state to "hand with a finger" one), because right now it's not easily discoverable by users they can click it (also because of point 1.)

image

3. I wouldn't show "~" in this case:
CleanShot 2024-05-21 at 14 17 57@2x
Since sats are much more granular than fiat, it's the actual price not approximation ^^

riccardobl:
Actually both units are visible
image
First one being ₿.

The tailing unit proposed in the design, as discussed before, is incredibly complex to implement in code as there is no standard component in html that allows tailing units.

I can make the other two changes.

stackingsaunter:
Hey, yeah both are visable and this is good, but when I enter the screen with input I see only one:

CleanShot 2024-05-24 at 16 13 00@2x


Only after I click on the unit the other appears:

CleanShot 2024-05-24 at 16 13 43@2x


Same happens on "send" screen. I don't know If this was intended or a bug?

riccardobl:
Oh ok, it was a bug. Should be fixed now, also i've made the other two changes.

> If it's possible it would be great If units itself would have hover unit (and change the cursor state to "hand with a finger" one), because right now it's not easily discoverable by users they can click it (also because of point 1.)

I used hover:text-neutral-400 on both, since they have the same styling and added the pointer cursor, is this is the correct way to adapt the original design?

riccardobl:
fixed

stackingsaunter (approved):
tACK

reneaaron:
@riccardobl Just tested it and that is an awesome improvement. :100: 

Can you take a look at the failing unit tests?

pavanjoshi914:
There's still and inconsistent behaviour. if i enter 0 and press backspace i can't remove the zero from there. also when entering lower amount via fiat. i can't enter for example i can enter 20 sats but not not 0.01 dollars in the fields

alsong regarding the tests. its weird that they fail. if i remove dual currency field from the screens. tests do work. can you check that ?  fixed the test for DualCurrency component itself

riccardobl:
> There's still and inconsistent behaviour. if i enter 0 and press backspace i can't remove the zero from there. also when entering lower amount via fiat. i can't enter for example i can enter 20 sats but not not 0.01 dollars in the fields

I can't reproduce these two issues, could you please double check and give me more info ?

pavanjoshi914:
@riccardobl 

1. in receive tab. when i enter amount and remove all numbers via backspace 0 still lies there and new numbers get entered before 0. i have to manually select whole number to cleanup the field
2. in site settings. if  you enter any amount and backspace 0 never vanishes. even after manually selection and trying to remove it + currency conversion seems broken i entered 0.11111 it automatically converted to 165 and it shows 165 dollars = 165 sats 
https://github.com/getAlby/lightning-browser-extension/assets/55848322/68012ae6-3688-47e1-b44c-23d19660d004

stackingsaunter:
@pavanjoshi914 @reneaaron could you review this? I think it's good to go

pavanjoshi914:
wow! great job! currency field is quite stable now.thanks for your valuable work :rocket:

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -1,22 +1,37 @@
...
   label: string;
   hint?: string;
   amountExceeded?: boolean;
   rangeExceeded?: boolean;
+  baseToAltRate?: number;
+  showFiat?: boolean;
+  onChange?: (e: DualCurrencyFieldChangeEvent) => void;
 };
 
 export default function DualCurrencyField({
```
pavanjoshi914, line 32: Now its a good enough complexity. maybe add comments wherere necessary why such and such thing is used

src/app/components/form/DualCurrencyField/index.tsx
```diff
@@ -53,19 +216,20 @@ export default function DualCurrencyField({
...
-      value={value}
+      value={inputValue ? inputValue : ""}
       autoFocus={autoFocus}
       autoComplete={autoComplete}
       disabled={disabled}
-      min={min}
-      max={max}
+      min={minValue}
+      max={maxValue}
+      step={useFiatAsMain ? "0.01" : "1"}
```
pavanjoshi914, line 232: what if we keep step value upto 4 decimal places. so that user can enter values for 10 20 sats in USD as well (it has no major use but just to keep it fully flexible)

pavanjoshi914 (approved):

src/app/components/form/DualCurrencyField/index.tsx
riccardobl: I've added some comments to the code and removed baseToAltRate since it was dead code from one of the previous iterations :+1:

pavanjoshi914:
unfortunately this is working when we input values in the input field itself. but we have this presets

but when we pass the value from outside for example this presets. dual currency field does nothing.

i noticed few things
1. value field in the component is not used anywhere. in value field such values are passed to the dualCurrencyCompontent. along with "inputValue" we also have to handle "value" here
2. values from presets are always in sats. so somehow we need to handle currency conversion + showing the converted value in dual currency field

image

src/app/screens/Keysend/index.tsx
```diff
@@ -46,15 +47,6 @@ function Keysend() {
       : +amountSat > (auth?.account?.balance || 0);
   const rangeExceeded = +amountSat < amountMin;
 
-  useEffect(() => {
```
pavanjoshi914, line 49: after passing value for such sats presets and converting them. if dualCurrencyField returns fiatValue then we don't need this useEffect. else we need to explicitly convert the value via such useEffect so that right fiatValue is shown in successCard

riccardobl:
This stems from the modified `DualCurrencyField` not using the `value` property internally, instead it uses the `inputValue ` state that is initialized with the value of the `value` property, so it won't react to changes of `value` even when passed as a react state.
This is all due to the fact that the `DualCurrencyField` presents itself always as a satoshi denominated input and so it has to perform the conversion internally and it needs to update the internal `inputValue`. 
There is also some code in alby that updates the `value` state passed to the `DualCurrencyField` with the value coming from the `DualCurrencyField` `onChange` callback, so a fix needs also to consider this to avoid a feedback loop.

I believe it is not an easy issue to solve, but i thought to it deeply and i've figured out the workaround in my last commit,  it is not the prettiest code, but it should work reliably.

stackingsaunter:
Could you review that workaround? @pavanjoshi914 and maybe @reneaaron or @bumi ?
This is constantly requested by many users

pavanjoshi914:
there are still few of the tests need to be fixed

PR #3092, opened 2024-03-19
fix: simplify styles & fix margins on homescreen

reneaaron:

Simplifying layouts & fixing margins on homescreen.

### DIFF
--- a/src/app/components/TransactionsTable/index.tsx
+++ b/src/app/components/TransactionsTable/index.tsx
@@ -37,13 +37,15 @@ export default function TransactionsTable({
   }
 
   return (
-    <div className="mt-4">
+    <div>
       {loading ? (
         <div className="w-full flex flex-col items-center">
           <Loading />
         </div>
-      ) : !transactions?.length && noResultMsg ? (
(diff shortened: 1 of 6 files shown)


src/app/components/TransactionsTable/index.tsx
```diff
@@ -37,13 +37,15 @@ export default function TransactionsTable({
...
+    <div>
       {loading ? (
         <div className="w-full flex flex-col items-center">
           <Loading />
         </div>
-      ) : !transactions?.length && noResultMsg ? (
-        <p className="text-gray-500 dark:text-neutral-400">{noResultMsg}</p>
+      ) : !transactions?.length ? (
+        <p className="text-gray-500 dark:text-neutral-400">
+          {t("no_transactions")}
```
bumi, line 47: we should remind ourself to always show something positive and maybe a call to action in such cases. 

empty states is the first thing the users sees and always somewhat sad.

reneaaron: I agree, but out of scope for this PR.

PR #3098, opened 2024-03-22
fix: account connection check

bumi:

so far we have checked if an account has an alias. but this is not very reliable.
This change now checks if we can get the balance from the account. if so the connection works

### Type of change

### How has this been tested?

manually, adding a NWC connection

### Checklist

- [x] Manual testing


src/app/screens/Onboard/TestConnection/index.tsx
```diff
@@ -29,7 +29,7 @@ export default function TestConnection() {
 
     try {
       const response = await api.getAccountInfo();
-      if (response.name && response.info.alias) {
+      if (response.name && response.balance?.balance) {
```
rolznz, line 32: if the balance is 0 this will return false

rolznz: should we also remove the name check?

reneaaron:
```suggestion
      if (response.name && response.balance?.balance !== 0) {
```

PR #3106, opened 2024-03-25
feat: add custom records in outgoing payments in lndhub

pavanjoshi914:
add custom records for outgoing payments for lndhub


reneaaron (approved):
utACK

PR #3124, opened 2024-04-05
fix: nostr permission inconsistencies

pavanjoshi914:

remove permission inconsistencies
1. add permission selectors and permisision modal to all nostr permissions
2. update prompt action logic for nostr permissions
3. remove block link from prompts
4. add deny permissions functionality 
5. update designs

followup: unity encrypt decrypt prompts

### Type of change

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


src/app/screens/Bitcoin/ConfirmSignPsbt/index.tsx
```diff
@@ -112,7 +112,7 @@ function ConfirmSignPsbt() {
             </div>
             <div className="flex w-full justify-center">
               <Hyperlink onClick={toggleShowAddresses}>
-                {showAddresses ? t("hide_details") : t("view_details")}
+                {showAddresses ? tCommon("details") : tCommon("details")}
```
reneaaron, line 115: Remove this conditional as it yields the same output in any case.

pavanjoshi914: resolved

src/app/screens/Nostr/ConfirmDecrypt.tsx
```diff
@@ -1,52 +1,64 @@
...
-    await msg.request("addBlocklist", {
-      domain: origin.domain,
-      host: origin.host,
-    });
-    alert(`Added ${origin.host} to the blocklist, please reload the website`);
-    msg.error(USER_REJECTED_ERROR);
+  function reject(e: React.MouseEvent<HTMLAnchorElement>) {
+    try {
+      setLoading(true);
+      msg.reply({
```
reneaaron, line 51: Possibly rejections have to be answered with msg.error(), that's probably why a signature is still delivered although user clicks "deny".

reneaaron (changes requested):

src/app/screens/Liquid/ConfirmSignPset.tsx
```diff
@@ -120,7 +120,7 @@ function ConfirmSignPset() {
             </div>
             <div className="flex w-full justify-center">
               <Hyperlink onClick={toggleShowAddresses}>
-                {showDetails ? t("hide_details") : t("view_details")}
+                {showDetails ? tCommon("details") : tCommon("details")}
```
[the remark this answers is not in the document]
reneaaron: Do we need to clean up the translations or are they used somewhere else?

src/app/screens/Bitcoin/ConfirmSignPsbt/index.tsx
pavanjoshi914: yes as you can see onClick  toggleShowAddress is triggered which decides popup is shown or not

src/app/screens/Liquid/ConfirmSignPset.tsx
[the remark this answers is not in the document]
pavanjoshi914: i replaced all the instances of view and hide details. also cleaned up other translation files. resolved

src/extension/background-script/actions/nostr/signEventOrPrompt.ts
```diff
@@ -74,6 +81,16 @@ const signEventOrPrompt = async (message: MessageSignEvent, sender: Sender) => {
           await addPermissionFor(permission, host, promptResponse.data.blocked);
         });
       }
+
+      if (promptResponse.data.confirm) {
```
reneaaron, line 85: Why the same code again? (see below)

pavanjoshi914: i structured it the same as other prompts are strctured

if(hasPermission){
 calculate and return directly
}else{
	open prompt
	
	if prompt response is positive
		calculate and return directly
}

yup it can be improved but its in other prompts as well and can be fixed for all in one go as a separate pr. is it okay?

reneaaron: Couldn't you just check for the confirm property and return if rejected? To me it seems the same code is being executed?

pavanjoshi914: when user cancels the permission they have option to store it as blocked permission. so we have to use same msg.reply() as we do for confirmed permissions. 

before the error was thrown when permission is cancelled from the screen itself. but with negative permissions and its permissions we need to pass aditional data.

so throwing errors and handling things are done in the action directly with the above mentioned approach.

pavanjoshi914:
tAck

PR #3134, opened 2024-04-25
feat: no limit warning

pavanjoshi914:

add warning for custodial users having no limit

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/app/screens/Home/DefaultView/index.tsx
+++ b/src/app/screens/Home/DefaultView/index.tsx
@@ -5,6 +5,7 @@ import {
   PopiconsArrowDownLine,
   PopiconsArrowRightLine,
   PopiconsBulbLine,
+  PopiconsCircleExclamationLine,
   PopiconsKeyLine,
 } from "@popicons/react";
 import dayjs from "dayjs";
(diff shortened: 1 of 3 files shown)


src/extension/background-script/connectors/alby.ts
```diff
@@ -122,17 +140,22 @@ export default class Alby implements Connector {
...
+    const limit = await this.getLimit();
+
+    if (cacheValue && cacheValue.data.limit === limit) {
       return cacheValue;
     }
 
     try {
       const info = await this._request((client) =>
         client.accountInformation({})
       );
```
bumi, line 153: can we use the new endpoint instead of this one? it should contain the same information and the additional account limit info. 

so we don't need to do a second request.

src/extension/background-script/connectors/alby.ts
```diff
@@ -406,4 +429,50 @@ export default class Alby implements Connector {
...
+
+    try {
+      const details = await this._genericRequest<UserDetails>(
+        url,
+        requestOptions
+      );
+      const limits = details.limits;
+      const custodial = details.custodial;
+
+      const hasZeroLimit = Object.values(limits).some((limit) => limit === 0);
```
bumi, line 454: we can do this on the server side.

pavanjoshi914: resolved

src/app/screens/Home/DefaultView/index.tsx
```diff
@@ -139,7 +141,27 @@ const DefaultView: FC<Props> = (props) => {
           </div>
         )}
 
-        <BalanceBox />
+        {accountLoading || nodeRequired ? (
```
reneaaron, line 144: Why show this hint during loading? :grimacing:

pavanjoshi914:
tAck

PR #3140, opened 2024-05-02
feat: communicate users to upgrade their old lndhub account

pavanjoshi914:
pr to show banner to upgrade alby lndhub account to alby oauth account


pavanjoshi914:
tAck

PR #3143, opened 2024-05-13
feat: unchain unlock call from enable screen

pavanjoshi914:

unchain unlock screen from enable screen

fixes triggering of enable screen for nostr when extension is locked, even if the provider was persisted before.
this also reduces the side effect of user needed to set permission again because of retriggering of enable screens

### Type of change

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout


stackingsaunter (approved):
tACK

src/app/router/Prompt/Prompt.tsx
```diff
@@ -60,6 +60,16 @@ function Prompt() {
     <Providers>
       <HashRouter>
         <Routes>
+          <Route
+            index
```
rolznz, line 64: Shouldn't this be path="/"? See https://reactrouter.com/en/main/route/route#index

Now there are two path elements on the same route, I'm not sure this is a good idea?

pavanjoshi914: https://reactrouter.com/en/main/start/concepts#index-routes

afaik that route navigates user to appropriate screen after screen is unlocked. eg. getPublicKey. before that route was inside <Route path="/"> which has RequireAuth guard. 

as for nostr we handle unlock screen explicitly and to avoid two popups for unlock screen. i kept it outside the main route ("/").  so yet it will work same as before. along with routes that don't require Auth guard

src/app/screens/Unlock/index.tsx
```diff
@@ -45,6 +47,7 @@ function Unlock() {
     setLoading(true);
     auth
       .unlock(password, () => {
+        if (navState.action === "unlock") msg.reply({ unlocked: true });
```
reneaaron, line 50: Shouldn't we return here after sending msg.reply?

src/extension/background-script/actions/nostr/enable.ts
```diff
@@ -25,8 +25,22 @@ const enable = async (message: MessageAllowanceEnable, sender: Sender) => {
 
   const enabledFor = new Set(allowance?.enabledFor);
 
+  if (!isUnlocked) {
```
reneaaron, line 28: Don't we have the same problem also for other providers? Should we change this behavior there as well?

src/app/screens/Unlock/index.tsx
pavanjoshi914: why return? unlock screen is open via openPrompt screen. when we msg.reply call is already completed and prompt is closed

src/extension/background-script/actions/nostr/enable.ts
pavanjoshi914: extended it for other providers

src/app/screens/Unlock/index.tsx
reneaaron: I just think it's a bit cleaner code-wise. Otherwise one could think the other code is still being executed (navigate(), setLoading())...


```suggestion
        if (navState.action === "unlock") {
            msg.reply({ unlocked: true });
            return;
        }
```

reneaaron:
Tested and worked like intended. :+1: 

There is still a bug when you manually lock the extension after you enabled it on this website before, which leads to an erroneous state in the extension (you need to manually reload it afterwards in order to use it again). This bug was already in there for years though, so I don't think we need to fix that in this PR.  

tACK

src/extension/background-script/actions/nostr/enable.ts
```diff
@@ -25,8 +25,22 @@ const enable = async (message: MessageAllowanceEnable, sender: Sender) => {
...
+    } catch (e) {
+      console.error(e);
+      if (e instanceof Error) {
+        return { error: e.message };
+      }
+    }
+  }
+
   if (
-    isUnlocked &&
```
bumi, line 29: don't we still need the unlock check? because if your new check fails for whatever reason then it's not unlocked and this action should not run.

src/app/screens/Unlock/index.tsx
pavanjoshi914: resolved

bumi (approved):

PR #3168, opened 2024-06-04
feat: account menu makeover

### LINKED ISSUE #3167: [Feature] Show active wallet in the wallet selection list
### Feature description

I find not having active wallet in your wallet selection list is confusing to users.

Even more confusing when you are newbie, create another wallet (or link the same Alby Account to new profile) and then see only one with the same name on the selection list - you might not even be aware that it is another wallet and are surprised why did your Nostr profile disappear :face_with_head_bandage:

### Describe the solution

It would be helpful to catch people on the idea that the extension can keep more than one wallet and to make clear which wallet is being used, to put it on the list and indicate that it's active.

image

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

reneaaron:

Fixes #3167 

**TODO**
 - [x] Fix colors
 - [x] Darkmode
 - [x] Unit tests
 - [x] Cleanup translations

### DIFF
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -2,13 +2,12 @@ import {
   PopiconsChevronBottomLine,
   PopiconsCirclePlusLine,
   PopiconsGlobeLine,
-  PopiconsWalletLine,
+  PopiconsSettingsMinimalLine,
 } from "@popicons/react";
 import { useEffect } from "react";
 import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import Avatar from "~/app/components/Avatar";
-import MenuDivider from "~/app/components/Menu/MenuDivider";
 import SkeletonLoader from "~/app/components/SkeletonLoader";
 import { useAccount } from "~/app/context/AccountContext";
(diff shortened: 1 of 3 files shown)


src/app/components/AccountMenu/index.tsx
```diff
@@ -92,62 +91,63 @@ function AccountMenu({ showOptions = true }: Props) {
...
+                        {accountLoading ? (
+                          <SkeletonLoader className="w-16" />
+                        ) : (
+                          balancesDecorated.accountBalance
+                        )}
+                      </span>
+                    </div>
+                  </div>
+                  <div className="flex flex-row gap-1 items-center">
+                    {(isAlbyLNDHubAccount(
```
pavanjoshi914, line 120: is it always true that albylndhub account users can manage their account on getalby.com. for eg. old lndhub users who just used signup form in extension we had before and got started with it.

pavanjoshi914 (approved):
tAck

src/app/components/AccountMenu/index.tsx
pavanjoshi914: i will merge this pr for now

pavanjoshi914:
tAck

PR #3171, opened 2024-06-12
feat: handle longer descriptions + buttons always visible in viewport

pavanjoshi914:
fix contentMessage component for longer descriptions
make sure buttons always visible in viewport

before: 
image

after: 
image


pavanjoshi914:
tAck

PR #3172, opened 2024-06-14
feat: add custom records to NWC connector

rolznz:
Adds custom records (mainly used for podcasting payments but could be other things) to the NWC connector.

There is no standard on how these should be passed. In the NIP-47 spec metadata can be returned, so I think it makes sense to keep the custom_records (custom TLVs) in a separate object inside the metadata object.

https://github.com/nostr-protocol/nips/blob/master/47.md

(I will link the NWC PR when it's ready)

Related: https://github.com/getAlby/nostr-wallet-connect-next/issues/444
Related: https://github.com/getAlby/nostr-wallet-connect/pull/251

image
image


reneaaron:
Great! Just one little thing I noticed in your screenshot? 

Is the encoding somewhat wrong here? (see App metadata)

rolznz:
@reneaaron it's the :zap: emoji. It seems to also be encoded wrongly using the LNDHub connector.

I also saw this check is wrong (it's checking the message but using the podcast) and I can't actually see where the message field is being used

```
{transaction.boostagram?.message && (
            <TransactionDetailRow
              title={t("boostagram.message")}
              content={transaction.boostagram?.podcast}
            />
          )}
```

reneaaron:
@rolznz I'll try to debug it :+1:

src/common/lib/string.ts
```diff
@@ -0,0 +1,10 @@
+// https://stackoverflow.com/questions/30106476/using-javascripts-atob-to-decode-base64-doesnt-properly-decode-utf-8-strings
```
bumi, line 1: 🙀

src/extension/background-script/connectors/nwc.ts
```diff
@@ -205,4 +217,21 @@ class NWCConnector implements Connector {
   }
 }
 
+function mapTLVRecords(
+  tlvRecords: TLVRecord[] | undefined
+): ConnectorTransaction["custom_records"] | undefined {
+  if (tlvRecords) {
```
bumi, line 223: if(!tlvRecords) { return; } 

wrapping the whole function body is not needed then.

PR #3180, opened 2024-06-25
feat: show lnaddress only when oauth  account settings are accessed

pavanjoshi914:

show lnaddress only when oauth account settings are accessed

### Checklist

- [X] Manual testing
- [X] Update Docs & Guides
- For UI-related changes
- [X] Darkmode
- [X] Responsive layout

### DIFF
--- a/src/app/screens/Accounts/Detail/index.tsx
+++ b/src/app/screens/Accounts/Detail/index.tsx
@@ -26,7 +26,7 @@ import Toggle from "~/app/components/form/Toggle";
 import { useAccount } from "~/app/context/AccountContext";
 import { useAccounts } from "~/app/context/AccountsContext";
 import { useSettings } from "~/app/context/SettingsContext";
-import { classNames } from "~/app/utils";
+import { classNames, isAlbyOAuthAccount } from "~/app/utils";
 import api, { GetAccountRes } from "~/common/lib/api";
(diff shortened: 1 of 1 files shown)


src/app/screens/Accounts/Detail/index.tsx
```diff
@@ -238,41 +237,38 @@ function AccountDetail() {
...
-                    <div className="flex flex-col sm:flex-row justify-between items-center">
-                      <div className="sm:w-9/12 w-full">
-                        <p className="text-gray-800 dark:text-white font-medium">
-                          {t("lnaddress.title")}
-                        </p>
-                        <p className="text-gray-600 text-sm dark:text-neutral-400">
-                          {lightningAddress}
-                        </p>
-                      </div>
+              {lightningAddress && account.connectorType == "alby" && (
```
bumi, line 240: can you add a comment above where we load the lightningAddress and here.

https://github.com/getAlby/lightning-browser-extension/pull/3180/files#diff-d7095991beda5bd1778ee83a05d8ba506d7585e20906a072039434163058722eL56

that describes that the lightningAddress is loaded from the currently active account.

and your check actually also will not work. because if I have two alby accounts in there it will show the wrong ln address.

bumi (changes requested):

pavanjoshi914:
tAck

PR #3194, opened 2024-07-15
fix: correct total amount in notifcations

### LINKED ISSUE #3192: [BUG] Browser notifications miscalculate

### Describe the bug

I'm sending from Alby to other wallets, amount X sats.
It shows on a transaction history that I paid X, but it actually charged me X plus FEES. 
Ok.

However, in Google Chrome notifications (they appear in system notifications) it displays that I paid amount X **minus** FEES, and that I also paid those fees (which sums to X, not to X + fees), which is incorrect.

Compare below last two transactions
left: Alby (extension and web app both display same amounts)
right: Chrome (Windows) notifications 
image

-----------
Btw, when I'm sending from another node (here: Blink),
then on transaction history it shows I've spent X+FEES (which is what was actually charged), and the Chrome notification shows "Successfully paid X".

### Screenshots [optional]

_No response_

### Steps To Reproduce

1. Enable notifications for Alby extension in your browser
2. Send 1000 sats to anoter node/wallet
3. Compare transaction records in Alby web or extension, with records in these notifications.

### Expected behavior

We don't tell users about fees in our interfaces, simply charge them from their balance. 
We tell them only (?) in these notifications - but it's incorrect.

### Alby information

3.8.1

### Device information

Windows, Chrome

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:
show totalamount including fees in the notifcations as we do in the transactions table Fixes #3192


src/extension/background-script/events/notifications.ts
```diff
@@ -22,7 +22,7 @@ const paymentSuccessNotification = async (
 
   const route = paymentResponseData?.data.route;
   const { total_amt, total_fees } = route;
-  const paymentAmount = total_amt - total_fees;
+  const paymentAmount = total_amt;
```
bumi, line 25: I am not sure if this is correct for all the connectors? 
is that correct for LND, NWC, alby?

pavanjoshi914: alby to lndhub -correct
lndhub to alby - fees is always zero. total amount is correct no issues
lnd to lnd (testnet) - correct 
nwc to alby oauth - totalAmount correct, fees shown 0
alby oauth to nwc - totalAmount correct, fees shown 8 sats

pavanjoshi914: for all the cases totalAmount is correct which is the amount that user paid. but in some cases even if fees is charged fees is not returned in the route which maybe corrected on server side

bumi: So I think LND returns the total_amt with the fees (which is why we deduct it here)
but all others return the paid amount without the fee.

which means your change is OK. but we might need to change the LND connector to return the sent amount.

PR #3198, opened 2024-07-16
youtube video battery

pavanjoshi914:
fix twitter and youtube video battries


src/extension/content-script/batteries/YouTubeVideo.ts
```diff
@@ -2,64 +2,89 @@ import getOriginData from "../originData";
...
-  }
-  // if there is no lnurl we check for a zap emoji with a lightning address
-  // we check for the @-sign to try to limit the possibility to match some invalid text (e.g. random emoji usage)
-  else if ((match = findLightningAddressInText(text))) {
-    lnurl = match;
-  } else {
-    // load the about page to check for a lightning address
-    const match = channelLink.href.match(
-      /^https:\/\/www\.youtube.com\/(((channel|c)\/([^/]+))|(@[^/]+)).*/
+  async function youtubeDOMChanged(
```
bumi, line 14: did we use some mutation observer before?

pavanjoshi914: yes we use it for twitter battery. as its required for twitter as well

src/extension/content-script/batteries/Twitter.ts
```diff
@@ -7,7 +7,7 @@ declare global {
   }
 }
 
-const urlMatcher = /^https:\/\/(mobile.)?twitter\.com\/(\w+).*/;
+const urlMatcher = /^https:\/\/(mobile.)?x\.com\/(\w+).*/;
```
bumi, line 10: this breaks things now on twitter.com 
it must support both

pavanjoshi914: how twitter.com is still accessible? resolved!

src/extension/content-script/batteries/YouTubeVideo.ts
```diff
@@ -2,64 +2,89 @@ import getOriginData from "../originData";
...
+    // we check for the @-sign to try to limit the possibility to match some invalid text (e.g. random emoji usage)
+    else if ((match = findLightningAddressInText(text))) {
+      lnurl = match;
+    } else {
+      // load the about page to check for a lightning address
+      const match = channelLink.href.match(
+        /^https:\/\/www\.youtube.com\/(((channel|c)\/([^/]+))|(@[^/]+)).*/
+      );
+      if (match) {
+        lnurl = await findLnurlFromYouTubeAboutPage(match[1]);
```
bumi, line 48: we should not do this on each mutation. I assume this will now cause a lot of requests.

src/extension/content-script/batteries/Twitter.ts
bumi: I am always on twitter.com

PR #3199, opened 2024-07-16
fix: scrollbar on extension popup

pavanjoshi914:
fix scrollbar and shift of view in popup for latest chromium browsers

before: 
image

now: 
image

### DIFF
--- a/static/views/popup.html
+++ b/static/views/popup.html
@@ -13,6 +13,7 @@
 
   <body
     class="w-96 max-w-full overflow-x-hidden h-[600px] no-scrollbar min-h-screen bg-gray-50 dark:bg-surface-00dp slashed-zero"
+    style="scrollbar-width: auto"
   >
     <div id="popup-root" class="h-full"></div>
   </body>


reneaaron:
This change effectively removes the scrollbar completely for chromium browsers? I'm not sure we want that as there is no indication anymore that you can scroll the view at all.  

According to the [docs](https://developer.chrome.com/docs/css-ui/scrollbar-styling#classic_and_overlay_scrollbars) `auto` should be the default anyway :thinking:

bitcoinuser:
> This change effectively removes the scrollbar completely for chromium browsers? I'm not sure we want that as there is no indication anymore that you can scroll the view at all.
> 
> According to the [docs](https://developer.chrome.com/docs/css-ui/scrollbar-styling#classic_and_overlay_scrollbars) `auto` should be the default anyway 🤔

I prefer no scrollbar.

pavanjoshi914:
i tested it on chrome + brave latest version. it removes scrollbar only from popup where we don't want any kind of scrollbar right? it won't effect prompt and options view

reneaaron:
> i tested it on chrome + brave latest version. it removes scrollbar only from popup where we don't want any kind of scrollbar right? it won't effect prompt and options view

Yes, but also in the popup there is no indication that you can actually scroll. On Firefox there still is one, and I wonder if it's a good practice to just hide the scrollbar completely. 

image

pavanjoshi914:
i would prefer not having scrollbar in the context of the popups.  by default we use no-scrollbar for popup's but it doesn't have any effect on latest browser. so I added additional rule to make it work. let me know if we shall go with this pr

PR #3211, opened 2024-07-25
fix: tailwindcss-3d plugin breaks scale animations

dylancom:

The menu's (such as UserMenu / AccountMenu) now feel "laggy" as there is no animation on the scaling. It jumps weirdly. I found out that tailwindcss-3d adjusts many classes and breaks the "scaling" transitions.

fix: remove the tailwindcss-3d plugin and create a custom plugin that provides the class that was actually used from the plugin: translate-z-0.

### Type of change

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,6 @@
     "puppeteer": "^21.4.1",
     "stream-browserify": "^3.0.0",
     "swc-loader": "^0.2.3",
-    "tailwindcss-3d": "^1.0.2",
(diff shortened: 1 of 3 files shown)


reneaaron (approved):
@dylancom Great catch, thanks for the PR! 

tACK

cc @pavanjoshi914

pavanjoshi914:
Thanks for the PR! tAck

PR #3220, opened 2024-08-03
Added note to SETUP.md

eesazahed:

I added a note to the top of SETUP.md. The note informs the reader that while the tutorial uses Yarn as the package manager, they can substitute it with other preferred managers like npm or pnpm, and should also update the package.json file accordingly.

### Type of change

### How has this been tested?

Tested locally with successful results.

### Checklist

- [ ] Manual testing
- [ x] Update Docs & Guides
- For UI-related changes
- [ ] Darkmode
- [ ] Responsive layout

### DIFF
--- a/doc/SETUP.md
+++ b/doc/SETUP.md
@@ -4,6 +4,8 @@
 
 > For Windows users, please use [WSL 2](https://learn.microsoft.com/en-us/windows/wsl/install).
 
+> **NOTE:** this tutorial uses `yarn`. If you prefer a different package manager, like `npm` or `pnpm`, you can replace Yarn with your preferred tool in all instructions. Just be sure to make the same changes in your local `package.json` file while in development.
+
 - Install dependencies\
   `$ yarn install`
 


reneaaron:
I am not sure this PR is actually solving a problem: 

If you are using this repository for the first time (and you have no knowledge about other package managers) you should probably stick with yarn. If you are already using other package managers I suppose you should also know how to use your preferred package manager instead. 

> Just be sure to make the same changes in your local `package.json` file while in development.

I also don't understand this hint. Yarn & npm both use package.json?

eesazahed:
So the problem I am trying to solve is that some new/inexperienced devs might not use yarn. They might only have npm installed. So I added that note at the top.

> I also don't understand this hint. Yarn & npm both use package.json?

In package.json, you can see `"build": "yarn build:chrome && yarn build:firefox && yarn build:opera",`

The note is to remind the reader that they can change it to `"build": "npm build:chrome && npm build:firefox && npm build:opera"` in addition to all of the other use cases of yarn in the file.

reneaaron:
I think unexperienced devs should just install yarn instead of manually editing the package.json file:

`npm install --global yarn`

eesazahed:
I get what you're saying. Final decision is up to you

pavanjoshi914:
I think the current version of documentation is consistent with yarn. and its not a greater side of usecase for anyone to suggest alternate package manager in docs. closing this

PR #3239, opened 2024-09-05
feat: add identifier in auth url

pavanjoshi914:
add identifier to the auth url. 
persist in account config
next time the refresh token expires, and user is logged in into different account on getalby.com we avoid weird behaviour and ask user to login again (potentially the correct account)

### DIFF
--- a/src/extension/background-script/connectors/alby.ts
+++ b/src/extension/background-script/connectors/alby.ts
@@ -34,6 +34,7 @@ interface Config {
   password: string;
   url: string;
   oAuthToken: OAuthToken | undefined;
+  userIdentifer: string | undefined;
 }
 
 export default class Alby implements Connector {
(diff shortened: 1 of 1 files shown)


src/extension/background-script/connectors/alby.ts
```diff
@@ -64,6 +65,25 @@ export default class Alby implements Connector {
...
+        const configData = decryptData(
+          accounts[this.account.id].config,
+          password
+        );
+        configData.userIdentifer = userIdentifier;
+        accounts[this.account.id].config = encryptData(configData, password);
+        state.setState({ accounts });
+        // make sure we immediately persist the updated accounts
+        await state.getState().saveToStorage();
+      }
```
bumi, line 83: we repeat this a few times it seems, can we put this into it's own function?

pavanjoshi914: here we 
1. decrypt
2. add or delete something in account config
3 encrypt it again. 

2nd step can always be anything that is why I didn't refactored it much. also currently we need userIdentifier only for alby connector so didn't make it universal. 

can we avoid doing any change here?

pavanjoshi914:
tACk

PR #3249, opened 2024-10-02
feat: setup your keys -> setup master key in default view cards

pavanjoshi914:
update setup keys card copy


pavanjoshi914:
tAck

PR #3253, opened 2024-10-16
fix: icon & copy for adding new wallets

reneaaron:

Some users seem not to be aware of the multi-account capabilities. 

 - Replaces circled plus icon with just the plus icon for for better recognizability
 - Replaced "Connect a Wallet" with just "Add"

image


bumi:
+1 what do we do with the translations?

reneaaron:
I removed existing translations and fixed a test.

pavanjoshi914:
tACk. thanks for updating other translation files as well!

PR #3254, opened 2024-10-16
fix: reload npub when account has been changed

reneaaron:

Reloads the npub when the account has changed.

### DIFF
--- a/src/app/screens/Home/DefaultView/index.tsx
+++ b/src/app/screens/Home/DefaultView/index.tsx
@@ -79,13 +79,13 @@ const DefaultView: FC<Props> = (props) => {
   useEffect(() => {
     (async () => {
       try {
-        const account = await api.getAccount();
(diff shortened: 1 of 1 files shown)


src/app/screens/Home/DefaultView/index.tsx
```diff
@@ -91,7 +91,7 @@ const DefaultView: FC<Props> = (props) => {
         if (e instanceof Error) toast.error(`Error: ${e.message}`);
       }
     })();
-  }, []);
+  }, [account]);
```
bumi, line 94: should that account variable in line 82 be called differently to not overlay the variable?

pavanjoshi914: it won't effect in any way as useEffect doesn't take variables as dependency defined within useEffect itself. 

but still i changed the name of the local variable

pavanjoshi914:
tAck

PR #3259, opened 2024-10-22
fix: remove translations for product names

reneaaron:

Removes translations of product names. Keys should still be translatable as for some cases it makes sense, e.g. `LND via LNC` --> The word "via" is translated)

image


pavanjoshi914:
this pr removes only those which are basically brand names and don't need to be translated

PR #3267, opened 2024-11-11
feat: update node_required copy

pavanjoshi914:
update node required copy

image


src/i18n/locales/en/translation.json
```diff
@@ -1126,6 +1124,7 @@
     "website": "Website",
     "wallet_settings": "Wallet Settings",
     "apps": "Apps",
+    "node_required": "Complete your wallet setup to start sending and receiving unlimited payments <0>here</0>.Until setup is complete, you can receive up to 50k sats as <1>fee credits</1> only.",
```
reneaaron, line 1127:
```suggestion
    "node_required": "Finish setting up your wallet to start sending and receiving unlimited payments <0>here</0>. Until then, you can receive up to 50,000 sats as <1>fee credits</1> only.",
```

PR #3276, opened 2024-11-21
feat: \using fee credits

pavanjoshi914:
image

shall be merged after https://github.com/getAlby/getalby.com/pull/1847 is deployed


pavanjoshi914:
again. we don't need using_discontinued_wallet banner for the users before march? this will be not shown to them right?

PR #3292, opened 2024-12-17
feat: show pending and failed transactions for nwc

### LINKED ISSUE #3291: [Feature] Show pending and failed transactions from the Hub
### Feature description

See https://github.com/getAlby/go/pull/208 and https://github.com/getAlby/hub/pull/835

### Describe the solution

Get failed and pending transactions via NWC and show them in the extension

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

No

### LINKED ISSUE #3299: Update transaction icon colors
I did slightly changes to the colors, darkmode now uses a bit different colors that look a bit nicer.
Here's a table with updated colors:
CleanShot 2025-01-08 at 17 49 46@2x

pavanjoshi914:
show failed and pending transactions for nwc

Fixes #3291 
Fixes #3299 
image
image
image
image
image
image
image
image


src/app/components/TransactionsTable/index.tsx
```diff
@@ -60,9 +64,19 @@ export default function TransactionsTable({
                 <div className="flex gap-3">
                   <div className="flex items-center">
                     {type == "outgoing" ? (
-                      <div className="flex justify-center items-center bg-orange-100 dark:bg-orange-950 rounded-full w-8 h-8">
-                        <PopiconsArrowUpSolid className="w-5 h-5 text-orange-400 dark:text-amber-600 stroke-[1px] stroke-orange-400 dark:stroke-amber-600" />
-                      </div>
+                      tx.state === "pending" ? (
+                        <div className="flex justify-center items-center bg-orange-100 dark:bg-orange-950 rounded-full w-8 h-8">
```
rolznz, line 68: Can you check the styles in Alby Go? it uses a pulse animation rather than orange

src/common/utils/helpers.ts
```diff
@@ -68,7 +68,7 @@ export function mergeTransactions(
   payments: ConnectorTransaction[]
 ): ConnectorTransaction[] {
   const mergedTransactions = [...invoices, ...payments].sort((a, b) => {
-    return b.settleDate - a.settleDate;
+    return (b.settleDate ?? Date.now()) - (a.settleDate ?? Date.now());
```
rolznz, line 71: why using Date.now?

src/extension/background-script/actions/ln/getTransactions.ts
```diff
@@ -9,19 +9,22 @@ const getTransactions = async (message: MessageGetTransactions) => {
...
-        const boostagram = utils.getBoostagramFromInvoiceCustomRecords(
-          transaction.custom_records
-        );
-        return {
-          ...transaction,
-          boostagram,
-          paymentHash: transaction.payment_hash,
-        };
-      });
+    const isNWC = connector.connectorType == "nwc";
```
rolznz, line 12: could this settled filtering be done at the connector level for other connectors to remove this conditional?

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -31,10 +31,12 @@ export interface ConnectorTransaction {
   /**
    * Settle date in UNIX milliseconds
    */
-  settleDate: number;
+  settleDate?: number;
+  creationDate?: number;
   totalAmount: number;
   displayAmount?: [number, ACCOUNT_CURRENCIES];
   type: "received" | "sent";
+  state?: string;
```
rolznz, line 39: can you use typing here for the possible states

src/common/utils/helpers.ts
pavanjoshi914: i made settleDate optional as for failed and pending its null. so i have to add Date.now just to make ts happy for other connectors

src/extension/background-script/actions/ln/getTransactions.ts
pavanjoshi914: changes in filtering logic are rare but if we wanna change something in future for this part of code we would have to reitrate over all the connectors. so i thought conditional is much more better than spreading filtering across all the connectors

src/app/components/TransactionsTable/index.tsx
pavanjoshi914: added animated pulse animation. for outgoin transactioons we use bg-orange. so i am using orange bg for sent as well as sending. sending have pulse animation now by default

for failed i changed things to red

src/common/utils/helpers.ts
pavanjoshi914: i was a bit wrong here. instead of making it optional it shall be number | null now. updated code accordingly

src/extension/background-script/actions/ln/getTransactions.ts
pavanjoshi914: moved it to connector level

src/extension/background-script/connectors/connector.interface.ts
pavanjoshi914: resolved

src/app/components/TransactionsTable/TransactionModal.tsx
```diff
@@ -56,17 +57,43 @@ export default function TransactionModal({
         <div>
           <div className="flex items-center justify-center">
             {getTransactionType(transaction) == "outgoing" ? (
-              <div className="flex justify-center items-center bg-orange-100 dark:bg-orange-950 rounded-full p-3">
-                <PopiconsArrowUpSolid className="w-10 h-10 text-orange-400 dark:text-amber-600 stroke-[1px] stroke-orange-400 dark:stroke-amber-600" />
-              </div>
+              transaction.state === "pending" ? (
+                <div className="flex justify-center items-center bg-orange-100 dark:bg-orange-950 rounded-full p-3 animate-pulse">
```
rolznz, line 61: @pavanjoshi914 can you please check the new colors in figma? pending should be blue

src/app/hooks/useTransactions.ts
```diff
@@ -16,13 +16,15 @@ export const useTransactions = () => {
...
         const transactions = getTransactionsResponse.transactions.map(
           (transaction) => ({
             ...transaction,
             title: transaction.memo,
-            date: dayjs(transaction.settleDate).fromNow(),
-            timestamp: transaction.settleDate,
+            date: dayjs(
+              transaction.settleDate || transaction.creationDate
+            ).fromNow(),
+            timestamp:
```
rolznz, line 26: why is this needed?

src/extension/background-script/actions/ln/getTransactions.ts
rolznz: I can still see the conditional? "isNWC"

src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -31,10 +31,12 @@ export interface ConnectorTransaction {
   /**
    * Settle date in UNIX milliseconds
    */
-  settleDate: number;
+  settleDate: number | null;
+  creationDate?: number;
```
rolznz, line 35: which connectors do not have a creation date? can't this be taken from the bolt11 invoice?

src/extension/background-script/connectors/lawallet.ts
```diff
@@ -128,7 +128,7 @@ export default class LaWallet implements Connector {
     return {
       data: {
         transactions: parsedTransactions.sort(
-          (a, b) => b.settleDate - a.settleDate
+          (a, b) => (b.settleDate ?? 0) - (a.settleDate ?? 0)
```
rolznz, line 131: I don't think this connector should be affected like this in this PR? if lawallet only has settled transactions, there is now extra code that does not even apply.

Can't you sort the transactions before mapping them?

src/extension/background-script/actions/ln/getTransactions.ts
pavanjoshi914: sry i forgot to push commit. resolved already

src/app/hooks/useTransactions.ts
pavanjoshi914: for pending transaction we don't have settleDate. so we show creationDate there. same is done in go app i think

src/extension/background-script/connectors/connector.interface.ts
pavanjoshi914: all other connectors only pass settled transactions and we don't even need creationDate for them except for nwc. so i kept it optional and passed this field only for nwc connector. maybe we can be consistent accross and pass it for all now

pavanjoshi914: for commando there is no creation date passed via api. https://docs.corelightning.org/reference/lightning-listinvoices

src/extension/background-script/connectors/lawallet.ts
pavanjoshi914: that was just to make typescript happy we have common type definitions for all the connectors. now settleDate can be either number | null (cause of nwc pending and failed transactions). any better approach here?

src/extension/background-script/connectors/connector.interface.ts
rolznz: it can be taken from the bolt11 invoice, can't it?

src/extension/background-script/connectors/lawallet.ts
rolznz: Can't you sort the transactions before mapping them?

src/app/hooks/useTransactions.ts
rolznz: to me the wording is strange, especially the "date" which is not a date

src/extension/background-script/connectors/connector.interface.ts
pavanjoshi914: resolved. added creationDate for all connectors.

src/extension/background-script/connectors/lawallet.ts
pavanjoshi914: how this will solve the problem of settledDate being  number | null. and avoid fallback value. its the type that is why we need fallback here. but it will never being triggered as all other transactions will returned settled transactions only

rolznz: because the settleDate cannot be null in the lawallet connector response. Only after you map it does it become nullable.

try sorting the `data` field.

src/extension/background-script/connectors/lawallet.ts
```diff
@@ -127,8 +129,8 @@ export default class LaWallet implements Connector {
 
     return {
       data: {
-        transactions: parsedTransactions.sort(
-          (a, b) => b.settleDate - a.settleDate
+        transactions: parsedTransactions.filter(
```
rolznz, line 132: why is this filter needed? it wasn't there before

rolznz: but I guess it's good to have

src/app/hooks/useTransactions.ts
```diff
@@ -16,13 +16,14 @@ export const useTransactions = () => {
...
           limit,
         });
-
         const transactions = getTransactionsResponse.transactions.map(
           (transaction) => ({
             ...transaction,
             title: transaction.memo,
-            date: dayjs(transaction.settleDate).fromNow(),
-            timestamp: transaction.settleDate,
+            date: dayjs(
```
rolznz, line 23: is it possible to rename this `date` field? to me it is not a date

src/extension/background-script/connectors/commando.ts
```diff
@@ -261,7 +272,7 @@ export default class Commando implements Connector {
     const transactions: ConnectorTransaction[] = mergeTransactions(
       incomingInvoicesResponse,
       outgoingInvoicesResponse
-    );
+    ).filter((transaction) => transaction.settled);
```
rolznz, line 275: is this needed?  because inside `mergeTransactions` there is also filtering? or should that be removed?

src/extension/background-script/connectors/lawallet.ts
pavanjoshi914: we moved filtering logic at connector level. filtering was present in getTransactions.ts background script. but for nwc we need to bypass thisf filter which require additional checks so we moved it to connector level

rolznz:
I tested the nwc connector - failed and pending states look good :raised_hands:

src/app/hooks/useTransactions.ts
pavanjoshi914: how about "timeAgo"  or "relativeDateTime" ?

src/extension/background-script/connectors/commando.ts
pavanjoshi914: i added filtering there as well to avoid fallback for settleDate. just a extra check to make typescript happy and function more secure.

src/app/hooks/useTransactions.ts
pavanjoshi914: changed to timeAgo

src/extension/background-script/connectors/commando.ts
rolznz: to me it seems a weird way to do it. Maybe we should remove the filtering in `mergeTransactions` as the function is about merging, not removing elements.

pavanjoshi914: resolved

pavanjoshi914:
tACk

PR #3306, opened 2025-02-05
feat: update depricated artifacts for workflows

pavanjoshi914:
upload and download artifacts for v3 are depricated needs to be uploaded to v4


reneaaron (approved):
utACK

PR #3308, opened 2025-02-19
feat: account-->wallet

pavanjoshi914:
rename account -->wallet


pavanjoshi914:
i already removed outdated translations still it showing warning of removing outdated translations. will debug it later

src/i18n/locales/en/translation.json
```diff
@@ -1116,7 +1107,7 @@
     "error": "Error",
     "general": "General",
     "settings": "Extension Settings",
-    "accounts": "Wallets",
+    "wallets": "Wallets",
```
reneaaron, line 1110: You changed the key here?

pavanjoshi914: aah it seems its not used anywhere. removed from all the translation files

pavanjoshi914: thinking to create flow to detect and remove unused keys for translations in separate pr

reneaaron: you can't because they are dynamically generated at some places, please don't invest any time and double check that this key really wasn't used :+1:

pavanjoshi914: yeah i double checked with search bar and if we remove key that is in used "yarn tsc:compile" will fail so yeah this is not being used anywhere

reneaaron: > "yarn tsc:compile" will fail so yeah

not if keys are dynamically generated (which I don't think is the case here)

pavanjoshi914: treid git blame as well. it was used in userMenu before. we don't put accounts in userMenu anymore. so we can safely remove this translations

https://github.com/getAlby/lightning-browser-extension/commit/141fef9de50963168359b74ef906bc004b679633#diff-18393fdc0e154ed1578d0a767e77e6abf3ccd7debbbf2fd7266bc568e5f937acR99

pavanjoshi914:
tAck

PR #3312, opened 2025-02-28
feat: hide balances if balance is returned as null (when hub connected to getalby.com is offfline)

pavanjoshi914:
before: 
image
image

after: 
image
image


pavanjoshi914:
NaN also belong to number object so we cannot use typeof x == "number"  to exlude it. so i used Number.isFinite() for checks

this helps hiding balances for all the cases where balance doesnt have actual balance returned

PR #3315, opened 2025-03-19
fix: remove e2e test instructions from setup docs

pavanjoshi914:
remove e2e test instructions from setup guide


reneaaron:
What sense does the next section (Run all tests) make if you remove the e2e section: 

https://github.com/getAlby/lightning-browser-extension/blob/master/doc/SETUP.md#run-all-tests

PR #3317, opened 2025-03-20
feat: accept input as raw string for signSchnorr

pavanjoshi914:


im-adithya (commented):
@pavanjoshi914 can you also check all other instances where this is being used and change the param names to `message` instead of `sigHash`?

For ex:
https://github.com/getAlby/lightning-browser-extension/blob/c95bc622c0a8c1b53111376bb5772532447185f6/src/extension/background-script/actions/nostr/signSchnorrOrPrompt.ts#L25

reneaaron (changes requested):
nACK

This constitutes a breaking change to public APIs. We won't implement it without serious justification.

PR #3327, opened 2025-04-03
feat: change keyboard shortcut to Alt+A (works in mv3)

pavanjoshi914:
update shortcut to Alt+A as this works for all chromium based browsers and mv3 manifest 

know more about issue: https://issues.chromium.org/issues/40071206


pavanjoshi914:
tAck

PR #3337, opened 2025-04-07
feat: use inbuilt fetch adapter for axios + upgrade to v1.8.4

pavanjoshi914:
* upgrade axios from v0 to v1
* use inbuilt axios fetch adapter and remove explicit fetch adapter packages which are now broken (cause of internal structure of axios package is changed)
* fix tests


pavanjoshi914:
tACk

PR #3350, opened 2025-04-28
fix: remove battries from extension

pavanjoshi914:
removes battries from the extension


reneaaron:
Please don't forget to update getalby.com, docs & other places.

pavanjoshi914:
updated alby guides to remove twitter and youtube usecases + getalby.com is clear and show integration only for meta tag.

PR #3356, opened 2025-05-07
feat: wallet settings improvements

### LINKED ISSUE #3300: Add "Wallet Settings" to the extension menu
### Feature description

"Wallet settings" are hard to find, many users struggle with finding that icon or knowing where it gets them

image

### Describe the solution

Add it to the 'main' menu, where everyone is looking for it. It could be accessed from both places

image

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:
Fixes #3300

### DIFF
--- a/src/app/components/AccountMenu/index.tsx
+++ b/src/app/components/AccountMenu/index.tsx
@@ -1,7 +1,7 @@
 import {
   PopiconsChevronBottomLine,
+  PopiconsCirclePlusLine,
   PopiconsGlobeLine,
-  PopiconsPlusLine,
   PopiconsSettingsMinimalLine,
 } from "@popicons/react";
 import { useEffect } from "react";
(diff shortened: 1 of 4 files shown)


src/app/components/AccountMenu/index.tsx
```diff
@@ -193,8 +196,10 @@ function AccountMenu({ showOptions = true }: Props) {
                   openOptions("accounts/new");
                 }}
               >
-                <PopiconsPlusLine className="h-4 w-4 mr-2 shrink-0" />
-                {t("options.account.add")}
+                <PopiconsCirclePlusLine className="h-4 w-4 mr-2 shrink-0 text-gray-800 dark:text-neutral-200" />
```
reneaaron, line 199: I know it's in the Figma like that, but I think both of the elements should have the same color, possibly just the default ones. (so you could remove all color classes here and below)

reneaaron (approved):
tACK

pavanjoshi914:
tACk

src/app/components/Modal/index.tsx
```diff
@@ -18,6 +19,7 @@ export default function Modal({
   contentLabel,
   title,
   position = "center",
+  padding = "p-5",
```
reneaaron, line 22: better use `className` here, we don't re-introduce properties usually

src/app/components/AccountMenu/index.tsx
```diff
@@ -55,153 +58,213 @@ function AccountMenu({ showOptions = true }: Props) {
...
+              const account = accounts[accountId];
+              return (
+                <Menu.ItemButton
+                  key={accountId}
+                  onClick={() => {
+                    selectAccount(accountId);
+                    if (window.location.pathname !== "/prompt.html") {
+                      navigate("/");
+                    }
+                    if (isPrompt) {
```
reneaaron, line 227: Instead of having an isPrompt variable I think it would be better to allow instances to pass an onClick function, that is called here. (more generic)

src/app/components/Modal/index.tsx
reneaaron: why do we need different paddings for modals at all? :thinking:

reneaaron (changes requested):
tACK

Works great, just some minor code-related things to fix. :+1:

PR #3360, opened 2025-05-21
feat: if successAction present don't close prompt automatically

### LINKED ISSUE #3348: [Feature] LUD-09: successAction field for payRequest missing
### Feature description

Can it really be that the Alby Extension does not support [LUD-09](https://github.com/lnurl/luds/blob/luds/09.md)? 🤯 

I use LNURLpay with successAction Message and URL. E.g. for offline payments, where the payer is shown a secret pin from the server and wanted to use the URL function now to make a secret Telegram group accessible as soon as the user pays a ticker 🎟️ of 10 Satoshi. Just a spam protection. The browser extension pays the ticket without any problems, but does not provide me with the link to the Telegram group. 

### Describe the solution

Spam protection is valuable and worthwhile. I would be happy if Alby would integrate LUD-09.

### Describe alternatives

_No response_

### Additional context

_No response_

### Are you working on this?

None

pavanjoshi914:
Fixes #3348

### DIFF
--- a/src/app/screens/LNURLPay/index.tsx
+++ b/src/app/screens/LNURLPay/index.tsx
@@ -49,6 +49,8 @@ function LNURLPay() {
   const navState = useNavigationState();
   const details = navState.args?.lnurlDetails as LNURLPayServiceResponse;
 
+  let paymentResponse: PaymentResponse;
+
   const {
     isLoading: isLoadingSettings,
     settings,
@@ -183,7 +185,7 @@ function LNURLPay() {
       }
 
       // LN WALLET pays the invoice, no additional user confirmation is required at this point
-      const paymentResponse: PaymentResponse = await msg.request(
+      paymentResponse = await msg.request(
         "sendPayment",
         { paymentRequest },
         {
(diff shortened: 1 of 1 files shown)


src/app/screens/LNURLPay/index.tsx
```diff
@@ -318,6 +319,7 @@ function LNURLPay() {
               <Button
                 onClick={() => {
                   if (successAction.url) utils.openUrl(successAction.url);
+                  msg.reply(paymentResponse);
```
reneaaron, line 322: Why reply when the user presses a button? :thinking: I am afraid this will lead to inconsistent behavior (whether user chooses to close the popup / clicks "close" button shouldn't make a difference)

@im-adithya Would it be an option to save info (such as the successAction) along with the transaction?

pavanjoshi914: if successAction is url. and when user clicks open url button it should redirect user to url as well as close the prompt. so i added it there

reneaaron: what happens if the user just closes the popup?

im-adithya: > Would it be an option to save info (such as the successAction) along with the transaction?

Ofc, it should be possible because we get the successAction along with the invoice itself 👌 Will add this to the ongoing metadata PRs

pavanjoshi914: what happens if the user just closes the popup?

we still resolve the promise with closePromise. that is why its fine

pavanjoshi914: https://github.com/getAlby/lightning-browser-extension/blob/f985da8ce4630852ae5f18cf7b2a88c97fbc657b/src/common/lib/utils.ts#L120

Issue #3015, opened 2024-02-05
[Feature] Context aware right click menus for lightning-related actions


reneaaron:
### Feature description

![image](https://github.com/getAlby/lightning-browser-extension/assets/100827540/eac32241-7b77-4061-913e-7b67856e8885)

### Describe the solution

Whenever a user selects some text Alby could check if the selected text can be handled (e.g. invoices, LNURL-withdraw, lightning addresses, etc) and display context menus. 

### Are you working on this?

None

stackingsaunter:
Shall we add npubs there? User could choose a client in the settings?

reneaaron:
No, let's keep that separate. It's 2 completely unrelated things that can be done separately.

PR #3382, opened 2025-06-01
feat(context-menu) : enable right-click menu to pay invoices and address

Dunsin-cyber:

Hi @pavanjoshi914   This PR introduces a new context menu in the `background-script/index.ts` file, allowing users to either:

* **Pay a Bitcoin or Lightning invoice**
* **Copy the selected invoice or address**

To support this feature, the necessary permissions were added in `manifest.json`.

Additionally, I updated the **Send** screen so that it automatically sets the invoice from the URL query parameter when the page mounts.

---

### Link this PR to an issue \[optional]

Fixes *#3015* 
This solves the proposed feature I suggested in my proposal from SoB (summer of Bitcoin)

---

### Type of change

* `feat`: New feature (non-breaking change which adds functionality)

---

### Screenshots of the changes \[optional]

Right-click on a Lightning invoice, Bitcoin address, LNURL, or pubkey to bring up this menu:
context-menu

Clicking **"Pay with Bitcoin or Lightning"** opens the send screen:
send-payment-from-context-menu

---

### How has this been tested?

* Tested manually in a Chromium-based browser.
* Verified that right-clicking on an invoice or address shows the context menu.
* Confirmed that clicking “Pay with Bitcoin or Lightning” opens the send screen with the invoice pre-filled.

---

### Checklist

* [x] Self-review of code changes
* [x] Manual testing completed
* [ ] Added automated tests (if applicable)
* [ ] Documentation or guides updated (if needed)
* [ ] Dark mode support (for UI changes)
* [ ] Responsive layout (for UI changes)

### DIFF
--- a/src/app/screens/Send/index.tsx
+++ b/src/app/screens/Send/index.tsx
@@ -5,7 +5,7 @@ import IconButton from "@components/IconButton";
 import TextField from "@components/form/TextField";
 import { PopiconsChevronLeftLine } from "@popicons/react";
 import lightningPayReq from "bolt11-signet";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useLocation, useNavigate } from "react-router-dom";
 import QrcodeAdornment from "~/app/components/QrcodeAdornment";
@@ -18,10 +18,34 @@ function Send() {
   const { t } = useTranslation("translation", { keyPrefix: "send" });
   const { t: tCommon } = useTranslation("common");
   const location = useLocation();
+
   // location.state used to access the decoded QR coming from ScanQRCode screen
   const [invoice, setInvoice] = useState(location.state?.decodedQR || "");
   const navigate = useNavigate();
   const [loading, setLoading] = useState(false);
+
+ 
+  // Extract invoice from URL parameters (if available)
+  // This is passed when a user selects "Pay with Alby" from the browser's context menu
+  useEffect(() => {
+    const params = new URLSearchParams(window.location.search);
+    const invoiceParam = params.get("invoice");
+    if (invoiceParam) {
+      setInvoice(invoiceParam);
+
+      // Clear query params without reloading the page
+      params.delete("invoice");
+      const newSearch = params.toString();
+      const newUrl =
+        window.location.origin +
+        window.location.pathname +
+        (newSearch ? `?${newSearch}` : "") +
+        window.location.hash;
+
+      // Update URL in address bar without page reload
+      window.history.replaceState(null, "", newUrl);
+    }
+  }, []);
 
   function isPubKey(str: string) {
     return str.length == 66 && (str.startsWith("02") || str.startsWith("03"));
--- a/src/extension/background-script/index.ts
+++ b/src/extension/background-script/index.ts
@@ -1,6 +1,7 @@
 import browser, { Runtime, Tabs } from "webextension-polyfill";
 import utils from "~/common/lib/utils";
 
+import { isBitcoinAddress } from "~/app/utils";
 import { isManifestV3 } from "~/common/utils/mv3";
 import { registerInPageContentScript } from "~/extension/background-script/registerContentScript";
 import { ExtensionIcon, setIcon } from "./actions/setup/setIcon";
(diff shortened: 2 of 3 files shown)


src/manifest.json
```diff
@@ -27,7 +30,9 @@
     "scripting",
     "storage",
     "tabs",
-    "unlimitedStorage"
+    "unlimitedStorage",
+    "contextMenus",
```
pavanjoshi914, line 34: hmm if this pr introduces additional permission we need to rethink the usecase of this. as we also have webln and website should use it. instead of right click and pay. wdyt @reneaaron

Dunsin-cyber: oh okay.. alright.

src/extension/background-script/index.ts
```diff
@@ -140,6 +141,66 @@ browser.tabs.onUpdated.addListener(extractLightningData);
...
 
+// create context menu items for lightning actions
+browser.runtime.onInstalled.addListener(() => {
+  browser.contextMenus.create({
+    id: "btc-lightning-pay",
+    title: "Pay with Bitcoin or Lightning",
+    contexts: ["selection"],
+  });
+
+  browser.contextMenus.create({
```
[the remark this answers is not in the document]
Dunsin-cyber: I agree, i have removed it in the recent push

src/manifest.json
Dunsin-cyber: hi @reneaaron and @pavanjoshi914, i have made a new push addressing these:

- I removed the Copy from the context menu since it is available on every browser
- I also removed the other permissions, so the added permission now is only `contextMenus`

### the new contextMenu now looks like this ⏬⬇⏬

updated-context-menu

 also about this

> Don't we have some way to initialize the extension on the send screen?

do you mean Alby should notice the invoice and pop up a helpful action button like:  “Detected invoice — click to pay!”

src/extension/background-script/index.ts
```diff
@@ -140,6 +141,46 @@ browser.tabs.onUpdated.addListener(extractLightningData);
 // When we subscribe to that event asynchronously in the init() function it is too late and we miss the event.
 browser.runtime.onInstalled.addListener(handleInstalled);
 
+// create context menu items for lightning actions
+browser.runtime.onInstalled.addListener(() => {
+  browser.contextMenus.create({
+    id: "btc-lightning-pay",
+    title: "Pay with Bitcoin or Lightning",
```
pavanjoshi914, line 148: currently it always show. is it possible to show it only when right invoice is selected. only if its not very complicated

src/app/screens/Send/index.tsx
```diff
@@ -18,11 +18,34 @@ function Send() {
...
   const location = useLocation();
+
   // location.state used to access the decoded QR coming from ScanQRCode screen
   const [invoice, setInvoice] = useState(location.state?.decodedQR || "");
   const navigate = useNavigate();
   const [loading, setLoading] = useState(false);
 
+  // Extract invoice from URL parameters (if available)
+  // This is passed when a user selects "Pay with Alby" from the browser's context menu
+  useEffect(() => {
```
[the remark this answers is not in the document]
Dunsin-cyber: okay. i'd think of a work around and give you an update.

src/extension/background-script/index.ts
Dunsin-cyber: Hi @pavanjoshi914,
I found a workaround by using a `content script` to listen for selected text and then sending a message to `background script` to show the context menu based on whether the text is valid.

I’ve just pushed the changes and  it now handles everything you requested. ✅

src/app/screens/Send/index.tsx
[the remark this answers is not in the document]
Dunsin-cyber: I pushed a cleanup for this. I no longer manually construct the URL, instead, I use `URLSearchParams` to remove only the invoice param after it's been used. I had to delete it because if it's left there, every time the user visits the send page, it keeps pre-filling that same invoice into the input field even when it's no longer relevant. 

All other params stay untouched; it's just the invoice param (which I added from the background script) that gets removed after it's consumed.

pavanjoshi914:
test it on firefox and firefox mobile as well!!!

pavanjoshi914:
you messed up git history a bit here, pr shouldn't introduce irrelevant commits such as weblate translations. i corrected it for now

Dunsin-cyber:
> test it on firefox and firefox mobile as well!!!

alright

Dunsin-cyber:
> you messed up git history a bit here, pr shouldn't introduce irrelevant commits such as weblate translations. i corrected it for now

oh... alright. noted. thanks

pavanjoshi914:
what happens when url is "https://example.com/page?invoice=...", "https://example.com/page?param1=xyz&invoice="



will it endput in "https://example.com/page?" or "https://example.com/page?&"




can you also update me on that

Dunsin-cyber:
> what happens when url is "https://example.com/page?invoice=...", "https://example.com/page?param1=xyz&invoice="
> 
> will it endput in "https://example.com/page?" or "https://example.com/page?&"
> 
> can you also update me on that

yes, the native `URL API + URLSearchParams.delete()` automatically handles joining the base URL, pathname and avoids trailing ? or &, so there’s no risk of malformed URLs  even if `invoice` is the only or last param.

For example:

- `?invoice=123 ➝ becomes ''"`

- `?param1=xyz&invoice=123 ➝ becomes ?param1=xyz`

the images below shows the URL at different stages, before the invoice is extracted and after the invoice has been extracted, 

i logged the output in  the console with a Prefix of **URL**


invoice-params-1
invoice-params-2

Dunsin-cyber:
> test it on firefox and firefox mobile as well!!!

still on this.. 

works well on firefox desktop but i'm trying to set up android studio on my machine. will get back to you as soon as i test the mobile

Dunsin-cyber:
Hi @pavanjoshi914 — just a quick follow-up on [this comment](https://github.com/getAlby/lightning-browser-extension/pull/3382#issuecomment-2985168127)

I tested the context menu feature on both Firefox Desktop and Firefox Mobile using Firefox Nightly (on Android versions <8.0 and >10.0) via `web-ext run` from my PC.

However, on mobile, the context menu didn't appear at all, not even when testing with a minimal, custom extension that only adds a basic contextMenus item. I confirmed this behavior across devices.

After some digging, I found that Firefox Mobile currently doesn't expose the context menu API to extensions in the same way as desktop does. The mobile long-press menu is native and not fully modifiable via the `contextMenus` API.

Here are a few supporting sources:

* [MDN discussion thread on context menu support](https://github.com/orgs/mdn/discussions/473)
* [Reddit thread highlighting the same limitation](https://www.reddit.com/r/firefox/comments/1bgdg3j/can_you_modify_the_contextmenu_of_firefox_in/)

src/extension/background-script/index.ts
```diff
@@ -140,6 +139,57 @@ browser.tabs.onUpdated.addListener(extractLightningData);
...
+// Handle menu clicks
+browser.contextMenus.onClicked.addListener((info, tab) => {
+  const text = info?.selectionText?.trim();
+  if (!text) return;
+
+  if (info.menuItemId === "btc-lightning-pay") {
+    const sendUrl = browser.runtime.getURL(
+      `popup.html?invoice=${encodeURIComponent(text)}#/send`
+    );
+    browser.windows.create({
```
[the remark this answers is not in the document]
Dunsin-cyber: how did i miss this beautiful helper function? 😁😁
i’ll test with it and get back to you shortly

pavanjoshi914:
this pr is taking more longer than required can you iterate through it and also do a self review of code. and then i will do final tests before merge

Dunsin-cyber:
> this pr is taking more longer than required can you iterate through it and also do a self review of code. and then i will do final tests before merge

apologies, i was testing different use cases to see how we can use the  `utils.openPrompt`

it works great for passing `args`, `origin`, and `action`, but I’m running into a routing issue when opening the prompt from the background script.

The prompt URL looks like this:

```perl
prompt.html?action=send
```
But since the extension uses `HashRouter` , the router expects something like:

```bash
prompt.html#/send
```
Without that #/send, I get an error:

```pgsql
router.js:247 No routes matched location "/send"
```

If we want to use the `utils.openPrompt`, we would have to modify it.

Dunsin-cyber:
> this pr is taking more longer than required can you iterate through it and also do a self review of code. and then i will do final tests before merge

yes i have done this and everything looks good.
i have also tested it on firefox and chrome, works well on both browsers.

pavanjoshi914:
tAck

PR #3387, opened 2025-06-06
feat: add metadata to transactions

pavanjoshi914:


src/extension/background-script/connectors/connector.interface.ts
```diff
@@ -37,6 +38,11 @@ export interface ConnectorTransaction {
   displayAmount?: [number, ACCOUNT_CURRENCIES];
   type: "received" | "sent";
   state?: "settled" | "pending" | "failed";
+  recipient_data?: {
```
rolznz, line 41: shouldn't this be inside the metadata below?

src/app/components/TransactionsTable/index.tsx
```diff
@@ -91,19 +119,9 @@ export default function TransactionsTable({
...
-                            : t(
-                                tx.state === "settled"
-                                  ? "sent"
-                                  : tx.state === "pending"
-                                  ? "sending"
-                                  : tx.state === "failed"
-                                  ? "failed"
-                                  : "sent"
-                              ))}
+                        {tx.title || tx.boostagram?.message || typeStateText}
```
im-adithya, line 122: Why is boostagram message taking the place of title or typeStateText? Shouldn't it be in the description?

im-adithya: Ah okay, because here we don't have descriptions. Can we unify that as well? So the date appears on the side of the title and description on the bottom?

pavanjoshi914: cool resolved. 

image

pavanjoshi914:
tACk

im-adithya:
Can you check the failing unit tests?

PR #3402, opened 2025-07-16
fix: use bolt11 field

### LINKED ISSUE #3400: [BUG] Creating new invoice fails with connected LNBits
Creating an invoice leads to an empty page due to a JS error when using the LNBits connector. 

TypeError: can't access property "toUpperCase", y.paymentRequest is undefined

reneaaron:

Fixes #3400

Should we add both fields? There are probably quite some instances out there that haven't been updated to the latest LNBits version?


pavanjoshi914:
why fallback is needed here for the api response? payment_request field its not longer returned in the api response right?

pavanjoshi914:
tested rest of the functions for this connector. they are working fine

reneaaron:
> why fallback is needed here for the api response? payment_request field its not longer returned in the api response right?

Yes, however there might be quite some outdated versions out there still returning the old field. Until all of them are upgraded we can just support both.

pavanjoshi914:
tAck

PR #3417, opened 2025-08-08
fix: remove alby cln proxy

pavanjoshi914:


pavanjoshi914:
closing as we decided to use nwc to connect cln to extension

PR #3428, opened 2025-09-11
feat: allow to connect  cln nodes through nwc 

pavanjoshi914:


bumi:
You have to think about existing users who have existing CLN wallets connected. 
Can we really remove the code and the dependency? - I guess in a first step we should just change the UI for new connections

pavanjoshi914:
I don't remember the exact discussion that happened before the conference but we were discussing this in meeting and decided to remove ln message as well. 

i followed the written notes: https://www.notion.so/getalby/Week-41-2025-15-08-25058e6cf56c80bbae62ce051d2cd511

@reneaaron do you know more. or shall we keep connector code for now?

reneaaron:
Since the ln-proxy is down for months (right?) and we never received any complains I assumed not many users are using this and would be tempted to remove it / replace it with NWC.

pavanjoshi914:
cc @bumi

bumi:
@reneaaron ok, then let's do it

pavanjoshi914:
tAck

PR #3434, opened 2025-09-15
feat: mnemonic generation for spark connector

pavanjoshi914:
1. we cant add conditionals directly in state level to check if mnemonic is present or not as it breaks other connectors 
2. update connector config to use mnemonic passed from account variable 
3. I have to introduce new script to encrypt mnemonic ( we store mnemonic encrypted and when account variable is saved it should have encrypted one and we have decrypt mechanism in settings page when its shown ) 
4. but spark don't accepts this encrypted mnemonic so the flow is:: generate and encrypt mnemonic at screen level --> pass to connector when initializing --> decrypt and pass it to initialized method  and then rest of the flows such as mnemonic showing in settings page will work

### DIFF
--- a/src/extension/background-script/connectors/spark.ts
+++ b/src/extension/background-script/connectors/spark.ts
@@ -269,7 +269,19 @@ class SparkConnector implements Connector {
   }
 
   async signMessage(args: SignMessageArgs): Promise<SignMessageResponse> {
-    throw new Error("Method not implemented.");
+    if (!this._wallet) {
+      throw new Error("Wallet not initialized");
+    }
+    const signature = await this._wallet.signMessageWithIdentityKey(
+      args.message
+    );
+
+    return {
+      data: {
+        message: args.message,
+        signature: signature,
+      },
+    };
   }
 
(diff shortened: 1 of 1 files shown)


src/app/screens/connectors/ConnectSpark/index.tsx
```diff
@@ -22,10 +23,16 @@ export default function ConnectSpark() {
   async function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
     event.preventDefault();
     setLoading(true);
+
+    const mnemonic = await api.generateMnemonic();
+    const encryptedMnemonic = await api.encryptMnemonic(mnemonic);
```
rolznz, line 28: is this OK? I saw we have an existing `setMnemonic` function which will also derive the nostr key

pavanjoshi914: yess! at this stage there is no account stored in db. so we can't use that script

rolznz: But what about the nostr key? do we need to do anything there?

rolznz (approved):
utACK

PR #3437, opened 2025-09-17
Feat: Add Spark connector

rolznz:
spark connector

image

### DIFF
--- a/src/i18n/locales/de/translation.json
+++ b/src/i18n/locales/de/translation.json
@@ -278,7 +278,8 @@
           },
           "errors": {
             "connection_failed": "Verbindung fehlgeschlagen. Ist Ihre NWC Wallet online und wurden App Verbindungen zugelassen?"
-          }
+          },
+          "instructions": "Fügen Sie eine NWC-Verbindungszeichenfolge aus <0>Alby Hub</0> oder aus anderen NWC-fähigen Wallets ein."
         },
         "title": "NWC (Nostr Wallet Connect)"
       },


rolznz:
@pavanjoshi914 I think we should fix this before merging. The link to setup the keys is also broken because it shows steps to create a new master key. 
image

image

I think it would be simpler to derive the nostr key too when generating the master key and extract the code so the same things happen when we generate a master key for Spark as what happens if you generate a master key normally.

CC @reneaaron @bumi

src/common/lib/api.ts
```diff
@@ -265,9 +269,9 @@ const setMnemonic = (id: string, mnemonic: string | null): Promise<void> =>
     mnemonic,
   });
 
-const encryptMnemonic = (mnemonic: string): Promise<string> =>
-  msg.request("encryptMnemonic", {
-    mnemonic,
+const encryptKey = (key: string): Promise<string> =>
```
rolznz, line 272: I think this should be called `encryptValue = (value: string)` right?

src/extension/background-script/actions/nostr/generatePrivateKey.ts
```diff
@@ -22,16 +23,19 @@ const generatePrivateKey = async (message: MessageNostrPrivateKeyGenerate) => {
       };
     }
     const mnemonic = new Mnemonic(decryptData(account.mnemonic, password));
-    const nostrPrivateKey = mnemonic.deriveNostrPrivateKeyHex();
-
-    return {
-      data: nostrPrivateKey,
-    };
+    nostrPrivateKey = mnemonic.deriveNostrPrivateKeyHex();
+  } else if (message.args?.mnemonic) {
```
rolznz, line 27: I'm just wondering if this is the best approach since we add more complexity to these functions to both handle accounts and non-accounts.

But I can't see any other way except actually creating the account before we validate it :thinking:

rolznz:
@bumi @reneaaron we introduce something new here with the mnemonic generation before the account creation, and it creates a bit of complexity in the code. Not sure if this is OK for now or there is a better way we could do it?

PR #3450, opened 2025-10-15
feat: kind description

pavanjoshi914:
image


reneaaron (commented):
Can you maybe try to do an info circle icon next to the kind instead? I don't like the many different font styles on this page already, and let's not make it worse.

pavanjoshi914:
we need to use <0><0></0></0> while nesting. explained here: https://github.com/i18next/react-i18next/issues/928#issuecomment-2580575003

other ways didn't worked for me. 

image

reneaaron (approved):
One minor comment regarding the translation component. Otherwise this is great! 👍  

tACK

PR #3455, opened 2025-10-23
feat: handle sub sat amounts

pavanjoshi914:
bolt-11 signet doesnt set "invoice.satoshis" if sub sat amount is passed. in that case we have only use millisatoshi. 

this pr adds support to use msats if satoshis field is not present and hence we show proper amount while paying invoices 

image

not tested end to end


reneaaron:
@copilot review this PR based on the bolt11 spec: https://www.bolt11.org/

pavanjoshi914:
@reneaaron copilot created empty pr instead of reviewing it

PR #3463, opened 2025-12-11
fix: disable context menus on android

pavanjoshi914:
contextMenus web api not supported on firefox on android

### DIFF
--- a/src/common/utils/os.ts
+++ b/src/common/utils/os.ts
@@ -3,6 +3,7 @@ function getOS() {
   if (userAgent.indexOf("Win") !== -1) return "Windows";
   if (userAgent.indexOf("Mac") !== -1) return "MacOS";
   if (userAgent.indexOf("X11") !== -1) return "UNIX";
+  if (userAgent.indexOf("Android") !== -1) return "Android";
   if (userAgent.indexOf("Linux") !== -1) return "Linux";
 }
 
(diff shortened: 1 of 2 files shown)


pavanjoshi914:
tACk

IsabelSydowISComplicated:
> contextMenus web api not supported on firefox on android

I apologize for my ignorance, I am very NOT tech savvy and I don't know how to use this forum. 
I was brough here by following a reply to a question which described a problem similar to what I have been experiencing: 

"...As of last week, Alby extension on Android Firefox on a Samsung tablet was working fine.
It's been now a few days since it stopped working.
I had to uninstall it it, re-install it, and nothing.
Also, when you go to the extension settings page, it just brings you to a white page, where no matter how many times you reload, clear cache, do headstands, and downward dogs, nothing happens. Basically unable to use any nostr application since Alby extension was the log in..."

If I read this correctly: "@pavanjoshi914
[fix: _**disable context menus on android**_](https://github.com/getAlby/lightning-browser-extension/pull/3463/commits/d96aac576f70052f1c23d97b04d5bf0aa9fbc6f9)"

Sounds like that would fix the issue I am experiencing, But I don't understand/know the steps to accomplish doing such. 
Thank you!!!

pavanjoshi914:
the issue is fixed but we are actively working on the new release of extension which will include this fix as well. we will do release as soon as possible and with new version you should be able to use extension again. thankyou for adding your concern here

IsabelSydowISComplicated:
> the issue is fixed but we are actively working on the new release of extension which will include this fix as well. we will do release as soon as possible and with new version you should be able to use extension again. thankyou for adding your concern here

Will the new release be announced somewhere?

PR #3470, opened 2026-01-03
Fix/blink lightning address

### LINKED ISSUE #2993: [BUG] Receive window when connected to Blink wallet doesn't show the lightning address qrcode

### Describe the bug

Receive window when connected to Blink wallet doesn't show the lightning address qrcode

Captura de tela de 2024-01-14 17-43-10

### Screenshots [optional]

_No response_

### Steps To Reproduce

1- Connect Blink wallet on Alby
2- Go to Alby receive page

### Expected behavior

See the Blink lightning address qrcode

### Alby information

Alby: 3.6.0

### Device information

_No response_

### Additional context

_No response_

### Are you working on this?

No

keshav0479:
Fixes #2993

## Problem
Previously, the "Receive" screen only displayed a Lightning Address QR code for Alby OAuth accounts. Blink (Galoy) wallets did not show a QR code because the `lightningAddress` was:
1. Not returned by the `Galoy` connector during setup.
2. Not persisted to `IndexedDB` in the `add` action.
3. Not exposed to the frontend via the `info` API endpoint.

### Visual Comparison

| Before Fix (Blink Wallet) | After Fix (Blink Wallet) |
| :---: | :---: |
| Before Fix | After Fix |
| *No QR code displayed* | *Lightning Address QR code is visible* |

## Solution
This PR implements a full-stack fix to persist and display the address for Galoy accounts:

### Backend / Connector
* **`src/types.ts`**: Extended `Account` interface to include `lightningAddress`.
* **`galoy.ts`**: Updated `getInfo()` to correctly derive the address (e.g., `user@blink.sv`) from the username and globals.
* **`add.ts`**: Updated account creation logic to fetch `connector.getInfo()` and persist the `lightningAddress` to storage immediately.
* **`info.ts`**: Updated the account info endpoint to bubble up the `lightningAddress` to the UI API response.

### Frontend
* **`Receive/index.tsx`**: Removed the strict `isAlbyOAuthUser` check. The component now displays the QR code for *any* account that has a valid `lightningAddress` available.

## Testing
* Verified that adding a new Blink wallet correctly saves the address to `IndexedDB` and Sync Storage.
* Verified that the Receive screen displays the QR code immediately after creation.
* Verified that refreshing the extension does not overwrite or lose the address data.

### DIFF
--- a/src/app/screens/Receive/index.tsx
+++ b/src/app/screens/Receive/index.tsx
@@ -15,7 +15,6 @@ import QRCode from "~/app/components/QRCode";
 import SkeletonLoader from "~/app/components/SkeletonLoader";
 import toast from "~/app/components/Toast";
 import { useAccount } from "~/app/context/AccountContext";
-import { isAlbyOAuthAccount } from "~/app/utils";
 import api from "~/common/lib/api";
 import { IconLinkCard } from "../../components/IconLinkCard/IconLinkCard";
 
@@ -28,7 +27,6 @@ function Receive() {
   const [loadingLightningAddress, setLoadingLightningAddress] = useState(true);
 
   const [lightningAddress, setLightningAddress] = useState("");
-  const isAlbyOAuthUser = isAlbyOAuthAccount(auth.account?.connectorType);
 
   useEffect(() => {
     (async () => {
@@ -60,7 +58,7 @@ function Receive() {
       <div className="pt-4">
         <Container justifyBetween maxWidth="sm">
           <div className="flex flex-col gap-2 md:gap-3">
-            {isAlbyOAuthUser && (
+            {lightningAddress && (
               <div className="bg-white dark:bg-surface-01dp border-gray-200 dark:border-neutral-700 rounded border p-4 flex flex-col justify-center items-center gap-1 text-gray-800 dark:text-neutral-200">
                 <>
                   <div className="relative flex flex-grid">
--- a/src/extension/background-script/connectors/galoy.ts
+++ b/src/extension/background-script/connectors/galoy.ts
@@ -94,6 +94,9 @@ class Galoy implements Connector {
               id
               username
           }
+          globals {
+            lightningAddressDomain
+          }
         }
       `,
     };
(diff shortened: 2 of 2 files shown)


pavanjoshi914:
this pr has many issues. 
1. introducing uneccessary types, and interface
2. complexity
3. introducing lightning address at toplevel of account even though we have it on info.data which can be used to render lightning address 
4. no need of changes in add account and info account
5. additional complexity in receive screen

i will create a fresh pr for this feature

src/app/components/QRCode/index.tsx
```diff
@@ -19,17 +23,40 @@ export type Props = {
...
   return (
-    <ReactQRCode
-      value={value}
-      size={size}
-      fgColor={fgColor}
-      bgColor={bgColor}
-      className={classNames("w-full h-auto rounded-md", className ?? "")}
-      level={level}
-    />
+    <div
```
pavanjoshi914: we don't need this as we show lnaddess and copy button below qr. less code is always better

pavanjoshi914:
once we do the following changes we don't require to change any of the files. 

here is the draft pr for solution: https://github.com/getAlby/lightning-browser-extension/pull/3473

thank for bringing this in front of. showing lnaddress related to other connector also adds a addition to UX.

keshav0479:
Hey @pavanjoshi914, thanks for taking the time to review this and for the detailed feedback.
You're right ,i definitely overcomplicated things here. After looking at your draft PR #3473, i can see much cleaner approach there:
*  Just return `lightning_address` (snake_case) in the Galoy connector
*  The existing `api.ts` layer already handles the conversion
*  Frontend only needs to swap `isAlbyOAuthUser` --> `lightningAddress` check

I got caught up trying to persist the address at the account level and ended up touching way more files than needed. learned lesson... should've studied the existing patterns more carefully before diving in..

Happy to update my PR to match the minimal approach if that would be helpful, or feel free to proceed with #3473 ,whatever works best for you.

pavanjoshi914:
It would be great that you update your pr and we can merge yourse as you
put efforts to draft the pr

keshav0479:
Done! I stripped out all the extra stuff so it's now just the minimal changes exactly like in your draft, also added the mapping for `pay.bbw.sv` -> `blink.sv` so it shows the clean domain.Thanks for the guidance on this!

pavanjoshi914:
closing as already fixed

PR #3473, opened 2026-01-16
feat: show lnaddress and qr code for lnaddress for galoy instances

pavanjoshi914:
show lnaddress and qr for lnaddress on receive screen for galoy instance (blink, bitcoinjungle) 

image

## Summary by CodeRabbit

* **New Features**
  * Receive payment screen now available to any account with a configured lightning address (previously limited by account type).
  * Backend now supplies lightning address domain info and returns a computed lightning address when possible, improving how receiving addresses are displayed.

### DIFF
--- a/src/extension/background-script/connectors/galoy.ts
+++ b/src/extension/background-script/connectors/galoy.ts
@@ -115,7 +115,7 @@ class Galoy implements Connector {
         ? data.me.username.substr(0, 10)
         : data.me.id.substr(0, 8);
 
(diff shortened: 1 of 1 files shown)


pavanjoshi914:
pay.bbw.sv is legacy address and is what returned as lnaddress domain via api. should be map this to blink.sv 

both are supported for now. 

> LN address: All users have a Lightning address such as user@blink.sv. (BTW, the legacy user@pay.bbw.sv addresses will continue to work)

https://www.blink.sv/blog/introducing-blink

Issue #3489, opened 2026-02-13
Missing input validation in account creation flow


abhishek-8081:
When a new account is added through the extension, there is no validation on the incoming data before saving it.

Looking at [src/extension/background-script/actions/accounts/add.ts](cci:7://file:///Users/abhishekkumar/lightning-browser-extension/src/extension/background-script/actions/accounts/add.ts:0:0-0:0), the code takes whatever is passed in `message.args` and directly saves it to storage without checking anything. There are no checks for:

- Whether the connector type is actually valid
- Whether the config object has all the required fields
- If the account name is empty or too long
- If the config data is in the expected format

There are even TODO comments left in the code saying `// TODO: add validations` and `// TODO: make sure a password is set`.

Right now bad or incomplete data can get saved into the extension's storage which can cause confusing errors later. For example if someone passes a wrong connector type, things will break only when they try to use the account, not when they create it.

**Steps to reproduce:**
1. Look at [src/extension/background-script/actions/accounts/add.ts](cci:7://file:///Users/abhishekkumar/lightning-browser-extension/src/extension/background-script/actions/accounts/add.ts:0:0-0:0)
2. Notice there is no validation before the account data is encrypted and saved

**Expected behavior:**
The add account function should validate the input data before saving — check that required fields exist, connector type is supported, and config format is correct. If validation fails it should return a clear error message.

PR #3490, opened 2026-02-13
fix: add input validation for account creation

abhishek-8081:
Fixes the missing input validation when adding a new account.

## What changed

The `add` function in `src/extension/background-script/actions/accounts/add.ts` had no validation
on the incoming account data before encrypting and saving it. There were TODO comments saying
`// TODO: add validations` and `// TODO: make sure a password is set`.

This PR adds three checks before saving:

1. **Connector type** — must be one of the known connector types (lnd, lndhub, lnbits, alby, nwc, etc.)
2. **Config** — must be a non-empty string
3. **Name** — must be a non-empty string

Each check returns a clear error message if it fails.

## Tests

Added 3 new test cases in `add.test.ts`:
- Returns error for invalid connector type
- Returns error for missing config
- Returns error for missing name

All 5 tests pass (2 existing + 3 new).

Fixes : #3489 

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for account creation with error messages for invalid connector type, missing account config, and missing account name.

* **Tests**
  * Added test cases to verify error handling for invalid account creation inputs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,6 @@ We welcome and appreciate new contributions! 🎉 To get started, [Refer to SETU
 
 #### 🎨 Designer
 
-- Have a look at our [Open source Design guide](https://github.com/getAlby/lightning-browser-extension/wiki/Open-source-Design)
 - Check out the issues that have specifically been [marked with "design"](https://github.com/getAlby/lightning-browser-extension/issues?q=is%3Aopen+is%3Aissue+label%3A%22design%22)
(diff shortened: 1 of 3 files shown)


pavanjoshi914:
@abhishek-8081 where this validations are really needed? we have fron't end forms. where user can't submit unless we get required information. there is validateAccount call that's being runned  to check if anything is wrong. 



Connector type - TypeScript enforces  -  all callers use valid hardcoded strings
Config present	 -All callers always pass config -validateAccount already runs first 
Name present	-All callers always pass a name - getUniqueAccountName handles edgecases

i guess we can close this unless we need more defensive code there cc @reneaaron

PR #3586, opened 2026-07-27
Correct allowance budget checks for WebLN payments

reneaaron:

The invoice amount used for the per-host budget check was read from the decoded `satoshis` field. `bolt11` only populates that field for whole-satoshi amounts, so an invoice with millisatoshi precision (e.g. 1000.5 sat) was compared against the budget as `0`.

The same class of problem existed on the other side of the budget: after a payment the allowance was debited by the connector's `route.total_amt`, which NWC and Eclair compute with `Math.floor`. A 999 msat invoice passed the (rounded-up) check as 1 sat but debited 0, so a site could pay repeatedly without ever touching the budget.

- adds `getPaymentRequestAmountSats()` to `src/common/utils/paymentRequest.ts`, which returns `ceil(millisatoshis / 1000)` and `null` when the invoice has no amount (or an explicit zero amount)
- uses it in `checkAllowance` **and** in `updateAllowance`, so the check and the debit always use the same number; keysends keep using `total_amt` since they carry no invoice
- uses it for the `total_amt` reported by the lnbits, lndhub, galoy and NWC connectors, and in the `ConfirmPayment` / `ConfirmPaymentAsync` prompts so the displayed amount and budget suggestion match the background
- rounds LNC and Eclair `total_amt` up instead of persisting fractional or floored budgets
- also checks `allowance.enabled` and `enabledFor` before paying from a budget as defense in depth (page-originated calls already pass through `enable`, which enforces both, so this is not a behaviour change)
- amountless invoices now open the confirm prompt instead of being checked against the budget

## Tests

- `src/common/utils/__tests__/paymentRequest.test.ts` — amount derivation, including explicit zero amounts
- `src/extension/background-script/actions/webln/__tests__/sendPaymentOrPrompt.test.ts` — prompt vs. auto-pay for whole, sub-satoshi and amountless amounts; the disabled-allowance case now isolates the `enabled` flag (verified by mutation: removing the gate fails the test)
- `src/extension/background-script/events/__test__/updateAllowances.test.ts` — the debit uses the invoice amount even when the connector reports `total_amt: 0`
- the invoice generator used by these tests lives in `src/fixtures/paymentRequests.ts`

`yarn lint:js`, `yarn tsc:compile`, `yarn format` and `yarn test:unit` all pass (182 passed, 2 pre-existing skips).

## Summary by CodeRabbit

* **New Features**
  * Improved payment amount handling for sub-satoshi invoices with consistent upward rounding.
  * Added consistent invoice amount processing across supported payment connections.

* **Bug Fixes**
  * Amountless invoices now always prompt for payment instead of being automatically paid.
  * Automatic payments require an enabled allowance with WebLN permission.
  * Allowance balances, payment confirmations, notifications, and history now use the invoice amount when available.
  * Improved handling of invoices where payment amounts are omitted by the service.


im-adithya (approved):
tACK

PR #3598, opened 2026-08-24
fix: scope nostr schnorr signing to messages the user can read

reneaaron:

`window.nostr.signSchnorr(sigHash)` took a caller-supplied 32-byte digest and signed it with the account's nostr key. A digest carries no information about what it stands for, so the confirmation screen could only show 64 hex characters — and a nostr event id *is* such a digest, which means a signature over one is a valid event signature produced outside `signEvent` and its per-kind permissions.

This PR narrows the schnorr signing surface to messages the extension can actually show the user.

- **`signSchnorr(sigHash)` is removed** from the nostr provider and from the signer. `hashAndSignSchnorr(message)` is now the only schnorr entry point. The background action rejects a `sigHash` argument too, so posting straight to the content script does not reach the signer either — the provider object was never the boundary.
- **Messages that are NIP-01 event serializations are rejected**, with an error pointing at `signEvent`. The digest of a serialization is the event id, so signing the serialization is equivalent to signing the event; without this, removing the raw-digest entry point would have changed nothing.
- **NIP-26 delegation strings get their own screen.** `nostr:delegation:<pubkey>:<conditions>` is recognised and the delegatee and the conditions are shown as separate fields with an explanation, instead of one long opaque line. A token with empty conditions is labelled as unlimited rather than falling back to the generic screen.
- **Everything else is shown in a scrollable box** rather than `ContentMessage`'s eight-line clamp, so a message padded with leading newlines cannot push its payload out of view.
- **"Block and ignore <host>" is offered on the screen.** Nothing is remembered any more, so without it a page that keeps asking has no remedy short of deleting the allowance.
- **Schnorr signing always asks.** `addPermissionFor` no longer stores an approval for `nostr/signSchnorr`, so a bulk "don't ask again for any nostr requests" cannot cover it either. Blocks are still stored and still honoured. The permission selector is gone from the screen, and a migration drops approvals stored by older versions so the permission list stops showing a grant that has no effect.

New `src/common/utils/nostrSigning.ts` holds the two classifiers (`isEventSerialization`, `parseDelegation`), shared by the background action and the confirmation screen so both describe a request the same way.

## Breaking change

**`window.nostr.signSchnorr()` no longer signs anything.** A stub is kept so callers get `nostr.signSchnorr was removed. Pass the message to nostr.hashAndSignSchnorr, or use nostr.signEvent to sign an event.` instead of a bare `TypeError`. Callers that were signing a digest they computed themselves need to pass the message to `hashAndSignSchnorr()` instead and let the extension hash it; callers that were signing nostr events need `signEvent()`. This is not part of NIP-07 — it is an Alby-specific method — but it is public API and sites do use it, so this wants a heads-up in the release notes.

## User-visible change

**Anyone who previously chose "Don't ask again for sign schnorr" will be asked again.** The stored approval is deleted on upgrade and no new one can be created; every schnorr signature now needs a confirmation. Sites that relied on the grant being silent will start showing a prompt per request.

## Screenshots

The same NIP-26 delegation request, before and after.

| Before | After |
| --- | --- |
| before | after |

## Tests

- `src/common/utils/__tests__/nostrSigning.test.ts` — serialization detection and delegation parsing, including the near misses
- `src/extension/background-script/actions/nostr/__tests__/signSchnorr.test.ts` — rewritten: rejects serializations, always prompts, never stores an approval, still honours a block
- `src/extension/background-script/permissions/__tests__/addPermissionFor.test.ts` — the approval guard, and that blocks still persist
- `src/extension/background-script/migrations/__tests__/index.test.ts` — the upgrade drops approvals and keeps blocks

`yarn lint`, `yarn tsc:compile` and `yarn test:unit` pass (189 passed, 2 pre-existing skips). Also checked by hand in a Chromium profile with a production build: the delegation screen renders as above, and a page that calls the removed method gets the deprecation error rather than a signature.


im-adithya:
Closing in favor of https://github.com/getAlby/lightning-browser-extension/pull/3611

PR #3599, opened 2026-08-24
feat: authenticated at-rest encryption for account secrets

### LINKED ISSUE #3612: Update encryption format for stored account credentials
## Description

`encryptData` in `src/common/lib/crypto.ts` wraps connector configs (macaroons, NWC URIs), the mnemonic, and the Nostr key with crypto-js OpenSSL AES: key+IV from EVP_BytesToKey (MD5, 1 iteration), AES-256-CBC, no authentication. Those blobs live in `browser.storage.sync`

Reading the ciphertext already requires access to the extension profile on-device. Hence it is not that big of an issue to tackle immediately.

### Possible Fix

• Versioned AEAD (AES-GCM via WebCrypto or @noble/ciphers)
• Strong KDF (PBKDF2-SHA256 ≥ 600k iterations, or argon2)
• Re-encrypt existing blobs on next unlock
• Password strength checks (PasswordForm currently accepts numeric-only)

reneaaron:
Fixes #3612 

Account secrets — connector configs (macaroons, NWC URIs, LNDHub credentials), the mnemonic, and the nostr private key — are encrypted at rest before being written to `browser.storage.sync`. This moves that encryption from crypto-js's passphrase mode to a modern, authenticated format with a strong key-derivation function:

- **`src/common/lib/crypto.ts`** now writes a versioned envelope: **PBKDF2-SHA256 (600,000 iterations)** for key derivation and **AES-256-GCM** for encryption. Both primitives come from `@noble/*`, already in the dependency tree (used by nip44). The functions keep their **synchronous** signatures, so the ~25 call sites (connectors and actions) are unchanged.
- **GCM authenticates the ciphertext**: a modified blob now fails to decrypt instead of silently yielding altered plaintext.
- **Backwards read compatibility**: `decryptData` detects the format and still reads the previous crypto-js blobs, so data written by older versions — including blobs that sync down from another device that hasn't updated yet — keeps working. New writes are always the new format.
- **On unlock**, any secrets still in the old format are re-encrypted to the new one (`actions/accounts/upgradeEncryption.ts`). It is best-effort and wrapped so it can never block unlock: any blob it can't re-encrypt is left byte-for-byte unchanged.
- Derived keys are **memoised per (password, salt, iterations)** (bounded cache, cleared on lock/reset/password-change) so decrypting the same stored blob repeatedly within a session only pays the ~0.45s derivation once.

---

## ⛔ Read this before merging: a rollback after release causes permanent user lockout

**The new format is write-once-forward.** A v2 blob **cannot be read by any build ≤ 3.14.5** — those builds have no v2 reader at all, and their crypto-js path throws on it (verified).

That creates an asymmetry that is easy to miss:

- Old blob read by a new build → **fine** (this PR reads both formats).
- **New blob read by an old build → permanent lockout for that user.** The account data is intact, but that build cannot decrypt it and will report "invalid password".

**Why this is a live risk, not a hypothetical:** releases get rolled back for reasons that have nothing to do with this change — an unrelated regression shipped in the same version. If a release containing v2 *writing* goes out and is then rolled back, **every user who unlocked even once in the meantime has had their secrets rewritten to v2 and is locked out of the rolled-back build.** There is no recovery path from the rolled-back build; the user must go forward to a v2-capable version again.

The same asymmetry applies across devices: once any device writes v2, the `accounts` key syncs, and a device still on ≤ 3.14.5 cannot unlock until it updates.

## Release sequencing — a maintainer decision, not made here

This PR currently **reads and writes v2 in the same release**. That is one of two options, and the choice is yours:

**Option A — reader-first (two releases).** Split this PR: ship only the v2 *reader* (plus legacy read/write) first, let it propagate, then enable v2 *writing* and the on-unlock upgrade in a later release.
- *Cost:* the weak format keeps being written until the second release, so remediation is delayed by roughly one release cycle. Requires splitting this PR and tracking the follow-up.
- *Benefit:* a rollback of either release is safe. No user can be locked out.

**Option B — ship as-is (one release).** Remediation lands immediately for everyone on the new version.
- *Cost:* accepts the rollback risk above for the window between this release and the point where a rollback is no longer plausible.
- *Benefit:* the weak format stops being written on day one.

**My recommendation is Option A**, on the grounds that the failure mode is permanent loss of wallet access for real users, triggered by an unrelated regression — while the cost is a delay of roughly one release cycle for a storage hardening improvement whose benefit only applies to someone who already holds a copy of the user's profile or sync data. But this is a release-management call with context I don't have — the extension's rollback history and staged-rollout policy should decide it, and I've deliberately not made that decision in code.

---

## Behaviour / performance notes

- **Change-password** re-encrypts every account's three secret fields; each encryption now runs a 600k-iteration KDF (~0.45s each). For a handful of accounts this is a few seconds — worth a spinner if there isn't one.
- **MV3 cold start**: the key cache lives in the service worker, so the first decrypt after a worker restart pays one ~0.45s PBKDF2 (previously ~1ms). This is the cost of a strong KDF on a synchronous surface; a native `crypto.subtle` path would be faster but is async and would touch all ~25 call sites. Happy to pursue that separately.
- The Alby connector re-encrypts its config on token refresh; that now costs one KDF derivation, in the background.
- No new dependency is added; crypto-js remains only for reading legacy blobs.

## Migration placement

Re-encryption runs on **unlock** rather than as a startup migration in `migrations/index.ts`, because the startup migration runner executes before the password is available.

## Hardening included

- Salt/nonce/ciphertext are base64, not hex, to stay within the `storage.sync` per-item quota for accounts holding several secrets.
- The stored iteration count is validated (integer within a bounded range) before key derivation, so a corrupted or out-of-range envelope is rejected as invalid rather than sending PBKDF2 into a long synchronous stall the unlock error path can't catch. Raising the default past that bound is a breaking format change and is commented as such.
- The derived-key cache is cleared on `lock()`, `reset()` and `changePassword`, so keys don't outlive the unlocked session or a replaced password. Encryption keys (fresh salt per write, single-use) are never cached.

## Tests

- `src/common/lib/__tests__/crypto.test.ts` — v2 round-trip, versioned-envelope shape, fresh salt/nonce per call, wrong-password rejection, **tamper rejection (AEAD tag)**, rejection of out-of-range iteration counts, cache clearing, legacy-blob read, and all `reEncryptToLatest` paths (upgrade, no-op on v2, safe pass-through on wrong password / null).
- `actions/accounts/__tests__/unlock.test.ts` — extended to assert a legacy-format config is re-encrypted to the new format on unlock.

`yarn tsc:compile` and `yarn lint:js` pass clean; `yarn test:unit` passes 76 suites / 181 tests (2 pre-existing skips). `prettier --check src` reports one pre-existing failure in `src/app/styles/index.css`, which is untouched by this PR and also fails on `master`.

## Summary by CodeRabbit

- **New Features**
  - Newly protected data now uses stronger authenticated encryption.
  - Existing encrypted data remains readable and is automatically upgraded after unlock.
  - Added protection against tampering, incorrect passwords, and invalid encryption parameters.

- **Security**
  - Cached encryption keys are cleared when locking, resetting, or changing the password.

- **Bug Fixes**
  - Encryption upgrade failures no longer block unlocking.
  - Password setup and changes reject passwords shorter than eight characters or containing only numbers.
  - Password changes now require matching confirmation.

### DIFF
--- a/src/app/components/PasswordForm/index.tsx
+++ b/src/app/components/PasswordForm/index.tsx
@@ -3,6 +3,12 @@ import type { KeyPrefix } from "i18next";
 import React, { useState } from "react";
 import { useTranslation } from "react-i18next";
 import PasswordViewAdornment from "~/app/components/PasswordViewAdornment";
+import {
+  UNLOCK_PASSWORD_MIN_LENGTH,
+  getPasswordError,
+  type PasswordConfirmationError,
+  type PasswordError,
+} from "~/common/utils/validations";
 
 export type PasswordFormData = {
   password: string;
@@ -19,13 +25,10 @@ export type Props<T extends PasswordFormData = PasswordFormData> = {
   autoFocus?: boolean;
 };
 
-type errorMessage =
-  | ""
-  | "enter_password"
-  | "confirm_password"
-  | "mismatched_password";
-
-const initialErrors: Record<string, errorMessage> = {
+const initialErrors: {
+  passwordErrorMessage: PasswordError;
+  passwordConfirmationErrorMessage: PasswordConfirmationError;
+} = {
   passwordErrorMessage: "",
   passwordConfirmationErrorMessage: "",
 };
@@ -36,7 +39,7 @@ export default function PasswordForm<
   formData,
   setFormData,
   i18nKeyPrefix,
-  minLength,
+  minLength = UNLOCK_PASSWORD_MIN_LENGTH,
   confirm = true,
   autoFocus = true,
 }: Props<T>) {
@@ -71,17 +74,15 @@ export default function PasswordForm<
   }
 
   function validate() {
-    let passwordErrorMessage: errorMessage = "";
-    let passwordConfirmationErrorMessage: errorMessage = "";
+    let passwordConfirmationErrorMessage: PasswordConfirmationError = "";
 
-    if (!formData.password) passwordErrorMessage = "enter_password";
     if (confirm && !formData.passwordConfirmation) {
       passwordConfirmationErrorMessage = "confirm_password";
     } else if (confirm && formData.password !== formData.passwordConfirmation) {
       passwordConfirmationErrorMessage = "mismatched_password";
     }
     setErrors({
-      passwordErrorMessage,
+      passwordErrorMessage: getPasswordError(formData.password, minLength),
       passwordConfirmationErrorMessage,
     });
   }
@@ -115,7 +116,7 @@ export default function PasswordForm<
         />
         {errors.passwordErrorMessage && (
           <p className="mt-1 text-red-500">
-            {t(`errors.${errors.passwordErrorMessage}`)}
+            {t(`errors.${errors.passwordErrorMessage}`, { min: minLength })}
           </p>
         )}
       </div>
--- a/src/app/screens/Onboard/SetPassword/index.tsx
+++ b/src/app/screens/Onboard/SetPassword/index.tsx
@@ -9,6 +9,7 @@ import { useTranslation } from "react-i18next";
 import { useNavigate } from "react-router-dom";
 import toast from "~/app/components/Toast";
 import msg from "~/common/lib/msg";
+import { isValidUnlockPassword } from "~/common/utils/validations";
 
 const initialFormData = {
   password: "",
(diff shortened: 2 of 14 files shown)


im-adithya:
Follow-up commit on this branch:

New unlock passcodes must be at least 8 characters and must not be numbers only. Enforced in `PasswordForm` (onboarding + settings) and in the `setPassword` / `changePassword` actions. Existing passcodes still unlock.

`lock()` and `reset()` now clear the derived-key cache again after their async work finishes, so an in-flight decrypt cannot leave a key cached after the session ends.

im-adithya (approved):
tACK

im-adithya:
Closing this for now as this issue is tracked in #3612

PR #3600, opened 2026-08-24
fix: scope the nostr permission presets to what each one describes

reneaaron:

Each Connect preset now grants a defined set of permissions, and the Connect screen lists them. Previously the three presets were described only as "Auto-sign all requests (except payments)" / "Auto-approve most common requests" / "Do not sign anything without asking me!", and "most common requests" covered rather more than the phrase suggests.

**What each preset grants now**

- **Let's be reasonable** — notes, reposts, reactions, zap requests, lists, long-form and app data. Events that change the account's profile, contacts or relay list, that speak privately as the user, or that authenticate the user to a relay or a website are confirmed individually.
- **I fully trust it** — signs anything and can send encrypted messages as the user.
- **Reading encrypted messages is not granted by any preset**, including "I fully trust it". Each decryption is confirmed individually.

The lists live in `src/common/utils/nostrPresets.ts` so the background script, the Connect screen and the migration share one definition.

**Also in here**

- The preset's permissions are now awaited before `enable()` resolves. They were written by an un-awaited `forEach(async …)`, and by a `Promise.all` over a callback that returned nothing — so `enable()` could resolve before any of them were persisted.
- A migration brings existing connections in line with what the presets now describe. It runs after the nip04/nip44 rename migrations, which normalise the method names it matches on. Blocked entries are left in place, since those are denials rather than grants.

## Behaviour changes worth noting

**Users will be prompted again for things that used to be silent.** Anyone who connected a site before this change loses the withdrawn grants on upgrade and is asked the next time that site needs one.

**The revoke is unconditional.** A permission row records only its method, not which preset or prompt created it. A grant the user made deliberately through "don't ask again" for one of these methods is removed along with the preset's. `nostr/encrypt` is deliberately *not* in that set, because "I fully trust it" still grants it.

**Relay and website login prompts may be noisy.** NIP-42 relay auth (kind 22242) and NIP-98 HTTP auth (kind 27235) now confirm individually, and `openPrompt` opens one window per request with no queueing. A client that authenticates to a dozen relays on load will stack a dozen prompts, and again on every reconnect. I have kept them out of the preset because both authenticate as the user, but if that proves too noisy the alternatives are to put 22242 back in the preset or to give it its own opt-in line on the Connect screen — happy to do either.

**Not addressed here:** the `DONT_ASK_ANY` branch in `signEventOrPrompt`, `signSchnorrOrPrompt`, `encryptOrPrompt`, `decryptOrPrompt` and `getPublicKeyOrPrompt` still grants every method in bulk, including `nostr/decrypt`, and still does so with the same un-awaited `forEach(async …)`. That path is being handled separately; left alone here to avoid conflicting with it.

## Screenshots

| Before | After |
| --- | --- |
| before | after |

## Tests

- `src/extension/background-script/actions/nostr/__tests__/enable.test.ts` — what each preset grants and does not grant, and that the grants are persisted before `enable()` resolves
- `src/extension/background-script/migrations/__tests__/revokeWithdrawnNostrPresetPermissions.test.ts` — what the migration removes and keeps, that it leaves blocked entries alone, and that it runs after the rename migrations

Also exercised by hand in a Chromium profile against a production build: the reasonable preset still signs notes, reposts, reactions, zap requests and long-form with no prompt, so posting and zapping are unaffected.

`yarn lint`, `yarn tsc:compile` and `yarn test:unit` pass (77 suites, 179 passed, 2 pre-existing skips).

Non-English `description` strings for the three presets are reset, since the English copy changed.

## Summary by CodeRabbit

- **New Features**
  - Nostr permission presets now clearly show granted capabilities and requests requiring approval.
  - “Reasonable,” “Fully trusted,” and “Paranoid” modes have updated permission behavior and descriptions.
  - Added an option to ask about every request.

- **Bug Fixes**
  - Existing permissions are safely updated when preset rules change, while encryption permissions remain preserved.

- **Localization**
  - Updated English preset descriptions and streamlined preset translations across supported languages.

- **Tests**
  - Added coverage for preset behavior, permission persistence, and migration handling.


reneaaron:
Closing this. The presets are intentionally opinionated defaults with sensible choices, not an exhaustive enumeration of event kinds, and I'd rather keep them that way than turn the Connect screen into a spec sheet. The un-awaited permission writes in enable() are worth fixing on their own; I'll open that separately.

PR #3601, opened 2026-08-24
Reserve the allowance budget before sending a WebLN payment

reneaaron:

Builds on #3586 — **please merge that first**; this PR targets `fix/allowance-msat-bypass`, so only the commits on top are mine.

The per-host budget was read before a WebLN payment was dispatched and written back only after it settled, from a pubsub subscriber. Two things follow from that gap:

- several payments from the same origin can each be measured against the same budget before any of them is written back, so together they spend more than it allows
- the write-back is a read-modify-write, so overlapping payments overwrite each other's result and the budget ends up higher than what was actually spent

### What changed

- `checkAllowance` now takes the amount out of the budget in the same Dexie transaction that reads it. A payment that passes the check has already been debited before the connector is called.
- The settle-time `updateAllowance` subscriber is removed. It was the second debit for budget payments and the only debit for prompt-confirmed ones.

### Behaviour worth noting

- The budget is debited pessimistically. The amount is never put back: if the payment fails, throws, never returns, or the MV3 service worker is torn down mid-payment, it stays deducted. The failure mode is a budget that is too small until the user tops it up, rather than one that can be spent twice.
- **Prompt-confirmed payments no longer count against the site's budget.** Previously a payment the user approved by hand on a site with an allowance was also debited at settle time. The budget now only caps what the site can spend *without* asking. `lastPaymentAt` on the allowance is likewise only updated by budget payments.

### Not addressed here

- `persistSuccessfulPayment` still throws when a connector settles without reporting a route (LNDHub keysend). Pre-existing, separate fix.
- `actions/allowances/add.ts` does not set `enabledFor`, so a budget set through the confirm prompt on a row that predates `enabledFor` is not honoured. That gate comes from #3586 and is noted on it.
- Routing fees still escape the budget (the connectors report `total_fees: 0`). Tracked separately.

## Tests

- `sendPaymentOrPrompt.test.ts` — the amount is taken out before paying, and five concurrent payments against a 500 sat budget reach the connector twice and prompt three times
- `updateAllowances.test.ts` removed with the subscriber

`yarn lint:js`, `yarn tsc:compile` and `yarn test:unit` pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Allowance balances are now debited atomically before payments are sent, preventing concurrent payments from exceeding the available budget.
  - Remaining budgets and payment timestamps are saved immediately when an allowance is used.
  - Invalid payment amounts, including negative or non-finite values, are rejected.
  - Payment allowance accounting no longer performs a separate update after successful payment notifications, preventing duplicate budget deductions.


reneaaron:
Superseded by #3613, which is rebuilt against master without the reservation flag.

PR #3606, opened 2026-08-31
Give each provider scope its own message channel

reneaaron:

The inpage providers and their content scripts communicated over `window.postMessage` on the page window. Everything sent that way — the correlation id, the arguments and the reply — is delivered to every other `message` listener in the frame, and the first reply carrying a matching id settles the call.

Each scope now gets its own `MessageChannel`. The content script keeps one end and transfers the other to the inpage world; requests, replies and events travel over that channel rather than the page window.

### What changed

- `content-script/messagePortServer.js` (new) creates the channel and hands the port over. It is called at module top, ahead of the asynchronous should-inject/blocklist decision, which now only gates whether requests get serviced. Requests that arrive before the scope registers a handler are buffered and flushed.
- `providers/postMessage.ts` requests its port and retries until it arrives — either world can initialise first — then routes replies to the waiting caller by id.
- A reply for an id that has already been settled is logged and dropped rather than applied to the result.
- `accountChanged` now reaches the providers over the channel instead of being read off the page window.
- The five content scripts (`webln`, `nostr`, `webbtc`, `liquid`, `alby`) move to the shared helper, which removes a fair amount of duplicated listener code.

### Scope of the change, and what it does not do

The port handover is negotiated with window messages, so the channel is private only from scripts that were not already running when the port was transferred. On MV2 the inpage script is injected inline at `document_start` and that ordering holds. On MV3 the main-world script is registered separately and the ordering is **not** guaranteed, so a script that is already running in the frame can still take part in the handover.

Treat this as a narrowing of the shared page-window surface, not as a guarantee that provider results are authenticated to the page. The security boundary remains the extension-rendered confirmation UI, which the page cannot draw. Worth deciding separately (not in this PR):

- whether to make provider results self-verifying where the cryptography allows it — a `sendPayment` preimage can be checked against the BOLT11 payment hash by the integrator
- whether to pin the provider objects on `window`. `Object.freeze()` on the instances is not viable as-is: `providerBase` assigns `this._isEnabled` after construction and these are ES modules, so freezing would make `enable()` / `isEnabled()` throw. A non-configurable `window.webln` / `window.nostr` would prevent replacement but risks breaking coexistence with other WebLN/NIP-07 extensions.

## Tests

- `providers/__tests__/postMessage.test.ts` — a request goes over the port and its reply resolves the call; a reply of the old shape replayed on the page window does not resolve it and the genuine reply still does; a second reply for a settled id does not overwrite the result; errors reject; events arrive over the port and are not picked up from the window.
- `content-script/__tests__/messagePortServer.test.ts` — a request over the port is serviced and answered on it, requests before a handler are buffered, the port is transferred once, another scope's handshake is ignored, events go over the port.
- JSDOM has no `MessageChannel`. `tests/unit/helpers/fakeMessageChannel.ts` supplies a minimal stand-in, installed per test file rather than globally, because msw depends on the real one.

`yarn lint`, `yarn tsc:compile` and `yarn test:unit` pass (77 suites, 179 passed, 2 pre-existing skips), and `yarn build:chrome` succeeds.

### DIFF
--- a/src/extension/content-script/__tests__/messagePortServer.test.ts
+++ b/src/extension/content-script/__tests__/messagePortServer.test.ts
@@ -1,0 +1,182 @@
+import { installFakeMessageChannel } from "../../../../tests/unit/helpers/fakeMessageChannel";
+
+let restoreMessageChannel: () => void;
+beforeAll(() => {
+  restoreMessageChannel = installFakeMessageChannel();
+});
+afterAll(() => {
+  restoreMessageChannel();
+});
+
+// The isolated-world side of the provider transport: it hands one end of a
+// channel to the inpage world and services requests that arrive over it.
+
+type PortLike = {
+  onmessage: ((ev: { data: unknown }) => void) | null;
+  postMessage: (data: unknown) => void;
+  start: () => void;
+};
+
+// Every createScopePort() adds a window listener that lives for the rest of the
+// file, so each test uses its own scope name and only its own server answers.
+let scopeCounter = 0;
+const nextScope = () => `webln-${++scopeCounter}`;
+
+async function loadServer() {
+  let mod!: typeof import("../messagePortServer");
+  await jest.isolateModulesAsync(async () => {
+    mod = await import("../messagePortServer");
+  });
+  return mod;
+}
+
+// Play the inpage side: ask for the port and take the one that is transferred.
+function requestPort(scope: string): PortLike {
+  const posted: unknown[][] = [];
+  const spy = jest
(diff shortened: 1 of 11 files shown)


reneaaron:
Closing without merging. A script running in the page's own realm can replace or wrap `window.webln` / `window.nostr` directly, so a private transport between the inpage providers and the content scripts does not make the page's view of the providers any more trustworthy, and it adds permanent complexity for that. The confirmation prompt rendered by the extension remains the trust boundary, which matches how MetaMask and other injected-provider wallets are built.

PR #3610, opened 2026-09-08
feat: remove webln.request to reduce complexity

im-adithya:

Drops the public webln.request API so websites can no longer forward arbitrary LND/LNC RPCs through the extension. `getInfo()` now lists only first-class WebLN methods. Existing `webln/lnd|lnc|commando` permissions are cleaned up by a one-shot migration.

### Type of change

### How has this been tested?

Tested manually using an LND connection

### Checklist

- [x] Manual testing
- [x] Update Docs & Guides
- For UI-related changes
- [x] Darkmode
- [x] Responsive layout

## Summary by CodeRabbit

* **Removed Features**
  * Removed support for generic WebLN RPC requests through the provider.
  * Existing generic-request permissions are removed during migration.
  * Removed related permission descriptions and documentation.

* **Maintenance**
  * Removed unused SnakeCase dependencies.
  * Connector method availability is now explicitly listed rather than generated dynamically.
  * Updated permission migration handling to cover matching permissions across connectors.


im-adithya:
Related: https://github.com/getAlby/webln-guide/pull/9 (removes webln `request` docs from guides)

src/extension/background-script/migrations/index.ts
```diff
@@ -92,6 +92,21 @@ const migrations = {
 
     console.info("Migration migrateDecryptPermission complete.");
   },
+
+  migrateRemoveWeblnRequestPermissions: async () => {
+    // webln.request permissions were stored as `webln/<connector>/<method>`.
+    // The connector segment came from `connector.constructor.name`, which is
+    // mangled in production builds, so match on the shape instead of the name.
+    // Other webln permissions (e.g. `webln/sendpayment`) have no third segment.
+    const weblnRequestMethod = /^webln\/[^/]*\/[^/]+$/;
```
reneaaron, line 101: **Why this regex matches on shape instead of connector names**

The removed `webln.request` handler stored its "always allow" grants as

```
webln/${connector.constructor.name.toLowerCase()}/${method}
```

`constructor.name` is a runtime identifier, and our production webpack config runs Terser with mangling enabled and no `keep_classnames`. In the shipped `background.bundle.js` no connector class name survives: `Lnd` becomes something like `class JI{`, and subclasses such as `NativeLnd` compile to anonymous `class extends` expressions whose name is the empty string. So real user databases contain rows like `webln/ji/getinfo` or `webln//getinfo`, never `webln/lnd/getinfo`. An explicit `lnd|lnc|commando` list would have matched only dev builds.

Because the mangled identifier can change between releases, these grants also silently reset on updates and left a fresh orphaned row each time. Long-time users may hold several generations of them.

What is stable is the shape: `webln.request` rows are the only permissions with three segments. Every other permission key is a string constant (`nostr/signMessage`, `bitcoin/getAddress`, `liquid/getAddress`, …), and WebLN payment grants live in `allowances`, not here. Matching `^webln/<anything>/<method>$` therefore removes exactly the orphaned `webln.request` rows and nothing else.

reneaaron (approved):
utACK

PR #3611, opened 2026-09-08
fix: always prompt for nostr schnorr signing

im-adithya:

Confirm every signSchnorr and hashAndSignSchnorr call. Drop the remember-this-method selector, show that confirmation is always required, and offer Block and ignore to add the host to the blocklist. Remove nostr/signSchnorr from PermissionMethodNostr so Trust Fully and "don't ask for any" no longer grant it. Rename the handler from signSchnorrOrPrompt to signSchnorr.

### Type of change

### How has this been tested?

Manual testing

### Checklist

- [x] Manual testing
- [x] Update Docs & Guides
- For UI-related changes
- [x] Darkmode
- [x] Responsive layout

## Summary by CodeRabbit

- **New Features**
  - Users can block a requesting site directly from the Schnorr signing confirmation prompt; blocked requests are rejected and added to the blocklist.
  - Nostr confirmation choices are applied consistently across supported operations.

- **Updates**
  - Schnorr signing requests now use the streamlined `signSchnorr` operation.
  - Confirmation prompts display an “always confirm” notice instead of per-request permission controls.
  - Trusted Nostr settings no longer explicitly exclude Schnorr signing from automatic approval.


src/i18n/locales/en/translation.json
```diff
@@ -562,7 +562,7 @@
       "presets": {
         "trust_fully": {
           "title": "I fully trust it",
-          "description": "Auto-sign all requests (except payments)"
+          "description": "Auto-sign all requests (except payments and Schnorr signing)"
```
reneaaron: I don't think this helps any users who reads it. Since usage of signSchnorr is anyways very limited, I'd just ignore it and leave the message as is.

src/extension/background-script/actions/nostr/signSchnorr.ts
```diff
@@ -0,0 +1,55 @@
...
+import { MessageSignSchnorr, Sender } from "~/types";
+
+import state from "../../state";
+
+const signSchnorr = async (message: MessageSignSchnorr, sender: Sender) => {
+  const host = getHostFromSender(sender);
+  if (!host) return;
+
+  const nostr = await state.getState().getNostr();
+  const sigHash = message.args.sigHash;
```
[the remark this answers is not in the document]
im-adithya: This is a commonly occurring pattern and is only caused due to a broken caller, which is not an extension issue. The fallout is also just an error and not a security hole, so this can be left unfixed.

src/i18n/locales/en/translation.json
im-adithya: Done ✅

src/i18n/locales/en/translation.json
```diff
@@ -899,6 +899,7 @@
       "allow_sign_event": "Allow {{host}} to sign a <0>{{kind}}<0></0></0> event",
       "block_and_ignore": "Block and ignore {{host}}",
       "block_added": "Added {{host}} to the blocklist, please reload the website.",
+      "always_confirm": "This request always requires confirmation",
```
reneaaron: can we get rid of this? i don't think it's really needed?

reneaaron: Removed in 95d67e64d, both the translation key and the paragraph in ConfirmSignSchnorr.

reneaaron (approved):
utACK

PR #3613, opened 2026-09-08
fix: debit the allowance budget before sending a WebLN payment

reneaaron:

Replaces #3601, rebuilt against `master` so it does not depend on #3586.

The per-host budget was read before a WebLN payment was dispatched and written back only after it settled, from a pubsub subscriber. Two things follow from that gap:

- several payments from the same origin can each be measured against the same budget before any of them is written back, so together they spend more than it allows
- the write-back is a read-modify-write, so overlapping payments overwrite each other's result and the budget ends up higher than what was actually spent

### What changed

- `checkAllowance` now takes the amount out of the budget in the same Dexie transaction that reads it. A payment that passes the check has already been debited before the connector is called. `keysendOrPrompt` uses the same function, so keysend is covered too.
- The settle-time `updateAllowance` subscriber is removed. It was the second debit for budget payments and the only debit for prompt-confirmed ones.

### Behaviour worth noting

- The budget is debited pessimistically. The amount is never put back: if the payment fails, throws, never returns, or the MV3 service worker is torn down mid-payment, it stays deducted. The failure mode is a budget that is too small until the user tops it up, rather than one that can be spent twice.
- **Prompt-confirmed payments no longer count against the site's budget.** Previously a payment the user approved by hand on a site with an allowance was also debited at settle time. The budget now only caps what the site can spend *without* asking. `lastPaymentAt` on the allowance is set whenever a payment for the host settles, budget or prompt-confirmed, so the Publishers list keeps sorting by the last actual payment.

### Not addressed here

- The `enabled` / `enabledFor` and msat checks in `checkAllowance` come from #3586. That PR touches the same function, so whichever lands second needs a trivial rebase.
- `persistSuccessfulPayment` still throws when a connector settles without reporting a route (LNDHub keysend). Pre-existing, separate fix.
- Routing fees still escape the budget: the fee is only known after the payment settles, so a site could drain more than its budget through fees. Out of scope here. Fee limits need to be enforced on the node side.

## Tests

`sendPaymentOrPrompt.test.ts` — the amount is taken out before paying and stays out when the payment fails, and five concurrent payments against a 500 sat budget reach the connector twice and prompt three times.

`yarn lint:js`, `yarn tsc:compile` and `yarn test:unit` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Allowance balances are reserved before WebLN payments and keysend requests proceed.
  * Failed payments no longer restore reserved allowance funds, preventing budget reuse.
  * Concurrent requests now respect available allowance funds.
  * Disabled or WebLN-incompatible allowances fall back to payment confirmation.
  * Amountless invoices bypass allowance debits and proceed through confirmation.
  * Fractional-satoshi amounts are rounded up when debiting allowances.
  * Allowance balances are no longer updated separately after successful payment events.

* **Tests**
  * Added coverage for budget limits, failed payments, concurrent spending, amountless invoices, rounding, and allowance eligibility.


src/extension/background-script/events/index.ts
```diff
@@ -20,8 +19,6 @@ const subscribe = () => {
 
     // @ts-expect-error typed as ln.sendPayment.success | ln.keysend.success
     PubSub.subscribe(`ln.${type}.success`, persistSuccessfulPayment);
-    // @ts-expect-error typed as ln.sendPayment.success | ln.keysend.success
-    PubSub.subscribe(`ln.${type}.success`, updateAllowance);
   });
```
im-adithya: **[bug]** Dropping the `ln.sendPayment.success` / `ln.keysend.success` `updateAllowance` subscriber means the only `remainingBudget` write is the reservation in `checkAllowance`. That is correct for silent WebLN/keysend, but ConfirmPayment and ConfirmKeysend still do `addAllowance` (sets `remainingBudget = totalBudget`) and then `api.sendPayment` / `keysend` — they never call `checkAllowance`. The payment that creates or resets the budget is no longer subtracted, so a user who pays 100 sats and sets the default 10× budget of 1000 is left with 1000 silent sats, not 900. The same hole applies when remember-me overwrites an existing allowance. LNURL pay and popup Send also go straight to `sendPayment` and will no longer touch an existing site budget; that matches the PR's stated model, but the remember-me UI (`BudgetControl`: "until it is exhausted") was built around the settle-time debit and was not updated. `lastPaymentAt` is likewise only bumped on reserved silent payments, so prompted/LNURL activity no longer sorts the publishers list.

**Suggestion:** Keep a debit for payments that were not already reserved. The smallest complete fix is: when ConfirmPayment/ConfirmKeysend save a budget, write `remainingBudget = totalBudget - amountPaid` (and set `lastPaymentAt`), and/or have the settle subscriber subtract only if this payment did not go through `checkAllowance`. Do not leave remember-me depending on a subscriber that no longer exists.

src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
```diff
@@ -25,13 +25,33 @@ const sendPaymentOrPrompt = async (message: Message, sender: Sender) => {
...
+    }
+
+    await db.allowances.update(allowance.id, {
+      remainingBudget: allowance.remainingBudget - amount,
+      lastPaymentAt: Date.now(),
+    });
+    return true;
+  });
+
+  if (debited) await db.saveToStorage();
```
im-adithya: **[suggestion]** The IDB debit is serialized, but `saveToStorage()` runs after the transaction and rewrites the entire `browser.storage.local` snapshot (allowances, payments, blocklist, permissions) with no mutex. Two overlapping reservations can persist an earlier, higher `remainingBudget` after a later one has already committed to IDB. `db.ts` `loadFromStorage` treats that snapshot as the restore source whenever IndexedDB comes up empty — which this codebase explicitly handles — so a concurrent pair of budget payments can undo the overspend fix after a service-worker restart that drops IDB. Failed/hung payments are correctly not refunded in IDB; this is only the persist-layer race.

**Suggestion:** Serialize `saveToStorage` (single-flight queue) or persist only the changed allowance row under a lock so a stale full-table write cannot revive a larger budget.

src/extension/background-script/actions/webln/__tests__/sendPaymentOrPrompt.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    await db.allowances.clear();
+    await db.allowances.bulkAdd(allowanceFixture);
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test("takes the amount out of the budget before paying", async () => {
+    await sendPaymentOrPrompt(message(100), sender);
```
im-adithya: **[suggestion]** "takes the amount out of the budget before paying" only asserts `remainingBudget === 400` after `sendPaymentOrPrompt` resolves. That passes if the debit happened after `sendPayment` returned. The failure test mocks a throw, but real `ln/sendPayment` catches connector errors and returns `{ error }` — same no-refund outcome, just not what the test exercises. The `Promise.all` case is the useful one: without an in-check debit all five 200-sat payments would reach the connector; with fake-indexeddb's overlapping-write queue it also checks that later reads see earlier writes. It still does not force a yield between read and write, so it is a weak proof of IDB isolation versus "debit happens in `checkAllowance` at all."

**Suggestion:** In the `sendPayment` mock, read `db.allowances` before resolving and assert the 100-sat reservation is already gone. Optionally delay the mock so the five concurrent calls overlap the read/write window. Add a case where `sendPayment` resolves `{ error: "boom" }` without throwing.

src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
```diff
@@ -25,13 +25,33 @@ const sendPaymentOrPrompt = async (message: Message, sender: Sender) => {
   }
 };
 
+// Checks the budget and takes the amount out of it in a single transaction,
```
im-adithya: **[suggestion]** The three-line comment restates the new control flow (check + debit in one transaction, no refund) and narrates why the old race existed. That is design history, not a non-obvious constraint a later reader could not get from the code and the `Number.isFinite` / `remainingBudget - amount` update.

**Suggestion:** Drop it, or keep a single line on the policy that is not obvious from the code: reserved sats are not credited back on failure.

im-adithya:
One major problem I saw with this was that it shows random amounts (sum of all previous failed attempts) in budget spent even without the relevant txs, do we wish to also list them here or deduct the allowance budget spent on failure?

Screenshot 2026-09-09 at 8 33 55 PM

src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
```diff
@@ -25,13 +25,33 @@ const sendPaymentOrPrompt = async (message: Message, sender: Sender) => {
...
+// before the payment is sent. The amount is not put back if the payment fails.
+// Concurrent payments would otherwise all read the same budget and each spend
+// it.
 async function checkAllowance(host: string, amount: number) {
```
bumi: can we rename the function maybe `checkAndDebitAllwoance()` ?

bumi (commented):
I think it's OK. 
there should not be failed payments.

src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
```diff
@@ -25,13 +25,33 @@ const sendPaymentOrPrompt = async (message: Message, sender: Sender) => {
...
-  const allowance = await db.allowances
-    .where("host")
-    .equalsIgnoreCase(host)
-    .first();
+  if (!Number.isFinite(amount) || amount < 0) return false;
 
-  return allowance && allowance.remainingBudget > amount; // check that the budget is higher than the amount. amount can be 0
+  const debited = await db.transaction("rw", db.allowances, async () => {
```
reneaaron, line 43: is this transaction needed here?

reneaaron: Yes, it is what makes the fix work. Without it the read and the update run as two separate Dexie calls, so concurrent payments all read the same budget before any of them writes.

I verified by removing the `db.transaction("rw", ...)` wrapper and rerunning the test file:

| | with transaction | without |
|---|---|---|
| 5 concurrent 200 sat payments vs 500 sat budget | 2 reach the connector, 3 prompt | all 5 reach the connector |

IndexedDB serialises overlapping readwrite transactions on the same store, so wrapping the read and the update in one transaction is what makes the check-and-debit atomic.

src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
reneaaron: Done in 2597adf11, renamed to `checkAndDebitAllowance` (also the import in `keysendOrPrompt.ts`).

src/extension/background-script/actions/webln/__tests__/sendPaymentOrPrompt.test.ts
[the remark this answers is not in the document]
reneaaron: Applied in 0d2b65942, the mock now asserts the budget is 400 while the payment is in flight.

src/extension/background-script/actions/webln/sendPaymentOrPrompt.ts
bumi: 🙏

src/extension/background-script/events/persistPayments.ts
```diff
@@ -44,6 +44,9 @@ const persistSuccessfulPayment = async (
     allowanceId: allowance ? (allowance.id ?? "").toString() : "",
     paymentRequest: "",
   });
+  if (allowance?.id) {
+    await db.allowances.update(allowance.id, { lastPaymentAt: Date.now() });
+  }
```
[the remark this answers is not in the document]
reneaaron: Intentional. `lastPaymentAt` means "last time this site was paid" and is used to sort the Publishers list, so it has to cover prompt-confirmed payments as well. Only the budget is restricted to auto-pay. Updated the PR description to match.

src/extension/background-script/actions/webln/keysendOrPrompt.ts
```diff
@@ -20,7 +20,7 @@ const keysendOrPrompt = async (message: Message, sender: Sender) => {
       error: "Destination or amount missing.",
     };
   }
-  if (await checkAllowance(host, parseInt(amount as string))) {
+  if (await checkAndDebitAllowance(host, parseInt(amount as string))) {
```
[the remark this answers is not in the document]
reneaaron: Pre-existing: `keysendOrPrompt` parsed the amount with `parseInt` before this PR too, and the keysend action does the same parse before handing it to the connector, so the debited amount and the paid amount always match. This PR only renamed the call and added a guard that rejects NaN and negative amounts. Input validation for fractional or suffixed strings is a separate WebLN concern.

PR #3615, opened 2026-09-08
fix: mirror migrated tables to browser.storage.local for every migration

reneaaron:

Migrations write to IndexedDB, but the extension keeps a copy of every table in `browser.storage.local` and re-seeds IndexedDB from that copy whenever it starts empty (`db.loadFromStorage()`). That copy is only refreshed by explicit `db.saveToStorage()` calls, and only one of the four migrations made one. So after a migration deleted or renamed permission rows, the pre-migration rows were still in the mirror and came back the next time IndexedDB was unavailable, while the migration stayed marked as applied. This affects every start for users on the in-memory fallback (Firefox private browsing, Tor Browser) and Firefox site-data clears.

This moves the mirror write into the migration runner: after any migration runs, `db.saveToStorage()` is called once, before `setMigrated()`, so a crash in between reruns the migration instead of leaving a stale mirror behind. The four `if (shouldMigrate(...))` blocks collapse into `runMigration(name)` calls. Migration bodies are unchanged; the explicit call inside `migrateRemoveWeblnRequestPermissions` is now redundant but left as is.

Adds tests for the migration runner covering the new schnorr migration (IndexedDB, mirror, and re-seed from mirror) and one of the 2023 migrations.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

- New `migrations/__tests__/index.test.ts`: the mirror and re-seed assertions fail on master and pass with this change.
- `yarn test:unit`, `yarn lint:js`, `yarn tsc:compile` all green.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration reliability by ensuring permission changes are synchronized across local storage.
  * Corrected permission migration behavior for decryption permissions.
  * Removed obsolete Schnorr signing permissions from stored permission data.

* **Tests**
  * Added coverage for permission cleanup, migration behavior, storage synchronization, and data reseeding scenarios.


im-adithya (approved):
tACK

